PR #10, opened 2021-02-11
Switch back to only supporting legacy signature formats.

dlorenc:
We shouldn't try to be too fancy here, it's best to stick with one
until more registries support the index format.


jonjohnsonjr (approved):
This seems fine, given that it's more likely to be broadly compatible with all registries, but the purist inside of me weeps a bit.

dlorenc:
Same :(

PR #29, opened 2021-02-21
WIP: Add timestamp signatures as a flag.

dlorenc:


dlorenc:
Still need to think through the UX of how to verify these.

You need to know the public key or cert for a TSA if you want them to have any value.

PR #39, opened 2021-03-02
Switch to OCI media types.

dlorenc:


font (approved):

dlorenc:
cc @jdolitsky we should figure out how to get this in.

dlorenc:
Would it make sense to register a media type somewhere?

jdolitsky:
Strictly speaking, the OCI artifacts project says to register one with IANA. In reality, I don't know of any project has gone through with that process..

That being said, I do think it makes sense for this project to define a custom mediatype. See https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md#defining-a-unique-artifact-type

So something along the lines of:
```
application/vnd.sigstore.cosign.config.v1+json
```

This could just be published in the cosign docs somewhere.

For registries that only allow Docker, perhaps provide a bool flag such as `--as-docker`, or make it even more customizable with a string flag/env var.

dlorenc:
We fixed this in #300

PR #43, opened 2021-03-03
Update to prevent error for base64 encoded signature passed in via arg

font:
I was getting a `no such file or directory` error when passing in the raw base64 encoded signature via command line arg as shown https://github.com/sigstore/cosign#upload-a-generated-signature. It's because the error signals to interpret the argument as a raw base64 encoded signature instead of a file or `stdin`. So this clears the error to avoid inadvertently returning an error and exiting the program.

### DIFF
--- a/cmd/cli/upload.go
+++ b/cmd/cli/upload.go
@@ -34,7 +34,7 @@ import (
 func Upload() *ffcli.Command {
 	var (
 		flagset   = flag.NewFlagSet("cosign upload", flag.ExitOnError)
-		signature = flagset.String("signature", "", "path to the signature or {-} for stdin")
+		signature = flagset.String("signature", "", "the signature, path to the signature, or {-} for stdin")
 		payload   = flagset.String("payload", "", "path to the payload covered by the signature (if using another format)")
 	)
 	return &ffcli.Command{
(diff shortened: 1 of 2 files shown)


cmd/cli/upload.go
```diff
@@ -61,6 +61,7 @@ func UploadCmd(ctx context.Context, sigRef, payloadRef, imageRef string) error {
 		b64SigBytes, err = ioutil.ReadAll(os.Stdin)
 	} else if _, err = os.Stat(sigRef); os.IsNotExist(err) {
 		b64SigBytes = []byte(sigRef)
+		err = nil
```
jonjohnsonjr, line 64: I think this is too tricky. I'd rather see this code cleaned up a bit to make it clearer what's going on.

dlorenc: oof yea I apologize for this code.

dlorenc:
@font - let me know if you want to pair for a bit tomorrow on cleaning this up. I'd also appreciate talking it through to make sure I even understand what I was trying to do :)

cmd/cli/upload.go
font: No worries!

font:
@jonjohnsonjr @dlorenc I've pushed an update that attempts to clean this up and added test cases for it. Let me know if you had something else in mind. Thanks @dlorenc - happy to discuss further tomorrow.

dlorenc (approved):
Oh wow, much nicer!

PR #52, opened 2021-03-05
Switch from ed25519 to ecdsa

### LINKED ISSUE #46: Switch from ed25519 to ecdsa?
We would use sha256 or 512, and P-384.

Basically until ed25519 gets into FIPS, most KMS systems and CAs aren't going to support ed25519 curves.

The keys are larger, but still smaller than RSA keys. For a sample:
```
-----BEGIN PUBLIC KEY-----
MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE0Gq6G8yJClUU8KhnzNs2Cer1C3Ro3yyQ
bNUkEnzOXzC11y3mjLdGXFjY8/RE22GtpTph8Z3nKH6WeNLxJbOGtjWdeAI22Vzu
gUEL+iz5J36ZI1fPu5lRmMaAU/u7qto7
-----END PUBLIC KEY-----
```

vs.
-----BEGIN PUBLIC KEY-----
MCowBQYDK2VwAyEAhJNkGxncSZ3R2ijORaJ9m1BHxzx5nJ2Q/GTLNfYt9UM=
-----END PUBLIC KEY-----

It's still manageable to squeeze one of these onto one line for easy CLI operations, but ugly.

priyawadhwa:
ed25519 won't work with most KMS systems for a while, so we need to use ecdsa until it will.

I chose the P521 elliptic curve because it seems like the most secure one, although I believe the tradeoff is larger keys.
It might be worth it to choose a different curve if the keys are too big & we want people to be able to pass in public keys via command line.

example public key:

```
-----BEGIN PUBLIC KEY-----
MIGbMBAGByqGSM49AgEGBSuBBAAjA4GGAAQBXqYiRrNiB8cScdM8yLCX9T9tfX3V
jVdFsUUzkSCvjzTScB7G0gmwlvtrVpbnnRJ9mp3AGZu4rgXJ3nrdlZ05fb8B/cOb
M/Od5EqXLia5hrc+DZUtRu/Qn+JfXdkbHy9GYwuh0j8KAapOIYcunFEddBqxtq+Q
9V9K5tkascZWeit9c4k=
-----END PUBLIC KEY-----
```

Fixes #46


dlorenc:
Ugh now I'm starting to waffle on this one. WDYT about keeping ed25519 where we can (we control the local keys), and using ecdsa only for remote KMS's? It gets us the best of both worlds without having to add new flags.

dlorenc:
Meh nvm, let's just get this in for now.

PR #63, opened 2021-03-10
verify: Iterate through signatures in parallel

priyawadhwa:
I noticed that once I was trying to verify an image that had been signed a bunch of times (~10 times is a good test), verify was taking 15 seconds on my machine & seemed to take around a second for each additional signature. My guess is the calls to `remote.Layer`/compression take some time.

Anyway, parallelizing going through the signatures reduced time from 15 seconds to around 5 seconds consistently for an image that had been signed 11 times on my machine.


pkg/cosign/fetch.go
```diff
@@ -66,31 +67,43 @@ func FetchSignatures(ref name.Reference) ([]SignedPayload, *v1.Descriptor, error
...
+				errs = append(errs, err)
+				return
+			}
+			signatures[i] = SignedPayload{
+				Payload:         payload,
+				Base64Signature: base64sig,
+			}
+		}(i, desc)
+	}
+	wg.Wait()
```
dlorenc, line 104: Would an "errgroup" help here: https://pkg.go.dev/golang.org/x/sync/errgroup

priyawadhwa: yes ty! switched to that

pkg/cosign/fetch.go
```diff
@@ -68,42 +70,39 @@ func FetchSignatures(ref name.Reference) ([]SignedPayload, *v1.Descriptor, error
...
-	var wg sync.WaitGroup
-	wg.Add(len(descriptors))
+
+	g, _ := errgroup.WithContext(context.Background())
+
 	for i, desc := range descriptors {
-		go func(i int, desc v1.Descriptor) {
-			defer wg.Done()
+		i, desc := i, desc
+		g.Go(func() error {
```
jonjohnsonjr, line 78: We might want to bound this to some reasonable number of concurrent in flight requests?

jonjohnsonjr (commented):
This implementation is bad (already, not in your PR) let's chat about why synchronously :)

jonjohnsonjr:
Sent https://github.com/sigstore/cosign/pull/66 for the other reasons this is slow.

pkg/cosign/fetch.go
priyawadhwa: good idea! limited to the # of CPUs

pkg/cosign/fetch.go
```diff
@@ -56,32 +60,45 @@ func FetchSignatures(ref name.Reference) ([]SignedPayload, *v1.Descriptor, error
...
 
-	signatures := []SignedPayload{}
-	for _, desc := range m.Layers {
-		base64sig, ok := desc.Annotations[sigkey]
-		if !ok {
-			continue
-		}
-		l, err := sigImg.LayerByDigest(desc.Digest)
-		if err != nil {
+	g, ctx := errgroup.WithContext(context.Background())
```
jonjohnsonjr, line 63: Not a problem with this PR, but we should be plumbing a context through the entire library so that we can cancel stuff, e.g. https://golang.org/pkg/os/signal/#NotifyContext

priyawadhwa: it was easy enough to plumb through so I just added it in

dlorenc (approved):

PR #66, opened 2021-03-10
Make verify faster

jonjohnsonjr:
Before, we were calling the token exchange several times for no reason.
This reuses the v1.Image from remote.Image for fetching layers, which
can reuse the token from the original exchange. For me, this caused
verify to go from 5s to 1s overall.

Signed-off-by: Jon Johnson <jonjohnson@google.com>


dlorenc:
Nice!

PR #81, opened 2021-03-12
add gosec to CI and resolve known issues

bobcallaway:


.github/workflows/tests.yaml
```diff
@@ -28,3 +28,9 @@ jobs:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: v1.29
           timeout: 5m
+      - name: Run Gosec Security Scanner
```
dlorenc, line 31: Maybe a nit - but I think there's actually a way to get this to run as part of the golangci-lint job above. Let me check...

dlorenc: Ugh, it's a little more work. I think we have to add a golangci-lint.yml config. I'll follow up with that.

PR #53, opened 2021-03-06
WIP fulcio integration!

dlorenc:

PR #83, opened 2021-03-12
Fulcio integration!

dlorenc:
This is a rebased/squashed version of https://github.com/sigstore/cosign/pull/53 - getting closer to ready for review.


KEYLESS.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+That's it! No keys!
+The rest of the flags (annotations, claims, tlog, etc.) should all work the same.
+
+## Overview
+
+This uses ephemeral keys and certificates, which are signed automatically by the `fulcio` root CA.
+Signatures are stored in the `rekor` transparency log, which automatically provides an attestation
+as to when the signature was created.
```
lukehinds, line 42: could add mention of the CTL as well now (albeit a nit)

cmd/cosign/cli/sign.go
```diff
@@ -132,11 +138,28 @@ func SignCmd(ctx context.Context, keyPath string,
...
+	case keyPath != "":
 		signature, publicKey, err = sign(ctx, get, keyPath, payload, pf)
-	}
-	if err != nil {
-		return errors.Wrap(err, "signing")
+		if err != nil {
+			return errors.Wrap(err, "signing payload")
+		}
+	default: // Keyless!
+		fmt.Fprintln(os.Stderr, "Generating ephemeral keys...")
```
lukehinds, line 147: do you have a  logging framework yet?

pkg/cosign/fulcio/fulcio.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	resp, err := fcli.Operations.SigningCert(params, bearerAuth)
+	if err != nil {
+		return "", err
+	}
+	return resp.Payload.Certificate, nil
+}
+
+// This is the root in the fulcio project.
+const rootPem = `-----BEGIN CERTIFICATE-----
```
lukehinds, line 97: I know this is a prototype,but it might be useful to stick the root cert into a file at some point, and pass it in via a CLI flag, that way anyone using their own certificate transparency project for development, can just pass in their own.

pkg/cosign/remote.go
```diff
@@ -34,11 +47,15 @@ func Upload(signature, payload []byte, dstTag name.Reference) error {
 		}
 	}
 
+	annotations := map[string]string{
+		sigkey: base64.StdEncoding.EncodeToString(signature),
```
lukehinds, line 51: maybe check for an err here?

lukehinds (approved):
I made some minor comments, but I am honestly good to see this merge and get folks using it and feeding back.

pkg/cosign/fulcio/fulcio.go
lukehinds: second thoughts, maybe not a flag (confusing for end users).

pkg/cosign/remote.go
dlorenc: Where? you can't error encoding to base64.

pkg/cosign/fulcio/fulcio.go
dlorenc: We need to make this configurable eventually. Opened https://github.com/sigstore/cosign/issues/87

cmd/cosign/cli/sign.go
dlorenc: No logging yet (it should all be debug when we add it). Right now this is all output directly intended for users. I usually think of 3 kinds of output in CLI tools:

- Results (parseable, to stdout)
- Messages (text, to stderr)
- Logging (debug, hidden by default, used to diagnose problems)

pkg/cosign/remote.go
lukehinds: ack, understood (still getting used to go idioms)

cmd/cosign/cli/sign.go
lukehinds: fair point! the user CLI is better without standard logging

pkg/cosign/verify.go
```diff
@@ -223,15 +266,39 @@ func Verify(ref name.Reference, co CheckOpts) ([]SignedPayload, error) {
...
 			}
-			if err := sp.VerifyTlog(rekorClient, pubKeyPem); err != nil {
+
+			// Find the uuid then the entry.
+			uuid, err := sp.VerifyTlog(rekorClient, pub)
+			if err != nil {
 				validationErrs = append(validationErrs, err.Error())
 				continue
 			}
+			// if we have a cert, we should check expirt
```
priyawadhwa, line 289: nit: expiry

test/e2e_test.go
```diff
@@ -118,6 +121,39 @@ func TestMultipleSignatures(t *testing.T) {
...
+	mustErr(cli.VerifyBlobCmd(ctx, pubKeyPath1, "", "badsig", blob), t)
+	mustErr(cli.VerifyBlobCmd(ctx, pubKeyPath2, "", "badsig", blob), t)
+
+	// Now sign the blob with one key
+	sig, err := cli.SignBlobCmd(ctx, privKeyPath1, bp, true, passFunc)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Now verify should work with that one, but not the other
+	must(cli.VerifyBlobCmd(ctx, pubKeyPath1, "", string(sig), bp), t)
```
priyawadhwa, line 154: this might also need

```
mustErr(cli.VerifyBlobCmd(ctx, pubKeyPath2, "", string(sig), bp), t)
```
here to make sure it only works with the first but not the second

test/e2e_test.go
```diff
@@ -118,6 +121,39 @@ func TestMultipleSignatures(t *testing.T) {
 	must(verify(pub2, imgName, true, nil), t)
 }
 
+func TestSignBlob(t *testing.T) {
+
+	var blob = "someblob"
+	td1 := t.TempDir()
+	td2 := t.TempDir()
+	defer func() {
```
priyawadhwa, line 129: nit: could replace this with

```
t.Cleanup(os.RemoveAll(td1))
t.Cleanup(os.RemoveAll(td2))
```

priyawadhwa (approved):
LGTM! Just a few nits, no need to address them if you wanna merge!

test/e2e_test.go
dlorenc: Close!

dlorenc:
Thanks!

PR #84, opened 2021-03-12
Use a .golangci.yml file for linter settings, and remove the gosec ci…

dlorenc:
… job.

This way it'll run automatically as part of golangci-lint, which is nice
for local checking.


dlorenc:
Follow up to #81

priyawadhwa (approved):

PR #89, opened 2021-03-15
return an error when both -key and -kms are used

### LINKED ISSUE #78: ux: using both -key and -kms should error out
it seems -key and -kms are mutually exclusive but they are not causing an
error in the CLI at the moment.

chrnorm:
Closes #78. I also added a test around the functionality.


priyawadhwa (approved):
Thank you for fixing this!

ahmetb:
I think "verify" had the same issue as well.

PR #90, opened 2021-03-15
CI: lock Go version and run unit tests

chrnorm:
Locks the Go version to 1.15.10 (the latest 1.15 release at the time of commit), which is consistent with `go.mod`.

Also runs go tests as part of the pipeline.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -15,6 +15,23 @@ jobs:
     steps:
       # Checks out a copy of your repository on the ubuntu-latest machine
       - uses: actions/checkout@v2
+
+      # https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds
+      - uses: actions/cache@v2
+        with:
+          # In order:
+          # * Module download cache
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -27,4 +53,3 @@ jobs:
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: v1.29
-          timeout: 5m
```
dlorenc, line 30: We needed this before for a bit, things were timing out. Any reason to remove it?

chrnorm: @dlorenc I assumed it was mistakenly added, as `timeout` isn't a valid argument for `golangci/golangci-lint-action@v2`. I've added [`timeout-minutes`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes) to the lint step, set to 5 minutes.

dlorenc: Ah, thanks!

PR #93, opened 2021-03-15
Add fuzzing test for getting password

priyawadhwa:
I tried to write a fuzzing target to be used by oss-fuzz -- can be tested by running this on linux:

```
go-fuzz-build -libfuzzer ./test/
clang -fsanitize=fuzzer reflect-fuzz.a -o cosign.libfuzzer
./cosign.libfuzzer
```


cmd/cosign/cli/generate_key_pair.go
```diff
@@ -105,3 +102,18 @@ func getPass(confirm bool) ([]byte, error) {
 	}
 	return pw1, nil
 }
+
+func setReadPasswordFn() {
+	if Read != nil {
+		return
+	}
+	// Handle piped in passwords.
+	Read = func() ([]byte, error) {
```
dlorenc, line 111: should this get defined up top? I don't love having this sit around as nil. We can still override it in here if it's not a terminal

dlorenc (approved):
One question

cmd/cosign/cli/generate_key_pair.go
priyawadhwa: yah that's better, fixed

dlorenc (approved):

PR #103, opened 2021-03-18
Prepare for the first release!

dlorenc:
This adds a github action that will run after merge and publish artifacts.
It also adds keys to the repo that we'll sign with.


.gitignore
```diff
@@ -15,8 +15,7 @@
 # vendor/
 
 # cosign stuff
-cosign.key
-cosign.pub
+cosign.sig
```
priyawadhwa, line 18: Maybe we keep `cosign.key` and `cosign.pub` here for users, and move the keys below into `.github/workflows` if they're needed for the release?

priyawadhwa (commented):
Just a couple comments, also`cmd/cosign/cosign.pub` is an empty file checked in so you probably need to delete that

.gitignore
dlorenc: Ah, good call.

PR #112, opened 2021-03-19
Separate out the root/chain from the cert we get.

dlorenc:
This is the last change to fix up the next version of the fulcio API.


font (approved):
I was hitting this and then saw this PR. :smile:

dlorenc:
> I was hitting this and then saw this PR. 😄

Yeah there were a few breaking changes that snuck into fulcio all at once. We should get a bit stricter about that once there's a cosign release.

PR #125, opened 2021-03-20
`cosign validate`: return an error when kms and key arguments are both provided

### LINKED ISSUE #78: ux: using both -key and -kms should error out

chrnorm:
My second pass at fixing this, closes #78. Thankyou to @priyawadhwa for catching that the issue wasn't fully resolved.

In order to add a test around the change I have refactored [`cli/verify.go`](https://github.com/chrnorm/cosign/blob/3f83da7da087513ad9d98bf35e6b551126cc0dff/cmd/cosign/cli/verify.go) to be closer to an [idiomatic ffcli example](https://github.com/peterbourgon/ff/tree/master/ffcli/examples/objectctl): I've added a `VerifyCommand` struct with an `Exec` receiver method which executes the actual command. 

Using a struct to hold configuration and writing to the struct using `flagset.StringVar` allows tests to be written around command execution (in this case, I've added a test checking the behaviour of when `-key` and `-kms` are both passed). See [`cli/verify_test.go`](https://github.com/chrnorm/cosign/blob/3f83da7da087513ad9d98bf35e6b551126cc0dff/cmd/cosign/cli/verify_test.go) for an example.

We may want to consider the pattern presented above for other commands added in future over using an anonymous function for `ffcli.Command.Exec`, as shown the below snippet

```go
return &ffcli.Command{
		Name:       "mycommand",
		ShortUsage: "mycommand run -value <value>",
		ShortHelp:  "An example command",
		FlagSet:    flagset,
		Exec: func(ctx context.Context, args []string) error {
                       // difficult to test any behaviour here
                }
```

Instead replacing this with
```go
type MyCommand struct {
	MyConfigValue string
}

func NewMyCommand() *ffcli.Command {
	cmd := MyCommand{}
	flagset := flag.NewFlagSet("mycommand run", flag.ExitOnError)

	flagset.StringVar(&cmd.MyConfigValue, "value", "", "example command configuration value")

	return &ffcli.Command{
		Name:       "mycommand",
		ShortUsage: "mycommand run -value <value>",
		ShortHelp:  "An example command",
		FlagSet:    flagset,
		Exec:       cmd.Exec,
	}
}

func (c *MyCommand) Exec(ctx context.Context, args []string) error {
	fmt.Println("example command execution")
	return nil
}
```

Additionally:
- split logging of verification results into a separate function
- update end-to-end test to more closely represent the real CLI command under test, by calling cmd.Exec (which includes parsing and validation logic)

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -29,89 +29,107 @@ import (
 	"github.com/sigstore/cosign/pkg/cosign/fulcio"
 )
 
+// VerifyCommand verifies a signature on a supplied container image
+type VerifyCommand struct {
+	KmsVal      string
+	Key         string
+	CheckClaims bool
+	Annotations map[string]string
+}
+
+// Verify builds and returns an ffcli command
 func Verify() *ffcli.Command {
-	var (
-		flagset     = flag.NewFlagSet("cosign verify", flag.ExitOnError)
-		kmsVal      = flagset.String("kms", "", "verify via a public key stored in a KMS")
-		key         = flagset.String("key", "", "path to the public key")
-		checkClaims = flagset.Bool("check-claims", true, "whether to check the claims found")
-		annotations = annotationsMap{}
-	)
+	cmd := VerifyCommand{}
+	flagset := flag.NewFlagSet("cosign verify", flag.ExitOnError)
+	annotations := annotationsMap{}
+
+	flagset.StringVar(&cmd.Key, "kms", "", "verify via a public key stored in a KMS")
+	flagset.StringVar(&cmd.KmsVal, "kms", "", "verify via a public key stored in a KMS")
+	flagset.BoolVar(&cmd.CheckClaims, "check-claims", true, "whether to check the claims found")
+
+	// parse annotations
 	flagset.Var(&annotations, "a", "extra key=value pairs to sign")
+	cmd.Annotations = annotations.annotations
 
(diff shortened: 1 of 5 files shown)


cmd/cosign/cli/verify.go
```diff
@@ -29,89 +29,107 @@ import (
...
-		kmsVal      = flagset.String("kms", "", "verify via a public key stored in a KMS")
-		key         = flagset.String("key", "", "path to the public key")
-		checkClaims = flagset.Bool("check-claims", true, "whether to check the claims found")
-		annotations = annotationsMap{}
-	)
+	cmd := VerifyCommand{}
+	flagset := flag.NewFlagSet("cosign verify", flag.ExitOnError)
+	annotations := annotationsMap{}
+
+	flagset.StringVar(&cmd.Key, "kms", "", "verify via a public key stored in a KMS")
```
dlorenc, line 46: nit: i think the help text on this one is incorrect/copy-pasted.

dlorenc (commented):
One nit on the help text, otherwise this looks great!

cmd/cosign/cli/verify.go
chrnorm: Oops - great pickup, thanks @dlorenc. Fixed now.

dlorenc (approved):
Thanks!

PR #132, opened 2021-03-21
Whoops, sign with the linux binary always in our CI builds.

dlorenc:


dlorenc:
An unfortunate casualty of not being able to access secrets during pull requests.

PR #135, opened 2021-03-22
Add functions to get the public-key from the private key.

dlorenc:


cmd/cosign/cli/public_key.go
```diff
@@ -0,0 +1,94 @@
...
+func PublicKey() *ffcli.Command {
+	var (
+		flagset = flag.NewFlagSet("cosign public-key", flag.ExitOnError)
+		key     = flagset.String("key", "", "path to the private key")
+		kmsVal  = flagset.String("kms", "", "sign via a private key stored in a KMS")
+	)
+
+	return &ffcli.Command{
+		Name:       "public-key",
+		ShortUsage: "cosign public key gets a public key from the key-pair [-kms KMSPATH]",
```
priyawadhwa, line 43: nit: "cosign public-key" instead of "cosign public key" might be easier to read

priyawadhwa (approved):
lgtm!

cmd/cosign/cli/public_key.go
dlorenc: Yup! I always have a fear that I typo'd these and wrote "pubic". It's bound to happen.

PR #145, opened 2021-03-23
Add a COSIGN_PASSWORD environment variable.

dlorenc:
It's pretty much impossible to pass in a PW through a Github action any other way.


priyawadhwa (approved):
😎

Issue #122, opened 2021-03-20
Add a "COSIGN_REPOSITORY" flag/env var


dlorenc:
This would act as a replacement for the image repository, in case you want to sign/store/verify from an alternate location.

We should use this before we sign distroless. The first round will be:

```
export COSIGN_REPOSITORY=gcr.io/distroless-signatures
```
Causing:

`gcr.io/distroless/static -> gcr.io/cosign-signatures/static:sha256-3b3507aa7ce16b085215915d6334f140d8f50ef35bbf3251ee52d3102296f60e.cosign`

PR #146, opened 2021-03-23
Add support for COSIGN_REPOSITORY env var

priyawadhwa:
This env var will allow users to specify which repo they want cosign signatures stored in!

I had to do some kinda sketchy string replacement to replace just the subrepo -- if we just directly replace the registry, we'd end up with something like this (which would be cleaner in the code, but isn't as pretty):

```
export COSIGN_REPOSITORY=gcr.io/distroless-signatures
gcr.io/cosign-signatures/static -> gcr.io/distroless/cosign-signatures/static
```

fixes https://github.com/sigstore/cosign/issues/122

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,18 @@ Multiple signatures are stored in a list which is unfortunately "racy" today.
 To add a signtaure, clients orchestrate a "read-append-write" operation, so the last write
 will win in the case of contention.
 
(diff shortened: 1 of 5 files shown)


dlorenc (approved):
nit: add this to the readme?

priyawadhwa:
good idea!

PR #167, opened 2021-03-24
pass fulcio.Roots properly

ahmetb:
previously the roots were hardcoded in TrustedCert() function as
fulcio.Roots (and the passed value was usuned), but I assume this is not
what was intended.


dlorenc (approved):

dlorenc:
I keep forgetting copasetic is in here :)

Issue #108, opened 2021-03-19
Yubikey support!


dlorenc:
We should be able to support storing keys in yubikeys! 

https://github.com/go-piv/piv-go
https://github.com/go-piv/go-ykpiv

dlorenc:
https://developers.yubico.com/python-fido2/

dlorenc:
We need to do a little bit of thinking on the design here before going too much further. I'll explain some background:

Yubikeys support 3 different levels of protection:

- a 6 digit PIN
- an 8 digit PUK
- a 24 byte management key

The defaults can be found here:
![image](https://user-images.githubusercontent.com/1714486/112653484-764e3f00-8e1c-11eb-9a63-a6a70501f80b.png)

There's a command line tool from yubikey to configure these, but it might be nice if we allow users to set it up directly in our tooling.

Here's what an ideal setup workflow might be:

```shell
$ cosign generate-key-pair --yubikey
Please enter the device management key, or press enter for the default:
Default management key selected. Would you like to change this? [y/N]:
Changing management key. Enter new key, up to 24 characters:
Confirm new management key:
Management key reset.

Would you like to set a key PIN? [y/N]: y
Please enter a 6 digit key PIN:
Confirm key pin:
Generating a private key on the device...

Key generated!
Public key written to cosign.pub
Device attestation certificate written to yubikey.crt
```

We could also allow configuring a PUK here, which is used to unlock the device if the pin is entered incorrectly too many times. It might be better to just bail out at that point and send them to the real yubikey tool though.

Same for resetting the entire thing.

dlorenc:
I want to get the balance right between "I just took a yubikey out of a box and want to start using it to sign things" with more advanced yubikey users/workflows where they might be reusing the same hardware for FIDO2/U2F etc., or they have specially provisioned corporate yubikeys that have already been setup with a fixed management key.

dlorenc:
Also, I'd hope most of this logic ends up in the http://github.com/sigstore/sigstore client/libraries so it can be used for all artifact types!

lukehinds:
I see what you mean.

On the one hand it's nice to be able to set this up users, that is a really good UX. On the other, if we get them to set up using the upstream tool, we want get any blowback if they can no longer log into their bank.

Perhaps a warning?

```
WARNING: Setting a pin may cause loss of current key configurations
Would you like to set a key PIN? [y/N]: y
Please enter a 6 digit key PIN:
Confirm key pin:
Generating a private key on the device...
[...]

PR #169, opened 2021-03-25
WIP Yubicosign

dlorenc:
It works on my box!

Ref #108

### DIFF
--- a/cmd/cosign/cli/generate_key_pair.go
+++ b/cmd/cosign/cli/generate_key_pair.go
@@ -16,6 +16,8 @@ package cli
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
@@ -38,11 +40,12 @@ func GenerateKeyPair() *ffcli.Command {
 	var (
 		flagset = flag.NewFlagSet("cosign generate-key-pair", flag.ExitOnError)
 		kmsVal  = flagset.String("kms", "", "create key pair in KMS service to use for signing")
+		ykVal   = flagset.Bool("yubikey", false, "create key pair in Yubikey")
 	)
 
 	return &ffcli.Command{
 		Name:       "generate-key-pair",
-		ShortUsage: "cosign generate-key-pair [-kms KMSPATH]",
+		ShortUsage: "cosign generate-key-pair [-kms KMSPATH], [-yubikey]",
 		ShortHelp:  "generate-key-pair generates a key-pair",
 		LongHelp: `generate-key-pair generates a key-pair for signing.
 
@@ -58,12 +61,12 @@ CAVEATS:
   the COSIGN_PASSWORD environment variable to provide one.`,
 		FlagSet: flagset,
 		Exec: func(ctx context.Context, args []string) error {
-			return GenerateKeyPairCmd(ctx, *kmsVal)
+			return GenerateKeyPairCmd(ctx, *kmsVal, *ykVal)
 		},
 	}
 }
 
-func GenerateKeyPairCmd(ctx context.Context, kmsVal string) error {
+func GenerateKeyPairCmd(ctx context.Context, kmsVal string, ykVal bool) error {
 	if kmsVal != "" {
 		k, err := kms.Get(ctx, kmsVal)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/cosign/cli/generate_key_pair.go
```diff
@@ -82,6 +88,36 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string) error {
...
+			return err
+		}
+		defer yk.Close()
+
+		pk, err := genYubikey(yk)
+		if err != nil {
+			return err
+		}
+
+		// cert, err := yk.Attest(piv.SlotSignature)
```
ahmetb, line 103: Leftover

cmd/cosign/cli/sign.go
```diff
@@ -141,6 +143,7 @@ func SignCmd(ctx context.Context, keyPath string,
 	var signature []byte
 	var publicKey *ecdsa.PublicKey
 	var cert, chain string
+	fmt.Println("KEYPATH:", keyPath)
```
ahmetb, line 146: Leftover?

cmd/cosign/cli/sign.go
```diff
@@ -155,6 +158,33 @@ func SignCmd(ctx context.Context, keyPath string,
 		if err != nil {
 			return errors.Wrap(err, "getting public key")
 		}
+	case keyPath == "yubikey":
+		fmt.Println("KEYPATH")
```
ahmetb, line 162: Leftover?

ahmetb (commented):
Not sure how these hardware work, but would this apply to public-key command too?

I assume the private key is never extracted from the device? If so, why do we implement generate-key-pair (and write to the file)? I don't think we do that in -kms?

Also, how does it apply to "verify" cmds? Would anyone actually verify with Yubikey?

cmd/cosign/cli/sign.go
dlorenc: Yup!

dlorenc:
> Not sure how these hardware work, but would this apply to public-key command too?
> 
Yes, added!
> I assume the private key is never extracted from the device? If so, why do we implement generate-key-pair (and write to the file)? I don't think we do that in -kms?
We have to create the key initially. I need to add some TODOs here to pin protect this and handle that other stuff (the admin pin, the pin for the touch).

> 
> Also, how does it apply to "verify" cmds? Would anyone actually verify with Yubikey?

Yeah, it works there too. You can verify either with the extracted pub or with the key itself.

cmd/cosign/cli/generate_key_pair.go
```diff
@@ -82,6 +85,31 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string) error {
 		}
 		fmt.Fprintln(os.Stderr, "Public key written to cosign.pub")
 		return nil
+	} else if ykVal {
+		yk, err := cosign.GetYubikey()
+		if err != nil {
+			return err
```
ahmetb, line 91: I know these are nice and simple but we already have github.com/pkg/errors imported, so it would be beneficial to wrap these with more meaningful messages (i.e. what we were trying to do). That sets the tone for the rest of the codebase.

ahmetb:
Does this work with Yubico gnubby we get at Google? I have one plugged  but getting `error: no yubikey found`. (I'll try to find my personal yubikey and try that too.)

image

cmd/cosign/cli/generate_key_pair.go
```diff
@@ -82,6 +85,31 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string) error {
...
+		}
+		pemBytes := pem.EncodeToMemory(&pem.Block{
+			Type:  "PUBLIC KEY",
+			Bytes: der,
+		})
+		if err := ioutil.WriteFile("cosign.pub", pemBytes, 0600); err != nil {
+			return err
+		}
+		fmt.Fprintln(os.Stderr, "Public key written to cosign.pub")
```
ahmetb, line 111: this goes against the cmd help message which says ("generates key pair") but it generates only a .pub here. not sure what to do with it.

copasetic/main.go
```diff
@@ -170,6 +170,7 @@ func main() {
 				return nil, err
 			}
 
+			yk
```
mnm678, line 173: Is this a typo?

ahmetb:
I tried again with these two keys and piv.Cards() still return nil.

image

cmd/cosign/cli/sign.go
```diff
@@ -155,6 +157,32 @@ func SignCmd(ctx context.Context, keyPath string,
...
+			return err
+		}
+		publicKey = pub.PublicKey.(*ecdsa.PublicKey)
+		auth := piv.KeyAuth{PIN: piv.DefaultPIN}
+		priv, err := yk.PrivateKey(piv.SlotSignature, publicKey, auth)
+		if err != nil {
+			return err
+		}
+		signer := priv.(crypto.Signer)
+		signer.Public()
```
ahmetb, line 178: seems redundant?

copasetic/main.go
dlorenc: doh, yes!

cmd/cosign/cli/generate_key_pair.go
dlorenc: It's a little subtle I guess. It does generate a key pair, but the only file it writes to disk is the public key. The private one is generated on the key and cannot be exported.

dlorenc: I improved the output here a bit.

There is still A LOT to be done here :)

cmd/cosign/cli/sign.go
dlorenc: Thanks!

dlorenc:
> Does this work with Yubico gnubby we get at Google? I have one plugged but getting `error: no yubikey found`. (I'll try to find my personal yubikey and try that too.)

I actually couldn't get that to work either. I think they might be firmware flashed, but my Yubikey was actually pretty old so I didn't try too hard.

I ordered a few of these: https://www.yubico.com/product/yubikey-5c/

Issue #40, opened 2021-03-03
Registry Support


dlorenc:
Our primary goal is broad registry support. Right now we're unsure of where we are:

- GCR (tested, mostly works)
- Google Artifact Registry (works)
- Quay - works in latest version, not in hosted service yet
- Dockerhub?
- Azure Container Registry?
- Amazon Container Registry?
- Others?

We have some options we can try to increase support, but they're kind of ugly. I'd first like to understand how much support we have vs. how much we would gain by doing terrible things with media types.

cc @jonjohnsonjr @font
(slack here: https://github.com/google/go-containerregistry/blob/93228a70849651ba98cdee6f0654f623d7cdcbdb/pkg/v1/manifest.go#L27)

jonjohnsonjr:
https://github.com/google/go-containerregistry/pull/954

dlorenc:
I just tested on DockerHub and things appear to work.

dlorenc:
Azure support here: https://github.com/sigstore/cosign/pull/61

fubarhouse:
Would be interested in any support for [Harbor](https://github.com/goharbor/harbor)

dlorenc:
@fubarhouse is there an easy for me to try it out? It probably will "just work", but I'd need to check to be sure.

fubarhouse:
@dlorenc I might give it a try later - would be great if it would already be supported. I don't really know of any public registries I could test on...

lukehinds:
> @dlorenc I might give it a try later - would be great if it would already be supported. I don't really know of any public registries I could test on...

should you do so, please open a separate issue so we can track, cheers

mchmarny:
Any technical reasons why cosign would not work with JFrog's container registry?

dlorenc:
> Any technical reasons why cosign would not work with JFrog's container registry?

Nope! The only registry we've tried that it hasn't worked on so far is Quay. It works on the newest versions, but not the one running in the hosted service so far.

dlorenc:
BTW, if anyone has access to these repos and can provide test credentials I'd love to get some CI setup...

Feel free to email me privately and we can figure out a way to handle the credentials.

zuBux:
hey! I just tested JFrog's Artifactory registry with cosign (signing and signature verification) and it works without any issues! Thought you'd like to know :)

rosstimson:
Just tested AWS Elastic Container Registry (ECR) and all seems fine there.

dlorenc:
[...]

PR #174, opened 2021-03-26
Add Gitlab Container Registry

eminks:
I have tested cosign for gitlab container registry and work! 

My verify result:
```
$ cosign verify -key cosign.pub registry.gitlab.com/eminks/denemler:latest
  - WARNING - THE CERTIFICATE EXPIRY WAS NOT CHECKED. set COSIGN_EXPERIMENTAL=1 to check!
{"Critical":{"Identity":{"docker-reference":""},"Image":{"Docker-manifest-digest":"sha256:c93fa5b332de47c9f031284c7aa7a95988536e515571b73fa6294a0f54b289aa"},"Type":"cosign container signature"},"Optional":null}
```
Based on: #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Today, `cosign` has been tested and works against the following registries:
 * Azure Container Registry
 * JFrog Artifactory Container Registry
 * The CNCF distribution/distribution Registry
+* Gitlab Container Registry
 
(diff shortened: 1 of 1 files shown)


dlorenc:
Thanks! This is great!

PR #180, opened 2021-03-27
Remove support for keys as base64 strings on the command line.

dlorenc:
This was nice when we used shorter ed25519 ones, it's not worth the complexity now with multi-line strings.


dlorenc:
I think I cleaned up all the docs here.

ahmetb (approved):

PR #185, opened 2021-03-28
Update README.md

danpopSD:
added details and instructions on installing crane.  

Signed-off-by: Dan POP <dan.papandrea@sysdig.com>


danpopSD:
one more time with feeling!

Issue #188, opened 2021-03-28
Idea: Base image validation


dlorenc:
Using the proposed annotations: https://github.com/opencontainers/image-spec/issues/821

We could identify the base image seam. We can't **trust** this annotation, but we can use it as a hint. We can do the following:

* Fetch the *actual* base image by combining the digest and location annotations. Basically, strip the tag from the `org.opencontainers.image.base.ref.name` and add the digest from `org.opencontainers.image.base.digest`. 
* Compare the layers from the base image with the layers from the image, pre-seam. They should match exactly.

From there, we can fetch signatures on the base image and verify those against keys/identities.

dlorenc:
@ImJasonH @jonjohnsonjr does my annotation-munging scheme make sense?

dlorenc:
We could do this in `cosign` and/or `copasetic`.

In copasetic it would look something like `oci.base_image_ref` and `oci.base_image_digest` just for the raw annotation values, and `oci.verified_base_image_ref` for fully-qualified ref of the base image (by digest), after verifying the layers.

In `cosign`, it could look something like `cosign verify-base` or `cosign verify --base`, and then everything else would be the same.

Dentrax:
This morning, @developer-guy sent me a link about the [OCI Base Image Annotations](https://github.com/ImJasonH/ImJasonH/tree/main/articles/oci-base-image-annotations). And just want to look around to see if there is an active discussion issue about this.

---

What would the cosign's behavior in case there are _multiple_ base images? Let's assume we have the following _minimal_ Dockerfile:

```dockerfile
# base image 1
FROM foo:v1 as foo
# base image 2
FROM bar:v1 as bar
# base image 3
FROM baz:v1 as baz

WORKDIR /src

# Create a group and user
RUN addgroup -S qux && adduser -S qux -G qux
RUN chown -R qux:qux /src

USER qux

COPY --chown=qux:qux --from=foo /usr/bin/foo /usr/bin/foo
COPY --chown=qux:qux --from=bar /usr/bin/bar /usr/bin/bar
COPY --chown=qux:qux --from=baz /usr/bin/baz /usr/bin/baz
```

Sorry, I'm a bit confused here, maybe I'm missing some point about the spec. Actually, is there one _base image_, which is `FROM foo:v1 as foo`?

jonjohnsonjr:
> What would the cosign's behavior in case there are multiple base images?

[...]

PR #189, opened 2021-03-28
WIP: Hypotenuse - a tool for extracting the base images out of a Dockerfile.

dlorenc:
I'm sending this one up for fun and discussion! This should not be merged as is :)

This introduces a small tool that can be used to parse a Dockerfile and extract the base images from inside it. The tool fully resolves those base image entries to a digest (hopefully the same one that would be used in the `docker build` that's about to follow. These references are then passed over stdout, one per line, where they can be consumed by other tools. Like `cosign`!

You can use this to verify the base images in a Dockerfile have all been signed before you do a build. This is related and complementary to #188.

See this gif!

The name `hypotenuse` is probably my worst pun yet.

hypotenuse

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM us-central1-docker.pkg.dev/dlorenc-vmtest2/test/golang:1.16.2 AS builder
+
+ADD go.mod go.sum $APP_ROOT/src/
+RUN go mod download
+
+# Add source code
+ADD ./ $APP_ROOT/src/
+
+RUN go build ./cmd/server
+
+# Multi-Stage production build
+FROM us-central1-docker.pkg.dev/dlorenc-vmtest2/test/distroless as deploy
+
+# Retrieve the binary from the previous stage
+COPY --from=builder /opt/app-root/src/server /usr/local/bin/server
+
+# Set the binary as the entrypoint of the container
+CMD ["server", "serve"]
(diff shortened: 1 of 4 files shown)


ahmetb:
Several different experiences come to mind:

1. `hypotenuse` command works as you illustrated, but we make `cosign verify` accept the input for IMAGE from stdin (one image reference per line)

2. we directly integrate this into `cosign verify`, e.g. `-f Dockerfile` since cosign is meant to be end-user friendly. if we want a print-only mode, we can do `-print-only` that just resolves image references. This way, user doesn't need to see a new `hypotenuse`  command.

dlorenc:
> 2\. we directly integrate this into `cosign verify`, e.g. `-f Dockerfile` since cosign is meant to be end-user friendly. if we want a print-only mode, we can do `-print-only` that just resolves image references. This way, user doesn't need to see a new `hypotenuse`  command.

Yup! @jonjohnsonjr had some other ideas on using this to rewrite (or pin the Dockerfiles), or intercept and pass back into docker build . -f <(...)

font:
Neat @dlorenc ! I was thinking of something like this, but was hoping it was possible to do this from an already built image. I didn't immediately find anything meaningful though.

dlorenc:
@font i have some ideas here with @ImJasonH. When do you need something by? I can hack something up pretty quickly as a POC.

dekkagaijin:
Implemented in https://github.com/sigstore/cosign/pull/395

Issue #127, opened 2021-03-20
Refactor into using crypto.Signer interface


dlorenc:
We have three key types now:

- Certs
- Fixed keys
- KMS

If we add Yubikey support (#108) we'll be up to 4. It's probably time to refactor these out into crypto.Signers.

ahmetb:
Agred. Right now 4 commands ({sign,verify}{,-blob}) accept the same set of key type flags (`-key`, `-kms`).

If we were to have more types e.g. 

- `-google-auth`, `-github-auth`, `-microsoft-auth` (or more generally `-oidc google`)
- `-yubikey` (or `-fido2`?)
- `-hsm`

these flags will appear on all these commands, crowding the space quite a bit. But I guess that's acceptable (and we might be able to group these options together as "KEY TYPES").

Although, I'm inclined to remove `-key STRING_LITERAL` option (ref #71).

dlorenc:
+1, drop the key literal.

dlorenc:
cc @ahmetb - you're working on this right?

ahmetb:
Yeah. Got distracted with pre-refactoring code reading yesterday.

dekkagaijin:
Oof, didn't realize you'd already started working in it :x

dlorenc:
@dekkagaijin I think this is done and can be closed now right?

dekkagaijin:
yup!

PR #194, opened 2021-03-28
Initial creation of Signer and Verifier interfaces.

dekkagaijin:
We should be abstracting Signer and Verifier (and Public Key provider) sooner rather than later. This gets us a step in that direction, but there's a lot more retooling to do on the interfaces

### DIFF
--- a/cmd/cosign/cli/sign_blob.go
+++ b/cmd/cosign/cli/sign_blob.go
@@ -96,7 +96,6 @@ func SignBlobCmd(ctx context.Context, keyPath, kmsVal, payloadPath string, b64 b
 			return nil, errors.Wrap(err, "signing blob")
 		}
 		pemBytes = cosign.KeyToPem(pub)
-
 	case kmsVal != "":
 		k, err := kms.Get(ctx, kmsVal)
 		if err != nil {
@@ -145,4 +144,17 @@ func SignBlobCmd(ctx context.Context, keyPath, kmsVal, payloadPath string, b64 b
 		}
 	}
 	return signature, nil
+}
+
+func sign(ctx context.Context, keyPath string, payload []byte, pf cosign.PassFunc) (signature []byte, publicKey *ecdsa.PublicKey, err error) {
+	k, err := loadKey(keyPath, pf)
+	if err != nil {
+		return nil, nil, err
+	}
+	publicKey = &k.Key.PublicKey
+	signature, err = k.Sign(ctx, payload)
+	if err != nil {
+		return nil, nil, err
+	}
+	return signature, publicKey, nil
 }
--- a/pkg/cosign/payload.go
+++ b/pkg/cosign/payload.go
@@ -20,20 +20,22 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
-func Payload(img v1.Descriptor, a map[string]string) ([]byte, error) {
+type ImagePayload struct {
+	Img         v1.Descriptor
+	Annotations map[string]string
+}
+
+func (p *ImagePayload) MarshalJSON() ([]byte, error) {
 	simpleSigning := SimpleSigning{
 		Critical: Critical{
 			Image: Image{
-				DockerManifestDigest: img.Digest.String(),
+				DockerManifestDigest: p.Img.Digest.String(),
 			},
 			Type: "cosign container signature",
 		},
-		Optional: a,
+		Optional: p.Annotations,
 	}
-
-	b, err := json.Marshal(simpleSigning)
-	if err != nil {
-		return nil, err
-	}
-	return b, err
+	return json.Marshal(simpleSigning)
 }
+
+//TODO: Unmarshal JSON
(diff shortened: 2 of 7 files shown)


dlorenc:
Ref #127 and cc @ahmetb

ahmetb:
Y'all are working on the weekends! I had a half baked patch that used crypto.Signer, but I also found out that we can simplify a great deal of the logic and use common key/pub marshaling code by doing this.

dekkagaijin:
@ahmetb When else to try and restructure core libraries 😅

I'm planning on having Signer, Verifier, and PublicKey...er interfaces that know how to interpret with their own formats and allow for `cosign` to provide the glue

dlorenc:
> Y'all are working on the weekends!

Someone should tell Jake's manager :)

pkg/cosign/sign.go
```diff
@@ -75,3 +79,38 @@ type Identity struct {
 type Image struct {
 	DockerManifestDigest string `json:"Docker-manifest-digest"`
 }
+
+type Signer interface {
```
dlorenc, line 83: Just so I understand - what other types of signer might there be? Is it per algorithm (ecdsa vs. rsa, etc.) or per "strategy" (key from disk, KMS, ephemeral, etc.)?

I mostly ask because I'm hoping we can avoid ever having to support multiple "algorithms" at the same time for signing, but could definitely see allowing for multiple types during verification.

dlorenc:
Nice, looks pretty good! A few small nits/questions.

pkg/cosign/sign.go
dekkagaijin: > Is it per algorithm (ecdsa vs. rsa, etc.) or per "strategy" (key from disk, KMS, ephemeral, etc.)?

I was thinking both, but I'm also happy to limit things to ECDSA+SHA256 for now. Until generics are a thing, and possibly even after then, it's hard to have a middle ground between "this thing" and "anything", which is presumably why the core crypto libs just gave up.

dekkagaijin: I'm not sure the core crypto portion would be the right level to enforce opinionated-ness as to the forms of the signed payloads. I think we can just copy the web in this respect-- declare what crypto algos we accept but let the payload itself be a black box more-or-less

dekkagaijin: We can consider fulcio and cosign somewhat independent, but we'll also want to nail down exactly what (if anything) fulcio should prescribe, as well, since that'll guide what we do for cosign

dlorenc: Yup, agree everywhere!

pkg/cosign/verify.go
```diff
@@ -83,20 +83,31 @@ func LoadPublicKey(keyRef string) (*ecdsa.PublicKey, error) {
...
-	}
+type Verifier interface {
+	Verify(ctx context.Context, payload, signature []byte) error
+}
+
+type ECDSAVerifier struct {
+	PubKey *ecdsa.PublicKey
+}
 
+func (v *ECDSAVerifier) Verify(_ context.Context, payload, signature []byte) error {
```
dlorenc, line 94: maybe just drop the unused context here since we dont even pass it below?

pkg/cosign/payload.go
```diff
@@ -20,20 +20,22 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
-func Payload(img v1.Descriptor, a map[string]string) ([]byte, error) {
+type ContainerImagePayload struct {
```
dlorenc, line 23: Maybe drop Container? We call it Image below.

cmd/cosign/cli/sign_blob.go
```diff
@@ -146,3 +145,16 @@ func SignBlobCmd(ctx context.Context, keyPath, kmsVal, payloadPath string, b64 b
 	}
 	return signature, nil
 }
+
+func sign(ctx context.Context, keyPath string, payload []byte, pf cosign.PassFunc) (signature []byte, publicKey *ecdsa.PublicKey, err error) {
+	k, err := loadKey(keyPath, pf)
+	if err != nil {
+		return
```
dlorenc, line 152: I won't complain tooooo hard but I personally don't love the naked returns. @priyawadhwa likes them too so I'm outnumbered.

pkg/cosign/payload.go
dekkagaijin: done

cmd/cosign/cli/sign_blob.go
dekkagaijin: clothed the returns

pkg/cosign/verify.go
dekkagaijin: `syntax error: mixed named and unnamed function parameters`

cmd/cosign/cli/sign_blob.go
priyawadhwa: honestly I don't love them either I think I did that back when I was returning like four things & thought it might be easier for someone new to understand

Issue #152, opened 2021-03-23
cmd: public-key writing to a particular file


ahmetb:
```
$ cosign public-key -kms $KMS
Public key written to cosign.pub
```

I think this experience is problematic because
it always writes to $CWD and with a fixed file name.
That'll make scripting harder. I think we should
write to stdout.

Happy to refactor.

dlorenc:
Yes, great point

PR #202, opened 2021-03-30
cmd: add option to specify public key output file

asraa:

Adds a flag for specifying the output file for the public key when using `cosign public-key`:
`cosign public-key -key <key> -outfile somefile.pub`
By default, it writes to standard out now.

Example:
```
$ ./cosign public-key -key cosign.key -outfile a.txt
Public key written to  a.txt
$ cat a.txt
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAETvioBh+xbZH4IMlltKuIGuVEYjqF
HdA7wGAHvdkB12TSULi6v/0PTdRU8Vap94/vspnNcD1SY9NkXavR5lt+Mw==
-----END PUBLIC KEY-----
$ ./cosign public-key -key cosign.key 
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAETvioBh+xbZH4IMlltKuIGuVEYjqF
HdA7wGAHvdkB12TSULi6v/0PTdRU8Vap94/vspnNcD1SY9NkXavR5lt+Mw==
-----END PUBLIC KEY-----
```

Docs: Added usage explanation in the docstring.
Tests: Added e2e test and unit tests.

Fixes https://github.com/sigstore/cosign/issues/152

------------------------------------------------------------------
Maybe the option should also exist to specify placement for `generate-key-pair`? But maybe will wait until an ask.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -215,8 +215,6 @@ $ cosign download us-central1-docker.pkg.dev/dlorenc-vmtest2/test/taskrun
 KMS:
 ```
 $ cosign public-key -kms gcpkms://projects/someproject/locations/us-central1/keyRings/foo/cryptoKeys/bug
-Public key written to cosign.pub
-$ cat cosign.pub
 -----BEGIN PUBLIC KEY-----
 MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEgrKKtyws86/APoULh/zXk4LONqII
 AcxvLtLEgRjRI4TKnMAXtIGp8K4X4CTWPEXMqSYZZUa2I1YvHyLLY2bEzA==
@@ -227,8 +225,6 @@ Private Key:
 ```
 $ ./cosign public-key -key cosign.key
 Enter password for private key:
-Public key written to cosign.pub
-$ cat cosign.pub
 -----BEGIN PUBLIC KEY-----
 MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEjCxhhvb1KmIfe1J2ceT25kHepstb
 IDYuTA0U1ri4F0CXXazLiftzGlyfse1No4orr8w1ZIchQ8TJlyCSaSuR0Q==
--- a/cmd/cosign/cli/public_key.go
+++ b/cmd/cosign/cli/public_key.go
@@ -19,6 +19,7 @@ import (
 	"crypto/ecdsa"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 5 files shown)


cmd/cosign/cli/public_key.go
```diff
@@ -54,12 +61,34 @@ EXAMPLES
...
+			var reader io.Reader
+			if *key != "" {
+				cl := filepath.Clean(*key)
+				if _, err := os.Stat(cl); os.IsNotExist(err) {
+					return fmt.Errorf("missing or invalid key path: %s", cl)
+				} else if err != nil {
+					return err
+				}
+				reader, err := os.Open(cl)
+				if err != nil {
```
asraa, line 74: it seems like this would already cover the missing or invalid path error above:
https://golang.org/pkg/os/

ahmetb:
Why don't we print to stdout? Users can always > FILE.

dlorenc:
> Why don't we print to stdout? Users can always > FILE.

That would be fine on this command but wouldnt work for generate-key-pair since we have two output files. What about stdout if there's no value set on the flag (drop the cosign.pub default), but support the same flag for consistency with generate-key-pair?

ahmetb:
Yeah I am not sure if we need that symmetry between the two.

We can make generate command ask for a base name (just like ssh-keygen command) and it would also have an -outfile/-f that defaults to "$PWD/cosign"?

cmd/cosign/cli/public_key.go
```diff
@@ -54,12 +61,39 @@ EXAMPLES
...
+			// Open output file for public key if specified.
+			if *outFile != "" {
+				writer.Name = *outFile
+				var err error
+				f, err = os.OpenFile(*outFile, os.O_WRONLY|os.O_CREATE, 0600)
+				if err != nil {
+					return err
+				}
+				writer.Writer = f
+				defer f.Close()
```
asraa, line 86: I read online that you probably shouldn't defer close on write-able files. Can fix if that's the case.

dlorenc:
> All set with public-key defaults to standard out, optionally can specify an outfile. Mind if I make a separate follow-up to generate-key-pair?

Ah of course! Didn't mean to imply it had to be in here.

cmd/cosign/cli/public_key.go
```diff
@@ -54,12 +61,39 @@ EXAMPLES
 			if (*key == "" && *kmsVal == "") || (*key != "" && *kmsVal != "") {
 				return &KeyParseError{}
 			}
-			return GetPublicKey(ctx, *key, *kmsVal, GetPass)
+			// Get private key file.
+			var reader io.Reader = (*os.File)(nil)
```
dlorenc, line 65: I think this can be simplified to `var reader io.Reader`

USAGE.md
```diff
@@ -225,9 +225,8 @@ AcxvLtLEgRjRI4TKnMAXtIGp8K4X4CTWPEXMqSYZZUa2I1YvHyLLY2bEzA==
 
 Private Key:
 ```
-$ ./cosign public-key -key cosign.key
+$ ./cosign public-key -key cosign.key -outfile cosign.pub
```
dlorenc, line 228: heh, you could also drop this and then the cat right below (same as above)

dlorenc (commented):
Two last nits I think!

dlorenc:
I like it!

dlorenc:
Do you want to rebase or should I click the button?

dlorenc (approved):

PR #213, opened 2021-04-03
Yubikey management CLI

dlorenc:
This adds a new set of subcommands for working with PIV tokens. They currently live under: `cosign piv-tool`

Right now we have:

`cosign piv-tool set-management-key`
`cosign piv-tool set-puk`
`cosign piv-tool set-pin`
`cosign piv-tool unblock`
`cosign piv-tool generate-key`
`cosign piv-tool attestation`

They seem to work pretty well! Before merging, I want to do a bit more work on the UX itself, including:

- [ ] Better prompting to avoid needing to pass the PINs, PUKs and management keys on the command line when possible
- [ ] The management key is 24 bytes long, we should probably allow people to pass in a base64 encoded string, or pipe in /dev/random or something
- [x] We might want a mode to generate a random management key in the tool rather than using a user-provided one (and then optionally display it or throw it away!)
- [x] We might want a mode to generate a random management key in the tool rather than using a user-provided one (and then optionally display it or throw it away!)
- [x] If we throw the key away from above - we should probably have an all-in-one command to set the key to a random value, use that to generate a signing key and then throw it away
- [x] I think we still need a full "reset" command

Then a "nice to have" would be a way to disable OTP on the keys :) They get angry whenacciccccccrikcvulenljchtvbbetejhevulebrvbrrkrkevdentally pressed.

Also, tests and stuff probably.


dlorenc:
Removing the WIP! This is good to go.

Wee can track the remaining TODOs (including actual signing support) in #108

test/piv_test.go
```diff
@@ -0,0 +1,129 @@
...
+// This requires the "resetyubikey" tag to be passed to go test.
+
+package test
+
+import (
+	"context"
+	"testing"
+
+	// Import the functions directly for testing.
+	. "github.com/sigstore/cosign/cmd/cosign/cli/pivcli"
```
priyawadhwa, line 30: cool i haven't seen this before!

test/piv_test.go
```diff
@@ -0,0 +1,129 @@
...
+	// Now this should fail
+	mustErr(GenerateKeyCmd(ctx, "", false), t)
+	// Unless we use the right key
+	must(GenerateKeyCmd(ctx, "mynewkey", false), t)
+
+	// Now if we use a random key it should set a new one
+	must(GenerateKeyCmd(ctx, "mynewkey", true), t)
+	// The old one shouldn't work.
+	mustErr(GenerateKeyCmd(ctx, "mynewkey", false), t)
+}
```
priyawadhwa, line 129: would it be useful to have a `TestAttestation` which basically tests the flow described in `TOKENS.md`?

priyawadhwa (approved):
LGTM!! 🗝️ 💯

test/piv_test.go
dlorenc: ooh good call. I didn't want to add a test without having any way to check it against something, but that's perfect!

dlorenc: BOOM

priyawadhwa: WOOT

Issue #212, opened 2021-04-02
Document signature format and specifications


dlorenc:
We should more formally document the signature storage format, location and specification for the signature format itself.

PR #214, opened 2021-04-03
Signature specification

dlorenc:
This mostly reflects the reality today. Some things I'd like to do:

- [ ] Comply with the simple signing format a little bit closer to reduce our extra semantics. Maybe reuse "atomic container signature"?
- [ ] Register mediaTypes with `OCI artifacts` and IANA?
- [ ] Better naming for `certificate` and `chain`
- [x] We don't really have a spot for hash algorithm today. Do we need one?
- [ ] Drop the .cosign suffix and replace with .sig?


dlorenc:
Ref #212

dlorenc:
cc @jonjohnsonjr please help me

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Signature Schemes
+
+Implementations must support at least the following schemes:
+
+* ECDSA-P256 with the SHA256 Hash Algorithm
+
+No information about the signature scheme is included in the object.
+Clients must determine the hash algorithm and signature scheme out-of-band during the verification process.
+This is an intentional decision to reduce the risk of [algorithm-confusion attacks](https://news.ycombinator.com/item?id=24346317).
```
OR13, line 239: I don't agree with this decision...

The cited issue is `alg=none`... which is a real JOSE issue... but that does not make JOSE by itself bad, just many, many implementations that blindly rely on it.

Above you say "ECDSA-P256 with the SHA256 Hash Algorithm" is required to be supported... why not say the signature formal will be Detached JWS with Unencoded payload and one of the following alg: "ES256, EdDsa", etc...

There are a lot of different signature standards, the only thing worse is no way to tell which one is used, when someone hands you a signed payload and no context.

consider the more popular signature suite specs, you might base this spec on:

https://tools.ietf.org/html/rfc4880#section-5.2.3

https://tools.ietf.org/html/rfc7515#section-4.1.1

^ notorious...

> The JWS Signature value is not
>  valid if the "alg" value does not represent a supported algorithm or
>  if there is not a key for use with that algorithm associated with the
>  party that digitally signed or MACed the content.

somehow ^ from this sentence, we got to "just turn off security with the `alg:none`"...

The commenters on the thread are correct.... if you are building a security library, don't give your callers a foot gun.... 

IMO, the solution is: 

- throw if algorithm is not defined
- throw if algorithm is not from allow list 
- update the allow list regularly based on security best practices ( https://safecurves.cr.yp.to/, etc)

This way you can know what software was signed with RSA 512 / MD5.... and not need to negotiate with someone out of band to learn that you should not trust that package.

We should assume that software will be published with crypto suites that are later broken, and make it easy to warn users.

dlorenc: Thanks for the feedback! I think there are two parts to it, and we actually agree on the important parts. This spec was written to describe how things are today, not necessarily how they should be. I want to fix some of it before merging.

For the first part - we only support ecdsa-256 with sha256 right now. We should support more, especially for verification. I'm still leaning towards this implementation only supporting one type of key for signing, but we should be able to verify signatures of other types.

I'm not sure about this part though:

> There are a lot of different signature standards, the only thing worse is no way to tell which one is used, when someone hands you a signed payload and no context.

The signature is definitely meaningless if all you have is the signed payload and signature. You need a public key to verify against. I think what I'm trying to say is that we should include the information on the signing algorithm used with the public key, not the payload. 

We have a few components we need overall for the signature to be meaningful:

- Payload bytes
- Signature bytes
- Hash algorithm (sha256, etc.)
- Signature algorithm (ecdsa-p256, etc.)
- Public Key Bytes

In the current implementation, we have the following breakdown:

- Payload bytes (stored in registry, next to target)
- Signature bytes (stored in registry, next to target)
- Hash algorithm (sha256, etc.) (hardcoded to sha256)
- Signature algorithm (ecdsa-p256, etc.) (stored in PKIX public key, delivered out of band)
- Public Key Bytes (stored in PKIX public key, delivered out of band)

I agree the JOSE spec isn't bad, but I don't think it actually buys us anything. We don't need anything to be URL safe, and we don't need to concatenate everything into one token. It also would require us to duplicate the algorithm info (outside of the hash function), which is my real concern. If you have a public key you want to verify things against, you already know the algorithm.

So I guess the main question is - do we need to support multiple hash algorithms? If we do, let's figure out a place to include that info.

OR13: While its not common, there are scenarios where having the public key is not enough,  for example: secp256k1 with ECDSA or Schnorr, or bls12381 with blake2b or shake256... Or Ed25519 to Curve25519 conversions, etc...

Most signatures schemes assume a hash algorithm, but it's also pretty common to hash with something different before signing.

My recommendation would be provide a data structure that allows you to specify everything needed to verify a signature without guessing.

Building on an existing signature suite and serialization scheme obviously helps make the software easier to implement in other languages with off the shelf tooling, in an interoperable way.

I would recommend building on JOSE, COSE, or PGP, and not inventing a new signature format.

But if you are inventing a new signature suite, it should cover the algorithmic agility, and assume more than one signature, hash and public key representation.

I think test vectors and a registry would be essential for this, something like:

```
0 ES256 (ECDSA with P256 and SHA256) on sha256(raw binary)
1 ES384 (ECDSA with P384 and SHA256) on sha256(raw binary)
```
^ saddly these will not produce stable test vectors due to non deterministic signatures.

You may also want to cite FIPS / NIST directly, since you can't rely on IANA JOSE  / COSE registries:

- https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf
- https://www.iana.org/assignments/jose/jose.xhtml
- https://www.iana.org/assignments/cose/cose.xhtml

dlorenc: To throw another signature spec into the mix: https://github.com/secure-systems-lab/signing-spec/blob/master/protocol.md

I think that one is closest philosophically to what I want to do here. It has support for "hinting" at the key id and algorithm alongside the payload, but it's very clear that this information can't be trusted and users are expected to obtain this information out of band.

dlorenc: The only real difference between the current cosign implementation and the SSL spec is whether or not the payload type is "protected". Cosign uses the mediaType element to store the payload type (simple signing), and this media type is not protected by a signature.

If we used the SSL spec, we would instead set a media type corresponding to the SSL spec. Then the signed component would be:

` PAE(UTF8(SimpleSigning),$JSON)`

I'm not sure I understand the attack/threat model on putting this payload type into the signature.

OR13: There are political considerations to promoting PAE, PGP, or really anything but JOSE / COSE... eventually the big guys will start attacking you since they are kinda locked to JOSE / COSE... 

I'd recommend sticking to JSON / JOSE... we should avoid innovation at the signature suite layer, and should focus on the input to the signature and the anchoring of the signature... I might go as far as saying the tool should support BOTH PGP and JOSE formatted signatures and keys.

meet developers where they are, with the formats they are familiar with.

OR13: @dlorenc here is a really young spec, from the IoT space... https://www.ietf.org/archive/id/draft-devault-bare-01.txt

Maybe its worth considering as well.

dekkagaijin:
> We don't really have a spot for hash algorithm today. Do we need one?

FWIW this'd probably be good, if only as a signal for implementers. e.g. SHA-512(-384, 512/256) is not only more secure but actually faster than SHA-256 on 64-bit architectures and has more-or-less universal support at this point.

dekkagaijin:
IMHO it's also better for both security and supportability to be opinionated about algorithms- lower barriers for entry, easier maintainability, and less/no possibility of downgrade attacks. If a "practical" attack surfaces, then just increment the spec and adopt whatever is best practice at the time.

dlorenc:
OK, I think I figured out the hash algorithm bit - we can lean on the registries!

By storing the raw payload as a `blob`, the registries already perform hashing to enter the blob into their CAS. The blobs are referenced by their `sha256:$digest` from the OCI manifest already, which also happens to be the algorithm we use. Registries also store the size of this object, which is what we use to reference it via descriptor. So by leaning on the registry spec we get server and client side validation of the digest/size of the payload already.

If we specify that signatures of registry objects should use the same algorithm the registry uses for references, we eliminate another moving part and get improved validation because registries and clients already understand that field.

dlorenc:
Just to make that last comment clear, we have the following hierarchy of objects:

- The actual "target" we are signing is any object in an OCI registry
- The "payload" we sign, which includes a reference to that object by digest (sha256:) is a Red Hat Simple Signing object
- All "signatures" for the "target" object are stored in a single, detached OCI object, referenced via naming convention
- Each individual "signature" is embedded into that OCI object as a "layer", formatted as a "descriptor".
  - The "payload" that we sign is stored raw as "blob" in the registry
  - The "payload" is then hashed using the standard registry algorithms (today sha256) and embedded back into the "layer descriptor" by that digest (and importantly, size)
- The hash algorithm used for the signature should match that of the registry (today sha256 but this could change in the future)
- The rest of the signature scheme is embedded in the PKIX public key (ecdsa, rsa, ed25519, etc)

dekkagaijin:
> The hash algorithm used for the signature should match that of the registry (today sha256 but this could change in the future)

I like it, makes sense

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+  Because signatures are __detached__, the payload MUST contain the digest of the image it references, in a well-known location.
+  This location is dependent on the payload format - see that section below for more details.
+
+* payload `mediaType` string
+  This REQUIRED property contains the media type of the payload.
+
+* `signature` string
+  This REQUIRED property contains the base64-encoded signature.
+  This signature MUST be generated by a supported scheme.
+  For more details on supported schemes, see the `Signature` section below.
```
ImJasonH, line 30: Link Signature?

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+  
+  Example `signature`:
+    `MEYCIQDXmXWj59naoPFlLnCADIPLKgLG3LyFtKrbjpnkYiGNGgIhAJ/eNx5zr/l1MJKSFpFMjPKKr4fjh5RHEtT2DhMamZuT`
+
+* `certificate` string
+  This OPTIONAL property contains a PEM-encoded, DER-formatted, ASN.1 x509 certificate.
```
ImJasonH, line 36: Link PEM, DER, ASN.1?

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+KoZIzj0EAwMDaAAwZQIwC15Gtd9F6W9lmJuoXMym9DfWlBpK5HEPak38WPXqowRp
+6p+2/3jSLkFT5Nn5fuISAjEAouVlX4zH2rlkfg45HnDJax7o6ZV+E0/6BdAms44D
+Ej6T/GLK6XJSB28haSPRWB7k
+-----END CERTIFICATE-----
+```
+
+* `chain` string
+  This OPTIONAL property contains a PEM-encoded, DER-formatted, ASN.1 x509 certificate chain.
+  The `certificate` property MUST be present if this property is present.
+  This chian MAY be used by implementations to verify the `certificate` property.
```
ImJasonH, line 64: typo: `chain`

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+1. Start with `gcr.io/dlorenc-vmtest2/demo:latest`
+2. Resolve this to a digest: `sha256:97fc222cee7991b5b061d4d4afdb5f3428fcb0c9054e1690313786befa1e4e36`
+3. Follow the encoding rules: `sha256-97fc222cee7991b5b061d4d4afdb5f3428fcb0c9054e1690313786befa1e4e36.cosign`
+
+Implementations MAY store signatures objects in the same OCI repository as the target image or a different one.
+
+### Object Types
+
+This section describes the way the properties from above are embedded into OCI objects that can be stored in a registry.
+Implementations MUST support storing signatures inat least the following object types:
```
ImJasonH, line 137: typo: "in at"

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+
+This section describes the way the mandatory and optional signature properties are embedded into an Image Manifest V2 Schema 2 object.
+
+Only one image manifest is created for every signed object.
+Multiple signatures can be embedded in one image manifest.
+
+##### Payload and mediaType
+
+The `payload` bytes are uploaded to an OCI registry as a `blob`, and are referenced by `digest`, `size` and `mediaType.`
+The digest is embedded into the `Image` manifest as a `layer`, via a `Descriptor`.
```
ImJasonH, line 151: Link "descriptor"?

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+The `mediaType` property for the `payload` is included in the same descriptor.
+
+Example `payload`:
+
+```
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+  "config": {
+    <ommitted for brevity>
```
ImJasonH, line 162: typo: omitted

SPEC.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Simple Signing
+
+The Simple Signing payload format is [specified here](https://github.com/containers/image/blob/a5061e5a5f00333ea3a92e7103effd11c6e2f51d/docs/containers-signature.5.md#json-data-format)
+
+The following additional semantics are applied:
+
+* The `mediaType` used to identify this payload format is: `application/vnd.dev.cosign.simplesigning.v1+json`.
+* The `critical.type` value used to identify `cosign` signatures is: `cosign container signature`.
+* The `critical.identity.docker-reference` field is not supported.
```
ImJasonH, line 228: What does it mean not to be supported? A spec can implementations MUST reject, or MAY ignore, or SHOULD discourage payloads that contain this information, but I'm not sure what it means to "not support" it.

SPEC.md
dlorenc: Good call.

SPEC.md
dlorenc: Sorry yeah this is basically me not knowing what we should do. Clarified a bit and added it to the top here.

SPEC.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Simple Signing
+
+The Simple Signing payload format is [specified here](https://github.com/containers/image/blob/a5061e5a5f00333ea3a92e7103effd11c6e2f51d/docs/containers-signature.5.md#json-data-format)
+
+The following additional semantics are applied:
+
+* The `mediaType` used to identify this payload format is: `application/vnd.dev.cosign.simplesigning.v1+json`.
+* The `critical.type` value used to identify `cosign` signatures is: `cosign container signature`.
+* The `critical.identity.docker-reference` field is not ignored.
```
mtrmac, line 231: Did you mean
```suggestion
* The `critical.identity.docker-reference` field is ignored.
```
?

SPEC.md
```diff
@@ -0,0 +1,342 @@
...
+* Optional user-specified claims may be included in the `Optional` section.
+
+## Signature Schemes
+
+Implementations must support at least the following schemes:
+
+* ECDSA-P256
+
+No information about the signature scheme is included in the object.
+Clients must determine the signature scheme out-of-band during the verification process.
```
mtrmac, line 241: Uh, how are they going to _actually do that_?

(In a single public key root of trust, I agree that supported schemes/algorithms are a property of that root of trust along with the raw public key, and should be a part of the same root-of-trust configuration set up by users. But with certificate chains that gets more difficult; x.509 certificates AFAICS include a public key signature algorithm, but not really a specific signature form/encoding scheme.)

I don’t have a better proposal, just noting that this is difficult.

Essentially this hinges on the hash algorithm choice; given that, things like the OpenPGP design include the algorithms inside the signed data, so maliciously modifying that would require breaking the signer-chosen public key algorithm or hash (both trusted by assumption), substituting the public key algorithm for a weaker one (fixed in client configuration for simple roots of trust, or included in X.509 certificates), or substituting the hash algorithm for a weaker one (no defense?).

SPEC.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Hashing Algorithms
+
+Signers and verifiers must know the hash algorithm used in addition to the signature scheme.
+In an attempt to avoid specifying a particular hashing algorithm, we require that digest be calculated using the SAME algorithm as the OCI registry.
```
mtrmac, line 246: That’s practical short-term but problematic longer-term if we ever need to transition from SHA256:
- This makes it impossible to use signatures with a stronger scheme to mitigate SHA256 weaknesses
- If we copy an image from a SHA256 registry to a post-SHA256 registry, or vice versa, even if the manifest+layers+config were exactly identical (I guess with the manifest listing multiple hash algorithms during the transition period), a signature is only valid in one of the registries.

SPEC.md
dlorenc: In practice the scheme is as you described - they're either stored a property of the public key itself or somewhere adjacent to it.

SPEC.md
dlorenc: I think I agree - we should think through and document what the migration from sha256 to something else might look like.

My understanding though is that copying an image from a sha256 registry to a post-sha256 one would result in a new image. Since there are unique images for each registry, we would need unique signatures for each registry. But exactly what that copy/migration process would look like is really a guess at this point.

SPEC.md
mtrmac: I’d argue fairly hard against designing a signature crypto system, or choosing an immature one. It’s not trivial, and it’s a huge distraction from container signing.

Find an industry standard that allows the container signing implementation to call a function with (root of trust in some widely-accepted form, signature blob, maybe (but probably not) an untrusted blob format string), and get a simple “signature valid/invalid” outcome (along with the signed payload if signature is valid).

That doesn’t _entirely_ simplify the discussion because choosing that industry standard still requires avoiding the incompetently-designed ones, and making a bit of a bet on its future viability, but it should at least getting the container signing implementation as far as possible out of the business of handling raw private keys, digest computations and parsing absolutely untrusted byte strings and https://golang.org/pkg/crypto/subtle/ .

With the MIME types in descriptors and the like it’s ultimately not _that big_ a deal to choose a wrong industry standard.

SPEC.md
mtrmac: (As discussed elsewhere, I do think that this field must be present and must be validated, so that users who want `vendor1/mysql:30.0` don’t get `vendor1-beta/mysql:0.0.9`. The downside is that enforcing name matches does require extra configuration of name mapping when mirroring images across registries.)

SPEC.md
dlorenc: +1 to that. This spec really deals with how the fields are stored in a container registry rather than any actual crypto details.

SPEC.md
dlorenc: Right, IMO the downside around copying content is significant enough that we should keep this as a SHOULD rather than a MUST.

dlorenc: Well, I guess we can say that the field MUST be set by the signer, but verification of that field is OPTIONAL for clients?

SPEC.md
mtrmac: My hypothesis is that, in order to avoid a flag day and breaking digest references entirely, there will be some mechanism to represent manifests that contain both the SHA256 digest and the new digest of each blob (e.g. by using SHA256 in the designated descriptor field, and the new digest in an annotation). So it should in principle be possible to have a byte-for-byte identical image in two different registries, accessible using the same `@sha256:…` digest on both, and using some `@new-digest:…` on the updated registry — and it would be nice to keep signatures working when copying the image across.

OTOH it might be possible for the signer to create two separate signatures using the two hash algorithms, and publish both along with the image, so this would not be a deal-breaker.

---

Another concern in such a situation is that the hash algorithm is essentially registry-chosen: given a `repo:tag` reference, I understand “digest be calculated using the SAME algorithm as the OCI registry” to mean “use the algorithm returned by the registry in the `Docker-Content-Digest` header of https://github.com/distribution/distribution/blob/main/docs/spec/api.md#existing-manifests ”, or something like that. The signature scheme should protect against malicious registries, and giving registries this point of influence feels risky.

SPEC.md
mtrmac: That would work fine for me — my primary concern here is interoperability between implementations, i.e. the file formats/APIs, not as much what the implementation ends up doing.

We ultimately can’t force users to verify signatures at all :)

SPEC.md
dlorenc: > Another concern in such a situation is that the hash algorithm is essentially registry-chosen: given a `repo:tag` reference, I understand “digest be calculated using the SAME algorithm as the OCI registry” to mean “use the algorithm returned by the registry in the `Docker-Content-Digest` header of https://github.com/distribution/distribution/blob/main/docs/spec/api.md#existing-manifests ”, or something like that. The signature scheme should protect against malicious registries, and giving registries this point of influence feels risky.

This is a very very good point. I was imagining the algorithm would actually be user controlled in this scenario, but this reinforces that we need to think through it with the OCI folks a bit more. The responsibility on digest calculation is sort of split between users and registries today - the Header from the registry is one place that is registry controlled, I was imagining it would be the one set by clients on upload though: https://docs.docker.com/registry/spec/api/#pushing-an-image

SPEC.md
dlorenc: Awesome, opened https://github.com/sigstore/cosign/issues/229 to start filling it in.

dlorenc: and fixed the typo!

SPEC.md
mtrmac: Quickly skimming docker/distribution , it seems that the implementation can’t look up by arbitrary digests (that would be hard), it essentially chooses its own digest algorithm during blob upload, the clients then must use that one for references.

But that’s not very relevant to the future migration scenarios; many implementations probably hard-code SHA256 in various places, and we don’t know for sure which components will be responsible for which choices in the future.

More to the point, even if the process will end up as uploader-controlled, a malicious registry can just lie and return a different digest (or an attacker with unwanted write access to the repo can upload an image with the uploader-chosen different digest).

dekkagaijin: FWIW OCI clients are in control of digest alg during blob uploads (the final step of an upload is always a PUT with the digest as a parameter). Additionally, clients are required to support digest algs on pulls since content verification is expected. A registry could theoretically lie and simply substitute a different blob (and derivative image config, manifest), but the alteration would propagate up the Merkle tree and be detectable on a pull. Note that such malicious tampering is only possible on levels higher than blobs, since blobs are only referenced by digest (currently) 

@jonjohnsonjr can you think of any monkey business that a registry could get up do that wouldn't be transparent to clients with a simple pull-after-push verification?

jonjohnsonjr: We'll almost certainly need to do some significant legwork if we're going to migrate from sha256. It is hardcoded almost everywhere.

Aligning with what OCI is doing today and upgrading along with them seems like the best choice. We will probably not solve this problem on this comment thread.

SPEC.md
```diff
@@ -0,0 +1,342 @@
...
+Then The digest of the object (of the form `sha256:abcdef...`) is encoded into a tag name using the following rules:
+
+* Replace the `:` character with a `-`
+* Append the `.cosign` suffix
+
+Example digest->tag mapping:
+
+1. Start with `gcr.io/dlorenc-vmtest2/demo:latest`
+2. Resolve this to a digest: `sha256:97fc222cee7991b5b061d4d4afdb5f3428fcb0c9054e1690313786befa1e4e36`
+3. Follow the encoding rules: `sha256-97fc222cee7991b5b061d4d4afdb5f3428fcb0c9054e1690313786befa1e4e36.cosign`
```
priyawadhwa, line 133: for clarity you could add:

4. Signature can be found at gcr.io/dlorenc-vmtest2/demo:sha256-97fc222cee7991b5b061d4d4afdb5f3428fcb0c9054e1690313786befa1e4e36.cosign

SPEC.md
```diff
@@ -0,0 +1,342 @@
...
+We use the tag based mechanism because it's the only option we can think of.
+It meets the two hard requirements: works **everywhere** today and requires no extra services.
+It also does not mutate the signed object (like an attached signature or index might).
+
+Support for multiple signatures works but is racy.
+
+### Alternatives Considered
+
+Notary v1/Grafeas both require another database.
+A few other proprietary APIs exist, but they're not. cross-registry
```
priyawadhwa, line 312: nit: extra period

SPEC.md
```diff
@@ -0,0 +1,342 @@
...
+
+Most common signature schemes support customizable hash algorithms.
+These are typically stored with the signature for convenience, presenting a possible attack/confusion vector.
+
+We decided to pin to the registry hash algorithm.
+This removes an entire moving part, without sacrificing agility.
+
+The registry/client-tooling already perform hash validation as part of the CAS.
+While their spec does not completely pin to specific algorithm, SHA256 is ubiquitous in practice.
+This means that our signed payload object references the actual image "target" by a sha-256 digest - he entire signature already relies on the strength of this algorithm.
```
priyawadhwa, line 324: nit : he entire signature/ the entire signature

SPEC.md
mtrmac: > can you think of any monkey business that a registry could get up do that wouldn't be transparent to clients with a simple pull-after-push verification?

Note that a malicious registry can return different content to different victims.

SPEC.md
```diff
@@ -0,0 +1,354 @@
...
+
+* Tag-based Discovery
+
+#### Tag-based Discovery
+
+In this scheme, signatures are stored in an OCI registry in a predictable location, addressable by tag.
+The location of signatures corresponding to a specific object can be computed using the digest of the object.
+
+If the object is referenced by tag, the tag must first be resolved to a digest.
+Then The digest of the object (of the form `sha256:abcdef...`) is encoded into a tag name using the following rules:
```
asraa, line 124: nit: capital The

SPEC.md
dekkagaijin: > Note that a malicious registry can return different content to different victims.

This is mitigated by pulling by digest, which is an independent UX issue. Assuming clients verify signatures, we'd only need to ensure that the content being pushed to the registry is what we're actually signing.

mtrmac: Neither the signatures nor their metadata are protected by the image digest. (And any end-user of the signed image doesn’t know the digest already, otherwise signatures wouldn’t be necessary.)

dlorenc: > (And any end-user of the signed image doesn’t know the digest already, otherwise signatures wouldn’t be necessary.)

I don't think I agree with this part - signatures are useful even when pulling by digest.

mtrmac: My mistake, the non-repudiation property of the signature (and perhaps any associated metadata) could still be useful, apart from the minimal (signer, name/purpose/identity, binding) essential to the signature.

dlorenc:
Dropping the WIP. I've opened tracking issues for all the current major points of discussion. This is ready for review as an initial spec that describes how things work today, with the improvements we want to make.

SPEC.md
```diff
@@ -0,0 +1,355 @@
...
+  * Simple Signing
+
+### Simple Signing
+
+The Simple Signing payload format is [specified here](https://github.com/containers/image/blob/a5061e5a5f00333ea3a92e7103effd11c6e2f51d/docs/containers-signature.5.md#json-data-format)
+
+The following additional semantics are applied:
+
+* The `mediaType` used to identify this payload format is: `application/vnd.dev.cosign.simplesigning.v1+json`.
+* The `critical.type` value used to identify `cosign` signatures is: `cosign container signature`.
```
dekkagaijin, line 231: Right now it's `cosign container image signature` in sigstore/sigstore, should I change it?
https://github.com/sigstore/sigstore/blob/main/pkg/signature/payload/payload.go#L26

dlorenc: Nah, I'll fix it here.

dekkagaijin (approved):
lgtm

SPEC.md
MarkLodato: Note: I am an author of the [SSL Signing Spec](https://github.com/secure-systems-lab/signing-spec), and I just filed #259 related to this topic.

> The only thing I don't really like about that spec is the usage of PAE.

@dlorenc Could you please explain your reservations?

> I'm not sure I understand the attack/threat model on putting this payload type into the signature.

If you have a signing oracle that uses the same key for multiple message types, then an attacker could potentially get the oracle to sign a message as one type and then have a verifier interpret it as another. Here's a very detailed example showing how to do that in a CI/CD scenario:
https://colab.research.google.com/github/secure-systems-lab/signing-spec/blob/master/hypothetical_signature_attack.ipynb#scrollTo=CIF11AmzXJuQ

Happy to explain more if that's not clear.

dlorenc: My main concern with PAE is that there's very limited library, language, and tooling support for serialization and deserialization. I haven't actually been able to find any for Golang for example.

MarkLodato: PAE itself doesn't need a library. It's just a one-statement concatenation. You never deserialize it. I don't know go, but here it is in Python:

```python
def PAE(payloadType: str, payload: bytes) -> bytes:
  return b''.join([struct.pack('<Q', 2),
                   struct.pack('<Q', len(payloadType)),
                   payloadType.encode('utf-8'),
                   struct.pack('<Q', len(payload)),
                   payload])
```

Now, I agree that SSL Signing Spec needs a reference implementation, but I wanted to make sure there are no security concerns with PAE. It sounds like not.

dlorenc: No security concerns, just interoperability. It's important we can do all these verifications in as much standard tooling as possible. This includes things like bash/openssl.

MarkLodato: Thanks for the feedback. I filed https://github.com/secure-systems-lab/signing-spec/issues/27 to consider a simpler PAE.

Issue #181, opened 2021-03-28
cosign sign --bundle flag


dlorenc:
We should have a way to bundle ALL proofs for into the signature object. This requires trusting all the signing keys through our system:

For all modes:
- Rekor trillian key

For keyless mdoe:
- Fulcio root CA
- CT Log key

The signing keys CT log keys will rotate more quickly than the root CA, so we should sign the other public keys with the Fulcio root CA and bake fulcio root CA trust into all the tools.

Then, a bundle could contain:

- The signature itself
- The signed inclusion proof (and timestamp) of the signature in the transparency log
- [Keyless Mode] The public certificate (containing the public key) corresponding to the signature
- [Keyless Mode] The signed inclusion proof of the certificate in the CT log

Combined with the client's Fulcio Root CA, this bundle could verify everything.

dlorenc:
cc @mnm678 @SantiagoTorres

dlorenc:
@trishankkarthik

trishankkarthik:
@trishankatdatadog

priyawadhwa:
I'm gonna start sketching out a design for this, if anyone has ideas or wants to collab let me know!

trishankatdatadog:
> I'm gonna start sketching out a design for this, if anyone has ideas or wants to collab let me know!

Sounds good! How can I help?

priyawadhwa:
@trishankkarthik I'll ping you on Slack and we can start chatting about it!

priyawadhwa:
As a first step, I'm going to implement this for not-keyless since it's way simpler. For this, we just need to add the following as an annotations:
- the signed inclusion proof that the signature was added to Rekor
- the UUID of the entry in the log

& that should be everything we need to verify the entry is in the log without having to actually access it

dlorenc:
Doc here: https://docs.google.com/document/d/1rVuwr0Acvq_0L1FIZn19bwQ5V3oWoEs_5CxX8H_R8Jc/edit?resourcekey=0-3CtYIjkUlZADfYoUwNvjWQ#heading=h.p4l5zh6qb4zs

dlorenc:
Fixed!

PR #224, opened 2021-04-06
WIP: Bundle rekor inclusion proof into signature

priyawadhwa:
This WIP PR starts to address https://github.com/sigstore/cosign/issues/181 and allows us to verify an entry in rekor without having to check the tlog!

Basically it adds the following as annotations:
- signed inclusion proof
- UUID of the log entry

and then verifies. I've been checking with this and it seems to be working:

```
$ cosign sign -key cosign.key gcr.io/priya-wadhwa/test

$ REKOR_SERVER=notreallol cosign verify -key cosign.pub gcr.io/priya-wadhwa/test

Verification for gcr.io/priya-wadhwa/test --
  - The claims were present in the transparency log
  - The signatures were integrated into the transparency log when the certificate was valid
{"Critical":{"Identity":{"docker-reference":""},"Image":{"Docker-manifest-digest":"sha256:af477e5d17fe3b451853679b31b2ac39872db395dd4bbaf4aa143cd10bd42d96"},"Type":"cosign container signature"},"Optional":null}
```

still have to figure out how to do this for keyless

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -226,11 +226,12 @@ func SignCmd(ctx context.Context, keyPath string,
 			}
 		}
 	}
-	inclusionProof, err := cosign.UploadTLog(signature, payload, pemBytes)
+	inclusionProof, uuid, err := cosign.UploadTLog(signature, payload, pemBytes)
 	if err != nil {
 		return err
 	}
-	uploadOpts.RekorInclusionProof = base64.RawStdEncoding.EncodeToString([]byte(inclusionProof))
+	uploadOpts.RekorInclusionProof = inclusionProof
+	uploadOpts.RekorUUID = uuid
 	return cosign.Upload(dstRef, uploadOpts)
 }
 
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -230,8 +230,10 @@ func SignCmd(ctx context.Context, keyPath string,
 	if err != nil {
 		return err
 	}
-	uploadOpts.RekorInclusionProof = inclusionProof
-	uploadOpts.RekorUUID = uuid
+	uploadOpts.Bundle = &cosign.Bundle{
+		InclusionProof: inclusionProof,
+		UUID:           uuid,
+	}
 	return cosign.Upload(dstRef, uploadOpts)
 }
 
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -231,8 +231,10 @@ func SignCmd(ctx context.Context, keyPath string,
 		return err
 	}
 	uploadOpts.Bundle = &cosign.Bundle{
+		// TODO: this will be SignedInclusionProof
 		InclusionProof: inclusionProof,
 		UUID:           uuid,
+		// TODO: get the rekor intermediate cert and store it here
 	}
 	return cosign.Upload(dstRef, uploadOpts)
 }
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -226,11 +226,11 @@ func SignCmd(ctx context.Context, keyPath string,
 			}
 		}
 	}
-	index, err := cosign.UploadTLog(signature, payload, pemBytes)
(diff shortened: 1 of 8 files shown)


cmd/cosign/cli/sign.go
```diff
@@ -223,12 +226,13 @@ func SignCmd(ctx context.Context, keyPath string,
...
 		}
 	}
-	index, err := cosign.UploadTLog(signature, payload, pemBytes)
+	inclusionProof, uuid, err := cosign.UploadTLog(signature, payload, pemBytes)
 	if err != nil {
 		return err
 	}
-	fmt.Println("tlog entry created with index: ", index)
-	return nil
+	uploadOpts.RekorInclusionProof = inclusionProof
```
asraa, line 233: Should this only be uploaded when bundling is specified?

pkg/cosign/upload.go
```diff
@@ -93,16 +92,30 @@ func UploadTLog(signature, payload []byte, pemBytes []byte) (string, error) {
 			}
 			fmt.Println("Signature already exists. Displaying proof")
 
-			return FindTlogEntry(rekorClient, cs.Base64Signature, cs.Payload, pemBytes)
+			uid, err := FindTlogEntry(rekorClient, cs.Base64Signature, cs.Payload, pemBytes)
```
asraa, line 95: should the log entry proof be grabbed from here to get inclusion proofs for existing entries?

cmd/cosign/cli/sign.go
```diff
@@ -223,12 +226,13 @@ func SignCmd(ctx context.Context, keyPath string,
 			}
 		}
 	}
-	index, err := cosign.UploadTLog(signature, payload, pemBytes)
+	inclusionProof, uuid, err := cosign.UploadTLog(signature, payload, pemBytes)
```
asraa, line 229: i could be totally wrong here but I this this needs to be signed by rekor and it's currently not

i had to walk myself through this:
why does it need to be signed if the merkle tree hashes prove inclusion? cause maybe someone lied about the root hash and is giving a fake inclusion proof into a garbage merkle tree.

i think this needs to be signed by rekor... head hurts :P

cmd/cosign/cli/sign.go
priyawadhwa: Yes! We check for that [here](https://github.com/sigstore/cosign/pull/224/files#diff-52fed6da2e81889fa98c519123e97900541f69ec5ad0d490b164c684da5b51feR211-R213)

asraa: Ahhh I guess it comes down to network access vs space in the image if they're being experimental

cmd/cosign/cli/sign.go
priyawadhwa: i think you're right! just did a quick skim through rekor code & i think we'd have to add that in (but i might have missed it if it's already there)

pkg/cosign/verify.go
```diff
@@ -343,3 +350,35 @@ func correctAnnotations(wanted, have map[string]string) bool {
...
+	}
+
+	hashes := [][]byte{}
+	for _, h := range ip.Hashes {
+		hb, _ := hex.DecodeString(h)
+		hashes = append(hashes, hb)
+	}
+
+	rootHash, _ := hex.DecodeString(*ip.RootHash)
+	leafHash, _ := hex.DecodeString(bundle.UUID)
```
letmaik, line 376: Couldn't I use an existing bundle from any other image and it would validate just fine? This assumes the image uploader wants to pretend they logged the image but actually hasn't.

dlorenc: @priyawadhwa is on vacation this week but I think this entire change is getting rearchitected somewhere else. I'll wait for her to get back and confirm to close it though.

Thanks for taking a look though!

Cc'ing @asraa in case she has a link to the new idea somewhere.

priyawadhwa: Yup, this PR is a bit out of date now. We've decided to add the following to rekor first:
- Return a signed timestamped "promise" that an entry was included in the log, similar to how a CT log returns an SCT

This promise will be included in the bundle. Then, cosign should be able to validate this promise offline. 

I'm going to go ahead and close this PR for now, and then probably open a new one once the new stuff in rekor has been merged!

Issue #243, opened 2021-04-12
Fix cross compilation


dlorenc:
We need to build for OSX and Linux right now, with support for the go-piv on both. This means we need CGO, making cross-compilation really tricky.

jonjohnsonjr:
I'm fairly convinced that zig can help with this.

https://github.com/google/ko/issues/261

https://github.com/ziglang/zig/pull/7377/files#diff-a26d0400c46377165e3ae8362be4528a4e9e5708cc70897a4c00905801a1d581

I just haven't had the time to experiment, and I'm sure there will be non-trivial challenges that I haven't considered.

dlorenc:
i havent been able to get zig working yet with the extra system dependency we need (libsclite-dev).

pearcec:
Are Windows binaries being considered?

dlorenc:
> Are Windows binaries being considered?

I'd be happy to publish them! I don't like publishing things we can't test at all though, and I don't have access to a windows machine. If someone is up for helping us get them tested in CI and published that would be amazing!

PR #244, opened 2021-04-12
Switch the CI builds to use OSX machines for mac builds and linux machines for linux builds.

dlorenc:


dlorenc:
Ref #243

PR #253, opened 2021-04-13
Allow keyless option in experimental mode

font:
Getting an error when using experimental mode with the keyless option. This fix allows it while still checking that if we're in experimental mode, we cannot provide both `-key` and `-kms`.

Signed-off-by: Ivan Font <ifont@redhat.com>


dlorenc:
Hmm, the test failure is related: https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign_test.go#L33

cmd/cosign/cli/sign.go
```diff
@@ -100,6 +100,10 @@ EXAMPLES
 				if !oneOf(*kmsVal, *key) {
 					return &KeyParseError{}
 				}
+			} else {
```
dlorenc, line 103: Ah, I think the issue is that the test was covering the duplicated check below, not up here.

dlorenc (approved):

PR #260, opened 2021-04-15
consolidate keyRef resolution logic

dekkagaijin:


dlorenc:
This is awesome. Looks like you missed the e2e_test.go stuff since that doesn't run by default though.

dlorenc (approved):

PR #261, opened 2021-04-15
Add flag to log all output to a file

font:
This is helpful for running cosign within a container when you do not have access to a shell to redirect output to a file. For example, using the cosign container within CI systems e.g. as a task in Tekton, makes redirecting all output to a file very useful for easier post-processing to analyze the results.

### DIFF
--- a/cmd/cosign/main.go
+++ b/cmd/cosign/main.go
@@ -23,15 +23,16 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/logs"
 	"github.com/peterbourgon/ff/v3/ffcli"
+	"github.com/pkg/errors"
 
 	"github.com/sigstore/cosign/cmd/cosign/cli"
 	"github.com/sigstore/cosign/cmd/cosign/cli/pivcli"
 )
 
 var (
-	rootFlagSet = flag.NewFlagSet("cosign", flag.ExitOnError)
-	debug       = rootFlagSet.Bool("d", false, "log debug output to stderr")
-	verbose     = rootFlagSet.Bool("v", false, "increase log verbosity")
+	rootFlagSet    = flag.NewFlagSet("cosign", flag.ExitOnError)
+	debug          = rootFlagSet.Bool("d", false, "log debug output")
+	outputFilename = rootFlagSet.String("output-file", "", "log output to a file")
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/cosign/main.go
```diff
@@ -47,22 +49,39 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "error: %v\n", errors.Wrapf(err, "Error creating output file %s", *outputFilename))
+			os.Exit(1)
 		}
-		fmt.Fprintf(os.Stderr, "error: %v\n", err)
-		os.Exit(1)
+		stdout := os.Stdout
+		stderr := os.Stderr
+		defer func() {
+			os.Stdout = stdout
+			os.Stderr = stderr
```
dlorenc, line 65: Do you think it makes sense to always capture both stdout and stderr? In general I've tried to be careful and put actual "output" in stdout, and user-centered "messages" to stderr.

cmd/cosign/main.go
```diff
@@ -47,22 +49,39 @@ func main() {
...
-		}
-		fmt.Fprintf(os.Stderr, "error: %v\n", err)
-		os.Exit(1)
+		printErrAndExit(err)
+	}
+}
+
+func printErrAndExit(err error) {
+	if *verbose {
+		fmt.Print("verbose!")
```
dlorenc, line 83: I think this can get deleted :) It was a copy/paste error IIRC.

dlorenc:
Big +1 to this overall, one question on stderr

cmd/cosign/main.go
font: I'm not sure I'm clear on the distinction. For example, https://github.com/SigStore/cosign/blob/2befb2893ee5ac5e697ed09fd4c146eef98a6ab7/cmd/cosign/cli/verify.go#L127 contains print statements to `stderr` for results of verifying claims, tlog entry, etc., while logging error messages to `stdout`. I think I would expect the opposite in that case and so I wasn't able to find a consistent pattern throughout to understand a distinction. As a result I found needing to see both stdout and stderr messages. But that's not necessarily a reason to log both to a file, and in principle I agree and think it's useful to separate them or at least duplicate them if nothing else.

We can just log to `stdout` for now (leave `stderr` untouched), file an issue to make `stdout`/`stderr` logging convention consistent and perhaps we'll need to consider revamping the entire print approach to use a more abstracted logging mechanism. WDYT?

cmd/cosign/main.go
font: I didn't know if it was part of more elaborate plans. :smile: I'll remove in the next push and remove the flag as well.

cmd/cosign/main.go
dlorenc: > We can just log to `stdout` for now (leave `stderr` untouched), file an issue to make `stdout`/`stderr` logging convention consistent and perhaps we'll need to consider revamping the entire print approach to use a more abstracted logging mechanism. WDYT?

Perfect! I *think* that code you linked is actually working as I intended it to, but I admit the intention isn't very clear and could probably use improvements.

dlorenc (dismissed):

cmd/cosign/main.go
font: Cool! Filed https://github.com/sigstore/cosign/issues/263.

font:
@dlorenc One more review needed - pushed an update to the flag usage.

dlorenc (approved):

Issue #274, opened 2021-04-21
Dedupe signatures


dekkagaijin:
We currently blindly append a new signature:

```
$ docker run -it gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064 -- verify -key https://raw.githubusercontent.com/GoogleContainerTools/distroless/main/cosign.pub gcr.io/distroless/base:latest
Unable to find image 'gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064' locally
gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064: Pulling from projectsigstore/cosign/ci/cosign
5dea5ec2316d: Already exists
bb771d6dc9a1: Already exists
9127c3610b7e: Already exists
72164b581b02: Already exists
6fe218878cac: Pull complete
Digest: sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064
Status: Downloaded newer image for gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064

Verification for gcr.io/distroless/base:latest --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.

[...]

PR #280, opened 2021-04-27
Dedupe signature uploads

### LINKED ISSUE #109: Use the location header from rekor
When this goes in we can simplify our upload handling:
https://github.com/sigstore/rekor/pull/226/files

dekkagaijin:
Fixes #109 and #274
 
### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -163,20 +163,23 @@ func SignCmd(ctx context.Context, so SignOpts,
 	}
 
 	var signer signature.Signer
+	var dupeDetector signature.Verifier
 	var cert, chain string
 	switch {
 	case so.Sk:
-		sk, err := pivkey.NewSigner()
+		sk, err := pivkey.NewSignerVerifier()
 		if err != nil {
 			return err
 		}
 		signer = sk
+		dupeDetector = sk
 	case so.KeyRef != "":
-		k, err := signerFromKeyRef(ctx, so.KeyRef, so.Pf)
+		k, err := signerVerifierFromKeyRef(ctx, so.KeyRef, so.Pf)
 		if err != nil {
 			return errors.Wrap(err, "reading key")
 		}
 		signer = k
+		dupeDetector = k
 	default: // Keyless!
 		fmt.Fprintln(os.Stderr, "Generating ephemeral keys...")
 		k, err := fulcio.NewSigner(ctx)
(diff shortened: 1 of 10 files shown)


cmd/cosign/cli/sign.go
```diff
@@ -172,11 +173,12 @@ func SignCmd(ctx context.Context, so SignOpts,
 		}
 		signer = sk
 	case so.KeyRef != "":
```
dlorenc, line 175: Any reason to not just do the dupe detection on every upload? I think it's probably safe and cheap enough to do it on every upload, even though it wouldn't really be hit ever for keyless ones.

dlorenc: Or maybe flip the polarity so the default is to detect dupes, but keyless mode skips that. (since I think we do want dupe detection for security keys, which are left out here).

dekkagaijin: in order to detect dupes, we need a Verifier unless we can rely on deterministic signatures (which isn't standard practice). I can add support for security keys in this PR

> I think it's probably safe and cheap enough to do it on every upload, even though it wouldn't really be hit ever for keyless ones.

imho we shouldn't _require_ a dupe detector if it can be known that an upload should be performed. It'd just result in no-op boilerplate to avoid dereferencing nil

dekkagaijin: > I can add support for security keys in this PR

done

dlorenc: Ah yeah makes sense.

pkg/cosign/remote.go
```diff
@@ -45,20 +49,53 @@ func Descriptors(ref name.Reference) ([]v1.Descriptor, error) {
...
+		mt: SimpleSigningMediaType,
+	}
+
+	base, err := SignatureImage(dst, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	if err != nil {
+		return nil, err
+	}
+
+	if dupeDetector != nil {
+		signedDigest, err := l.Digest()
```
dlorenc, line 80: style: maybe move this out to a helper?

if dupeDetector != nil {
  detectDupes(dupedetector, base)
}

dlorenc (commented):
One more nit.

pkg/cosign/remote.go
dekkagaijin: done

dlorenc (approved):

Issue #281, opened 2021-04-28
Use rekor signedEntryTimestamp if present


dlorenc:
We should figure out where to store these in the OCI registry (an annotation?), update the SPEC, start including them and then start looking for them during verification.

We can always use one of these if present (no opt-in/out needed, and do a force check of the log (--verify-inclusion)? as opt-in.

This should mean we can eliminate read operations from experimental for now (moving them to default). We still need a way to get the write calls out of experimental.

priyawadhwa:
Yah I think we can store the SET as part of the "bundle" annotation (which will probably end up including other stuff like certs for keyless mode).

dlorenc:
Is this done now?

priyawadhwa:
Yup!

PR #285, opened 2021-04-28
Use rekor signedEntryTimestamp if present

priyawadhwa:
If the SET is present, we store the logEntry as a bundle annotation. On verify, we first try to verify it against the rekor public key (not experimental). if that fails, then we hit the tlog. 

Starts to address https://github.com/sigstore/cosign/issues/281, still need to add the `--verify-inclusion` flag to force checking the tlog


SPEC.md
```diff
@@ -99,7 +99,20 @@ Gyp4apdU7AXEwysEQIb034aPrTlpmxh90SnTZFs2DHOvCjCPPAmoWfuQUwPhSPRb
...
 ```
 
-* `bundle` TODO
+* `bundle` string
+  This OPTIONAL property contains a JSON formatted `bundle` type, which can be used for offline verification.
+  Example `bundle`:
+
+```json
+{
+  "body": "eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJzcGVjIjp7ImRhdGEiOnsiaGFzaCI6eyJhbGdvcml0aG0iOiJzaGEyNTYiLCJ2YWx1ZSI6IjQwZGI5ZmE1MjkxNDM0NjU4NDNkYzI2MTZjMDI0YmQ2MzA2Mjg4NzUzMzJhM2UyMjg1MTA1MzczYmFjZWZlMDIifX0sInNpZ25hdHVyZSI6eyJjb250ZW50IjoiTUVRQ0lEcnRNMDZ4c2szNTBKWEtyd1lYb0VPU3Q5QndncmNqdDl0djdReUdiOFNUQWlCYW9IKzFIVDlSb0dVK05XVHJXdFFQbVJZZElFZTBIb1lhUjFLeVZNUm43dz09IiwiZm9ybWF0IjoieDUwOSIsInB1YmxpY0tleSI6eyJjb250ZW50IjoiTFMwdExTMUNSVWRKVGlCUVZVSk1TVU1nUzBWWkxTMHRMUzBLVFVacmQwVjNXVWhMYjFwSmVtb3dRMEZSV1VsTGIxcEplbW93UkVGUlkwUlJaMEZGV0dVMVlYRjRibVpHY21KNFpUQnJkM1UxU0ZOa09VUTFhUzlJVkFwVVdUQjVSelJEWVd4NVJHNHlWbVpGT1hSVlYzbHZVVlIzU0V0dlZVVlpZamc1VW5WQk5tMU1SM05wV1hCMVQyNW9lR3RQUkUxcmJXOW5QVDBLTFMwdExTMUZUa1FnVUZWQ1RFbERJRXRGV1MwdExTMHRDZz09In19fSwia2luZCI6InJla29yZCJ9",
```
dlorenc, line 108: I'll be honest, I'm a little disappointed you didn't hide something fun in here.

pkg/cosign/keys.go
```diff
@@ -38,10 +38,11 @@ import (
...
 const (
-	pemType  = "ENCRYPTED COSIGN PRIVATE KEY"
-	sigkey   = "dev.cosignproject.cosign/signature"
-	certkey  = "dev.sigstore.cosign/certificate"
-	chainkey = "dev.sigstore.cosign/chain"
+	pemType   = "ENCRYPTED COSIGN PRIVATE KEY"
+	sigkey    = "dev.cosignproject.cosign/signature"
+	certkey   = "dev.sigstore.cosign/certificate"
+	chainkey  = "dev.sigstore.cosign/chain"
+	BundleKey = "dev.sigstore.cosign/bundle"
```
dlorenc, line 45: This name looks fine to me, anyone else have an opinion?

pkg/cosign/rekor.pub
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
```
dlorenc, line 1: Want to leave a comment somewhere (not here probably) about how you got this file from Rekor, and how/when to update it?

cmd/cosign/cli/sign.go
```diff
@@ -241,10 +247,31 @@ func SignCmd(ctx context.Context, so SignOpts,
...
+	uo.Bundle = bundle(entry)
+	uo.AdditionalAnnotations = annotations(entry)
+	if _, err = cosign.Upload(ctx, uo); err != nil {
+		return errors.Wrap(err, "uploading")
+	}
 	return nil
 }
+
+func bundle(entry *models.LogEntryAnon) *cosign.Bundle {
+	if entry.Verification == nil {
```
dlorenc, line 265: it's good to check, but is this an expected case? Should we warn here?

pkg/cosign/verify.go
```diff
@@ -203,6 +214,11 @@ func Verify(ctx context.Context, ref name.Reference, co CheckOpts) ([]SignedPayl
 			}
 		}
 
+		verified, err := sp.VerifyBundle()
```
dlorenc, line 217: We should explain how this works a bit in the output too, in cli/verify.go I think.

dlorenc (commented):
This is so cool.

SPEC.md
dlorenc, line 108: We should explain the full contents of the bundle here too. That probably deserves it's own entire section.

pkg/cosign/remote.go
```diff
@@ -95,29 +105,51 @@ func findDuplicate(ctx context.Context, sigImage v1.Image, payload []byte, dupeD
 	return nil, nil
 }
 
-func Upload(ctx context.Context, signature, payload []byte, dst name.Reference, cert, chain string, dupeDetector signature.Verifier) (uploadedSig []byte, err error) {
+type Bundle struct {
+	*models.LogEntryAnon
```
dlorenc, line 109: (See my comment on the spec above too), we should think through this format a bit since it will become the public API. We may want to version it (though we can do that via the annotation), but we definitely want to explicitly place the fields to separate this type from the Rekor API. Nesting this portion down a bit (Bundle.Verification) might also help place the future keyless part.

cmd/cosign/cli/sign.go
priyawadhwa: I put it in there in case people were verifying old signatures that wouldn't have this! That's pretty much the only case I can think of.

pkg/cosign/remote.go
priyawadhwa: That makes sense! Maybe it would be better to store the payload, signature and SET in the bundle, and then we can verify those more easily (right now it actually generates the payload during verification instead)

cmd/cosign/cli/sign.go
dlorenc: Ah makes sense.

SPEC.md
priyawadhwa: 🤔

pkg/cosign/keys.go
priyawadhwa: we could do `dev.sigstore.cosign/bundle/v1` if we plan on versioning this

pkg/cosign/rekor.pub
priyawadhwa: Added as a comment in verify.go

SPEC.md
priyawadhwa: Added in some details about the payload + SET

pkg/cosign/keys.go
dekkagaijin: > we could do dev.sigstore.cosign/bundle/v1 if we plan on versioning this

Could also add a `.v2` later if that ever happens :p

pkg/cosign/verify.go
priyawadhwa: 👍🏽

pkg/cosign/remote.go
```diff
@@ -80,7 +83,14 @@ func findDuplicate(ctx context.Context, sigImage v1.Image, payload []byte, dupeD
 		return nil, err
 	}
 
+LayerLoop:
 	for _, layer := range manifest.Layers {
+		// if there are any new annotations, then this isn't a duplicate
```
priyawadhwa, line 88: @dekkagaijin does this look ok to you? basically I needed cosign to know that if we are trying to add a new bundle annotation then don't mark this new signature as a duplicate

dekkagaijin: hrm. we'd probably also want to verify that the values are the same (for annotations other than `signature`)

pkg/cosign/verify.go
```diff
@@ -275,6 +301,25 @@ func (sp *SignedPayload) VerifyClaims(d *v1.Descriptor, ss *payload.SimpleContai
...
+func (sp *SignedPayload) VerifyBundle() (bool, error) {
+	if sp.Bundle == nil {
+		return false, nil
+	}
+	rekorPubKey, err := PemToECDSAKey([]byte(rekorPub))
+	if err != nil {
+		return false, errors.Wrap(err, "pem to ecdsa")
+	}
+	// verify the SET against the public key
+	h := crypto.SHA256.New()
```
dekkagaijin, line 313:
```
hash := sha256.Sum256(sp.Bundle.CanonicalizedPayload)
if !ecdsa.VerifyASN1(rekorPubKey, hash[:], []byte(sp.Bundle.SignedEntryTimestamp)) {
   return false, fmt.Errorf("unable to verify")
}
```

pkg/cosign/verify.go
```diff
@@ -275,6 +301,25 @@ func (sp *SignedPayload) VerifyClaims(d *v1.Descriptor, ss *payload.SimpleContai
 	return nil
 }
 
+func (sp *SignedPayload) VerifyBundle() (bool, error) {
```
dekkagaijin, line 304: should probably just return `error`, unless `error == nil` would ever not imply a successful validation

dekkagaijin (dismissed):

pkg/cosign/verify.go
priyawadhwa: if Bundle isn't available for some reason, having both tells us that it was unverified but also there was no actual error to return. The bool is used later to decide if we still need to hit the tlog.

pkg/cosign/verify.go
priyawadhwa: thanks :)

pkg/cosign/remote.go
dlorenc: +1

pkg/cosign/verify.go
dekkagaijin: Ah, I just noticed the very next line. Carry on :p

pkg/cosign/remote.go
priyawadhwa: for sure

SPEC.md
```diff
@@ -99,7 +99,19 @@ Gyp4apdU7AXEwysEQIb034aPrTlpmxh90SnTZFs2DHOvCjCPPAmoWfuQUwPhSPRb
...
 
-* `bundle` TODO
+* `bundle` string
+  This OPTIONAL property contains a JSON formatted `bundle` type, which can be used for offline verification.
+  Example `bundle`:
+
+```json
+{
+  "SignedEntryTimestamp": "MEUCIB31Q9vol8f9UXz6ppNkyQeDv5qAEMLJdhlonWpB7VEnAiEA7t9rcHnfN9FO2rJMKK7g633EJk8CwN+TptwgnvRkyxM=",
+  "CanonicalizedPayload": "cHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlhcHJpeWFwcml5YXByaXlh"
```
dlorenc, line 109: lgtm

pkg/cosign/remote.go
```diff
@@ -95,29 +105,53 @@ func findDuplicate(ctx context.Context, sigImage v1.Image, payload []byte, dupeD
 	return nil, nil
 }
 
-func Upload(ctx context.Context, signature, payload []byte, dst name.Reference, cert, chain string, dupeDetector signature.Verifier, auth authn.Keychain) (uploadedSig []byte, err error) {
+type Bundle struct {
+	SignedEntryTimestamp strfmt.Base64
+	CanonicalizedPayload []byte
+}
+
+type UploadOpts struct {
```
dlorenc, line 113: I like moving to the opts but don't want it to become just a huge grab bag, it also gets weird for mandatory vs. optional ones. What about separating out the required ones and the optional ones?

Leave signature, payload and dst as actual args, and move cert, chain, dupeDetector, auth and bundle to the struct?

dlorenc (commented):
One nit on the opts struct.

pkg/cosign/remote.go
priyawadhwa: good call! I left auth in as well since I think we need it

dlorenc:
Ehl gee tee emm!

dlorenc (approved):

PR #294, opened 2021-05-01
Use matrix for build, include windows, and unify SHA256 hash with powershell

pearcec:
#243

### DIFF
--- a/.github/workflows/cross.yaml
+++ b/.github/workflows/cross.yaml
@@ -4,8 +4,8 @@ on:
       - main
       - release-*
     tags:
-      - '*'
-    pull_request:
+      - "*"
+  pull_request:
 name: Cross
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/cross.yaml
```diff
@@ -4,63 +4,51 @@ on:
...
+        uses: actions/checkout@v2
+      - name: install deps
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -yq libpcsclite-dev
+      - name: build
+        run: make cosign && mv ./cosign ./cosign-${{matrix.TARGET}}-amd64
+      - name: sign
+        run: |
+          echo -n "${{secrets[matrix.COSIGN_PASSWORD]}}" | ./cosign-${{matrix.TARGET}}-amd64 sign-blob -key ./.github/workflows/cosign.key ./cosign-${{matrix.TARGET}}-amd64 > cosign-${{matrix.TARGET}}-amd64.sig
+      - name: Print Info
```
dlorenc, line 41: Does this one need to be within an "if matrix.os..." clause like the one for ubuntu on 34?

dlorenc (commented):
Nice!

.github/workflows/cross.yaml
pearcec: No, you want this to run for all three since they each will have different signatures.

dlorenc (approved):

dlorenc (approved):

dlorenc:
This is awesome!

PR #300, opened 2021-05-04
Use OCI Image Manifest v1

### LINKED ISSUE #250: Spec: Switch to OCI media types
Ref: #214 

This is backwards compatible, so no big rush in getting it in. We'll want to be careful and test with all the supported registries again as part of the rollout.

dekkagaijin:
Fixes #250

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -156,11 +156,12 @@ Implementations MAY store signatures objects in the same OCI repository as the t
 This section describes the way the properties from above are embedded into OCI objects that can be stored in a registry.
 Implementations MUST support storing signatures in at least the following object types:
 
-* [Image Manifest V2 Schema 2](https://docs.docker.com/registry/spec/manifest-v2-2/)
+* [OCI Image Manifest V1](#oci-image-manifest-v1)
 
-#### Image Manifest V2 Schema 2
+#### OCI Image Manifest V1
 
-This section describes the way the mandatory and optional signature properties are embedded into an Image Manifest V2 Schema 2 object.
+This section describes the way the mandatory and optional signature properties are embedded into an
+[OCI Image Manifest V1](https://github.com/opencontainers/image-spec/blob/master/manifest.md) object.
 
 Only one image manifest is created for every signed object.
 Multiple signatures can be embedded in one image manifest.
@@ -177,8 +178,8 @@ Example `payload`:
 ```
 {
   "schemaVersion": 2,
-  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
     <omitted for brevity>
   },
   "layers": [
--- a/pkg/cosign/remote.go
+++ b/pkg/cosign/remote.go
@@ -52,7 +52,7 @@ func Descriptors(ref name.Reference) ([]v1.Descriptor, error) {
 	return m.Layers, nil
 }
 
-// SignatureImage
+// SignatureImage returns the existing destination image, or a new, empty one.
 func SignatureImage(dstTag name.Reference, opts ...remote.Option) (v1.Image, error) {
 	base, err := remote.Image(dstTag, opts...)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


dekkagaijin:
@jonjohnsonjr any guesses how/where this might not work? GCR is happy

SPEC.md
```diff
@@ -177,7 +178,7 @@ Example `payload`:
 ```
 {
   "schemaVersion": 2,
-  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
```
jonjohnsonjr, line 181: I suspect this gets dropped: https://github.com/google/go-containerregistry/blob/9cf3ed4ac182c640e59a82343512e4c2cacb6d68/pkg/v1/mutate/image.go#L132-L140

(There's also no way to set the config file mediaType... we should either add that to ggcr or infer it like we do the top-level mediaType.)

jonjohnsonjr (commented):
I'm honestly surprised GCR doesn't complain.

SPEC.md
dekkagaijin: We aren't using an image index, just a manifest. The term "index" has been used to describe the manifest which contains the signatures, but it's not a manifest list

> There's also no way to set the config file mediaType... we should either add that to ggcr or infer it like we do the top-level mediaType.

SGTM

dekkagaijin: never mind, I see what you mean now

dekkagaijin: what I did seems to work fine:
```
base = mutate.MediaType(empty.Image, types.OCIManifestSchema1)
m, _ := base.Manifest()
m.Config.MediaType = types.OCIConfigJSON
```

pkg/cosign/upload.go
```diff
@@ -34,10 +34,11 @@ import (
 )
 
 const (
-	ExperimentalEnv = "COSIGN_EXPERIMENTAL"
-	repoEnv         = "COSIGN_REPOSITORY"
-	ServerEnv       = "REKOR_SERVER"
-	rekorServer     = "https://api.rekor.dev"
+	ExperimentalEnv    = "COSIGN_EXPERIMENTAL"
+	repoEnv            = "COSIGN_REPOSITORY"
+	LegacyMediaTypeEnv = "COSIGN_LEGACY_MEDIA_TYPE"
```
dlorenc, line 39: Want to just call it COSIGN_DOCKER_MEDIA_TYPE instead of legacy to make it more clear?

dlorenc:
One nit, a one-line integration test of the legacy env var here would be good too: https://github.com/sigstore/cosign/blob/main/test/e2e_test_secrets.sh#L55

pkg/cosign/remote.go
```diff
@@ -52,15 +52,23 @@ func Descriptors(ref name.Reference) ([]v1.Descriptor, error) {
...
 			}
-			base = empty.Image
+			if !LegacyMediaType() {
+				base = mutate.MediaType(empty.Image, types.OCIManifestSchema1)
+				m, err := base.Manifest()
+				if err != nil {
+					// should never happen...?
+					return nil, err
+				}
+				m.Config.MediaType = types.OCIConfigJSON
```
jonjohnsonjr, line 70: Oh that's interesting. I don't think we intended for that return value to be mutable, but I guess it is...

cc @imjasonh I would consider this a bug.

pkg/cosign/upload.go
dekkagaijin: done

pkg/cosign/remote.go
dlorenc: Hahha.

dekkagaijin:
> One nit, a one-line integration test of the legacy env var here would be good too: https://github.com/sigstore/cosign/blob/main/test/e2e_test_secrets.sh#L55

done

pkg/cosign/remote.go
dekkagaijin: let us have our fun :p

dlorenc (approved):

PR #308, opened 2021-05-06
add `make cosign-static` target, use in CI `cross` for linux

dekkagaijin:


dlorenc:
So it does work well by default on OSX (and I think Windows...). Can we do this for linux only somehow?

dekkagaijin:
> So it does work well by default on OSX (and I think Windows...). Can we do this for linux only somehow?

done

dlorenc (approved):

PR #312, opened 2021-05-08
fix usage example for sign + verify subcommands

Dentrax:
`$ cosign verify -k cosign.pub ...` is the wrong usage since we need to verify with a private key and enter the password in prompt. In case if we give a public key, it throws: `unsupported pem type: PUBLIC KEY`

Signed-off-by: Furkan Turkal <furkan.turkal@trendyol.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ $ cosign generate-key-pair
 Enter password for private key:
 Enter again:
 Private key written to cosign.key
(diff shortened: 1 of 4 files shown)


dlorenc:
Thank you again @Dentrax!

dlorenc (approved):

Issue #140, opened 2021-03-22
Code coverage metric in CI


asraa:
Hi!

I was peering around CI (yay at signing CI builds!), and wondering whether a code coverage metric would be useful per push/PR to detect coverage regressions on new code. 
https://github.com/sigstore/cosign/blob/ca5a8aa06525cc6a732812aef243eaa7f6bff6ac/.github/workflows/tests.yaml#L3

I'm very new to Go so don't know its extend of coverage profiling (how easy it is to filter files/generate reports), but this would be cool to have!

dlorenc:
Yes! It's actually pretty easy to add coverage, but every continuous reporting system I've seen has been incredibly flaky/problematic. The one in Tekton never seems to work.

I'd be up for trying another one though if anyone has recommendations!

To do it in cosign now, you can run:

```
$ go test -tags=e2e -coverpkg=./... -coverprofile=cover.out ./...
$ go tool cover -html=cover.out
```

dlorenc:
![image](https://user-images.githubusercontent.com/1714486/116797523-3e39bc00-aaac-11eb-9219-c8442e881a29.png)

PR #314, opened 2021-05-08
Containerized dev environment

### LINKED ISSUE #114: Multi platform makefile
Arm, osx!

### LINKED ISSUE #243: Fix cross compilation

### LINKED ISSUE #305: Static binaries are broken
**Description**

The `cosign` binary isn't static so unfortunately grabbing the new release and trying to use it results in shared library errors.

```
dan@code:~$ wget https://github.com/sigstore/cosign/releases/download/v0.4.0/cosign-linux-amd64
--2021-05-05 13:55:56--  https://github.com/sigstore/cosign/releases/download/v0.4.0/cosign-linux-amd64
Resolving github.com (github.com)... 140.82.121.4
Connecting to github.com (github.com)|140.82.121.4|:443... connected.

<URL Verbosity>

HTTP request sent, awaiting response... 200 OK
Length: 29108032 (28M) [application/octet-stream]
Saving to: ‘cosign-linux-amd64’

cosign-linux-amd64  100%[===================>]  27.76M  6.13MB/s    in 4.9s

2021-05-05 13:56:01 (5.62 MB/s) - ‘cosign-linux-amd64’ saved [29108032/29108032]

dan@code:~$ mv cosign-linux-amd64 cosign
dan@code:~$ ./co^C
dan@code:~$ chmod +x ./cosign
dan@code:~$ ./cosign
./cosign: error while loading shared libraries: libpcsclite.so.1: cannot open shared object file: No such file or directory
dan@code:~$ file ./cosign
./cosign: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=0f8323346ee35dca8129cd8f12f28ec38b9ba7d9, for GNU/Linux 3.2.0, not stripped
dan@code:~$
```

crazy-max:
Fixes #114 
Fixes #243 
Fixes #305 

First of all thanks for this project!

I was going to use it through GitHub Actions but fails because the final binary is not statically linked (#305). I wanted to create a PR to fix that but looking at the workflow I think we can enhance this behavior using a containerized dev environment which is the goal of this PR.

So what is a containerized dev environment? Well as you may know, not everybody has the tool chain locally to be able to build the project. That's why I think using Docker to create artifacts, validate and test the project with a single command in an agnostic way for major platforms seems a good idea.

Here is what have been done in this PR.

In the [hack folder](https://github.com/crazy-max/cosign/tree/ctn-dev-env/hack) you will find the Dockerfiles to:
* Build the project (create multi-platform artifacts)
* Check license header
* Lint
* Test
* Validate and update vendors

These Dockerfiles are handled through[ buildx bake](https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md) which is a high-level build command for buildx through a HCL definition you can find in [`docker-bake.hcl`](https://github.com/crazy-max/cosign/blob/ctn-dev-env/docker-bake.hcl). I have enhanced the Makefile to be able to run the targets available in this definition.

The major change of this PR is to allow building statically linked and multi-platform artifacts through the `artifact-all` target which will produce the [following artifacts on release](https://github.com/crazy-max/cosign/releases/tag/v0.4.1):

image

I have also added a vendor validation because go sum is tampered so I think in the future we should check everything is ok before merging a PR.

I also updated and merged GitHub Actions workflows. You can see the result [on my fork](https://github.com/crazy-max/cosign/actions). I wanted to add code coverage with codecov but want your feedback on it before (#140).

I know this PR is quite huge. Just wanted to showcase the benefits but we can split it if you want.


.github/workflows/build.yml
```diff
@@ -0,0 +1,128 @@
+#
+# Copyright 2021 The Sigstore Authors.
```
dlorenc, line 2: nit: could we keep the .yaml extension here and on all the other github files to make history tracking/review easier?

crazy-max: @dlorenc Yeah agree. I will review other changes too to avoid breaking history.

crazy-max: @dlorenc Done, let me know if you need anything else.

PR #315, opened 2021-05-09
Add cosign upload wasm command!

dlorenc:
This probably isn't the ideal location/UX for the command, but I was able to get it working here when playing with https://www.kubewarden.io/.

We want to be helping sign/push WASM for other uses cases too, but they might use different storage mechanisms. This version (for kubewarden) used the one defined at https://github.com/engineerd/wasm-to-oci. There's another spec here: https://www.solo.io/blog/announcing-the-webassembly-wasm-oci-image-spec/

I'll need to understand the differences between these specs to see which (if any) might make sense to support here. For kubewarden, the policies need to be compiled to wasm and uploaded to a registry for retrieval. The documentation points to using the actual `wasm-to-oci` tool, but I was unable to get that working with GCR.

I read the spec and tried to reimplement it here, where it did work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,20 @@ tlog entry created with index:  5086
 Pushing signature to: us.gcr.io/dlorenc-vmtest2/demo:sha256-124e1fdee94fe5c5f902bc94da2d6e2fea243934c74e76c2368acdc8d3ac7155.sig
 ```
 
+#### WASM
+
+Web Assembly Modules can also be stored in an OCI registry, using this [specification](https://github.com/solo-io/wasm/tree/master/spec).
+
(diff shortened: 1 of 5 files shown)


font (commented):
Awesome!

+1 to supporting signing/verifying webassembly modules. These wasm policies are good place to start.

dlorenc:
cc @asraa 

I think we decided to move this to a subcommand - cosign wasm --- right?

dlorenc:
Sorry! I meant one more level deep - cosign wasm upload then cosign wasm sign

asraa:
Oh, yeah, that definitely makes sense.

asraa (approved):
I like how it sits under the upload command!

PR #317, opened 2021-05-11
implement `cosign copy`

### LINKED ISSUE #303: Cosign cp
**Description**

It's pretty easy to copy an image with crane, and you can use cosign triangulate to figure out how to copy signatures, but it's not trivial to hook up in bash.

Let's add a cosign cp command to copy an image with it's signatures.

dekkagaijin:
Fixes #303

### DIFF
--- a/cmd/cosign/cli/copy.go
+++ b/cmd/cosign/cli/copy.go
@@ -1,0 +1,127 @@
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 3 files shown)


cmd/cosign/cli/copy.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"github.com/sigstore/cosign/pkg/cosign"
+)
+
+func Copy() *ffcli.Command {
+	var (
+		flagset     = flag.NewFlagSet("cosign copy", flag.ExitOnError)
+		srcImgFlag  = flagset.String("source", "", "the image (and associated signatures) to copy")
```
dlorenc, line 32: Do you think we need to be explicit about src and dst with named flags rather than positionals? I could see it either way.

cmd/cosign/cli/copy.go
```diff
@@ -0,0 +1,109 @@
...
+	)
+	return &ffcli.Command{
+		Name:       "copy",
+		ShortUsage: "cosign copy -source <source image> -destination <destination image>",
+		ShortHelp:  `Copy the supplied container image and signatures.`,
+		LongHelp: `Copy the supplied container image and signatures.
+
+EXAMPLES
+  # copy a container image and signature
+  cosign copy -source example.com/foo:latest -destination example.com/bar:oldest
```
dlorenc, line 45: This might not be the best example since it's the same repo. It would "work" but isn't really any different than a crane cp, since the signature doesn't actually move. Maybe two repos?

dlorenc (commented):
Nice! two nits.

cmd/cosign/cli/copy.go
dekkagaijin: made it positional `src dst`

cmd/cosign/cli/copy.go
dekkagaijin: done

dlorenc (approved):

PR #332, opened 2021-05-20
make builds static (and yubikey support disabled) by default

dekkagaijin:
Added `make cosign-pivkey` to offer support

Fixes: https://github.com/sigstore/cosign/issues/305


dlorenc:
ughhhhhhh yeah this is fine, let's not give up on trying to actually get this statically compiled quite yet though...

dekkagaijin:
I don't think we'll be able to as long as libc is a transitive dependency

dlorenc (approved):

Issue #86, opened 2021-03-12
Cosign/TUF integration


mnm678:
As mentioned in the README, cosign can be used to store [TUF](https://theupdateframework.io/) metadata on a registry. This TUF metadata can secure the tag to digest relationships on the registry, define trust delegations, and provide compromise resilience in the case of a registry or infrastructure compromise.  It would be nice to get a working prototype of this integration, possibly leveraging existing TUF implementations such as [go-tuf](https://github.com/theupdateframework/go-tuf). 

As I envision it, a TUF/cosign integration would involve:
[x] Storing TUF metadata on a registry using cosign
[] Ensuring easy querying of that TUF metadata by users, ideally using a human-readable name

In the future, this could lead to more sigstore integration, for example:
[] [Allowing TUF targets metadata to delegate to a fulcio identity](https://github.com/theupdateframework/taps/pull/141)
[] Supporting TUF root key discovery through a threshold of fulcio servers
[] Publishing TUF metadata to rekor for an immutable history

dlorenc:
This sounds awesome! How can we help?

mnm678:
I'll play more with storing TUF metadata using cosign this week, but I think the biggest issue there will be the discoverability of the metadata. So one place I could use some help is a mechanism to query cosign objects by some kind of deterministic name (ie 'root' for root metadata).

For the future steps, allowing storage of the TUF metadata in rekor would require support for either arbitrary data added to the TL or support specifically for TUF metadata (I imagine the former would be easier).

lukehinds:
Could you outline a bit more @mnm678 what would constitute arbitrary data , we might be able to accommodate that with rekor types, but I guess it depends on how arbitrary we are talking?

dlorenc:
Sorry! Missed this reply.

Depending on how much TUF metadata there is, we can store it as a custom object in cosign. Right now we hardcode the "cosign signature object" type in the redhat simple signing object, but don't really do much else to look at that field.

We could make that configurable and queryable pretty easily, if the TUF metadata can fit into the optional section.

mnm678:
The metadata should be able to fit into the optional section (I'm working on a PoC for that). It's just json with a specific set of fields (example from the [TUF spec](https://theupdateframework.github.io/specification/latest/#file-formats-root)). 

The only caveat is that not every piece of TUF metadata would be directly associated with an image, so the 'image' field in the simple signing object doesn't make as much sense.

dlorenc:
[...]

PR #366, opened 2021-06-10
tuf demo code

asraa:
This is some WIP TUF code. 

The major thing I want to do is to not rely on my hand-grown store. I'd like to remove a lot of it in favor of actually using methods exposed by [Repo](https://github.com/theupdateframework/go-tuf/blob/aee6270feb5596036edde4b6d7564fa17db811cb/repo.go#L56) , which I think is possible. Here were some fixes to undraftify it:

* Get go-tuf implementation to be OK with adding public keys (like python impls' "add_verification_key")
* Generalize adding targets in go-tuf where we can add non-file-based targets
* Remove my hand-grown signing. cosign keys. Dependent on https://github.com/sigstore/sigstore/pull/69#issuecomment-854693360
* But mostly add tests, and ensure that updates actually work

https://github.com/sigstore/cosign/issues/86


lukehinds:
How do you see this working with https://docs.google.com/document/d/12BBOXMhA99J3KzNaotT1J4OTSDlNnK110_9FZkPv-A0/

I have started to put a WIP here: https://github.com/sigstore/cosign/pull/366

I am wondering if I should wait for this work to complete first and in case we are driving at the same thing / going to conflict.

Do you have a design doc or proposal so I can see how this will ultimately look and function?

dlorenc:
This one came from this discussion and doc: https://github.com/sigstore/cosign/issues/86
https://docs.google.com/document/d/1oyMQ-a0Uwyl9Pew7ISYUdKFfnqEp-qfk1psVFdd-o8Y/edit

lukehinds:
> This one came from this discussion and doc: #86
> https://docs.google.com/document/d/1oyMQ-a0Uwyl9Pew7ISYUdKFfnqEp-qfk1psVFdd-o8Y/edit

Gotcha, it makes sense we work together on this then.

From the doc:

> cosign tuf init-repo gcr.io/dlorenc-vmtest2 --threshold 2 --roots asraa@google.com, dlorenc@google.com, lhinds@redhat.com --signer mnm678@gmail.com --timestamper serviceaccount

The above is what the other PR does (minus the timestamper), although it also fixes it to a registry

priyawadhwa:
Hey @asraa , is this still a WIP or was it moved to the sigstore-root repo??

PR #369, opened 2021-06-14
add flags to specify slot, PIN, and touch policies for security keys

ddz:
This permits flexibility to use different PIN and touch policies
for container signing. Key generation will use the default PIN
and touch policies for the specified slot unless additional flags
are set to attempt to override the default. The security key is
not guaranteed to honor a particular PIN and touch policy for the
specified slot.

Signed-off-by: Dino A. Dai Zovi <ddz@theta44.org>

### DIFF
--- a/cmd/cosign/cli/pivcli/commands.go
+++ b/cmd/cosign/cli/pivcli/commands.go
@@ -23,6 +23,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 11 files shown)


dlorenc:
Nice! Just pulled your branch and ran the yubikey tests that don't work in CI:

```shell
$ go test ./test -tags=resetyubikey,e2e,pivkey -count=1
ok      github.com/sigstore/cosign/test 19.086s
```

dlorenc (approved):

ddz:
Thank you for the quick response and merge!

PR #371, opened 2021-06-14
verify the SET when getting a log entry

asraa:

I noticed that in non-bundle, we are still using the integratedTime to check expiration without verifying rekor's signature of the signedEntryTimestamp

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -135,6 +135,16 @@ func VerifyTLogEntry(rekorClient *client.Rekor, uuid string) (*models.LogEntryAn
 	if err := v.VerifyInclusionProof(*e.Verification.InclusionProof.LogIndex, *e.Verification.InclusionProof.TreeSize, hashes, rootHash, leafHash); err != nil {
 		return nil, errors.Wrap(err, "verifying inclusion proof")
 	}
+
+	// Verify rekor's signature over the SET.
+	rekorPubKey, err := PemToECDSAKey([]byte(rekorPub))
+	if err != nil {
+		return nil, errors.Wrap(err, "rekor public key pem to ecdsa")
+	}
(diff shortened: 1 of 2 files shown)


pkg/cosign/verify.go
```diff
@@ -135,6 +135,16 @@ func VerifyTLogEntry(rekorClient *client.Rekor, uuid string) (*models.LogEntryAn
 	if err := v.VerifyInclusionProof(*e.Verification.InclusionProof.LogIndex, *e.Verification.InclusionProof.TreeSize, hashes, rootHash, leafHash); err != nil {
 		return nil, errors.Wrap(err, "verifying inclusion proof")
 	}
+
+	// Verify rekor's signature over the SET.
+	rekorPubKey, err := PemToECDSAKey([]byte(rekorPub))
```
asraa, line 140: should i just get this from rekorClient instead of using the offline one?

priyawadhwa: personally i think this is fine

priyawadhwa (approved):

pkg/cosign/verify.go
asraa: agreed! i actually might change it over so we don't have to use the prod instance in the e2e test (which is why it's failing)

asraa:
Debugging e2e test. Sees to work fine with a prod rekor instance, but not with the test/local one. Working on it!

asraa:
OK it fails because rekor doesn't add a signature on on the signed entry timestamp when you get an existing entry. i don't know why we don't maintain the signature?

asraa:
Depends on the linked PR, will update when ready for review

asraa:
OK! PR is merged, this is fixed now!

dlorenc (approved):

PR #373, opened 2021-06-16
Added transparency log address to warning

zuBux:
### What issue does this PR address?

When pushing a signature to rekor originating from a private registry, the user gets the following warning message:

```
"warning: uploading to the public transparency log for a private image, please confirm [Y/N]: "
```

This message appears **even when you override the default public transparency log** with the `REKOR_SERVER` env. variable, which is confusing since the user is not sure which transparency log they are about to push to. 

### How does this PR address the issue?
In order to avoid this confusion, there were two potential avenues:

a) Override this check if `TlogServer() != rekorServer`
b) Display the transparency log address in the warning message

To avoid accidental pushes I decided to go with the b). Happy to switch to a) if you feel it's better. 

Signed-off-by: Paris Zoumpouloglou <pariszoump@gmail.com>

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -246,7 +246,7 @@ func SignCmd(ctx context.Context, so SignOpts,
 	uploadTLog := EnableExperimental()
 	if uploadTLog && !force {
 		if _, err := remote.Get(ref); err != nil {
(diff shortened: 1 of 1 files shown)


dlorenc:
Thank you! This is great!

dlorenc (approved):

PR #387, opened 2021-06-24
SBOM support in "cosign attach" and "cosign download sbom".

dlorenc:

### DIFF
--- a/cmd/cosign/cli/attach/sbom.go
+++ b/cmd/cosign/cli/attach/sbom.go
@@ -1,0 +1,112 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


cmd/cosign/cli/attach/sbom.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return err
+	}
+
+	m, err := img.Manifest()
+	if err != nil {
+		return err
+	}
+	// TODO: Fix me.
+	m.Config.MediaType = "application/vnd.sbom.config.v1+json"
```
dlorenc, line 85: @jonjohnsonjr what do i put here

jonjohnsonjr: Is there any useful configuration you'd want to put in there?

dlorenc: No. I could eventually see the need for multiple SBOMs (one for the overall image, then some for sub-parts), but that would get tackled via annotations on each layer to indicate which portion of the image the SBOM is intended to cover.

cmd/cosign/cli/attach/sbom.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	m, err := img.Manifest()
+	if err != nil {
+		return err
+	}
+	m.Config.MediaType = "text/spdx"
```
samj1912, line 83: Is it possible to support both cyclonedx and spdx here?

dlorenc: Yup! added

dlorenc:
Dropping the WIP, this is ready to go!

priyawadhwa (approved):

PR #398, opened 2021-06-30
signing&verifying container images based on Kubernetes Secrets

### LINKED ISSUE #347: Support signing and verification with cosign keys stored in k8s secrets
We can implement this once https://github.com/sigstore/cosign/pull/345 is merged

developer-guy:
Closes #347

Screen Shot 2021-07-01 at 00 20 54

Known issues about the implementation:
* We duplicated the Kubernetes Client logics (reducing both responsibilities to a single function?)
* It's a bit hard to distinguish `namespace/name` compared to other providers (adding a new -k8s flag for signing and verifying)
* Ref priority refactoring? (File -> KMS -> K8s) - we're checking K8s `namespace/name` in first-order ever since a file exists in that path
* We bring a brand-new `loadPublicKey` function to `cli/keys.go` due to `import cycle now allowed` problem

* We haven't been checked the values of the data map of the secret whether is nil or not nil.

* Use public key as a file
```bash 
$ kubectl get secrets my-secret -ojson | jq -r '.data["cosign.pub"]' | base64 -D > /tmp/cosign.pub
$ ./cosign verify -key /tmp/cosign.pub gcr.io/$(gcloud config get-value project)/hello-world:cosign --> #OK
```

* Use public key as a K8s Secret
 ```bash
$ ./cosign verify -key tmp/cosign.pub2 gcr.io/$(gcloud config get-value project)/hello-world:cosign
error: loading public key: checking if secret exists: secrets "cosign.pub2" not found
```
Screen Shot 2021-07-06 at 09 20 14

Screen Shot 2021-07-06 at 00 03 56

### DIFF
--- a/pkg/cosign/kubernetes/secret.go
+++ b/pkg/cosign/kubernetes/secret.go
@@ -22,12 +22,32 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/sigstore/cosign/pkg/cosign"
 	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/sigstore/cosign/pkg/cosign"
 )
+
+const KeyReference = "k8s://"
+
+func GetKeyPairSecret(ctx context.Context, k8sRef string) (*v1.Secret, error) {
+	namespace, name, err := parseRef(k8sRef)
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := Client()
+	if err != nil {
+		return nil, errors.Wrap(err, "new for config")
+	}
+
+	var s *v1.Secret
+	if s, err = client.CoreV1().Secrets(namespace).Get(ctx, name, metav1.GetOptions{}); err != nil {
+		return nil, errors.Wrap(err, "checking if secret exists")
+	}
+
+	return s, nil
+}
 
 func KeyPairSecret(ctx context.Context, k8sRef string, pf cosign.PassFunc) error {
 	namespace, name, err := parseRef(k8sRef)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by dlorenc** [comment]
This looks awesome so far! What do you think about using a "-key k8s://namespace/name" prefix or something like that to help with the signer type selection logic?

**Comment by developer-guy** [comment]
> This looks awesome so far! What do you think about using a "-key k8s://namespace/name" prefix or something like that to help with the signer type selection logic?

Thank you @dlorenc for giving this idea, we thought that too, to do so we should move this logic to `sigstore` project as a new provider called `k8s` maybe, wdyt?
cc: @Dentrax

**Comment by dlorenc** [comment]
That sounds great!

**Comment by Dentrax** [comment]
Exactly! Should we add a brand-new _kubernetes://_ provider into kms [init()](https://github.com/sigstore/sigstore/blob/a2db3ecdb6557e2a0d7c5691537210320a667435/pkg/kms/kms.go#L29) function? By doing so, we are able to clean [kubernetes](https://github.com/sigstore/cosign/tree/main/pkg/cosign/kubernetes) package from cosign into sigstore, I guess. 🤔

**Comment by dlorenc** [comment]
I think that would make sense - cosign would still need to keep the key generation code though right?

**Comment by developer-guy** [comment]
> I think that would make sense - cosign would still need to keep the key generation code, though, right?

The most suitable place for the Kubernetes provider is in the cosign project because of the key generation code. Because we have to generate key pairs before creating the actual Kubernetes secret, thus that logic belongs to cosign. Therefore, we have to keep that Kubernetes provider logic within the `cosign` project.

WDYT @Dentrax @dlorenc?

**Comment by dlorenc** [comment]
Good point. That makes sense to me. We can always move it around later if there's demand for it!

**Comment by developer-guy** [comment]
> Good point. That makes sense to me. We can always move it around later if there's demand for it!

@dlorenc, I added new changes to add the `k8s://` thing and update the description of the PR with some of the working examples screenshots, can you please take a look at it when time permits?

cc: @Dentrax

File: pkg/cosign/kubernetes/secret.go [hunk 1]
```diff
@@ -19,16 +19,39 @@ import (
...
 	"github.com/pkg/errors"
+	"github.com/sigstore/cosign/pkg/cosign"
 	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/sigstore/cosign/pkg/cosign"
 )
 
+var referenceRegex = regexp.MustCompile(`^k8s:\/\/[a-zA-Z0-9\-]+(\/{1}[a-zA-Z0-9\-]+)?$`)
```
dlorenc, line 32: I prefer to avoid regexes for stuff like this if possible... Is there anyway to parse this like a uri? or strip the prefix and then split on a /?

developer-guy: you are absolutely right, thank you so much for this suggestion, I updated the code according to your feedback.

dlorenc (approved):

dlorenc:
This is awesome!

PR #408, opened 2021-07-05
Make .sha256 file compatible with sha256sum

rgreinho:
Updates the output of the CI step generating the sha256 checksum in
order to make it compatible with the `sha256sum` tool.

With this update, the checksum can be verified using:

```
$ sha256sum -c cosign-darwin-amd64.sha256
cosign-darwin-amd64: OK
```

### DIFF
--- a/.github/workflows/cross.yaml
+++ b/.github/workflows/cross.yaml
@@ -38,7 +38,8 @@ jobs:
       - name: Print Info
         shell: pwsh
         run: |
-          $(Get-FileHash -Path ./${{matrix.TARGET}}).Hash | Tee-Object -Path ${{matrix.TARGET}}.sha256
+          $hash=Get-FileHash -Path ./${{matrix.TARGET}}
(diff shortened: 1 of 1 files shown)


dlorenc:
This looks fine to me but I admittedly know very little about Windows. @pearcec - any chance you can take a look?

cpanato:
humm, how do we generate this for windows during the release?

pearcec:
> This looks fine to me but I admittedly know very little about Windows. @pearcec - any chance you can take a look?

Code is fine.  Might break some people downstream since the file format is changing.

rgreinho:
@pearcec If changing the format is a problem, maybe we could describe the process to verify the SHA in the documentation then? However after a bit of searching, I found out that Windows user may have to copy/paste the SHA to a verification tool anyway (https://www.maketecheasier.com/verify-md5-sha-1-sha-256-checksum-windows10/), therefore the file format does not matter for them (I think 🤔).

dlorenc:
If the code works this is fine with me.

dlorenc (approved):

dlorenc:
Thanks!

PR #433, opened 2021-07-13
Fix/verify dockerfile parser

### LINKED ISSUE #430: Bug in verify-dockerfile
**Description**

It fails on the dockerfile we have in this repo:

```shell
$ ./cosign verify-dockerfile -key cosign.pub Dockerfile
Extracted image(s): build, gcr.io/distroless/base:debug
error: GET https://index.docker.io/v2/library/build/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/build Type:repository]]
```

Dentrax:
Fixes #430 

Parser now replace the `as` keyword to `AS` before moving into the splitting function. 

> _`AS` is the one of the special keywords that we allowed the write-in lowercase._

### DIFF
--- a/cmd/cosign/cli/verify_dockerfile.go
+++ b/cmd/cosign/cli/verify_dockerfile.go
@@ -122,10 +122,15 @@ func getImagesFromDockerfile(dockerfile io.Reader) ([]string, error) {
 }
 
 func getImageFromLine(line string) string {
-	line = strings.TrimPrefix(line, "FROM")     // Remove "FROM" prefix
-	line = os.ExpandEnv(line)                   // Substitute templated vars
-	line = strings.ReplaceAll(line, "as", "AS") // replace lower "as" with "AS"
-	line = strings.Split(line, " AS ")[0]       // Remove the "AS" portion of line
+	line = strings.TrimPrefix(line, "FROM") // Remove "FROM" prefix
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/verify_dockerfile.go
```diff
@@ -122,9 +122,10 @@ func getImagesFromDockerfile(dockerfile io.Reader) ([]string, error) {
 }
 
 func getImageFromLine(line string) string {
-	line = strings.TrimPrefix(line, "FROM") // Remove "FROM" prefix
-	line = os.ExpandEnv(line)               // Substitute templated vars
-	line = strings.Split(line, " AS ")[0]   // Remove the "AS" portion of line
+	line = strings.TrimPrefix(line, "FROM")     // Remove "FROM" prefix
+	line = os.ExpandEnv(line)                   // Substitute templated vars
+	line = strings.ReplaceAll(line, "as", "AS") // replace lower "as" with "AS"
```
dekkagaijin, line 127: this would substitute all instances of "as". I'm not sure if image names are case sensitive, but URL paths can be

dekkagaijin:
ping
I forgot that I could offer suggestions in review and pushed a commit to your branch, which broke presubmit

dekkagaijin (approved):

PR #436, opened 2021-07-14
make base image an arg, use distroless/static for releases

dekkagaijin:


dlorenc (approved):

cpanato:
@dekkagaijin @dlorenc `ko` add a feature to pass more goflags, is it something we want to push back or we will stay with dockerfile?

PR #439, opened 2021-07-16
SBOM specification!

dlorenc:

### DIFF
--- a/specs/SBOM_SPEC.md
+++ b/specs/SBOM_SPEC.md
@@ -1,0 +1,142 @@
+# Cosign SBOM Specifications
+
+This document aims to describe how `cosign` attaches SBOM (Software Bill of Materials) documents to containers.
+
+The goal is to specify the behavior well enough to promote other implementations and enable interoperability.
+SBOMs attached with `cosign` should be retrievable in other tools, and vice-versa.
+
+This document focuses on the layout of an SBOM within an [OCI Image Manifest V1](https://github.com/opencontainers/image-spec/blob/master/manifest.md) object.
+
+This document does not prescribe any specific SBOM format.
+Multiple formats can be used, and the `mediaType` property should describe the format of a particular SBOM document.
+
+Multiple SBOMs may be "attached" to one image.
+Each SBOM may refer to the entire image, or to a specific part of that image.
+Exactly what the SBOM refers to is called the "scope" of that SBOM.
+
+SBOMs stored in an OCI registry are generally assumed to refer to other objects stored in that same registry.
+This document does not specify how these "links" are created.
+A naming convention or the [in-progress OCI `references` API](https://github.com/opencontainers/image-spec/issues/827) are viable options.
+
+This document does not specify how clients should behave When multiple SBOMs are present for an image.
+Clients may list all the SBOMs, or may provide tooling to filter based on SBOM type or scope.
+
+## Overall Layout
+
+An SBOM object is represented an [OCI Image Manifest V1](https://github.com/opencontainers/image-spec/blob/master/manifest.md).
+
+Each individual SBOM is represented as a `layer`, using a standard `descriptor`.
+The `layers` list is ordered, but no order is assumed or important for the SBOM documents.
+
+Here is an example manifest containing one SBOM, in the [SPDX](https://spdx.org) format:
+
+```json
+{
+  "schemaVersion": 2,
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 233,
+    "digest": "sha256:83bd5fb5b39f65f28e50a86d48fa79c07880befc292d92eebdc18531054b070c"
+  },
+  "layers": [
+    {
+      "mediaType": "text/spdx",
+      "size": 246,
+      "digest": "sha256:ed3ad03d3b87843b5419d7dce9d50a3e0f45554b2ba93bf378611cae6b450cff",
+    }
+  ]
+}
+```
+
+Multiple SBOMs may be attached, using multiple formats.
+This example shows two SBOMs, one in the SPDX format and one in the CycloneDX format:
+
+```json
+{
+  "schemaVersion": 2,
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 233,
+    "digest": "sha256:83bd5fb5b39f65f28e50a86d48fa79c07880befc292d92eebdc18531054b070c"
+  },
+  "layers": [
+    {
+      "mediaType": "text/spdx",
+      "size": 246,
+      "digest": "sha256:ed3ad03d3b87843b5419d7dce9d50a3e0f45554b2ba93bf378611cae6b450cff",
+    },
+    {
+      "mediaType": "application/vnd.cyclonedx",
+      "size": 462,
+      "digest": "sha256:e0851a4aa13657fc8dcd01e0e5e08cb817123ccb82e2c604b34f9ec9c1755e3f",
+    }
+  ]
+}
+```
+
+Each individual SBOM may also be "scoped" to a part of the object it refers to.
+This is indicated via an annotation on the descriptor.
+In this example, the SBOM only refers to a single layer:
+
+```json
+{
+  "schemaVersion": 2,
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 233,
+    "digest": "sha256:83bd5fb5b39f65f28e50a86d48fa79c07880befc292d92eebdc18531054b070c"
+  },
+  "layers": [
+    {
+      "mediaType": "text/spdx",
+      "size": 246,
+      "digest": "sha256:ed3ad03d3b87843b5419d7dce9d50a3e0f45554b2ba93bf378611cae6b450cff",
+      "annotatons": {
+        "dev.sigstore.sbom.scope": "layer=sha256:a69d803ab2179a570eda27135989ee850de53bbd98efc8f0284f13700a94149f",
+      }
+    }
+  ]
+}
+```
+
+## MediaTypes
+
+The two main SBOM formats in use are [SPDX](https://spdx.org) and [CycloneDX](https://cyclonedx.org/).
+The `mediaTypes` for these should be indicated in the `descriptor` for each `layer`.
+
+The `mediaTypes` are:
+
+* `application/vnd.cyclonedx`
+* `text/spdx`
+
+These `mediaTypes` can contain format-specific suffixes as well. For example:
+
+* `application/vnd.cyclonedx+xml`
+* `application/vnd.cyclonedx+json`
+* `text/spdx+xml`
+* `text/spdx+json`
+
+## Scopes
+
+SBOMs may refer to an entire object, or to a specific part of that object.
+This is called the `scope` of the SBOM.
+
+The `scope` should be indicated via an annotation on the `Descriptor`, with the key of `dev.sigstore.sbom.scope`.
+
+A descriptor with no scope is assumed to refer to the entire object.
+This is the same as the scope of `all`.
+
+Well-known scopes include:
+
+* `all`: the SBOM refers to the entire object.
+* `layer=sha256:$DIGEST`: the SBOM refers to the layer with the appropriate digest.
+* `path=<foo>`: the SBOM refers to file at path `foo` in the flattened image.
+
+Scopes may be repeated, and are separated by the `,` character.
+This scope refers to two layers: `layer=sha256:$DIGEST,layer=sha256:$OTHERDIGEST`
+
+## Relationship
+
+While SBOMs typically relate directly to the contents of the object they refer to, in certain circumstances they may instead relate to the object indirectly.
+One exmaple here is that the SBOM could describe the environment the object was built in, rather than the contents of the object itself.
+This type of relationship will be tracked by this spec somehow, but we're not sure exactly how yet.


specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,101 @@
...
+      "size": 462,
+      "digest": "sha256:e0851a4aa13657fc8dcd01e0e5e08cb817123ccb82e2c604b34f9ec9c1755e3f",
+    }
+  ]
+}
+```
+
+## MediaTypes
+
+The two main SBOM formats in use are SPDX and CyloneDX.
```
imjasonh, line 76:
```suggestion
The two main SBOM formats in use are [SPDX](https://spdx.org) and [CycloneDX](https://cyclonedx.org/).
```

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,101 @@
...
+
+* `application/vnd.cyclonedx`
+* `text/spdx`
+
+## Scopes
+
+SBOMs may refer to an entire object, or to a specific part of that object.
+This is called the `scope` of the SBOM.
+
+The `scope` should be indicated via an annotation on the `Descriptor`.
```
imjasonh, line 89: Just for clarity, can you include an example like the one above where a layer specifies the scope?

imjasonh: > Keys SHOULD be named using a reverse domain notation - e.g. com.example.myKey.

(from https://github.com/opencontainers/image-spec/blob/main/annotations.md)

Should this be something like `dev.sigstore.sbom.scope` then?

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,101 @@
...
+This document does not prescribe any specific SBOM format.
+Multiple formats can be used, and the `mediaType` property should describe the format of a particular SBOM document.
+
+Multiple SBOMs may be "attached" to one image.
+Each SBOM may refer to the entire image, or to a specific part of that image.
+Exactly what the SBOM refers to is called the "scope" of that SBOM.
+
+SBOMs stored in an OCI registry are generally assumed to refer to other objects stored in that same registry.
+This document does not specify how these "links" are created.
+A naming-convention or the in-progress OCI `references` API are viable options.
```
imjasonh, line 19:
```suggestion
A naming-convention or the [in-progress OCI `references` API](https://github.com/opencontainers/image-spec/issues/827) are viable options.
```

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,101 @@
...
+
+A descriptor with no scope is assumed to refer to the entire object.
+This is the same as the scope of `all`.
+
+Well-known scopes include:
+
+* `all`: the SBOM refers to the entire object.
+* `layers=N`: the SBOM refers to the Nth layer.
+* `layers=N..M`: the SBOM refers to the layer range from N to M, exclusive of M.
+* `path=<foo>`: the SBOM refers to file at path `foo` in the flattened image.
```
imjasonh, line 101: Should this spec cover what implementations should do when confronted with two overlapping SBOMs?

E.g., if there's an SBOM for `all` and for layer `layers=N`, is the full bill of materials the union of the two? Does `layers=N` supercede the `all` in some way? Is that configuration to be considered invalid, and clients should error out?

specs/SBOM_SPEC.md
dlorenc: Thanks!

specs/SBOM_SPEC.md
dlorenc: Fixed and clarified!

specs/SBOM_SPEC.md
dlorenc: Added this to the top. IMO it should be up to clients. The SBOM landscape is too ambiguous right now to prescribe what should happen here for multiple SBOMs or overlapping ones.

dlorenc:
I'm not sure if this diagram helps or hurts, leaving it here for now but it could get merged into the spec:
https://gist.github.com/dlorenc/624a49e68cee0e4196663d0e54efc220

graphviz

justaugustus:
cc: @puerco

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,102 @@
...
+  ]
+}
+```
+
+## MediaTypes
+
+The two main SBOM formats in use are [SPDX](https://spdx.org) and [CycloneDX](https://cyclonedx.org/).
+The `mediaTypes` for these should be indicated in the `descriptor` for each `layer`.
+
+The `mediaTypes` are:
```
garethr, line 82: Both SPDX and CycloneDX can be representing in different formats, and different versions. Should the format be included as a subtype here?

eg.

* `application/vnd.cyclonedx+json`
* `application/vnd.cyclonedx+xml`

This would help clients who otherwise have to determine the format based on parsing along.

dlorenc: Probably! I couldn't find anything other than text/spdx registered with IANA for that format. This makes sense for cyclonedx though.

Any ideas what would make sense for SPDX?

garethr: These are in addition to the registered types, technically [Structured Syntax Suffixes](https://datatracker.ietf.org/doc/html/draft-ietf-appsawg-media-type-suffix-regs-02#section-3.1), so should be applicable to both. So I think:

* `text/spdx+json`
* `text/spdx+xml`

@stevespringett @coderpatros would be good to check for CycloneDX, and @kestewart for SPDX I think too.

stevespringett:
Confirming the CycloneDX media types that @garethr mentioned.

* [application/vnd.cyclonedx+xml](https://www.iana.org/assignments/media-types/application/vnd.cyclonedx+xml)
* [application/vnd.cyclonedx+json](https://www.iana.org/assignments/media-types/application/vnd.cyclonedx+json)

Both are officially registered with IANA.

specs/SBOM_SPEC.md
stevespringett: Confirming the CycloneDX media types that @garethr mentioned.

* [application/vnd.cyclonedx+xml](https://www.iana.org/assignments/media-types/application/vnd.cyclonedx+xml)
* [application/vnd.cyclonedx+json](https://www.iana.org/assignments/media-types/application/vnd.cyclonedx+json)

Both are officially registered with IANA.

dlorenc: Thanks, added!

coderpatros: @garethr Do you know if structured syntax suffixes are valid offical media types without being specified in the relevant RFC and media type registration? Text formats like SPDX tag/value have very different security considerations to something like XML.

dlorenc: @jonjohnsonjr do you know?

jonjohnsonjr: > Do you know if structured syntax suffixes are valid offical media types without being specified in the relevant RFC and media type registration?

I'm not sure I understand the question. Are you asking if you can make up suffixes that aren't registered? Or if you can use registered suffixes with arbitrary unregistered media types?

coderpatros: @jonjohnsonjr whether or not you can use a registered suffix with a registered media type without registering the combination.

jonjohnsonjr: https://datatracker.ietf.org/doc/html/rfc6839#section-2

>    Each of the structured syntax suffixes defined in this document is
>    appropriate for use when the media type identifies the semantics of
>    the protocol payload.  That is, knowing the semantics of the specific
>    media type provides for more specific processing of the content than
>    that afforded by generic processing of the underlying representation.
> 
>    At the same time, using the suffix allows receivers of the media
>    types to do generic processing of the underlying representation in
>    cases where
> 
>       they do not need to perform special handling of the particular
>       semantics of the exact media type, and
> 
>       there is no special knowledge needed by such a generic processor
>       in order to parse that underlying representation other than what
>       would be needed to parse any example of that underlying
>       representation.

Given that it's intended to be useful when received by a "generic processor", I would imagine it's fine to append a suffix to something without registering it, assuming that the content conforms to the definition of that suffix.

One example I have in mind is when displaying something -- it's useful to know if it's JSON or XML in order to correctly pretty-print some content, even if you don't know how to interpret the fields.

Also, with the vnd tree, you don't need to register a mediaType _anyway_, so that's at least one example of the suffix being valid even if the combination isn't registered.

https://datatracker.ietf.org/doc/html/rfc6838#section-3.2

>   While public exposure and review of media types to be registered in
>    the vendor tree are not required, using the media-types@iana.org
>    mailing list for review is encouraged, to improve the quality of
>    those specifications.  Registrations in the vendor tree may be
>    submitted directly to the IANA, where they will undergo Expert Review
>    [RFC5226] prior to approval.

puerco: From what I've seen you have to register the subtype with the media type to make it official, for example, `svg+xml`:
https://www.iana.org/assignments/media-types/image/svg+xml

puerco: Oh interesting @jonjohnsonjr . Three things come to mind:

1. I wonder if RDF would fall under `+xml` under this clause: "_in cases where they do not need to perform special handling of the particular semantics_"
2. `+yaml` is not defined in RFC 6839, would that require a specific registration?
3. I guess the real corner case is SPDX requiring an additional (new) domain-specific media type like, `+tagvalue`.

jonjohnsonjr: YAML isn't registered as a media type or suffix, but probably should be. There's a thread here about doing so, but it looks like it never actually resolved: https://mailarchive.ietf.org/arch/msg/media-types/e9ZNC0hDXKXeFlAVRWxLCCaG9GI/

I would (boldly) suggest that it would be reasonable to use +yaml even if it's not registered, assuming the ecosystem understands what it means, but it would certainly be better if it were actually registered. I'm not sure who would be the best person to follow up with re: yaml registration.

The actual registries aren't the RFCs, but maintained here:
* https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml
* https://www.iana.org/assignments/media-types/media-types.xhtml

I'm not familiar enough to speak to the tagvalue thing -- is that specific to spdx?

puerco: Exactly, while the doc can be expressed in json, yaml, rdf, etc tag value is like its own format: https://spdx.github.io/spdx-spec/appendix-IX-file-tags/#format

coderpatros: Regarding using unregistered suffixes, like `+yaml`, this is from RFC6838 (noting a conflict is unlikely)

> "+suffix" constructs for as-yet unregistered structured syntaxes SHOULD NOT be used, given the possibility of conflicts with future suffix definitions.

And if it is not an official registered media type it should use the `x.` prefix to indicate it is in the unregistered tree as per RFC6838.

We did this for CycloneDX Protobuf format, for the above reasons, `application/x.vnd.cyclonedx+protobuf`.

Regardless, I don't think this is the place to be determining the SPDX media types. Just my 2c :)

jonjohnsonjr: > And if it is not an official registered media type it should use the x. prefix to indicate it is in the unregistered tree as per RFC6838.

My reading is that this should basically never be used for public things:

https://datatracker.ietf.org/doc/html/rfc6838#section-3.4

>    Subtype names with "x." as the first facet may be used for types
>    intended exclusively for use in private, local environments.  Types
>    in this tree cannot be registered and are intended for use only with
>    the active agreement of the parties exchanging them.
> 
>    However, with the simplified registration procedures described above
>    for vendor and personal trees, it should rarely, if ever, be
>    necessary to use unregistered types.  Therefore, use of types in the
>    "x." tree is strongly discouraged.

Agreed that we SHOULD NOT use +yaml, but also we SHOULD register it, and that's not happening either :) 

> Regardless, I don't think this is the place to be determining the SPDX media types. Just my 2c :)

+1 -- I'm happy to spread my poor interpretation having skimmed a could RFCs, but this should be something the SPDX folks are responsible for.

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,109 @@
...
+* `text/spdx`
+
+These `mediaTypes` can contain format-specific suffixes as well. For example:
+
+* application/vnd.cyclonedx+xml
+* application/vnd.cyclonedx+json
+* text/spdx+xml
+* text/spdx+json
+
+## Scopes
```
samj1912, line 94: Could you add an example of scopes in the above manifest?

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,109 @@
...
+
+SBOMs may refer to an entire object, or to a specific part of that object.
+This is called the `scope` of the SBOM.
+
+The `scope` should be indicated via an annotation on the `Descriptor`, with the key of `dev.sigstore.sbom.scope`.
```
samj1912, line 99: Should `scope` be an array? If so, can we have an example of a delimiter separated `scope` array?

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,109 @@
...
+
+A descriptor with no scope is assumed to refer to the entire object.
+This is the same as the scope of `all`.
+
+Well-known scopes include:
```
samj1912, line 104: Could we also add a scope called `type` which is either `build` or `launch`/`run` where the `type` indicates if the `bom` describes the components used to `build ` specific layer or path and if it is actually present in the final output image.

For eg. if you have a binary called `my-go-binary`, it could have a `launch` `bom` with `scope`=`path=/app/my-go-binary;type=launch` which describes the version of `my-go-binary`, its licenses etc.

But you could also have another `bom` with `scope`=`path=app/my-go-binary;type=launch` which describe the go compiler version etc. used to compile that binary?

Is that a valid enough use case?

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,109 @@
...
+
+```json
+{
+  "schemaVersion": 2,
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 233,
+    "digest": "sha256:83bd5fb5b39f65f28e50a86d48fa79c07880befc292d92eebdc18531054b070c"
+  },
+  "layers": [
```
samj1912, line 62: +1 on having multiple boms of different formats that can point to different layers -- would be super helpful on the buildpacks side for https://github.com/buildpacks/rfcs/pull/166

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,109 @@
...
+
+The `scope` should be indicated via an annotation on the `Descriptor`, with the key of `dev.sigstore.sbom.scope`.
+
+A descriptor with no scope is assumed to refer to the entire object.
+This is the same as the scope of `all`.
+
+Well-known scopes include:
+
+* `all`: the SBOM refers to the entire object.
+* `layers=N`: the SBOM refers to the Nth layer.
```
samj1912, line 107: Should this be layer digests instead of using the layer number?

`layers=sha1,sha2,sha3`
or `layers=shaStart..shaEnd`?

idk if that helps or makes sense

if not, it is also worth supporting

`layers=n1,n2,n3` ?

specs/SBOM_SPEC.md
dlorenc: That's much cleaner.

specs/SBOM_SPEC.md
dlorenc: Yeah I like this, but scope might not be the right place for it. What about something like `relationship` as another annotation?

specs/SBOM_SPEC.md
dlorenc: +!

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,136 @@
...
+A descriptor with no scope is assumed to refer to the entire object.
+This is the same as the scope of `all`.
+
+Well-known scopes include:
+
+* `all`: the SBOM refers to the entire object.
+* `layer=sha256:$DIGEST`: the SBOM refers to the Nth layer.
+* `path=<foo>`: the SBOM refers to file at path `foo` in the flattened image.
+
+Sopes may be repeated, and are separated by the `,` character.
```
samj1912, line 135:
```suggestion
Scopes may be repeated, and are separated by the `,` character.
```

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,136 @@
...
+
+The `scope` should be indicated via an annotation on the `Descriptor`, with the key of `dev.sigstore.sbom.scope`.
+
+A descriptor with no scope is assumed to refer to the entire object.
+This is the same as the scope of `all`.
+
+Well-known scopes include:
+
+* `all`: the SBOM refers to the entire object.
+* `layer=sha256:$DIGEST`: the SBOM refers to the Nth layer.
```
samj1912, line 132:
```suggestion
* `layer=sha256:$DIGEST`: the SBOM refers to the layer with the appropriate digest.
```

specs/SBOM_SPEC.md
samj1912: I like that!

specs/SBOM_SPEC.md
dlorenc: Thanks!

specs/SBOM_SPEC.md
dlorenc: I'll let this one bake a bit more and track it as a TODO.

specs/SBOM_SPEC.md
```diff
@@ -0,0 +1,142 @@
...
+
+* `application/vnd.cyclonedx`
+* `text/spdx`
+
+These `mediaTypes` can contain format-specific suffixes as well. For example:
+
+* application/vnd.cyclonedx+xml
+* application/vnd.cyclonedx+json
+* text/spdx+xml
+* text/spdx+json
```
imjasonh, line 117:
```suggestion
* `application/vnd.cyclonedx+xml`
* `application/vnd.cyclonedx+json`
* `text/spdx+xml`
* `text/spdx+json`
```

dlorenc: Thanks!

dekkagaijin (approved):

PR #449, opened 2021-07-17
pivkey: Add support for RSA keys

jalseth:
In some environments, RSA keys are used to ensure compatibility with systems
that do not support elliptic curve cryptography. Since the PIV standard
supports both RSA and ECDSA it makes sense to support both in `cosign`.

Signed-off-by: James Alseth <james@jalseth.me>


jalseth:
Closing this in favor of the refactor in #451

PR #450, opened 2021-07-17
pivkey: Change default slot to Signature (9c)

jalseth:
Previously, the default slot was Authentication (9a) which is intended to be used
for authentication with SSH or similar systems. In the context of cosign, it makes
more sense to default to the Signature slot which is intended for use with message
and code signing.

Signed-off-by: James Alseth <james@jalseth.me>


dlorenc:
Thanks! We still don't have these tests running in CI so I'll pull in your branch and run them myself later today to make sure they don't break.

dlorenc (approved):

dlorenc:
Thanks! This Looks great.

PR #451, opened 2021-07-18
cmd/sign: Add `-cert` flag

jalseth:
The `-cert` flag allows for the certificate to be included in the
Signature object when signing with `-key` or `-sk`. Adding this
certificate ensures that a verifying party does not need to fetch the
certificate from some unknown location before being able to validate the
identity of the actor that created the signature.

---

This also includes a refactor of the `pivkey` library to make it more idiomatic and to resolve an issue where calling `NewPublicKeyProvider` or `NewSignerVerifier` would lock the PIV token until the program had exited, making future calls to the PIV token outside of the signer/verifier flow impossible.


dlorenc (approved):
Nice! Ran tests again locally.

PR #462, opened 2021-07-21
Infra flags for fulcio / rekor / oidc values

lukehinds:
Implements flags for fulco, rekor and OIDC values to allow a user to set their own custom values.

The defaults fall to the public good service

image

image

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -42,6 +42,7 @@ type VerifyCommand struct {
 	Sk          bool
 	Slot        string
 	Output      string
+	RekorURL    string
 	Annotations *map[string]interface{}
 }
 
@@ -50,6 +51,7 @@ func applyVerifyFlags(cmd *VerifyCommand, flagset *flag.FlagSet) {
 	flagset.StringVar(&cmd.KeyRef, "key", "", "path to the public key file, URL, KMS URI or Kubernetes Secret")
 	flagset.BoolVar(&cmd.Sk, "sk", false, "whether to use a hardware security key")
 	flagset.StringVar(&cmd.Slot, "slot", "", "security key slot to use for generated key (default: signature) (authentication|signature|card-authentication|key-management)")
+	flagset.StringVar(&cmd.RekorURL, "rekor-url", "https://rekor.sigstore.dev", "address of rekor STL server")
 	flagset.BoolVar(&cmd.CheckClaims, "check-claims", true, "whether to check the claims found")
 	flagset.StringVar(&cmd.Output, "output", "json", "output the signing image information. Default JSON.")
 
@@ -123,7 +125,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, args []string) (err error) {
 		co.ClaimVerifier = cosign.SimpleClaimVerifier
 	}
 	if EnableExperimental() {
-		co.RekorURL = TlogServer()
+		co.RekorURL = c.RekorURL
 	}
 	keyRef := c.KeyRef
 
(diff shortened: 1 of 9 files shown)


dlorenc:
Looks like you might need to plumb the flag through to the e2e tests to point at localhost

cmd/cosign/cli/verify.go
```diff
@@ -50,6 +51,7 @@ func applyVerifyFlags(cmd *VerifyCommand, flagset *flag.FlagSet) {
 	flagset.StringVar(&cmd.KeyRef, "key", "", "path to the public key file, URL, KMS URI or Kubernetes Secret")
 	flagset.BoolVar(&cmd.Sk, "sk", false, "whether to use a hardware security key")
 	flagset.StringVar(&cmd.Slot, "slot", "", "security key slot to use for generated key (default: signature) (authentication|signature|card-authentication|key-management)")
+	flagset.StringVar(&cmd.RekorURL, "rekor-url", "https://rekor.sigstore.dev", "address of rekor STL server")
```
font, line 54: `s/rekor-url/rekor-server/`

pkg/cosign/fulcio/fulcio.go
```diff
@@ -165,7 +153,7 @@ type Signer struct {
 	*signature.ECDSASignerVerifier
 }
 
-func NewSigner(ctx context.Context, idToken string) (*Signer, error) {
+func NewSigner(ctx context.Context, idToken, oidcIssuer string, oidcClientID string, oidcClientSecret string, fulcioClient string, rekorClient string) (*Signer, error) {
```
font, line 156: You can combine all these argument names and use one `string` argument type e.g. `oidIssuer, oidcClientID, oidcClientSecret, fulcioClient, rekorClient string` to be more Go idiomatic. This comment applies above as well.

font: Any reason not to keep these named `fulcioURL` and `rekorURL` for consistency?

cmd/cosign/cli/verify.go
lukehinds: I honestly prefer url, as it is a url. Are we using rekor-server anywhere else in cosign?

pkg/cosign/fulcio/fulcio.go
lukehinds: > You can combine all these argument names and use one string argument type e.g. oidIssuer, oidcClientID, oidcClientSecret, fulcioClient, rekorClient string to be more Go idiomatic. This comment applies above as well.

Yup, good call.

> Any reason not to keep these named fulcioURL and rekorURL for consistency?

I think I was sticking to the convention already set in the struct, I will have check.

cmd/cosign/cli/verify.go
font: `rekor-server` was used everywhere else in this PR except here, so I assumed that's what you intended here as well.

dekkagaijin: IMHO we should probably `s/rekor-server/rekor-url/`

pkg/cosign/fulcio/fulcio.go
dekkagaijin: I'd be in favor of changing these to "URL" or "Endpoint" to be more descriptive

cmd/cosign/cli/verify.go
font: +1 to using `rekor-url`.

lukehinds: I am bit lost here, there is no current 'rekor-server' vars

https://github.com/sigstore/cosign/search?q=rekor-server

The only rekorServer was a const used in a helper function and this PR removes that

https://github.com/sigstore/cosign/blob/7393e9667083744d13686fbe2ff0d511f5fe5063/cmd/cosign/cli/util.go#L34-L54

image

pkg/cosign/fulcio/fulcio.go
lukehinds: Looking into it we don't need rekorClient there(so removed it) and the fulcioClient is an connection instance

dlorenc (approved):

PR #407, opened 2021-07-05
[WIP] Policy init

lukehinds:
Work in progress, feeding this over in bite sized pieces

Signed-off-by: Luke Hinds <lhinds@redhat.com>

PR #469, opened 2021-07-22
Add a policy-init using TUF metadata and Fulcio signers

asraa:

Hoping to iterate on this. This only initializes the root policy. We need to:
* Add the sign command
* Add ISSUER information to trust?

Replaces/uses code from
- https://github.com/sigstore/cosign/pull/366
- https://github.com/sigstore/cosign/pull/407

Current usage:
`./cosign policy-init --namespace gcr.io/asra-ali/busybox --maintainers asraa@google.com,lukehinds@redhat.com --threshold 2 --out root.json`

```
./cosign policy-init --namespace gcr.io/asra-ali/busybox --maintainers asraa@google.com,lukehinds@redhat.com --threshold 2 --out root.json
Uploading file from [root.json] to [gcr.io/asra-ali/busybox/root.json:latest] with media type [text/plain]
File [root.json] is available directly at [gcr.io/v2/asra-ali/busybox/root.json/blobs/sha256:973a21f41a3884b9f40e70df6bfdd19b04c3ddcbd8935b82a09fd9bdac2ed8c8]
Uploaded image to:
gcr.io/asra-ali/busybox/root.json@sha256:6e6251e7f656c207e265d0629e9cf9d1fb9539b12f3505ceebb90bf16be9befc
```

The output file is (currently unsigned)
```
{
  "signatures": null,
  "signed": {
    "_type": "root",
    "consistent_snapshot": true,
    "expires": "2021-12-29T19:48:52Z",
    "keys": {
      "94e61dba379a148be37639c8bf7ece0087df6dd59733d0a3fd21da7ad1a0ac29": {
        "keyid_hash_algorithms": [
          "sha256",
          "sha512"
        ],
        "keytype": "sigstore-oidc",
        "keyval": {
          "identity": "lukehinds@redhat.com",
          "issuer": ""
        },
        "scheme": "https://fulcio.sigstore.dev"
      },
      "d9e0fe253edc4349d26dcdac691dad02c142be05e959b11577d151551c34ab8c": {
        "keyid_hash_algorithms": [
          "sha256",
          "sha512"
        ],
        "keytype": "sigstore-oidc",
        "keyval": {
          "identity": "asraa@google.com",
          "issuer": ""
        },
        "scheme": "https://fulcio.sigstore.dev"
      }
    },
    "namespace": "gcr.io/asra-ali/busybox",
    "roles": {
      "root": {
        "keyids": [
          "d9e0fe253edc4349d26dcdac691dad02c142be05e959b11577d151551c34ab8c",
          "94e61dba379a148be37639c8bf7ece0087df6dd59733d0a3fd21da7ad1a0ac29"
        ],
        "threshold": 2
      }
    },
    "spec_version": "1.0",
    "version": 1
  }
}

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,7 @@ require (
 	github.com/secure-systems-lab/go-securesystemslib v0.1.0
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/spf13/cobra v1.2.1
+	github.com/tent/canonical-json-go v0.0.0-20130607151641-96e4ba3a7613
 	github.com/urfave/cli v1.22.5 // indirect
 	go.opentelemetry.io/contrib v0.22.0 // indirect
 	go.opentelemetry.io/proto/otlp v0.9.0 // indirect
--- a/cmd/cosign/cli/commands.go
+++ b/cmd/cosign/cli/commands.go
@@ -124,6 +124,7 @@ func New() *cobra.Command {
 
 	// Add sub-commands.
 	addPublicKey(cmd)
+	addPolicyInit(cmd)
 	addGenerate(cmd)
 	addSign(cmd)
 	addSignBlob(cmd)
--- a/cmd/cosign/cli/commands.go
+++ b/cmd/cosign/cli/commands.go
@@ -124,7 +124,7 @@ func New() *cobra.Command {
 
 	// Add sub-commands.
 	addPublicKey(cmd)
-	addPolicyInit(cmd)
+	addPolicy(cmd)
 	addGenerate(cmd)
 	addSign(cmd)
 	addSignBlob(cmd)
--- a/cmd/cosign/cli/options/policy.go
+++ b/cmd/cosign/cli/options/policy.go
@@ -1,0 +1,45 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package options
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// PolicyInitOptions is the top level wrapper for the policy-init command.
+type PolicyInitOptions struct {
+	ImageRef    string
+	Maintainers []string
+	Threshold   int
+	OutFile     string
+}
+
+var _ Interface = (*PolicyInitOptions)(nil)
+
+// AddFlags implements Interface
+func (o *PolicyInitOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.ImageRef, "namespace", "ns",
+		"registry namespace that the root policy belongs to")
+
+	cmd.Flags().StringVar(&o.OutFile, "out", "o",
+		"output policy locally")
+
+	cmd.Flags().IntVar(&o.Threshold, "threshold", 1,
+		"threshold for root policy signers")
+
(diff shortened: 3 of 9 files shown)


lukehinds:
nice, thanks @asraa . I am not to familiar with the go-tuf client, so excuse my greenness. Is there anyway this root policy is fixed to the registry namespace (in the signed envelope), what would stop someone lifting this and putting into a different registry namespace? I was just thinking if we should stash the registry in the `signed` section.

lukehinds:
> nice, thanks @asraa . I am not to familiar with the go-tuf client, so excuse my greenness. Is there anyway this root policy is fixed to the registry namespace (in the signed envelope), what would stop someone lifting this and putting into a different registry namespace? I was just thinking if we should stash the registry in the `signed` section.

one more question, the `cert` is the fulcio cert and we can pull the email from there, how is the `keyid` generated?

asraa:
> one more question, the cert is the fulcio cert and we can pull the email from there? how is the keyid generated?

Oops! The public value is the email, I accidentally hex encoded it. I'll update that and the output. The keyid is the hash over all of the key data (type, public value, scheme). This is to prevent someone a double signing if someone uses a different fulcio server.

> Is there anyway this root policy is fixed to the registry namespace (in the signed envelope), what would stop someone lifting this and putting into a different registry namespace? I was just thinking if we should stash the registry in the signed section.

That indeed is a TODO in the code! It seemed like the targets that are being signed (whose paths are specified) are always relative to the root... The root keys are supposed to be provided out of band, to establish initial trust. I have to give this some more thought.

lukehinds:
Thanks @asraa , happy to chat it over anytime.

go.mod
```diff
@@ -27,3 +27,5 @@ require (
 	k8s.io/apimachinery v0.21.3
 	k8s.io/client-go v0.21.3
 )
+
+replace github.com/theupdateframework/go-tuf => /home/asraa/git/go-tuf
```
lukehinds, line 31: understand why you have this (local fork), could it point to a the online fork instead (I can play around with it then).

asraa: Sure! Just FYI the go-tuf changes need to be pulled in

asraa:
> Thanks @asraa , happy to chat it over anytime.

BTW some thoughts on this: the signing event sent to rekor should include the namespace (verifying that the signing happened in that registry namespace). verifying a policy added to a random registry should fail when verifying the signing event on rekor. WDYT?

lukehinds:
> > Thanks @asraa , happy to chat it over anytime.
> 
> BTW some thoughts on this: the signing event sent to rekor should include the namespace (verifying that the signing happened in that registry namespace). verifying a policy added to a random registry should fail when verifying the signing event on rekor. WDYT?

Yep, I agree. If the namespace is already in signed envelope that would mean it ends up in the log anyway?

asraa:
@lukehinds PR is updated so it at least points to me local fork with the changes in go-tuf that are needed

Still working on:
* Updating the key field to actually display an email
* Adding a namespace to Root signed payload

asraa:
@lukehinds I decided for ease of getting this in, I'll simplify and remove go-tuf dependency. PTAL just does the metadata generation, not the signing.

asraa:
@lukehinds let me know if the CLI I implemented was wrong -- I think we actually want a subcommand for `policy` with commands like `init`, `sign`, etc like

```
cosign policy init ...
cosign policy sign ...
```

lukehinds:
Looks good @asra

Yep , we need a sign operation next. Should we leverage an existing signer / json parse in go-tuf or implement our own?

I just noticed a little bug (from my code I passed over to you). If bombs if there is a whitespace between the maintainers array passed as an arg.

lukehinds (approved):

lukehinds:
It's an improve from me, but i don't normally review a lot in cosign, so happy for another core to review as well @dlorenc @dekkagaijin and others to pitch in.

asraa:
> Should we leverage an existing signer / json parse in go-tuf or implement our own?

I have a Sign method: https://github.com/sigstore/cosign/blob/b2ffe2cc7e3d420eba99b779d573c4e5e2f28343/pkg/cosign/tuf/signer.go#L91

I actually will punt this to the sign command PR

asraa:
Cleaned up, tests added, subcommand added so you run `cosign policy init ...` and I'll add `cosign policy sign` in a follow up

dekkagaijin (approved):

PR #474, opened 2021-07-25
fixing verify-dockerfile command in such cases that arg instruction comes before the from instruction

### LINKED ISSUE #435: verify-dockerfile: support for ARG instructions to define base images
Sometimes, we might be using `ARG` instruction to define their base images within the `FROM` instruction within the Dockerfiles like the following one:

```Dockerfile
ARG IMAGE=gcr.io/fancy/test/image
FROM $IMAGE AS fancy
```

I did some tests to see what is going to happen and I got the following error from the method [getImagesFromDockerfile](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify_dockerfile.go#L124):

```shell
panic: runtime error: index out of range [-1] [recovered]
panic: runtime error: index out of range [-1]
```

So, we have to tackle this problem.

cc: @Dentrax

developer-guy:

Fixes #435

### DIFF
--- a/cmd/cosign/cli/verify_dockerfile.go
+++ b/cmd/cosign/cli/verify_dockerfile.go
@@ -117,7 +117,11 @@ func getImagesFromDockerfile(dockerfile io.Reader) ([]string, error) {
 	fileScanner := bufio.NewScanner(dockerfile)
 	for fileScanner.Scan() {
 		line := strings.TrimSpace(fileScanner.Text())
+		fileScanner.Scan()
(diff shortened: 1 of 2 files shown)


dekkagaijin:
`ARG`s can be separated by an arbitrary amount of lines from the `FROM` lines they're used in. It'd be better to handle `ARG` lines separately, e.g. by applying default values to ENV (if unset) before `os.ExpandEnv`. Might also want to use [`os.Expand`](https://pkg.go.dev/os#Expand) with a custom mapper to avoid any potential for disaster by playing with Env

Issue #87, opened 2021-03-15
Managing/configuring root certs for fulcio


dlorenc:
Right now this is hardcoded in the PR (#83 ). We should make this configurable.

dlorenc:
I propose we do something like:

- Keep the fulcio root pinned here until we have something further back to pin to
- Extra roots can be configured at some user-level path ($HOME/.sigstore/certs)?
  - Env var for another path: sigstore_cert_path
  - These roots are all appended to the fulcio root for verification
- To disable the fulcio root, pass `--no-default-root` or something similar

PR #477, opened 2021-07-26
Allow custom root PEM

lukehinds:
This change allows use of a custom env `COSIGN_ROOT` for those
who wish to use their own root CA in an off public net fulcio
instance.

Ultimately we might want to do this with a flag, but it requires
a fair amount of rewrite, so this seems a good pragamtic bridge
without being to disruptive prior to 1.0

Resolves: #87 

Signed-off-by: Luke Hinds <lhinds@redhat.com>


pkg/cosign/fulcio/fulcio.go
```diff
@@ -46,6 +47,7 @@ const (
 	FlowNormal = "normal"
 	FlowDevice = "device"
 	FlowToken  = "token"
+	altRoot    = "COSIGN_ROOT"
```
dlorenc, line 50: nit: maybe "SIGSTORE_ROOT_DIR" instead? These are really the roots for the sigstore project and aren't cosign specific.

dlorenc (commented):
One nit on the env var name. We should also get this into the README.md somewhere.

pkg/cosign/fulcio/fulcio.go
lukehinds: agree on both, I will add the other flags added too.

pkg/cosign/fulcio/fulcio.go
```diff
@@ -198,7 +200,16 @@ var Roots *x509.CertPool
 
 func init() {
 	cp := x509.NewCertPool()
-	if !cp.AppendCertsFromPEM([]byte(rootPem)) {
+	rootEnv := os.Getenv(altRoot)
+	if rootEnv != "" {
+		raw, err := ioutil.ReadFile(rootEnv)
```
dlorenc, line 205: Oh sorry - I just realized this is a file, not a directory. I think we have two options:

- Keep it as a directory and concatenate all the pem files in it using something like filepath.Glob (this is pretty standard for roots, you can easily have multiple)
- Rename it to "_FILE" and leave it this way. 

My hunch in the long term is that we'll want the ability to have multiple (adding yours in addition to the public good one)

dlorenc (approved):

PR #479, opened 2021-07-26
Add cosign/ to useragent for remote calls

### LINKED ISSUE #478: Add cosign info to user agent
Add `cosign/$VERSION` to whatever ggcr is already sending.

Ends up looking like: `User-Agent: cosign/version go-containerregistry/version`

details: https://github.com/google/go-containerregistry/blob/f37e4debb1523f20848e7b43837d809a3aee2f97/pkg/v1/remote/options.go#L169

loosebazooka:
Signed-off-by: Appu Goundan <appu@google.com>

fixes #478

### DIFF
--- a/cmd/cosign/cli/util.go
+++ b/cmd/cosign/cli/util.go
@@ -90,5 +90,6 @@ func DefaultRegistryClientOpts(ctx context.Context) []remote.Option {
 	return []remote.Option{
 		remote.WithAuthFromKeychain(authn.DefaultKeychain),
 		remote.WithContext(ctx),
+		remote.WithUserAgent("cosign/" + VersionInfo().GitVersion),
 	}
 }


cmd/cosign/cli/util.go
```diff
@@ -90,5 +90,6 @@ func DefaultRegistryClientOpts(ctx context.Context) []remote.Option {
 	return []remote.Option{
 		remote.WithAuthFromKeychain(authn.DefaultKeychain),
 		remote.WithContext(ctx),
+		remote.WithUserAgent("cosign/" + VersionInfo().GitVersion),
```
loosebazooka, line 93: maybe this isn't enough? any way to know if it's a release or non-release version?

loosebazooka: nvm, looks like `git describe --tags --always --dirty` does the right thing

dlorenc (approved):

PR #483, opened 2021-07-26
Unexport pkg/cosign/remote.StaticLayer

imjasonh:
This type isn't used outside of the package, so there's no reason to export it. If other external packages come to depend on it you might anger people with breaking changes. Let's just unexport it now while (we think) it doesn't have many dependents. If we decide it's worth sharing we can make that decision later.


dlorenc:
SGTM I think @jonjohnsonjr was actually thinking of moving this into ggcr itself. Did that ever happen?

imjasonh:
> SGTM I think @jonjohnsonjr was actually thinking of moving this into ggcr itself. Did that ever happen?

It hasn't happened yet but I definitely think we should consider it.

dlorenc (approved):

dekkagaijin (approved):

PR #487, opened 2021-07-26
Avoid remote.Gets when the ref contains a digest

imjasonh:
@jonjohnsonjr noticed these and says this should make cosign a lot faster.

Some of these were TODOs, some just fell out of having `AttachedImageTag` take a digest instead of a descriptor, and looking around while fixing the resulting code breakage.

### DIFF
--- a/cmd/cosign/cli/triangulate.go
+++ b/cmd/cosign/cli/triangulate.go
@@ -20,9 +20,7 @@ import (
 	"flag"
 	"fmt"
 
-	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/peterbourgon/ff/v3/ffcli"
 
 	"github.com/sigstore/cosign/pkg/cosign"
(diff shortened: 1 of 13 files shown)


cmd/cosign/cli/triangulate.go
```diff
@@ -52,17 +53,28 @@ func MungeCmd(ctx context.Context, imageRef string) error {
 		return err
 	}
 
-	// TODO: just return the descriptor directly if we have a digest reference.
-	desc, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain), remote.WithContext(ctx))
-	if err != nil {
-		return err
+	// If the image ref contains the digest, use it.
+	// Otherwise, look up the digest the tag currently points to.
+	var h v1.Hash
```
dlorenc, line 58: I don't like that we have this same thing 4 places :( could we factor it out somehow?

imjasonh: For you Dan, anything.

dlorenc: What about for golangci-lint though?
image

imjasonh:
NB: this has the effect that a reference by digest can be signed/attached-to/whatever without cosign checking that the thing being signed/etc exists -- before, `remote.Get` would fail if it didn't exist, with this change no such check will be made.

That might be just fine, or it might be surprising. In theory it means you could upload a signature for a thing at the same time you're uploading the thing (or before), since the thing doesn't have to exist yet. It also means you can sign `my/image@sha256:0cd1c94a0c260ee12ced4a1f3a2f215287855870774107fcb6c61d50868a19b2` without any idea what that image is, or whether it exists.

dlorenc (approved):

Issue #437, opened 2021-07-15
 proposal: verify-manifest: Traversing Kubernetes Resources and Verify the Given Images


developer-guy:
## Abstract

Slightly the same with `verify-dockerfile,` but for Kubernetes Resources, responsible for creating containers such as `Deployments,` `ReplicaSets,` `DaemonSets.`
etc. The same logic we used to check `ImageRef`s in `verify-dockerfile` command can simply traverse the whole resource to check given `image`s actually _signed_.
There is an alternative way of doing the same with creating `ValidatingAdmissionWebhook` like a project [cosigned](https://github.com/dlorenc/cosigned). But with this command, we can verify the whole images as early as possible in the CI pipelines before applying them into the Kubernetes cluster.

## Implementation

We may have to write a Resource decider to correctly decide where we should get the _image_ ref from the manifest by looking at the _kind_ of the Resource. For example like the following examples:

**Kind: Deployment**
|
|-> `spec.template.spec.containers[*].image`

**Kind: Pod**
|
|-> `spec.containers[*].image`

After founding the images, we should verify them one by one against the public key or keyless.

## Usage

Maybe we can give this command the following name: `verify-manifest.`

```shell=
$ cosign verify-manifest -key cosign.pub <path/to/manifest>
```

To propose as simple as possible, I don't want to add all the other options that we can suğpport, but we can support all the other commands that `verify-dockerfile` currently supports.

cc: @Dentrax

Dentrax:
| KIND        | IMAGE PATH                                        |
|-------------|---------------------------------------------|
| Deployment  | spec.template.spec.containers[*].image      |
| DaemonSet   | spec.template.spec.containers[*].image      |
| ReplicaSet  | spec.template.spec.containers[*].image      |
| StatefulSet | spec.template.spec.containers[*].image      |
| CronJob     | spec.spec.template.spec.containers[*].image |

dlorenc:
This is awesome! I've also heard another use case that might be good to capture here: "inlining" the signatures for each image nto the k8s object itself as an annotation. That would make verification by a policy system easier.

developer-guy:
oh nice, yes we would definitely consider that once we start to implement this, what do you suggest we use as an annotation key, is there any standard key for it ?

dlorenc:
No, not that I know of yet!

dlorenc:
Does anyone want to take a stab at the "inlining" aspect?

dlorenc:
Here's a possible UX:

[...]

PR #490, opened 2021-07-27
Add `cosign verify-manifest` command

joshes:
Adds support for verifying kubernetes resources (manifests) and addresses #437 

It's a naive approach, but wanted to get feedback early on, especially around YAML parsing via regex and JSON support as discussed in the ticket.

### DIFF
--- a/cmd/cosign/cli/verify_manifest.go
+++ b/cmd/cosign/cli/verify_manifest.go
@@ -1,0 +1,120 @@
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"github.com/pkg/errors"
+)
+
+// VerifyCommand verifies all image signatures on a supplied k8s resource
+type VerifyManifestCommand struct {
+	VerifyCommand
+}
+
+// Verify builds and returns an ffcli command
+func VerifyManifest() *ffcli.Command {
+	cmd := VerifyManifestCommand{VerifyCommand: VerifyCommand{}}
+	flagset := flag.NewFlagSet("cosign verify-manifest", flag.ExitOnError)
+	applyVerifyFlags(&cmd.VerifyCommand, flagset)
+
+	return &ffcli.Command{
+		Name:       "verify-manifest",
+		ShortUsage: "cosign verify-manifest -key <key path>|<key url>|<kms uri> <path/to/manifest>",
+		ShortHelp:  "Verify all signatures of images specified in the manifest",
+		LongHelp: `Verify all signature of images in a Kubernetes resource manifest by checking claims
+against the transparency log.
+
+Shell-like variables in the Dockerfile's FROM lines will be substituted with values from the OS ENV.
+
+EXAMPLES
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/verify_manifest.go
```diff
@@ -0,0 +1,120 @@
...
+	applyVerifyFlags(&cmd.VerifyCommand, flagset)
+
+	return &ffcli.Command{
+		Name:       "verify-manifest",
+		ShortUsage: "cosign verify-manifest -key <key path>|<key url>|<kms uri> <path/to/manifest>",
+		ShortHelp:  "Verify all signatures of images specified in the manifest",
+		LongHelp: `Verify all signature of images in a Kubernetes resource manifest by checking claims
+against the transparency log.
+
+Shell-like variables in the Dockerfile's FROM lines will be substituted with values from the OS ENV.
```
hectorj2f, line 49: I believe this is coming from the verify-dockerfile command description.

cmd/cosign/cli/verify_manifest.go
```diff
@@ -0,0 +1,135 @@
...
+		return errors.New("no images found in manifest")
+	}
+	fmt.Fprintf(os.Stderr, "Extracted image(s): %s\n", strings.Join(images, ", "))
+
+	return c.VerifyCommand.Exec(ctx, images)
+}
+
+func getImagesFromYamlManifest(manifest string) ([]string, error) {
+	var images []string
+	re := regexp.MustCompile(`image:\s?(?P<Image>.*)\s?`)
```
hectorj2f, line 116: suggestion: Perhaps it could be more robust to decode the manifest and validate that the extensions are among those identified in the issue (Pod, DaemonSet...), then use the decoded object to get the image value for all the containers (e.g. `obj.Spec.Containers[*].Image`). 
Otherwise you could have a some resources that might not be among the identified list, e.g. CRs.

cmd/cosign/cli/verify_manifest.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Exec runs the verification command
+func (c *VerifyManifestCommand) Exec(ctx context.Context, args []string) error {
+	if len(args) != 1 {
+		return flag.ErrHelp
+	}
+
+	manifestPath := args[0]
+
+	err := isExtensionAllowed(manifestPath)
```
hectorj2f, line 92: suggestion: try to decode the yaml file to verify it is a valid.

cmd/cosign/cli/verify_manifest.go
hectorj2f: Let me know if you want me to help you with this.

joshes: @hectorj2f yes, this was a consideration as mentioned in https://github.com/sigstore/cosign/issues/437#issuecomment-887084109 - I think it would be better, but as different Kinds come out, it binds us more to the K8s API for parsing unless we still just walk down the tree looking for "image" keys (but guaranteed valid YAML) which would address your other comment.

This regex also doesn't account for quoted string variants, so either way this naive approach needs to change in _some_ fashion.

cmd/cosign/cli/verify_manifest.go
joshes: Updated the help docs to reflect.

cmd/cosign/cli/verify_manifest.go
dekkagaijin: Agreed. There's prior art for ducktyping `PodSpec`-like Objects, but I don't think it's worth blocking merge for a "perfect solution

cmd/cosign/cli/verify_manifest.go
dekkagaijin: (YAML or JSON)

Eventually we should probably just use a canonical k8s parser

dekkagaijin (approved):
I'm OK squeaking this in for 1.0.0, with the mentioned caveats...

cmd/cosign/cli/verify_manifest.go
hectorj2f: > There's prior art for ducktyping PodSpec-like Objects. 

Yes, you could use runtime.Raw and decode the object to the list of supported types here.

Sure, the current implementation should not be blocked by this. Once this PR is merged, I could open a PR with my proposed changes, so you can have a look at them.

hectorj2f:
@joshes You need to sign off your commits.

joshes:
> @joshes You need to sign off your commits.

Signed and pushed, thanks.

dlorenc:
Amazing work! Let's get this in and follow up :)

PR #491, opened 2021-07-27
[W.I.P] feat(verify-list): traverse all image tags to verify all of them

### LINKED ISSUE #434: verify-dockerfile: does not recognize default parameter substitution
**Description**

Minimal reproduce:

* Set the `$BUILDPLATFORM` env to `arm/v7`.

Happy-path test case:
```
FROM --platform=${BUILDPLATFORM} golang:1.16.3-alpine3.13 AS builder
```

* Unset the `$BUILDPLATFORM` env.

Failing test case:
```
FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.16.3-alpine3.13 AS builder
```

i.e.:
```
{
	name:         "digest-platform",
	fileContents: `FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.16.3-alpine3.13 AS builder`,
	expected:     []string{"golang:1.16.3-alpine3.13"},
},
```

[getImageFromLine()](https://github.com/sigstore/cosign/blob/e68da418ca48b1c5bf860ee33bd7e8b85cd3edf3/cmd/cosign/cli/verify_dockerfile.go#L126) function returns `--platform=` instead of `--platform=linux/amd64`

**P.S:** This is a limitation of `os.ExpandEnv` functions. Go currently does not support this feature. We may need to use a [custom framework](https://github.com/a8m/envsubst#docs) to make a workaround to fix this.

cc: @developer-guy

Dentrax:
We (I and @developer-guy) implemented a new command called `verify-list`.

Fixes #434

Here is an animated image that shows how success will look like:

render1627415513371

### TODOs
- [x] verify-list (must)
- [x] passing multi-images
- [ ] e2e tests
- [ ] sign-list? (necessary?)
- [ ] concurrency for multi-images?

### Caveats
- no concurrency since we used `c.VerifyCommand.Exec` (i think we need to duplicate `cosign.Verify` or we should implement some goroutine support [here](https://github.com/sigstore/cosign/blob/14d1d0afeef5864943da4514494978a1bf308ffd/cmd/cosign/cli/verify.go#L148))
- see the _Risks & Mitigations_ section in the issue for the further concerns 

### DIFF
--- a/cmd/cosign/cli/verify_list.go
+++ b/cmd/cosign/cli/verify_list.go
@@ -1,0 +1,109 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"strings"
+)
+
+// VerifyListCommand verifies all signatures on a supplied container image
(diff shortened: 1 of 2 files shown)


dekkagaijin:
IMHO this isn't different enough from `verify` with a list of images to want to support in perpetuity

Dentrax:
Should I move this feature into `verify` command as a new flag called `--list`?

dekkagaijin:
> Should I move this feature into `verify` command as a new flag called `--list`?

ehhhh

The registry API kinda (really) sucks for listing images. It's actually not possible to list untagged images by default.

IMHO we should have code samples which show how to compose utilities to achieve this without accepting responsibility for maintaining the feature and foot-gunny UX

PR #492, opened 2021-07-28
add documentation and reduce the surface of `/pkg`

dekkagaijin:


dlorenc:
Looks like this broke some tests, I think we'll have to drop this one for the 1.0.

dekkagaijin:
@dlorenc  unborked :)

dlorenc (approved):

PR #504, opened 2021-07-28
add well-known attestation specs support to the attest command

### LINKED ISSUE #472: Feature: Add support for custom attestation predicate types
Right now we hardcode a specific one for cosign. We should add a flag to select a different one, and then have an index of "well-known" predicates that can be unmarshalled correctly/validated into a go struct

developer-guy:

Fixes #472

Screen Shot 2021-07-29 at 00 29 03
Screen Shot 2021-07-29 at 00 20 11

Known Limitations:
* We don't add support for SPDX predicate yet, because of [that](https://github.com/in-toto/in-toto-golang/blob/886210ae2ab9bfe02fc975213d9bc0cb7ae5f6b8/in_toto/model.go#L960-L966):
> Currently not implemented. Some tooling exists here:
https://github.com/spdx/tools-golang, but this software is still in
early state.

cc: @Dentrax

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -23,15 +23,12 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"io/ioutil"
-	"os"
-	"path/filepath"
-
 	"github.com/google/go-containerregistry/pkg/name"
 	ggcrV1 "github.com/google/go-containerregistry/pkg/v1"
-	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/peterbourgon/ff/v3/ffcli"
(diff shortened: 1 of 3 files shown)


dlorenc (commented):
This looks good once you get the lint errors cleaned up! Nice job with reflection :)

developer-guy:
> This looks good once you get the lint errors cleaned up! Nice job with reflection :)

thank you so much, but we copied over the code from the `in-toto-golang` module because that it is a private method, to be honest. Let's remember Kelsey Hightower's words: `Good developers copy, great developers paste! ` 😅😅😅 
cc: @Dentrax

dlorenc (approved):

PR #509, opened 2021-07-29
Verify ignore scratch images

### LINKED ISSUE #508: verify-dockerfile  Does not appropriately handle scratch images.
**Description**

`cosign verify-dockerfile -key $key Dockerfile` Does not appropriately handle scratch images.

```
error: GET https://index.docker.io/v2/library/scratch/manifests/latest: MANIFEST_UNKNOWN: manifest unknown; map[Tag:latest]
```

rjbrown57:
Fixes #508

### DIFF
--- a/cmd/cosign/cli/verify_dockerfile.go
+++ b/cmd/cosign/cli/verify_dockerfile.go
@@ -118,7 +118,12 @@ func getImagesFromDockerfile(dockerfile io.Reader) ([]string, error) {
 	for fileScanner.Scan() {
 		line := strings.TrimSpace(fileScanner.Text())
 		if strings.HasPrefix(line, "FROM") {
-			images = append(images, getImageFromLine(line))
+			switch image := getImageFromLine(line); image {
+			case "scratch":
+				fmt.Fprintln(os.Stderr, "- scratch image ignored")
+			default:
+				images = append(images, image)
+			}
 		}
 	}
 	if err := fileScanner.Err(); err != nil {


cmd/cosign/cli/verify.go
```diff
@@ -146,6 +146,12 @@ func (c *VerifyCommand) Exec(ctx context.Context, args []string) (err error) {
 	co.SigVerifier = pubKey
 
 	for _, imageRef := range args {
+
+		if imageRef == "scratch" {
```
dlorenc, line 150: Should we normalize the case here on imageRef to support SCRATCH? or is that handled somewhere else too

rjbrown57: I think it's unnecessary. A test with a `FROM SCRATCH` Dockerfile shows repositories must be lowercase.

```
failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to parse stage name "SCRATCH": invalid reference format: repository name must be lowercase 
```

dekkagaijin:
Is there a use-case where "scratch" is being verified directly with `verify`? Or is meant to fix an issue with `verify-dockerfile`? If the latter, it'd be better to patch that command. FWIW I thought I had already special-cased `FROM scratch` in `verify-dockerfile` but I was clearly mistaken

rjbrown57:
> Is there a use-case where "scratch" is being verified directly with `verify`? Or is meant to fix an issue with `verify-dockerfile`? If the latter, it'd be better to patch that command. FWIW I thought I had already special-cased `FROM scratch` in `verify-dockerfile` but I was clearly mistaken

At least when a user is running with docker and "scratch" defaults to "https://index.docker.io/v2/library/scratch" I don't think anyone will run verify scratch.

Would it make more sense to filter somewhere in https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify_dockerfile.go#L115?

rjbrown57:
Updated to ignore scratch in verify-dockerfile

dlorenc (approved):

PR #510, opened 2021-07-30
verify-manifest: decode and use kubernetes resources

hectorj2f:
Signed-off-by: Hector Fernandez <hectorj@gmail.com>

This PR is related to a discussion (https://github.com/sigstore/cosign/pull/490#discussion_r677526759) in this PR https://github.com/sigstore/cosign/pull/490. It aims to decode the manifest looking for multiple resources within a yaml file and decoding the list of core Kubernetes API types that has an image property. Considering the Kubernetes core API is in a stable state, we can assume there won't be many changes to these core types.

### DIFF
--- a/cmd/cosign/cli/verify_manifest.go
+++ b/cmd/cosign/cli/verify_manifest.go
@@ -15,17 +15,25 @@
 package cli
 
 import (
+	"bytes"
 	"context"
 	"flag"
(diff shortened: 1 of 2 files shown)


dlorenc (approved):

dlorenc:
Nice!

PR #519, opened 2021-08-03
Lazy initialize the fulcio Root

asraa:

Main motivation:
* I added a `cosign init` command that sets up the TUF repository locally so that trusted targets can be verified and retrieved from there.
* Unfortunately, it's kind of circular to init the fulcio root certificate before you do that, and that also means you can't run the init command to configure your TUF mirrors and trusted keys yourself if you wanted to.

So this lazily inits the fulcio roots
Sorry for breaking semantic versioning...


hectorj2f (approved):
Lgtm

dlorenc (approved):

Issue #389, opened 2021-06-24
Integrate sigstore's root of trust into `cosign`


dekkagaijin:
Allows for us to validate cert chains and signatures for arbitrary images and binaries.

TODO: decide on what the UX should look like for keyless attestation of signatures

dlorenc:
cc @asraa any thoughts here?

asraa:
Hm! Interesting good point. Here's what you *could* do:

With the client portion of the demo code https://github.com/sigstore/cosign/pull/366:
We can fetch and validate the targets the root of trust signed (e.g. fulcio root ca and rekor public key hardcoded in cosign) before using them. Where would we want to place them the TUF metadata though?
* on a registry like the demo PR?
* we could fetch directly with HTTP on GitHub as well

In order to facilitate all of that people would need to provide / trust some initial root keys they trust (to bootstrap all of this anyway). That needs a source of truth/provision as well.

asraa:
> We can fetch and validate the targets the root of trust signed (e.g. fulcio root ca and rekor public key hardcoded in cosign) before using them.

If it makes sense to pin the initial root in cosign and add verification for Fulcio/Rekor keys and certs, let me know! I can at least use an HTTP fetch.

dlorenc:
Don't know how I closed this one by accident!

@dekkagaijin could you link your doc here when it's ready?

PR #520, opened 2021-08-04
Add `cosign init` to initialize the SigStore root metadata

asraa:
* A command `cosign init` that initializes the local trusted SigStore root and pulls the trusted targets into `.sigstore/root`
* The initial root keys are added to `.sigstore/keys.json`. Clients need this to verify the root they pull
* The targets are used for the Fulcio cert and rekor public key, falling back for now to the previous embedded string in case `init` hasn't been run

Details:
* I implemented a "TUF remote store" located on a GCS bucket (for now it's a testing bucket I set up with the root metadata)
* We only download new metadata/targets from the remote location:
   - initially
   - if, when we retrieve a target, the local metadata no longer verifies
   - the local target does not match hash/size of the signed targets

Fixes
https://github.com/sigstore/cosign/issues/389

In practice, you can see if you run cosign without init'ing, you get log warnings asking "did you run `cosign init`?" otherwise, it's silent, we're using the local metadata.

```
asraa@asraa-glaptop:~/git/cosign$ COSIGN_EXPERIMENTAL=1 ./cosign verify gcr.io/asra-ali/base-builder-new
using embedded fulcio certificate. did you run `cosign init`? error retrieving target:  missing target metadata: tuf: no root keys found in local meta store
using embedded rekor public key. did you run `cosign init`? error retrieving target:  missing target metadata: tuf: no root keys found in local meta store

Verification for gcr.io/asra-ali/base-builder-new --
Certificate subject:  [asraa@google.com]
{"critical":{"identity":{"docker-reference":"gcr.io/asra-ali/base-builder-new"},"image":{"docker-manifest-digest":"sha256:291750a9758506b9b3cace9c43e29b1a0f81172f43e9df1d293cb486a5d90abc"},"type":"cosign container image signature"},"optional":null}
asraa@asraa-glaptop:~/git/cosign$ ./cosign init
asraa@asraa-glaptop:~/git/cosign$ COSIGN_EXPERIMENTAL=1 ./cosign verify gcr.io/asra-ali/base-builder-new

Verification for gcr.io/asra-ali/base-builder-new --
Certificate subject:  [asraa@google.com]
{"critical":{"identity":{"docker-reference":"gcr.io/asra-ali/base-builder-new"},"image":{"docker-manifest-digest":"sha256:291750a9758506b9b3cace9c43e29b1a0f81172f43e9df1d293cb486a5d90abc"},"type":"cosign container image signature"},"optional":null}

```

WIP because of tests and some defaults/globals I want to get rid of


trishankatdatadog:
Hihi, would you need a review? Happy to help! 🙂

asraa:
I have to do a bit of refactoring, but would love to get a review on the TUF code! I'll ping you

pkg/cosign/tlog.go
```diff
@@ -41,6 +44,19 @@ import (
 // rekor.pub should be updated whenever the Rekor public key is rotated & the bundle annotation should be up-versioned
 //go:embed rekor.pub
 var rekorPub string
+var rekorTargetStr = `rekor.pub`
+
+func GetRekorPub() string {
```
dekkagaijin, line 49: IMHO we should put this somewhere under `cli/` in order to keep logging and `cosign` binary-specific out of package

pkg/cosign/tlog.go
```diff
@@ -41,6 +44,19 @@ import (
...
+	ctx := context.Background() // TODO: pass in context?
+	buf := tuf.ByteDestination{Buffer: &bytes.Buffer{}}
+	err := tuf.GetTarget(ctx, rekorTargetStr, &buf)
+	if err != nil {
+		// The user may not have initialized the local root metadata. Log the error and use the embedded root.
+		fmt.Println("using embedded rekor public key. did you run `cosign init`? error retrieving target: ", err)
```
dekkagaijin, line 55:
```suggestion
		fmt.Fprintln(os.Stderr, "using embedded rekor public key. did you run `cosign init`? error retrieving target: ", err)
```

pkg/cosign/tuf/client.go
```diff
@@ -0,0 +1,199 @@
...
+func Init(ctx context.Context) error {
+	rootClient, err := RootClient(ctx)
+	if err != nil {
+		return errors.Wrap(err, "initializing root client")
+	}
+	rootKeys, err := getRootKeys()
+	if err != nil {
+		return errors.Wrap(err, "retrieving root keys")
+	}
+	if err := rootClient.Init(rootKeys, TufThreshold); err != nil {
```
dekkagaijin, line 138: Pls make `TufThreshold` and `TufLocalTargets` args to init and/or provide a function that allows for non-default location and threshold

dekkagaijin (changes requested):

pkg/cosign/tuf/client.go
asraa: Yep!! Exactly the WIP part, these are gonna be args to `cosign init` with those defaults

pkg/cosign/tlog.go
asraa: Ohhh hmmmm you mean keeping `fmt.Println` out? It's used elsewhere in this file as well

dekkagaijin: hrm yeah

pkg/cosign/tuf/client.go
asraa: the only problem one is the location of the default local store:
* if you run `cosign init --local path/to/some/other/directory`
* and then you run `cosign verify`: how is it supposed to carry over to the non-default path?

i'm not convinced you should be able to configure that part, unless it's pulled from an env var

dekkagaijin: Pulling from an env var is pretty standard practice, e.g. `KUBECONFIG=` and `DOCKER_CONFIG=`

dekkagaijin (approved):
I'm fine with this for now, can always move packages around as we integrate more tuf stuff

dekkagaijin:
@asraa squashing your local `go.sum` from upstream/main and running `go mod tidy` again should fix the conflict

asraa:
All set, a local test added, and the path to the remote repository is in production.
@trishankatdatadog want to take a quick look at the TUF code? should be pretty simple.
Only question I had was: should we be making a `tuf.db` file? it works fine without one, except for in the test.

PR #527, opened 2021-08-09
Move "verify-manifest" into "verify".

dlorenc:
We'll keep both commands for now, but deprecate verify-manifest.

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -42,6 +43,7 @@ type VerifyCommand struct {
 	Output      string
 	RekorURL    string
 	Annotations *map[string]interface{}
+	Files       StringSlice
 }
 
 func applyVerifyFlags(cmd *VerifyCommand, flagset *flag.FlagSet) {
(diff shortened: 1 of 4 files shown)


dlorenc:
Ref #437

dekkagaijin:
Dan and I chatted about things offline, but I think we probably don't want to be adding any additional file scanning capabilities to `cosign`. It's probably regrettable that we had it in the first place, instead of creating an independent utility which could be composed with `cosign` to achieve the desired functionality.

Going forward, we want to prioritize ease of use for signature and attestation validation over all

hectorj2f (approved):
lgtm

hectorj2f:
@dekkagaijin Please, could you disclosure more details about this new approach `creation of an independent utility which could be composed with cosign to achieve the desired functionality` ?

cmd/cosign/cli/verify.go
```diff
@@ -145,7 +149,26 @@ func (c *VerifyCommand) Exec(ctx context.Context, args []string) (err error) {
 	}
 	co.SigVerifier = pubKey
 
-	for _, imageRef := range args {
+	allImgs := []string{}
+	allImgs = append(allImgs, args...)
+
+	for _, f := range c.Files.slice {
+		err := isExtensionAllowed(f)
```
Dentrax, line 156: Will we distinguish the given file whether `Dockerfile` in this function? Currently, `isExtensionAllowed` only returns _true_ in case given file is `.yml` or `.yaml`. In the flag description, since we pointed out that `-f` flag accepts _Dockerfile_, we actually do not check here.

```golang
flagset.Var(&cmd.Files, "f", "files to validate (kubernetes manifests or Dockerfiles)")
```

Will we move the `verify-dockerfile` into `verify` command in the long term?

dlorenc:
Closing, we went a different way here!

PR #528, opened 2021-08-09
Explicitly disable auth for the sigstore-tuf-root.

dlorenc:
I had expired credentials that were causing this to fail. The bucket
is public, so we should just not use auth (which apparently requires being
explicit).


asraa (approved):
Thanks for catching this!

PR #530, opened 2021-08-10
'cosign init' minor enhancements (file or URL root, write to $HOME/.sigstore)

asraa:

Makes two minor changes to `cosign init` behavior:
* Writes the resulting metadata to the home directory
* Allows users to specify a file or URL that defaults to the github repository mirror for the initial trusted root metadata


cmd/cosign/cli/init.go
```diff
@@ -18,17 +18,44 @@ package cli
...
+	return raw, nil
+}
+
 func Init() *ffcli.Command {
 	var (
 		flagset = flag.NewFlagSet("cosign init", flag.ExitOnError)
 		// TODO: Support HTTP mirrors as well
 		mirror    = flagset.String("mirror", "sigstore-tuf-root", "GCS bucket to a SigStore TUF repository.")
-		root      = flagset.String("root", ".sigstore/keys.json", "path to trusted initial root.")
+		root      = flagset.String("root", "https://raw.githubusercontent.com/sigstore/root-signing/main/ceremony/2021-06-18/repository/1.root.json", "path to trusted initial root.")
```
dlorenc, line 58: wdyt about somehow sticking this into a GCR or GCS bucket, and hardcoding the hash of the resulting file into cosign somehow? This is the root root root, so the hash should never change.

asraa: ooh thanks! i like sticking the hash in. 
 (and actually i think there's http fetches for buckets too? that might make it comptaible with a url flag)

dlorenc: Yea! "storage.googleapis.com/<bucket>/<object>" should work.

asraa: is it really necessary to stick the sha in if it's validated in the snapshot.json?

pkg/cosign/tuf/client.go
```diff
@@ -35,17 +38,23 @@ import (
 const (
 	TufRootEnv        = "TUF_ROOT"
 	defaultLocalStore = ".sigstore/root/"
+	// SHA512 hash of the initial root.json. This is used to pin and bootstrap the initial root to cosign.
+	initialRootSha = "87d41965ac08a2e03a33aa2db1dbbf5a7d6616e49f0ca3b20afe1c46ae78e52e841e21cca347f7b77ec929b6a3687d86d320ee04ba5f75a1880974310958b9c2"
```
dlorenc, line 42: Do we still need this one?

dlorenc (approved):

dlorenc:
I think there's one unused variable, lgtm either way!

pkg/cosign/tuf/client.go
dlorenc: Ah, I guess it's still used (but since we embed maybe we don't need it?)

Issue #542, opened 2021-08-15
cosign verify does not check if image digest matches the current artifact


JimBugwadia:
This issue was reported by a Kyverno user and I've reproduced via the cosign CLI:

Currently its possible to change (or delete) the image without updating the signature. When verifying the image via the digest, cosign will not report an error.

1. Sign an push image signature. `cosign sign -key cosign.key ghcr.io/jimbugwadia/pause`.
2. Delete the image in the image registry (I used GitHub to do this), but leave the signature.
3. Create a new image with the same name and tag to impersonate the prior image. `docker tag nginx ghcr.io/jimbugwadia/pause:latest`.
4. Push the new image. `docker push ghcr.io/jimbugwadia/pause:latest`.
5. Run cosign verify with the previous (incorrect) digest appended. `cosign verify -key cosign.pub ghcr.io/jimbugwadia/pause@sha256:b31bfb4d0213f254d361e0079deaaebefa4f82ba7aa76ef82e90b4935ad5b105`.

```sh
cosign verify -key cosign.pub ghcr.io/jimbugwadia/pause@sha256:b31bfb4d0213f254d361e0079deaaebefa4f82ba7aa76ef82e90b4935ad5b105

Verification for ghcr.io/jimbugwadia/pause@sha256:b31bfb4d0213f254d361e0079deaaebefa4f82ba7aa76ef82e90b4935ad5b105 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.
{"critical":{"identity":{"docker-reference":"ghcr.io/jimbugwadia/pause"},"image":{"docker-manifest-digest":"sha256:b31bfb4d0213f254d361e0079deaaebefa4f82ba7aa76ef82e90b4935ad5b105"},"type":"cosign container image signature"},"optional":null}
```
Note that the verify will return an error if the digest is not used, which is the desired behavior:

```sh
cosign verify -key cosign.pub ghcr.io/jimbugwadia/pause
error: fetching signatures: getting signature manifest: GET https://ghcr.io/v2/jimbugwadia/pause/manifests/sha256-3f13b4376446cf92b0cb9a5c46ba75d57c41f627c4edb8b635fa47386ea29e20.sig: NAME_UNKNOWN: repository name not known to registry
```

PR #543, opened 2021-08-15
always check remote image

JimBugwadia:
Signed-off-by: Jim Bugwadia <jim@nirmata.com>

Fixes https://github.com/sigstore/cosign/issues/542

Remove fetch by digest, to always retrieval of the latest digest from the registry, in case an invalid digest is passed in or if the image is deleted and the signature is not.

With this change an invalid manifest error is returned when the incorrect digest is passed in:

```sh
cosign verify -key ~/cosign.pub ghcr.io/jimbugwadia/pause@sha256:b31bfb4d0213f254d361e0079deaaebefa4f82ba7aa76ef82e90b4935ad5b105
error: GET https://ghcr.io/v2/jimbugwadia/pause/manifests/sha256:b31bfb4d0213f254d361e0079deaaebefa4f82ba7aa76ef82e90b4935ad5b105: MANIFEST_UNKNOWN: manifest unknown
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -79,22 +79,12 @@ func Verify(ctx context.Context, signedImgRef name.Reference, co *CheckOpts) ([]
 		return nil, errors.New("one of verifier or root certs is required")
 	}
 
-	// If the image ref contains the digest, use it.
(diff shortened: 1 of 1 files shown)


dlorenc (approved):
LGTM, thanks!

PR #546, opened 2021-08-15
Fix verify and verify-attestation output flag

josedonizetti:
There are a few things on this PR.

1 - improve documentation, listing the possible options for `-output` -> '(json|text)'
2 - fix existing behavior, where `-output` is actually ignored, [using always `text`](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify.go#L166), even though the documentation says the [default is `json`](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify.go#L54).

A couple questions:

1 - I think it makes more sense to rename this flag to `-format`, instead of `-output` considering the options (json/text).
2 - The current default is `text`, but fixing this to follow the documentation, it becomes `json`, but I think text looks better on the command line.

wdyt?

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -51,7 +51,7 @@ func applyVerifyFlags(cmd *VerifyCommand, flagset *flag.FlagSet) {
 	flagset.StringVar(&cmd.Slot, "slot", "", "security key slot to use for generated key (default: signature) (authentication|signature|card-authentication|key-management)")
 	flagset.StringVar(&cmd.RekorURL, "rekor-url", "https://rekor.sigstore.dev", "address of rekor STL server")
 	flagset.BoolVar(&cmd.CheckClaims, "check-claims", true, "whether to check the claims found")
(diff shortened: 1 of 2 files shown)


dlorenc:
> 1 - I think it makes more sense to rename this flag to `-format`, instead of `-output` considering the options (json/text).

Could we keep output as an alias? I think you're right, but would prefer not to break stability right after 1.0 :(

josedonizetti:
@dlorenc What do you think about merging as is? It fixes the hardcoded option, allowing the user to choose a format, and I can follow up with a different PR for a new `format` flag alias of `output`.

developer-guy:
> @dlorenc What do you think about merging as is? It fixes the hardcoded option, allowing the user to choose a format, and I can follow up with a different PR for a new `format` flag alias of `output`.

yeah make sense to me, I'll approve the PR as is, as you said maybe you can create another PR for the alias thing, PTAL @dlorenc

developer-guy (approved):
LGTM

PR #558, opened 2021-08-22
Add a commented out list of OWNERS for transparency.

dlorenc:
We manage the CODEOWNERS via a GitHub team which unfortunately means the list
isn't present here. Let's keep them in sync manually until we come up with
something better.


lukehinds (approved):
Good idea, I will make an issue to do the other repos.

PR #563, opened 2021-08-24
fix nil pointer deref in cli/upload.BlobCmd

ansemjo:
I was playing around with the `cosign` cli and found that calling `cosign upload blob ...` with an image address but without any `-f` arguments leads to a nil pointer dereference. This patch adds a check that at least one file was provided.

The underlying cause is that calling `cremote.UploadFiles` with an empty `files` list never sets the `var img v1.Image` within and thus the returned `dgstr` is `nil`, I believe?

### DIFF
--- a/cmd/cosign/cli/upload/blob.go
+++ b/cmd/cosign/cli/upload/blob.go
@@ -17,6 +17,7 @@ package upload
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -83,7 +84,7 @@ EXAMPLES
   `,
 		FlagSet: flagset,
 		Exec: func(ctx context.Context, args []string) error {
-			if len(args) != 1 {
+			if len(args) != 1 || len(fmap.Files) < 1 {
 				return flag.ErrHelp
 			}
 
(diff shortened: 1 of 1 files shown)


cmd/cosign/cli/upload/blob.go
```diff
@@ -102,6 +102,10 @@ func BlobCmd(ctx context.Context, files []cremote.File, contentType, imageRef st
 	if err != nil {
 		return err
 	}
+	// TODO: error format?
+	if dgster == nil {
+		return fmt.Errorf("dgstr is nil, no files uploaded?")
```
dlorenc, line 107: I think you can use "errors.New" here instead of fmt.Errorf since there is no format parameter.

dlorenc:
Thanks for the fix!

ansemjo:
Sure! I wasn't sure if it was the right place to raise an error in `BlobCmd` or if `dgster.Digest` was more appropriate .. I force-pushed another edit using `errors.New` per your suggestion.

dlorenc (approved):

dlorenc:
Thanks!

Issue #550, opened 2021-08-18
Include a cosign custom and independent Kubernetes admission controller to verify images


hectorj2f:
**Question**

I believe sigstore should include under his own organization an implementation of a cosign admission controller for Kubernetes. This idea does not aim to replace any other possible implementation of verifiers or admission controllers that exist out there. The intention is to have a custom and independent admission controller that simply verifies Kubernetes resources using cosign. The community could maintain this cosign implementation under sigstore organization, and thereby eases its maintenance and testing as part of every release.

The initial idea would be to reuse https://github.com/dlorenc/cosigned implementation and upgrade it following common and best practices when developing a controller for Kubernetes. 

What do you think about it ?

dlorenc:
Thanks for filing! Ref #196 as well.

I'm open to moving that code over here if there's demand for it! I started it as a learning project and then kept it separate because I didn't want to slow down support for cosign in other full-featured policy systems like Gatekeeper or kyverno, but I think we're past that point now.

So, +1 on moving it in if people are interested in using i!

hectorj2f:
Thanks @dlorenc.

Yes. Even if anyone can still use the implementations from Gatekeeper, Kyverno or Connaisseur. I believe there is a benefit having a single-purpose admission controller that adapts to the on-going cosign development and lives under the sigstore umbrella.

dlorenc:
@developer-guy what do you think here? I think it would be easiest to keep in the same repo as cosign long term if there's interest.

developer-guy:
> I believe there is a benefit having a single-purpose admission controller that adapts to the on-going cosign development and lives under the sigstore umbrella.

Yeah, I agree with @hectorj2f, we should definitely move the cosigned project into the sigstore organization and start maintaining it from there, that would be awesome and we (w/@Dentrax) would love to get involved in it because we really like to work with Kubernetes extension points like AdmissionWebhooks. 

> I think it would be easiest to keep in the same repo as cosign long term if there's interest.

Yeah, makes sense 🤝

dlorenc:
OK, SGTM! Want to start moving code over @hectorj2f?

hectorj2f:
@dlorenc Good news 🎉 !. Yes, it sounds good to me. 

I would move your code to cosign repository under a gitmodule. So we could start to add enhancements/improvements once all the cosigned foundation is in there.

dlorenc:
[...]

PR #565, opened 2021-08-24
[wip]: cosigned admission controller

hectorj2f:
This PR adds the source code of an initial implementation of a sigstore admission controller for cosign. The code was initially inspired based on the source code developed by @dlorenc in https://github.com/dlorenc/cosigned. 

It is related to sigstore/cosign#550.

TODO: 
* We are still figuring it out other details such as the image name and the release process to create a specific docker image.

**Note:** The chart installation would be moved to https://github.com/sigstore/helm-charts/pull/9.


dlorenc (commented):
One other nit would be to add a .gitattributes to declare the generated code: https://github.com/tektoncd/chains/blob/main/.gitattributes

PR #567, opened 2021-08-24
Pin `k8s.io` dependencies to v0.20.7

priyawadhwa:
Tekton Chains depends on knative, which is pinned to v0.20.7. Unfortunately, Chains can't depend on both knative and cosign unless either knative upgrades or cosign is pinned to v0.20.7. 

If we don't want to downgrade, we can discuss other options in this issue I opened in knative: https://github.com/knative/pkg/issues/2232 


dlorenc (approved):

dlorenc:
Will this make dependabot not send updates?

priyawadhwa:
Looks like there's a PR open for that (https://github.com/dependabot/dependabot-core/pull/4140), but for now it probably will send updates :(

jerbob92:
> Looks like there's a PR open for that ([dependabot/dependabot-core#4140](https://github.com/dependabot/dependabot-core/pull/4140)), but for now it probably will send updates :(

PR has been merged :) Release should be on its way.

Issue #591, opened 2021-08-28
Verify SCTs


dlorenc:
Fulcio now returns SCTs as a header: https://github.com/sigstore/fulcio/pull/163

Let's verify those in cosign when getting a cert!

dlorenc:
cc @priyawadhwa you were looking at this kind of right?

priyawadhwa:
Yah, this was the last thing we needed for bundling to be complete!

We can verify the SCT when we get it, add it to the bundle, and verify again whenever `cosign verify` gets run?

PR #600, opened 2021-08-31
Verify SCTs returned by fulcio

priyawadhwa:
Added in the CT log public key for this verification.

up next, adding it to the bundle!

closes https://github.com/sigstore/cosign/issues/591


cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -53,11 +59,26 @@ const (
...
+	SCT      []byte
+}
+
 // This is the root in the fulcio project.
 //go:embed fulcio.pem
 var rootPem string
+
+// This is the public key for the fulcio CT log
+//go:embed ct.pub
+var ctPublicKey string
```
asraa, line 74: this is also found in the sigstore TUF root under `ctfe.pub`
similar to https://github.com/asraa/cosign/blob/4b3ec76463cdfa1dd9cb94fbc03ae11c07be5c07/cmd/cosign/cli/fulcio/fulcio.go#L214

priyawadhwa: thanks!

dekkagaijin:
We should have at least one place where "SCT" is explained :p

priyawadhwa:
Added a comment :)

dekkagaijin (approved):

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -109,17 +128,56 @@ func getCertForOauthID(priv *ecdsa.PrivateKey, scp signingCertProvider, connecto
...
+	return fr, nil
+}
+
+// verifySCT verifies the SCT against the Fulcio CT log public key
+// The SCT is a `Signed Certificate Timestamp`, which promises that
+// the certificate issued by Fulcio was also added to the public CT log within
+// some defined time period
+func verifySCT(fr Resp) error {
+	buf := tuf.ByteDestination{Buffer: &bytes.Buffer{}}
+	if err := tuf.GetTarget(context.TODO(), ctPublicKeyStr, &buf); err != nil {
```
asraa, line 161: not really sure if we should error out on this -- because it will fail if you haven't run cosign init and i don't think we want that to be failure behavior (rather preferrable behavior)

priyawadhwa: sure, i can add in a warning that SCTs won't be verified if the user doesn't run `cosign init` 

in the future i wonder if we could just automatically run `cosign init` for someone if they have experimental mode set!

dlorenc: Maybe a silly question, but If we have all the info to validate against in the binary to run "cosign init", could we just validate it there before putting it on disk?

asraa: "there"? 
i think i'm having trouble parsing the question! do you mean if you haven't run cosign init at this point, go ahead and run it?

PR #603, opened 2021-09-01
Try fixing the go.mod file by pinning the go syntax version to 1.16.

dlorenc:
We can still use go 1.17, but this tells it to generate backwards compatible changes.


dlorenc:
This is just a guess why things like #602 are failing

Issue #604, opened 2021-09-01
Automatically run `cosign init` for keyless mode


priyawadhwa:
Verifying SCTs (https://github.com/sigstore/cosign/pull/600) relies on the user running `cosign init` first so that the CT log public key is on disk for verification

We can automatically run `cosign init` for a user if they haven't already done it before verification! Once this is done, we can consider failing if SCT verifications fails (right now we just log a warning and move on)

dlorenc:
+1 to either option there Priya. We should embed enough stuff to either initialize the directory or verify in memory

priyawadhwa:
sgtm! I'll probably switch back to embedding the ct log public key and verifying in memory then, so that we don't have to depend on GCS to get the public key

asraa:
This is done.

PR #607, opened 2021-09-01
Embed CT log public key for SCT verification

priyawadhwa:
This way, we don't have to depend on users running `init` and we can error out if SCT verification fails.

Ref https://github.com/sigstore/cosign/issues/604


dlorenc:
Nice. Longer term we'll probably want a way to embed the entire latest TUF root when we start to use real delegations for this.

dlorenc (approved):

PR #619, opened 2021-09-03
Pinned the dockerfile to sha256

naveensrinivasan:
The dockerfile was referring to the tag. It is recommended to use pinned
dependencies based on SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

Changed the go:1.17 to SHA256 digest. This still works with dependabot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 
 ARG RUNTIME_IMAGE=gcr.io/distroless/base:debug
 
-FROM golang:1.17.0 as build
+FROM golang@sha256:c994ea4c0e524ea97ea7b4b21c19b968170a0c804b2fa7eee3c70c779fe84211 as build
 
 WORKDIR /go/src/cosign
 ADD . /go/src/cosign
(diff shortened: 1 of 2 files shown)


dlorenc:
Cc @cpanato can we just delete these Dockerfiles? Are they used anywhere?

naveensrinivasan:
> Cc @cpanato can we just delete these Dockerfiles? Are they used anywhere?

It is used https://github.com/sigstore/cosign/search?q=Dockerfile

cpanato:
@dlorenc @naveensrinivasan the file `Dockerfile.cosigned` is used now to build the `cosigned` image, I can follow up on that to instead having the dockerfile we have the similar approach that we have for `cosign` (using KO) 

the other `Dockerfile` we used to use that to build the image without KO.

cpanato:
yep we use here: https://github.com/sigstore/cosign/blob/main/release/cloudbuild.yaml#L69-L74

dlorenc (approved):

Issue #556, opened 2021-08-20
Allow pinning trust for verifying keyless signatures


xopham:
**Description**
Trust pinning should be implemented for Keyless signatures by adding an optional `-email` parameter to the `cosign verify`. This email can then be used as a trust anchor to validate the keyless signature and returned error code is 0 if the `Certificate subject` contains the email or 1 if not.
In case the optional `-email` is not specified, Cosign should always return 1 as the intent of cosign cannot be fulfilled (secure by default)

### Context
When verifying with Cosign, one provides a key and the signature data is validated against that key and the valid digest returned:

```bash
$ cosign verify -key cosign.pub xoph/demo:co-signed

Verification for xoph/demo:co-signed --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.
{"critical":{"identity":{"docker-reference":"index.docker.io/xoph/demo"},"image":{"docker-manifest-digest":"sha256:c5327b291d702719a26c6cf8cc93f72e7902df46547106a9930feda2c002a4a7"},"type":"cosign container image signature"},"optional":null}
```

For keyless signatures, we break with this flow and one just asks to verify an image upon which one gets both trust root and digest returned and now as a user I decide whether I like the email handle I see.

```bash
$ COSIGN_EXPERIMENTAL=1 cosign verify xoph/demo:co-signed

Verification for xoph/demo:co-signed --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - Any certificates were verified against the Fulcio roots.
Certificate subject:  [xoph@posteo.de]
{"critical":{"identity":{"docker-reference":"index.docker.io/xoph/demo"},"image":{"docker-manifest-digest":"sha256:e7e0f847265b6d729c718d0b778a4e15c0bce7078d7042b2a4b90e9998794e67"},"type":"cosign container image signature"},"optional":null}
```

It would be better to pin the trust to a trust anchor (and also move the certificate subject to the json):

```bash
$ COSIGN_EXPERIMENTAL=1 cosign verify -email xoph@posteo.de xoph/demo:co-signed

Verification for xoph/demo:co-signed --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - Any certificates were verified against the Fulcio roots.
[...]

PR #622, opened 2021-09-04
add `-cert-email` flag to specify an email for a `fulcio` cert  to be valid

dekkagaijin:
Fixes https://github.com/sigstore/cosign/issues/556
(at least partially)

In the long term we'll want to implement the ability to explicitly state expected cert subject(s), but the UX design for that was sticky enough that I wanted to make that change independently.

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -37,6 +37,7 @@ import (
 type VerifyCommand struct {
 	CheckClaims bool
 	KeyRef      string
+	CertEmail   string
 	Sk          bool
 	Slot        string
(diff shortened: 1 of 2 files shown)


dlorenc:
Nice!!! Wdyt about cert-subject instead, to handle the SPIFFE ones?

dekkagaijin:
> Wdyt about cert-subject instead, to handle the SPIFFE ones?

I wanted to add that later to make the distinction between the "email" and the "subject" of the cert. We'll also have to give some thought as to the UX for that

dlorenc (approved):

PR #623, opened 2021-09-05
Make sure generate-key-pair doesn't overwrite existing key-pair

chhetripradeep:
Also create public key with 0644 permission. Both of these changes are done using ssh-keygen as reference.

Signed-off-by: Pradeep Chhetri <pradeepchhetri4444@gmail.com>


dlorenc:
Looks like a lint error, otherwise lgtm!

dlorenc (approved):

PR #635, opened 2021-09-07
Drop the distinction between Create/Update.

mattmoor:
There was a bunch of boilerplate to differentiate between Create/Update, which wasn't particularly useful.  This drops it and streamlines a bunch of the logic.

I expect this will be the first cleanup of a number to follow...

/assign @dlorenc


dlorenc:
Nice cleanup!

dlorenc (approved):

PR #641, opened 2021-09-09
feat: add validation for predicates via cue and rego policy files support

### LINKED ISSUE #512: Feature: Add support for custom attestation predicate types but this time while verifying attestation
**Description**

We added this support for uploading attestation to the registries in PR #504  but we haven't added this support for verifying attestations against claims yet, so we have to add this support for it too. 

https://github.com/sigstore/cosign/blob/main/pkg/cosign/verifiers.go#L47

cc: @dlorenc @Dentrax

developer-guy:
Fixes #512


erkanzileli:
We thought there are two ways to implement it on the cli’s UX

First is users runs cosign for each attestation type of their image's existing attestations. This means if an image has both `spdx` and `slsa` attestations then users have to run cosign twice by changing `-type` flag of the commands.

For example:

```shell
$ cosign verify-attestation -key cosign-keys/cosign.pub -policy=demos/spdx.cue -type spdx devopps/alpine:3.8
$ cosign verify-attestation -key cosign-keys/cosign.pub -policy=demos/slsaprovenance.cue -type slsaprovenance devopps/alpine:3.8
```

Seconds is users run cosign just one time. Here is the example to describe it better.

```shell
cosign verify-attestation -key cosign-keys/cosign.pub -policy=slsaprovenance=demos/slsaprovenance.cue -policy=spdx=demos/spdx.cue devopps/alpine:3.8
```

We implemented the first way now.

WDYT @dlorenc
CC: @developer-guy @Dentrax

cmd/cosign/cli/verify_attestation.go
```diff
@@ -166,8 +191,65 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, args []string) (err
...
+				var spdxStatement in_toto.SPDXStatement
+				if err := json.Unmarshal(decodedPayload, &spdxStatement); err != nil {
+					return err
+				}
+				payload, _ := json.Marshal(spdxStatement.Predicate)
+				if err := cue.ValidateJSON(payload, c.Policies.entrypoints); err != nil {
+					return err
+				}
+			default:
+				continue
```
verdverm, line 250: Is `c.PredicateType` already validated at this point? Or should an error be thrown about an unknown predicate type? (for example when a user has a typo)

If not validated, this looks to silently pass the check.

cmd/cosign/cli/verify_attestation.go
```diff
@@ -166,8 +191,65 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, args []string) (err
...
+				if err := cue.ValidateJSON(payload, c.Policies.entrypoints); err != nil {
+					return err
+				}
+			case predicateSpdx:
+				var spdxStatement in_toto.SPDXStatement
+				if err := json.Unmarshal(decodedPayload, &spdxStatement); err != nil {
+					return err
+				}
+				payload, _ := json.Marshal(spdxStatement.Predicate)
+				if err := cue.ValidateJSON(payload, c.Policies.entrypoints); err != nil {
```
verdverm, line 246: each case looks to have the same three lines, can the be moved outside and after the switch statement?

developer-guy: yeah, you are absolutely right, fixed, thank you @verdverm 🙏

cmd/cosign/cli/verify_attestation.go
developer-guy: thank you so much @verdverm, it should be fixed right now.

Dentrax:
Just updated to 3a738f17f452fe90021375d606901a7d6d9ab2ad commit, added some sanity and error checks.

erkanzileli:
We think we are ready.

WDYT @dlorenc

dlorenc:
This is so cool and the code looks great! Do you mind if I leave the PR open for a bit to play with the cli, cue and the overall experience?

This is really new and exciting for all of the in-toto space!

pkg/cosign/cue/cue.go
```diff
@@ -0,0 +1,45 @@
...
+
+package cue
+
+import (
+	"cuelang.org/go/cue/cuecontext"
+	"cuelang.org/go/cue/load"
+	cuejson "cuelang.org/go/encoding/json"
+)
+
+func ValidateJSON(jsonBody []byte, entrypoints []string) error {
```
hectorj2f, line 24: nit: ValidateCueJSON as function name.

cmd/cosign/cli/verify_attestation.go
```diff
@@ -166,10 +200,92 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, args []string) (err
...
+			}
+			if err := cue.ValidateJSON(payload, c.Policies.EntryPoints); err != nil {
+				validationErrors = append(validationErrors, err)
+			}
+		}
+
+		if len(validationErrors) > 0 {
+			fmt.Println("Some errors occurred during the validation:")
+			for _, v := range validationErrors {
+				_, _ = fmt.Fprintf(os.Stderr, "- %v\n", v)
```
hectorj2f, line 284: suggestion: use fmt.Printf( instead of Fprintf

cmd/cosign/cli/verify_attestation.go
```diff
@@ -49,6 +75,8 @@ func applyVerifyAttestationFlags(cmd *VerifyAttestationCommand, flagset *flag.Fl
 	flagset.BoolVar(&cmd.CheckClaims, "check-claims", true, "whether to check the claims found")
 	flagset.StringVar(&cmd.FulcioURL, "fulcio-url", "https://fulcio.sigstore.dev", "[EXPERIMENTAL] address of sigstore PKI server")
 	flagset.StringVar(&cmd.RekorURL, "rekor-url", "https://rekor.sigstore.dev", "[EXPERIMENTAL] address of rekor STL server")
+	flagset.StringVar(&cmd.PredicateType, "type", "custom", "specify predicate type (default: custom) (slsaprovenance|link|spdx)")
+	flagset.Var(&cmd.Policies, "policy", "specify CUE files will be using for validation")
```
hectorj2f, line 79: i suggest to include in the description the statement "comma separeted list of CUE files".

pkg/cosign/cue/cue.go
erkanzileli: Thanks for the suggestion. The method is already in the `cue` package and it receives a normal JSON. It would be better to use it like this. WDYT?

cmd/cosign/cli/verify_attestation.go
erkanzileli: Do you mean printing errors to the STDOUT?

cmd/cosign/cli/verify_attestation.go
erkanzileli: Right now it's not working like that. You should specify all your Cue files with separate `-policy` flags.
For example: 
```shell
cosign verify-attestation -key cosign-keys/cosign.pub -policy=demos/spdx1.cue -policy=demos/spdx2.cue -type spdx devopps/alpine:3.8
```
 Is this command should work as you said?

developer-guy:
Hello @dlorenc, we thought that we could support verifying in-toto attestations by using Rego policies as well like the following:

Screen Shot 2021-09-12 at 22 38 05

As you already know, we can use OPA as a [go library](https://www.openpolicyagent.org/docs/latest/#5-try-opa-as-a-go-library). So the only thing that we have to do is detecting the extension of whether it is Cue or Rego, then calling the correct implementation. But there is a small workaround here to make the flags consistent with the cue implementation. We don't want to accept queries for Rego; instead, what we expect from users is that they have to write Rego policies in package "cosign," and they have to give "allow" names to rules like the above picture.

verdverm:
@developer-guy I was thinking that it would be nice if I could specify all my policies in a single CUE file and then select the ones I'd like to use for a given attestation. The `cue` CLI has this concept through the `-e` flag, or in Go with the `LookupPath` function. OPA looks to have similar ability to select out a subset of the full input.

Both CUE and Rego look to have the concept of modules and packages. It seems that both could be handled similarly, in that `cue.Load` was used to support multiple entrypoints which may include other files. I could, in theory, pass a directory in for a CUE entrypoint. I don't know if this is possible with Rego, but could cause issues with extension detection.

Generally, it looks like the code is intended to be that each policy flag holds a separate policy. I think it would be helpful to clarify the ideas of entrypoints from how policies are written. The CUE validation passes all policies together during validation, for each iteration of the loop. I'm not familiar enough with the details yet to know if this is implementing the desired outcomes. How do the policy flags relate to the PredicateType here:

https://github.com/sigstore/cosign/pull/641/files#diff-54166c9045199178b2f0a516fb36977ee5f3c21e25f80727b0f8b2cf24930ab0R238

verdverm:
Another thing to consider is required fields. There is an open proposal related to this in CUE: https://github.com/cue-lang/cue/issues/822

However, consider the following example modified from the one in #512 

`sigstore.go`
```
package main

import (
        "cuelang.org/go/cue"
        "cuelang.org/go/cue/cuecontext"
        cuejson "cuelang.org/go/encoding/json"
        "io/ioutil"
)

func main() {
        ctx := cuecontext.New()
        cueBs, _ := ioutil.ReadFile("policy.cue")
        v := ctx.CompileString(string(cueBs), []cue.BuildOption{}...)

        input := `{
              "foo": {
                      "bar": {
                              "baz": 10
                      }
               },
               "user": {
                       "email": "user@domain.com"
               }
         }`
        err := cuejson.Validate([]byte(input), v)
        if err != nil {
                panic(err)
        }
}
```

`policy.cue`
```
foo: bar: {
        baz: <11
        taz: int
}

user: role: "dev"
```

```
$ go run sigstore.go 
```

This has no output, which is passing the policy checks as I would assume as a user. Should this fail because the user lacks a role at all?

developer-guy:
> @developer-guy I was thinking that it would be nice if I could specify all my policies in a single CUE file and then select the ones I'd like to use for a given attestation. The `cue` CLI has this concept through the `-e` flag, or in Go with the `LookupPath` function. OPA looks to have similar ability to select out a subset of the full input.

**Actually, this one requires one additional flag to get the expression like **_"-e"_** and it is also helpful for the Rego implementation if we decide to support Rego because we need a query like an expression in Cue, so yeah this implementation would be helpful for us, WDYT @dlorenc?**

> Both CUE and Rego look to have the concept of modules and packages. It seems that both could be handled similarly, in that `cue.Load` was used to support multiple entrypoints which may include other files. I could, in theory, pass a directory in for a CUE entrypoint. I don't know if this is possible with Rego, but could cause issues with extension detection.

**Yeah, definitely we should support loading CUE or Rego policies from directories, thanks for the idea @verdverm. 🤝 If we decide to support Rego files we might use separate flags for these, like _"-cue"_ or _"-rego"_ to avoid this kind of problem like you said extension detection for traversing within the directory**

> Generally, it looks like the code is intended to be that each policy flag holds a separate policy. I think it would be helpful to clarify the ideas of entrypoints from how policies are written. The CUE validation passes all policies together during validation, for each iteration of the loop. I'm not familiar enough with the details yet to know if this is implementing the desired outcomes. How do the policy flags relate to the PredicateType here:
> https://github.com/sigstore/cosign/pull/641/files#diff-54166c9045199178b2f0a516fb36977ee5f3c21e25f80727b0f8b2cf24930ab0R238

**There are various types of [attestations](https://github.com/in-toto/attestation) provided by the in-toto community and cosign uses them internally. So, we thought that users might want to make a one-to-one relationship between the CUE policy file and the attestation's type to verify them like _"-policy slsaprovenance.cue -type slsaprovenance"_ because we need to know which policy is associated with which type of attestation. Additionally, you can store more than one attestation in the OCI registry for an image, so all of them will be validated with the same CUE policy file**

developer-guy:
> Another thing to consider is required fields. There is an open proposal related to this in CUE: [cue-lang/cue#822](https://github.com/cue-lang/cue/issues/822)
> 
> However, consider the following example modified from the one in #512
> 
> `sigstore.go`
> 
> ```go
> package main
> 
> import (
>         "cuelang.org/go/cue"
>         "cuelang.org/go/cue/cuecontext"
>         cuejson "cuelang.org/go/encoding/json"
>         "io/ioutil"
> )
> 
> func main() {
>         ctx := cuecontext.New()
>         cueBs, _ := ioutil.ReadFile("policy.cue")
>         v := ctx.CompileString(string(cueBs), []cue.BuildOption{}...)
> 
>         input := `{
>               "foo": {
>                       "bar": {
>                               "baz": 10
>                       }
>                },
>                "user": {
>                        "email": "user@domain.com"
>                }
>          }`
>         err := cuejson.Validate([]byte(input), v)
>         if err != nil {
>                 panic(err)
>         }
> }
> ```
> 
> `policy.cue`
> 
> ```cue
> foo: bar: {
>         baz: <11
>         taz: int
> }
> 
> user: role: "dev"
> ```
> 
> ```
> $ go run sigstore.go 
> ```
> 
> This has no output, which is passing the policy checks as I would assume as a user. Should this fail because the user lacks a role at all?

I think it would be better if @dlorenc takes this because IMHO it is ok, if you specify any field within the CUE file that is not in the attestation body, we should pass the validation but if we can support required fields, that would be good for the security perspective.

dlorenc:
Sorry I lost track of this one last week! I'll get some time to play with it this week and get it merged!

developer-guy:
> Sorry I lost track of this one last week! I'll get some time to play with it this week and get it merged!

It totally ok, we know you are busy 😋 but actually we are waiting for your thoughts about supporting to validate predicates by using Rego policies also, if you want us to do this, we are so excited to implement it 🥳🙋🏻‍♂️

dlorenc:
> It totally ok, we know you are busy 😋 but actually we are waiting for your thoughts about supporting to validate predicates by using Rego policies also, if you want us to do this, we are so excited to implement it 🥳🙋🏻‍♂️

Sure!

lukehinds:
> > Sorry I lost track of this one last week! I'll get some time to play with it this week and get it merged!
> 
> It totally ok, we know you are busy 😋 but actually we are waiting for your thoughts about supporting to validate predicates by using Rego policies also, if you want us to do this, we are so excited to implement it 🥳🙋🏻‍♂️

rego validation would be really useful as well, especially with its high adoption rate at present.

Dentrax:
We're not sure if it's good UX to allow to validate both CUE and Rego policies at the same time; we couldn't distinguish which UX is better:

```bash
# using specific flags; `foo.rego`, takes the precedence and will be executing first
$ cosign verify-attestation -cue foo.cue -rego bar.rego 

# passing engine flag
$ cosign verify-attestation -p foo.cue # default `-engine` is cue?
$ cosign verify-attestation -p foo.cue -engine cue   # valid
$ cosign verify-attestation -p bar.rego -engine rego # valid
$ cosign verify-attestation -p bar.cue -engine rego  # invalid

# directory example
$ cosign verify-attestation -p foo.rego -p bar.cue -p ./dir/to/rego/files/

# multi dir example (Regos and CUEs)
$ cosign verify-attestation -p foo.rego -p bar.cue -p ./dir/to/rego/files/ -p ./dir/to/cue/files/
```

Since we're traversing the `./dir/to/rego/files/` directory to load regos, what will happen if the directory contains a `CUE` file? And vice versa? (e.g., cue dir, but includes a rego)
@developer-guy

developer-guy:
Our changes are lost while migrating from `ffcli` to `cobra`, so we have to re-implement the whole logic again, FYI.

Maybe, we can talk a bit about how to do that again.

cc: @Dentrax @erkanzileli @dlorenc @n3wscott @mattmoor

n3wscott:
@developer-guy please take a look at this PR https://github.com/developer-guy/cosign/pull/1/

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -20,11 +20,14 @@ import (
...
 	"encoding/json"
 	"flag"
-	"io"
-
+	"fmt"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/pkg/errors"
+	"github.com/sigstore/cosign/pkg/cosign/rego"
+	"io"
```
n3wscott, line 28: looks like you need to sort the imports

dlorenc (approved):

PR #646, opened 2021-09-11
feat: allow to verify-blob from urls

caarlos0:
A minor quality of life improvement, allowing the blob and signature to come from a url as well, so we can, for instance:

```sh
cosign ./cmd/cosign/ verify-blob \
              -key https://raw.githubusercontent.com/caarlos0/goreleaserfiles/main/cosign.pub \
              -signature https://github.com/caarlos0/tasktimer/releases/download/v1.9.0/checksums.txt.sig \              https://github.com/caarlos0/tasktimer/releases/download/v1.9.0/checksums.txt
```

previous behaviors (from file and base64 string) still work as before.

### DIFF
--- a/cmd/cosign/cli/verify_blob.go
+++ b/cmd/cosign/cli/verify_blob.go
@@ -27,7 +27,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


lukehinds:
I would consider the security impacts here and consider using rekor instead (gather the signaturess by look up using digest). If someone compromises github.com/caarlos0 then they can dupe users into accepting a compromised container / blob and its respective pub key and signature etc with no open transparency for others to monitor.

lukehinds (changes requested):
would like to just hold for a bit while we discuss this over.

dlorenc:
There are no issues here IMO, I think you're misunderstanding the patch based on the example in the commit message. This would work with the transparency log if the experimental variable is set.

@lukehinds could you explain the actual threat you're worried about in more detail? I don't think this introduces anything new or makes anything worse.

This command would be run before a rekor-cli lookup, or use the built in transparency support if that's desired/important.

lukehinds:
You're right, did not articulate my comment to well, there is no new attack. For me it's a bit more about encouraging users into good habits to utilize rekor for public transparency (and detect key compromise etc). But I can see how this patch is a layer above that and its contingent upon cosign_experimental being set. 

> This command would be run before a rekor-cli lookup

I might be misunderstanding verify-blob here, is this not to verify the blog signing, if so I would have thought you would want to check its in the tlog first?

lukehinds (commented):
toggle off hold after speaking with Dan.

lukehinds (approved):

PR #652, opened 2021-09-12
Support GCP environments without workload identity (GCB).

dlorenc:
This requires the use of an environment variable called "GOOGLE_SERVICE_ACCOUNT_NAME"
to direct the client on which account to impersonate.


mattmoor (approved):
Should we document how to use this in `KEYLESS.md`, since it is a bit tricky to get set up?

mattmoor (approved):
Awesome, thanks for the docs!

PR #667, opened 2021-09-14
Move `LoadEcdsa...` into `pkg/cosign/keys.go`

mattmoor:
Chains is consuming this public function, but because `./cmd/cosign/cli` pulls in `glog` we hit the `-log_dir` problem.  With this change it can drop it's dependency on `cli`.

cc @priyawadhwa @dlorenc @dekkagaijin

### DIFF
--- a/cmd/cosign/cli/keys.go
+++ b/cmd/cosign/cli/keys.go
@@ -17,10 +17,6 @@ package cli
 import (
 	"context"
 	"crypto"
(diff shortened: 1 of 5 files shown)


mattmoor:
Whoops, left in some bonus (dead) code.

mattmoor:
Should be cleaned up.

dekkagaijin (approved):

Issue #311, opened 2021-05-07
Allow specifying an insecure repo for sign/verify 


priyawadhwa:
We'd have to do this if we wanted to run the tests that require a registry for tekton chains

Ref:  https://github.com/tektoncd/chains/pull/70

loosebazooka:
I can look at this

jonjohnsonjr:
Why does it have to be insecure?

loosebazooka:
I think I might want to try to get the test infrastructure to bring up a "trusted" registry to see if that can work?

priyawadhwa:
Yah so basically the Chains integration tests run on a GKE cluster but don't have credentials for GCR, so I set up an insecure registry in-cluster that we could push signatures to. We'd basically need this feature to be able to run `cosign verify` as a Task as the last step of the integration test.

Right now we get around needing this by port-forwarding the registry to a local port and using that for verification (instead of verifying in-cluster), and it seems to be working fine (although the port forwarding is pretty flaky). tbh since the port forwarding is a passable solution idk how necessary this is anymore.

jonjohnsonjr:
I'd suggest having a self-signed cert for the registry and getting cosign to trust it via `SSL_CERT_FILE` or something.

dlorenc:
Isn't this what crane has this for: https://github.com/google/go-containerregistry/blob/7e30746dde2f5a19b609f885dfef4e7b1d1503ad/pkg/name/registry.go#L35 ?

jonjohnsonjr:
Yes, and it's lazy, and you shouldn't propagate it.

PR #669, opened 2021-09-14
add `-allow-insecure-registry` flag to permit unsecured container registries

dekkagaijin:
Fixes https://github.com/sigstore/cosign/issues/311

### DIFF
--- a/cmd/cosign/cli/flags.go
+++ b/cmd/cosign/cli/flags.go
@@ -15,7 +15,15 @@
 
 package cli
 
-import "reflect"
+import (
+	"context"
+	"crypto/tls"
+	"flag"
+	"net/http"
+	"reflect"
+
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
 
 // oneOf ensures that only one of the supplied interfaces is set to a non-zero value.
 func oneOf(args ...interface{}) bool {
@@ -31,4 +39,20 @@ func nOf(args ...interface{}) int {
 		}
 	}
 	return n
+}
+
+type RegistryOpts struct {
+	AllowInsecure bool
+}
+
+func (co *RegistryOpts) GetRegistryClientOpts(ctx context.Context) []remote.Option {
+	opts := defaultRegistryClientOpts(ctx)
+	if co != nil && co.AllowInsecure {
+		opts = append(opts, remote.WithTransport(&http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}})) // #nosec G402
+	}
+	return opts
+}
+
+func ApplyRegistryFlags(regOpts *RegistryOpts, fs *flag.FlagSet) {
+	fs.BoolVar(&regOpts.AllowInsecure, "allow-insecure-registry", false, "whether to allow insecure connections to registries. Don't use this for anything but testing")
 }
--- a/cmd/cosign/cli/generate.go
+++ b/cmd/cosign/cli/generate.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"os"
 
-	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/peterbourgon/ff/v3/ffcli"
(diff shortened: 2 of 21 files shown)


priyawadhwa (commented):
lgtm! you could write a quick test by starting an insecure registry [in docker](https://docs.docker.com/registry/deploying/#run-a-local-registry), pushing an image, signing & verifying it (that's basically why we needed this in chains)

dekkagaijin:
@priyawadhwa 
```
jsand:cosign jsand$ cosign sign -key cosign.key localhost:5000/my-alpine
Pushing signature to: localhost:5000/my-alpine:sha256-69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a.sig
jsand:cosign jsand$ cosign verify -key cosign.pub localhost:5000/my-alpine

Verification for localhost:5000/my-alpine --
{"critical":{"identity":{"docker-reference":"localhost:5000/my-alpine"},"image":{"docker-manifest-digest":"sha256:69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a"},"type":"cosign container image signature"},"optional":null}
```

dekkagaijin:
@priyawadhwa 

```
jsand:cosign jsand$ ./cosign sign -key cosign.key localhost:443/my-image
error: signing [localhost:443/my-image]: getting remote image: Get "https://localhost:443/v2/": x509: certificate is not valid for any names, but wanted to match localhost; Get "http://localhost:443/v2/": net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
jsand:cosign jsand$ ./cosign sign -allow-insecure-registry -key cosign.key localhost:443/my-image
Enter password for private key: Pushing signature to: localhost:443/my-image:sha256-69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a.sig
jsand:cosign jsand$ ./cosign verify -allow-insecure-registry -key cosign.pub localhost:443/my-image

Verification for localhost:443/my-image --

[{"critical":{"identity":{"docker-reference":"localhost:443/my-image"},"image":{"docker-manifest-digest":"sha256:69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a"},"type":"cosign container image signature"},"optional":null}]
jsand-macbookpro3:cosign jsand$ 
```

cmd/cosign/cli/generate.go
```diff
@@ -59,18 +60,18 @@ EXAMPLES
...
 
-func GenerateCmd(_ context.Context, imageRef string, annotations map[string]interface{}, w io.Writer) error {
+func GenerateCmd(ctx context.Context, regOpts RegistryOpts, imageRef string, annotations map[string]interface{}, w io.Writer) error {
 	ref, err := name.ParseReference(imageRef)
 	if err != nil {
 		return err
 	}
 
-	get, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	get, err := remote.Get(ref, regOpts.GetRegistryClientOpts(ctx)...)
```
mattmoor, line 74: Is dropping the LHS intentional, since IIRC it's the default behavior?

mattmoor (approved):

cmd/cosign/cli/generate.go
dekkagaijin: ?

The effect of the change on this line specifically should be to a) propagate ctx, b) allow for broken TLS when the flag is set

cmd/cosign/cli/flags.go
```diff
@@ -32,3 +40,19 @@ func nOf(args ...interface{}) int {
 	}
 	return n
 }
+
+type RegistryOpts struct {
+	AllowInsecure bool
+}
+
+func (co *RegistryOpts) GetRegistryClientOpts(ctx context.Context) []remote.Option {
+	opts := DefaultRegistryClientOpts(ctx)
```
mattmoor, line 49: I wonder whether it's worth getting rid of this helper?  Just inline it so that folks use the appropriate new one?

cmd/cosign/cli/generate.go
mattmoor: I think what confused me was that the LHS wasn't using `DefaultRegistryClientOpts` and the implementation of this function delegates to that which hides the fact that it is a superset.  I left a comment on that below, let's kill `DefaultRegistryClientOpts` completely if we can.

dekkagaijin: roger

cmd/cosign/cli/flags.go
dekkagaijin: It _would_ gently nudge people in the right direction...

priyawadhwa (approved):

PR #674, opened 2021-09-15
Several cleanups to `SignCmd`

mattmoor:
1. Use the `DefaultRegistryClientOpts` which is a superset of what we use now, but sets `User-Agent`,
2. Pre-allocate the `toSign` array to have space for at least the number of images passed in (only more if recursive),
3. Use `Digest` to access the initial hash, which saves a `remote.Get` when we're passed a digest and non-recursive (at the cost of two requests when we are passed a tag in recursive mode),
4. Change the way we iterate over `toSign` to be conventional (not sure why we were doing it the other way before, but I'm guessing the two loops we now have were previously fused).

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -33,7 +33,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/peterbourgon/ff/v3/ffcli"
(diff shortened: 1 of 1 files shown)


cmd/cosign/cli/sign.go
```diff
@@ -259,12 +258,9 @@ func SignCmd(ctx context.Context, ko KeyOpts, annotations map[string]interface{}
 		}
 	}
 
-	remoteOpts := []remote.Option{
-		remote.WithAuthFromKeychain(authn.DefaultKeychain),
-		remote.WithContext(ctx),
-	}
+	remoteOpts := DefaultRegistryClientOpts(ctx)
```
dekkagaijin, line 261: https://github.com/sigstore/cosign/pull/669 :p

cmd/cosign/cli/sign.go
```diff
@@ -259,12 +258,9 @@ func SignCmd(ctx context.Context, ko KeyOpts, annotations map[string]interface{}
...
 
-	var toSign []name.Digest
+	toSign := make([]name.Digest, 0, len(imgs))
```
dekkagaijin, line 263: unnecessary, but not worse than what exists. @jonjohnsonjr assures me that the number of recursive images is unbounded

cmd/cosign/cli/sign.go
```diff
@@ -304,9 +304,7 @@ func SignCmd(ctx context.Context, ko KeyOpts, annotations map[string]interface{}
 		}
 	}
 
-	for len(toSign) > 0 {
```
dekkagaijin, line 307: good catch, didn't simplify this after refining the recursion

dekkagaijin (approved):

PR #680, opened 2021-09-15
Add a simple "DO NOT SUBMIT" workflow.

mattmoor:
#### Summary

This is a workflow I wrote for Knative, which is based on a presubmit check we had at Google.

Essentially, it fails the check (and flags the line) if it sees the words `DO NOT SUBMIT`.

So a comment practice with "work in progress" changes is to tag things that shouldn't get merged with:
```go
// TODO(mattmoor): DO NOT SUBMIT
```

Personally, I find this kind of workflow keeps me honest about things, so I thought I'd share!

#### Ticket Link

N/A

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/donotsubmit.yaml
+++ b/.github/workflows/donotsubmit.yaml
@@ -1,0 +1,44 @@
+name: Do Not Submit
+
+on:
+  pull_request:
+    branches: [ 'main', 'release-*' ]
+
+jobs:
+
+  donotsubmit:
+    name: Do Not Submit
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


mattmoor:
This is the source of truth:
https://github.com/knative-sandbox/.github/blob/main/workflow-templates/knative-donotsubmit.yaml#L18

I'll push a second commit and screenshot it here to show this in action, because I think it's pretty neat!

mattmoor:
Here's an example of what you get out the other side:
image

dekkagaijin (approved):

PR #681, opened 2021-09-15
Run `gofmt -s` and `goimports` against PRs.

mattmoor:
This change runs a few of the standard Go linting tools against PRs, and instead of just failing it annotates files that contain problems so users can run `gofmt -s -w <filename>` (or `goimports`).

The annotations look like this: https://github.com/sigstore/cosign/pull/680#issuecomment-920424768

This is a nice little hygiene check that we can use to keep things consistent.

This is based on the following, which we used in Knative: https://github.com/knative-sandbox/.github/blob/7d63e57ca2793794eef0b7a44035818f37f5f448/workflow-templates/knative-style.yaml#L26-L85

#### Ticket Link

N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -1,0 +1,68 @@
+name: Code Style
+
+on:
+  pull_request:
+    branches: [ 'main', 'release-*' ]
+
+jobs:
+
+  autoformat:
+    name: Auto-format and Check
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false # Keep running if one leg fails.
+      matrix:
+        tool:
+          - goimports
+          - gofmt
+
+        include:
+        - tool: gofmt
+          options: -s
+        - tool: goimports
+          importpath: golang.org/x/tools/cmd/goimports
+
+    steps:
+      - name: Set up Go 1.16.x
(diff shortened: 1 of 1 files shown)


mattmoor:
Hmm, this may be covered by `golangci-lint`.  Didn't see that since it's in `test.yaml` 🤔

dekkagaijin (approved):

mattmoor:
It's probably worth still doing this for the better errors, and `-s` on `gofmt`, but I'm happy to drop part/all of this.

Not sure why we don't run these in Knative this way 🤔

cpanato:
the gofmt is covered in the golangci-lint: https://golangci-lint.run/usage/linters/#gofmt

and goimport as well: https://golangci-lint.run/usage/linters/#goimports

not against but need to see if worth to have another job

mattmoor:
@cpanato yeah, I noticed that shortly after sending this (since golangci-lint was next on my list, and I didn't notice it because it was in `test.yaml` 😅 )

I'm not sure it adds value, or whether `gofmt` in `golangci-lint` does `-s` (which is nice), but we can kill whatever isn't useful 👍

PR #685, opened 2021-09-15
Enable `errorlint` checks.

mattmoor:
This mainly enables (and fixes) the `errorlint` checks.

This also fixes a handful of other things I noticed as I was passing through:
 * `fmt.Errorf` without formatting -> `errors.New`
 * `'%s'` -> `%q`

#### Ticket Link

N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/pkg/cosign/remote/remote.go
+++ b/pkg/cosign/remote/remote.go
@@ -70,7 +70,8 @@ func DockerMediaTypes() bool {
 func SignatureImage(ref name.Reference, opts ...remote.Option) (v1.Image, error) {
 	base, err := remote.Image(ref, opts...)
 	if err != nil {
-		if te, ok := err.(*transport.Error); ok {
+		var te *transport.Error
+		if errors.As(err, &te) {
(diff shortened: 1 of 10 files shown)


pkg/cosign/remote/remote.go
```diff
@@ -70,7 +70,8 @@ func DockerMediaTypes() bool {
 func SignatureImage(ref name.Reference, opts ...remote.Option) (v1.Image, error) {
 	base, err := remote.Image(ref, opts...)
 	if err != nil {
-		if te, ok := err.(*transport.Error); ok {
+		te := &transport.Error{}
```
dekkagaijin, line 73: Can these be inlined?

dekkagaijin (approved):

pkg/cosign/remote/remote.go
mattmoor: I tried that, but I think because of the `{}` it cannot 😞

mattmoor: Now they are vars, so really no.

PR #689, opened 2021-09-16
consolidate `require` directives, `tidy`

dekkagaijin:
Certain versions of Go (at least within 1.16) don't play well with multiple require directives.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,60 +3,57 @@ module github.com/sigstore/cosign
 go 1.16
 
 require (
+	cloud.google.com/go v0.94.1 // indirect
+	cloud.google.com/go/kms v0.1.0 // indirect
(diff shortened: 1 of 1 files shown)


dlorenc:
I don't actually think this helps, we ended up undoing this because it broke dependabot. We need to be building/releasing with go 1.17, and afaik the go1.16 directive at the top of the go.mod is all we need to indicate backwards compatibility: https://golang.org/ref/mod#go-mod-file-go

dekkagaijin:
@dlorenc hrm ok. The error I was seeing before explicitly called out multiple `require`s as the issue. I'll create the presubmit for 1.16 which will settle things

dekkagaijin:
looks like this is fine w/ 1.16 /shrug

Issue #679, opened 2021-09-15
Support `verify`ing arbitrary PodSpec-like k8s Objects 


dekkagaijin:
Right now, `verify-manifest` supports [a set of built-in Kubernetes Objects](https://github.com/sigstore/cosign/blob/39a1a16788b54f602f121165ac472a8b13a52a2f/cmd/cosign/cli/verify_manifest.go#L154-L209) which are known to define images. Ideally, we'd use some form of 🦆 typing to support verifying images in arbitrary PodSpec-like Objects. This will allow us to elegantly support custom resources and updates to the k8s API as it evolves over time.

PR #697, opened 2021-09-16
cosign verify-manifest now supports Jobs and CRD Types that fit within PodSpec, PodSpecTemplate, or JobSpecTemplate

n3wscott:
#### Summary
Rather than using the known Kubernetes types, to extract images from inbound manifests using `cosign verify-manifest`, we can use a simple duck type to union the types into just the fields we need and extract the image fields directly. 

As a result of this change, we can now verify the manifests of things like Knative Serving.

#### Ticket Link
Fixes https://github.com/sigstore/cosign/issues/679

#### Release Note

```release-note
verify-manifest now supports verifying Jobs, and any other type that implements the PodSpec, PodSpecTemplate, or JobSpecTemplate shapes.
```

### DIFF
--- a/cmd/cosign/cli/verify_manifest.go
+++ b/cmd/cosign/cli/verify_manifest.go
@@ -27,21 +27,15 @@ import (
 
 	"github.com/peterbourgon/ff/v3/ffcli"
 	"github.com/pkg/errors"
-	appsv1 "k8s.io/api/apps/v1"
-	batchv1 "k8s.io/api/batch/v1"
-	"k8s.io/api/batch/v1beta1"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
-// VerifyCommand verifies all image signatures on a supplied k8s resource
+// VerifyManifestCommand verifies all image signatures on a supplied k8s resource
 type VerifyManifestCommand struct {
 	VerifyCommand
 }
 
-// Verify builds and returns an ffcli command
+// VerifyManifest builds and returns an ffcli command
 func VerifyManifest() *ffcli.Command {
 	cmd := VerifyManifestCommand{VerifyCommand: VerifyCommand{}}
 	flagset := flag.NewFlagSet("cosign verify-manifest", flag.ExitOnError)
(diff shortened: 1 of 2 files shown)


mattmoor:
We've got `knative.dev/pkg` for `cosigned`, so you could just use `duckv1.Pod` and `duckv1.WithPod`.

mattmoor:
See: https://github.com/sigstore/cosign/blob/fb04df80b945209e4d186fb52f1391f2170d4cc9/cmd/cosign/webhook/main.go#L76-L82

n3wscott:
Ok awesome, let me switch to those.

n3wscott:
Na, duckv1.Pod/WithPod does not help here, we are only interested in the image. I am specifically union-ing the duck types for this extraction.

mattmoor:
honestly parsing it twice feels less gross than your crazy struct beast 🤣

n3wscott:
It would be parsing three times, or using the scheme. This is very simple: ```type imageContainer struct {
	Spec struct {
		Containers []struct {
			Image string
		}
		InitContainers []struct {
			Image string
		}
		Template struct {
			Spec struct {
				Containers []struct {
					Image string
				}
				InitContainers []struct {
					Image string
				}
			}
		}
		JobTemplate struct {
			Spec struct {
				Template struct {
					Spec struct {
						Containers []struct {
							Image string
						}
						InitContainers []struct {
							Image string
						}
					}
				}
			}
		}
	}
}```

I could break it apart if you like, I like the idea of the decoder getting to throw away the majority of the keys that come in.

cmd/cosign/cli/verify_manifest.go
```diff
@@ -116,95 +110,91 @@ func (c *VerifyManifestCommand) Exec(ctx context.Context, args []string) error {
...
+		addImage(c.Image)
+	}
 
-			for _, c := range containers {
-				images = append(images, c.Image)
+	return images
+}
+
+func getImagesFromYamlManifest(manifest []byte) ([]string, error) {
+	dec := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(manifest), 4096)
```
mattmoor, line 186: I'm sort of puzzled by how this works without the annotations 🤔

cmd/cosign/cli/verify_manifest.go
```diff
@@ -116,95 +110,91 @@ func (c *VerifyManifestCommand) Exec(ctx context.Context, args []string) error {
...
-	deserializer := serializer.NewCodecFactory(cScheme).UniversalDeserializer()
-	for {
-		ext := runtime.RawExtension{}
-		if err := dec.Decode(&ext); err != nil {
-			if errors.Is(err, io.EOF) {
-				break
+// imageContainer is the union type that match PodSpec, PodSpecTemplate, and
+// JobSpecTemplate; but filtering all keys except for `Image`.
+type imageContainer struct {
+	Spec struct {
```
mattmoor, line 116: I'd be curious what this did if you just inlined both of `duckv1.Pod` and `duckv1.WithPod` 🤔  😅  🤣

cmd/cosign/cli/verify_manifest.go
n3wscott: the marshaler assumes `camelCase` for the keys, the json keys are only needed when you want to control a behavior or  key name (like you want `init_containers` vs the default `initContainers`).

cmd/cosign/cli/verify_manifest.go
n3wscott: I did try, and you can't because both `duckv1.Pod` and `duckv1.WithPod` define the common metadata and only one wins. The decoder gets real confused 🤣 .

n3wscott: I could create my own union type and break the `duckv1.Pod` components up and make a new parent wrapper like I do here, but this usage just is not complicated enough to bother

mattmoor:
I don't really care, it just seems super verbose, and I was hoping for better.  We can always iterate later.

Do you want to add a ksvc test, since it should work?  Is there one already maybe?

mattmoor:
Also, I'd be curious if this worked with CronJob, which isn't PodSpecable 🙄

dekkagaijin:
> I was hoping for better

Very poor first showing from the new Scott. Really makes you miss the old Scott

cmd/cosign/cli/verify_manifest.go
mattmoor: Yeah, I wouldn't worry about it.

n3wscott:
> Also, I'd be curious if this worked with CronJob, which isn't PodSpecable 🙄

and I test for cron jobs: https://github.com/sigstore/cosign/pull/697/files#diff-4756f111371f1063a1bc2510070a8b51ebe6d2dd2c45b28cc0c785ff1eeacd34R205-R230


> I don't really care, it just seems super verbose, and I was hoping for better. We can always iterate later.
> 
> Do you want to add a ksvc test, since it should work? Is there one already maybe?

I test with a made up type: https://github.com/sigstore/cosign/pull/697/files#diff-4756f111371f1063a1bc2510070a8b51ebe6d2dd2c45b28cc0c785ff1eeacd34R121-R131

😎

n3wscott:
> > I was hoping for better
> 
> Very poor first showing from the new Scott. Really makes you miss the old Scott

ok I can do better. One sec.

n3wscott:
tell me you have used prow without telling me you used prow...

🤣

mattmoor (approved):

dekkagaijin (approved):
Welcome to the club!

PR #702, opened 2021-09-17
Define a shared `SignedEntity` interface.

mattmoor:
Starting to look at things like `remote.Get`, I realized that it will be possible for us to return either a `SignedImage` or `SignedImageIndex` and the caller can still access `Signatures()` and `Attestations()`.

#### Ticket Link

Related: https://github.com/sigstore/cosign/pull/701

#### Release Note
```release-note
NONE
```


dekkagaijin:
Arguably `Signature`s and `Attestation`s should implement signed entity, rather than the other way around

mattmoor:
Those are the signatures, the Image or Index is the entity that is signed... 🤔

dlorenc:
I like it!

dlorenc (approved):

PR #708, opened 2021-09-17
Skip empty values, not just missing keys.

mattmoor:
One of my previous changes switch from skipping `""` fetched from the map to using this syntax:

```go
blah, ok := themap[thekey]
if !ok {
...
}
```

However, it seems like there may be paths where we actually have these keys, but they contain empty string.

I'm seeing an index-out-of-bounds with both open PRs and HEAD trying to `cosign verify dockerfile` with `gcr.io/distroless/base`.

I'm sending this while I do a big more digging into the signatures there, but wanted to eliminate this delta quickly to avoid tripping others up.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/internal/oci/remote/remote.go
+++ b/internal/oci/remote/remote.go
@@ -116,8 +116,8 @@ func (s *sigLayer) Base64Signature() (string, error) {
 
 // Cert implements oci.Signature
 func (s *sigLayer) Cert() (*x509.Certificate, error) {
-	certPEM, ok := s.desc.Annotations[certkey]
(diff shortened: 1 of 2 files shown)


dekkagaijin (approved):

mattmoor:
Hmm, another fresh install of HEAD may have just worked for me, so the title may be more sensational than reality

mattmoor:
I'm going to give this a more accurate title and repush.  I think it's still worth doing this, but not as urgent as I thought.

PR #716, opened 2021-09-18
Fix go install method of installation

### LINKED ISSUE #588: Error when downloading cosign
**Description**

I'm getting this error when I try to download cosign.

```
$ go install github.com/sigstore/cosign/cmd/cosign@latest
go install github.com/sigstore/cosign/cmd/cosign@latest: github.com/sigstore/cosign@v1.1.0
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
````

### PR DESCRIPTION
**Pull request by luhring**
#### Summary

Cosign's installation instructions tell users to run `go install github.com/sigstore/cosign/cmd/cosign@latest`. This installation method appears to have been broken since fb04df8 (via #637), when a `replace` directive was added to `go.mod`.

The `replace` block was downgrading a few K8s modules (`k8s.io/api`, `k8s.io/apimachinery`, and `k8s.io/client-go`) to a slightly older version (from `v0.22.1` to `v0.21.4`). Presumably, this was done because #637 also introduced a dependency on the module `knative.dev/pkg`, which is not compatible with `v0.22.1` of the K8s modules.

A more stable alternative solution to this problem is to adjust the K8s module versions in the `require` block to the desired version (`v0.21.4`). Because of how [MVS](https://research.swtch.com/vgo-mvs) works, once none of the K8s modules in the module graph need a version as new as `v0.22.1`, Go will select version `v0.21.4` for these modules, which resolves the issue with `knative.dev/pkg`, which allows _all_ cosign builds to succeed.

Thus, this PR makes the above updates to the `require` block and removes the `replace` block.

#### Testing

This fix can be verified in a clean environment via Docker. This command installs Cosign using a testing branch in my fork of Cosign:

```
docker run --rm --env 'GOPRIVATE=github.com/luhring/cosign' golang:1.16 sh -c 'go install github.com/luhring/cosign/cmd/cosign@fix-go-install--testing && cosign --help'
```

This test:

- uses Go 1.16, but I've also tested with 1.17
- sets `GOPRIVATE` in order to bypass the Go proxy, to ensure the fork code is used directly
- uses a [slightly different branch](https://github.com/luhring/cosign/compare/fix-go-install...luhring:fix-go-install--testing) than this PR uses: `fix-go-install--testing`. This is because I needed to set the module name in `go.mod` to `github.com/luhring/cosign` so that `go install` doesn't error out early.
- should show Cosign's CLI help at the end if the installation was successful

#### Ticket Link

Fixes #588 

#### Release Note

```release-note
Fix go install method of installation
```

Signed-off-by: Dan Luhring <dan.luhring@anchore.com>


dlorenc:
Wow, thank you! This is great work!

dlorenc (approved):

PR #724, opened 2021-09-20
Move sget -key to --key

n3wscott:
#### Summary

In accordance with https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html POSIX flags and arguments are to be single dash short code, or double dash long code. `-key` is not a valid option. Replaced with `--key`

#### Ticket Link

Related to #709 

#### Release Note

```release-note
BREAKING API CHANGE: sget flags have been migrated to POXIS style, `-key` --> `--key`; `-output` --> `--output` or `-o`.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ Fetching contents without verifying them is dangerous, so we require the artifac
 $ sget gcr.io/dlorenc-vmtest2/artifact
 error: public key must be specified when fetching by tag, you must fetch by digest or supply a public key
 
-$ sget -key cosign.pub us.gcr.io/dlorenc-vmtest2/readme > foo
+$ sget --key cosign.pub us.gcr.io/dlorenc-vmtest2/readme > foo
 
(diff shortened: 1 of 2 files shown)


n3wscott:
This PR is to fix the e2e tests that only run post-submit.

n3wscott:
I have a better idea. Hold

n3wscott:
I added a deprecation warning while manually migrating the flags passed that are currently supported but in the wrong form.

dlorenc (approved):

dekkagaijin (approved):

dekkagaijin:
Hrm we'll want to do the same when we migrate `cosign`. Seems like we'll want to go ahead and refactor shared flags at the same time to make the job saner

n3wscott:
> Hrm we'll want to do the same when we migrate `cosign`. Seems like we'll want to go ahead and refactor shared flags at the same time to make the job saner

I am doing it now.

Issue #666, opened 2021-09-14
`cosign` as a library


mattmoor:
**Description**

tl;dr I propose the creation of a well-curated set of libraries for signing, that compose extremely well with and complement `github.com/google/go-containerregistry` for signing OCI images and indices.

`cosign` has gained considerable traction as a mechanism for signing OCI artifacts as a standalone executable.  I believe that to kick this into an even higher-gear, we want to encourage ecosystem tooling to natively integrate cosign _**as a library**_ to naturally extend image production to include signing.

`cosign` (along with a growing list of other tooling) builds around `github.com/google/go-containerregistry` (aka "GGCR") for interacting with OCI images and indices, so this isn't a huge reach, it is mostly refactoring the code we have today to better align with the way we do things in GGCR so that it is trivially simple for folks to adopt signing, if they've adopted GGCR.

---

One of the directions I'd like to see us push the library portion of this is to align with the ["philosophy" of GGCR](https://github.com/google/go-containerregistry#philosophy), in part to enable a wider variety of use cases.

For example, today `cosign` is very oriented around registry-based images.  Despite it's name GGCR actually allows interacting with images in a wide variety of formats, including things like `docker save`-tarballs, OCI-layouts, and more.

If cosign were to adhere to some of the principles of GGCR, we might see signing take the form of `(v1.Image, Options) -> v1.Image`, where the input image is what we want signed and the output image is the signature payload we want recorded somewhere.

With this separation of the signing from the storage, consumers can then leverage a variety of mechanisms to publish the resulting image in different ways.  They can choose to `remote.Write` that image to a registry following `cosign`'s naming convention, or they can `tarball.Write`.

---

GGCR is all about enabling cool new things thru composition, and so this is very much about making sure `cosign` becomes a tool compatible with that ecosystem of composition!

I am still wrapping my head around everything that `cosign` does, so this picture is starting to form in my head, but if folks are receptive to this (or better willing to help) then I'd love to discuss this direction.

cc @jonjohnsonjr @imjasonh @dekkagaijin @dlorenc

dlorenc:
@jonjohnsonjr who could have seen this coming?

mattmoor:
Only someone with vision 😏

dekkagaijin:
Gonna add another +1 here to the idea of composition :p

samj1912:
[...]

PR #729, opened 2021-09-20
Start to build up a `mutate` package.

mattmoor:
This is not yet used, but start to lay out some useful utilities for walking over and manipulating Images and Indices.

The first bit is `mutate.AppendManifests`, which builds around the utility with the same name in GGCR.  This utility produces an `oci.SignedImageIndex`, where (because of the mutation) the signatures are necessarily empty, but which provides access to the signatures of the entities contained within it.  This was mostly useful for constructing images for tests.

The second bit is `mutate.Map`, which surfaces a way to apply a `Mutator` function to the `oci.SignedEntity`s contained within a particular `oci.SignedEntity`.  Notable features:
 * `Mutator`s can control whether they descend into the children of indices (useful for recursive signing or signature verification).
 * `Mutator`s can filter entities by returning `nil` without `error`.
 * `Mutator`s are called before and (if changed) after its children are walked, which is detectable via helpers on `ctx`.
 * `Mutator`s can return their input entity if they are just a simple readonly walk.

#### Ticket Link

Related: https://github.com/sigstore/cosign/issues/666

#### Release Note
```release-note
NONE
```


internal/oci/mutate/mutate.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/sigstore/cosign/internal/oci"
+	"github.com/sigstore/cosign/internal/oci/empty"
+	"github.com/sigstore/cosign/internal/oci/signed"
+)
+
+// Appendable is our signed version of mutate.Appendable
+type Appendable interface {
+	oci.SignedEntity
+
+	MediaType() (types.MediaType, error)
```
dekkagaijin, line 33: Is there a reason for this to ever return an error?

internal/oci/mutate/map.go
```diff
@@ -0,0 +1,167 @@
...
+	return fn(after(ctx), result)
+}
+
+// This is used to associate which pass of the Map a particular
+// callback is being invoked for.
+type mapPassKey struct{}
+
+// before decorates the context such that IsBeforeChildren(ctx) is true.
+func before(ctx context.Context) context.Context {
+	return context.WithValue(ctx, mapPassKey{}, "before")
```
dekkagaijin, line 148: turn "before" and "after" into consts and/or enums?

internal/oci/mutate/map.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if !changed {
+		return parent, nil
+	}
+
+	// Construct a new ImageIndex from the new consituent signed images.
+	result := AppendManifests(mutate.IndexMediaType(empty.Index, im.MediaType), adds...)
+
+	// Since the children changed, give the callback a crack at the new image index.
+	return fn(after(ctx), result)
```
dekkagaijin, line 139: Should make sure we're also preserving manifest-level annotations: https://github.com/opencontainers/image-spec/blob/main/annotations.md

internal/oci/mutate/mutate.go
mattmoor: Rather than explain that this is GGCR copy/pasta I inlined the GGCR interface I copied it from 😉

internal/oci/mutate/map.go
mattmoor: I was going to say "but I'll just call them before/after", but I can't because that's what the methods are called. 🙃 

Given the proximity of the functions, and the tests to make sure they agree, I'm sort of inclined to leave it 😅

internal/oci/mutate/map.go
mattmoor: IDK if we have a function for this in GGCR, lemme look.

At a minimum I'll TODO it.

mattmoor: Ok, I found `mutate.Annotations` that has a weird interface, but it gets the job done.

dekkagaijin (approved):

PR #732, opened 2021-09-20
Drop VerifyOpts from CheckOpts

mattmoor:
These were only ever used to pass `options.WithContext(ctx)` and we have `ctx`, so we simply pass `ctx` to `VerifySignature` and pass `options.WithContext(ctx)` directly.

We were also inconsistent about whether we even passed `VerifyOpts` to `VerifySignature`, so now things are both simpler and more consistent.

#### Ticket Link

N/A

#### Release Note

```release-note
Retire VerifyOpts in CheckOpts, which was only used to pass `options.WithContext`
```

### DIFF
--- a/copasetic/main.go
+++ b/copasetic/main.go
@@ -186,7 +186,6 @@ func main() {
 			ctxOpt := options.WithContext(bctx.Context)
 			co := &cosign.CheckOpts{
 				SigVerifier:        pubKey,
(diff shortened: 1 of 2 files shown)


dekkagaijin:
Just stating for the record that this is a trust fall exercise because the promised functionality replacement should be good

PR #737, opened 2021-09-21
Eliminate `AttachedImageTag` in favor of `ociremote` variants.

mattmoor:
We were starting to accumulate a large number of variants of this method, which all do roughly equivalent things.

This change consolidates around a handful of methods in the `ociremote` package oriented around the kind of attachment they represent.
 * `ociremote.SignatureTag(ref, opts...)` produces the tag for referencing attached signatures.
 * `ociremote.AttestationTag(ref, opts...)` produces the tag for referencing attached attestations.
 * `ociremote.SBOMTag(ref, opts...)` produces the tag for referencing attached SBOMs.

These share some core logic to be efficient, but they:
1. Check whether they were passed a digest,
2. If not, resolves the digest using the supplied options,
3. Construct the final tag based on the options (e.g. target repo) and the resolved or passed digest.

#### Ticket Link
Related: https://github.com/sigstore/cosign/issues/666

#### Release Note
```release-note
NONE
```


mattmoor:
cc @n3wscott since you just introduced this 😅

mattmoor:
This also eliminates `TargetRepositoryForImage` and drops the error returned by `ociremote.Option`.  The last bit should be a non-issue since only this package can define or consume them since `*options` is private.

n3wscott:
I just moved it, I have no ... Attachment to it

mattmoor:
> I just moved it, I have no ... Attachment to it

ZING!

dekkagaijin (approved):

PR #741, opened 2021-09-21
Add `Annotations` to `oci.Signature`.

mattmoor:
This change has two main parts, the first is largely as the title indicates, which is to add an `Annotations()` method to `oci.Signature` (and our implementations thereof).  This is largely motivated by the second half of this change.

The second part of this change is starting to orient `UploadSignature` to operate in terms of our `oci` interfaces.  For a much longer-form version of the direction I'd like to drive this towards, see here: https://github.com/sigstore/cosign/issues/666#issuecomment-923609067.  This is a bit messier than I'd like, but I think it is progress towards the final form outlined in the issue since it starts to orient things like deduplication around the `oci.Signature` interface.

#### Ticket Link

Related: Related: https://github.com/sigstore/cosign/issues/666

#### Release Note
```release-note
NONE
```

### DIFF
--- a/internal/oci/remote/layer.go
+++ b/internal/oci/remote/layer.go
@@ -36,6 +36,11 @@ type sigLayer struct {
 
 var _ oci.Signature = (*sigLayer)(nil)
 
+// Annotations implements oci.Signature
+func (s *sigLayer) Annotations() (map[string]string, error) {
+	return s.desc.Annotations, nil
+}
+
 // Payload implements oci.Signature
 func (s *sigLayer) Payload() ([]byte, error) {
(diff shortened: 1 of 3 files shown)


cpanato:
nice issue number

mattmoor:
Looks like I broke something about signature deduplication. 😅

mattmoor:
It looks like the e2e tests expect the caller to have put `bundle` in `annotations`, and the that to trigger a fresh upload, so I'm reducing the scope of my prior fix to just exclude the signature key from the annotations check, since that is what we specifically check specially below this loop.

dekkagaijin (approved):

PR #747, opened 2021-09-21
WIP Update sigstore root ux

asraa:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

* Only makes network calls when embedded or cached root is expired


asraa:
This is ready for review now

cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
```diff
@@ -64,22 +60,17 @@ func initRoots() *x509.CertPool {
...
-				panic("error creating root cert pool")
-			}
-		} else {
-			// TODO: Remove the string replace when SigStore root is updated.
-			replaced := strings.ReplaceAll(buf.String(), "\n  ", "\n")
-			if !cp.AppendCertsFromPEM([]byte(replaced)) {
-				panic("error creating root cert pool")
-			}
+		if err := tuf.GetTarget(ctx, fulcioTargetStr, &buf); err != nil {
+			panic(errors.Wrap(err, "creating root cert pool"))
```
dekkagaijin, line 68: Is there a reason why we're using panics instead of errors in this package?

asraa: This really shouldn't happen, but I'm not sure -- kind of was following existing code. We really should be getting these targets for any cosign verification, most of the callers expect these targets to be available.

I can plumb through errors, but it'll touch a bit of code

n3wscott:
So sorry, I moved these files to keep them isolated. Can you rebase?

asraa:
updated!

dlorenc:
Looks like there's still a conflict :(

asraa:
because somehow i never pushed :P thanks!

dlorenc:
Looks like just a few lint errors now!

PR #765, opened 2021-09-23
Bump go-containerregistry and repair prior crimes against immutability.

### LINKED ISSUE #718: `cremote.UploadFile` improperly sets config media type
**Description**

Reading through the code, I noticed that `cremote.UploadFile` sets the media type here by accessing the `Manifest()` and then setting `m.Config.MediaType = mt`.

https://github.com/sigstore/cosign/blob/03f3f4dd9ec039e3a90e5b12193bb4f2dc8cb408/pkg/cosign/remote/index.go#L155-L159

It is plausible that this works (due to a bug in GGCR), but it should not.  The intent of abstractions like `v1.Image` is that they are immutable, so being able to access and change part of the image should NOT be possible.  For example, if the `v1.Image` actually lived on a registry, mutating things like this would be ineffective (fetching the manifest again would yield the original value).

The expectation is that when folks want to change some aspect of an image they go through the `mutate` package, as the lines above this do, e.g. these:
https://github.com/sigstore/cosign/blob/03f3f4dd9ec039e3a90e5b12193bb4f2dc8cb408/pkg/cosign/remote/index.go#L143-L152

cc @jonjohnsonjr @imjasonh for the potential GGCR bug (note: I haven't confirmed this code actually works, but someone thought it did at some point 😅 )

dlorenc:
Matt and Jon caught our dirty tricks to change media types, and fixed the bug :)
This bumps ggcr and fixes the way we were setting config media types.

#### Ticket Link

Fixes #718 

#### Release Note

```release-note

```


imjasonh (approved):

cpanato (approved):

mattmoor:
Awesome, thanks @dlorenc

dekkagaijin:
booo

**DIR TY TRICKS**
**DIR TY TRICKS**
DIR-T-TRICKS

PR #745, opened 2021-09-21
Add `sget` to the release artifacts

rgreinho:
During the release process, builds `sget` along cosign, signs, verifies
and publishes the binaries.

Fixes sigstore/cosign#725

PR #766, opened 2021-09-23
Fix the CI

rgreinho:
After pushing sigstore/cosign#745, the signing/verifying mechanism for
the created binaries was broken. This patch fixes the issue and removes
the useless duplication, making the CI workflow DRYer.

### DIFF
--- a/.github/workflows/cross.yaml
+++ b/.github/workflows/cross.yaml
@@ -27,13 +27,16 @@ jobs:
         os: [macos-latest, ubuntu-latest, windows-latest]
         include:
           - os: macos-latest
-            TARGET: cosign-darwin-amd64
+            COSIGN_TARGET: cosign-darwin-amd64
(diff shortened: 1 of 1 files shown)


rgreinho:
@cpanato @dlorenc sorry it looks like I broke the CI with my last PR. This one should fix it.

dlorenc (approved):
Thanks!

.github/workflows/cross.yaml
```diff
@@ -47,22 +50,29 @@ jobs:
       #   if: matrix.os == 'ubuntu-latest'
       #   run: sudo apt-get update && sudo apt-get install -yq libpcsclite-dev
       - name: build cosign
```
cpanato, line 52: maybe we can add `build cosign & sget` or something like that, but it is a non blocking

PR #776, opened 2021-09-24
Ensure we resolve tags once.

mattmoor:
As I've been refactoring things, I noticed a handful of places where a single command potentially used a `name.Tag` multiple times for remote access.  There is an exceedingly small window here for a race to happen where the first tag access points to one image, and the subsequent access points to another.

I've already fixed a handful of these, and this change fixes one more (in `sget` IIRC), but also tries to add some defensive logic in a few places where we were already doing the right thing.

The defensive logic I added is to clobber `ref` with `digest` after resolving the reference:
```go
       digest, err := ociremote.ResolveDigest(ref, regOpts.ClientOpts(ctx)...)
       if err != nil {
               return err
       }
       // Overwrite "ref" with a digest to avoid a race where we use a tag
       // multiple times, and it potentially points to different things at
       // each access.
       ref = digest // nolint
```

This ensures that regardless of which reference is used below resolution, we always get what we resolved at this point.

There are some other superficial cleanups lumped in, which I noticed as I skimmed the code.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/cmd/cosign/cli/attach/sbom.go
+++ b/cmd/cosign/cli/attach/sbom.go
@@ -86,11 +86,8 @@ func SBOMCmd(ctx context.Context, regOpts options.RegistryOpts, sbomRef, sbomTyp
 	}
 
 	fmt.Fprintf(os.Stderr, "Uploading SBOM file for [%s] to [%s] with mediaType [%s].\n", ref.Name(), dstRef.Name(), sbomType)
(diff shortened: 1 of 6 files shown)


dekkagaijin (approved):

dekkagaijin:
We might wanna alter ~most of the library to take digests (name.Digest, v1.Descriptor)

PR #795, opened 2021-09-25
Change `UploadFiles` to return a `name.Digest`.

mattmoor:
Previously this defined and returned a `Digester`, and constructed the digest using the ref that was passed in.  It simplifies things a bunch to just have `UploadFiles` return the digest of the thing it uploaded.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/cmd/cosign/cli/upload/blob.go
+++ b/cmd/cosign/cli/upload/blob.go
@@ -102,19 +102,13 @@ func BlobCmd(ctx context.Context, regOpts options.RegistryOpts, files []cremote.
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


mattmoor:
I guess this also makes it use `lastHash.String()` to construct the URL (instead of assuming the algorithm is `sha256`) and makes `UploadFiles` use the passed `remoteOpts` to write the index.

dekkagaijin (approved):
(┛◉Д◉)┛彡┻━┻

mattmoor:
The test failure was because I had thought about inverting a condition and hadn't fully reverted the predicate, so we only returned an index when `len(files) == 1` 🤦 

Should pass now!

Issue #784, opened 2021-09-24
`cosigned` webhook should resolve tags to digests


mattmoor:
**Description**

`tl;dr` This is similar to https://github.com/sigstore/cosign/issues/648 (albeit with a less sensational title); essentially, if we allow tags to pass through the `cosigned` webhook, even if they are verified, we are leaving folks open to problems.

One option here would be to simply outright reject non-Digest requests, which we probably should be doing in pure validation contexts.

Another option here would be to actually support running `cosigned` as a mutating webhook as well (trivial with the knative infra), and simply mutate the image references to be their resolve digests as we verify them.

In fact, if we simply set:
```go
c.Image, err = ociremote.ResolveDigest(...)
```

Then when this logic is run as a mutating webhook, the infra will automagically synthesize the appropriate patches to affect the digest resolution.

cc @dlorenc @mlieberman85 @dekkagaijin

PR #799, opened 2021-09-25
Reject tags in `cosigned`.

mattmoor:
This change makes `cosigned` reject tag references, which can drift after validation has occured rendering the validation we perform ineffective.

This also adds unit test coverage for the bulk of the `webhook` package.

A subsequent change will introduce a mutating webhook that resolves tags to digests (we still need to validating webhook due to webhook ordering).

Signed-off-by: Matt Moore <mattomata@gmail.com>

#### Ticket Link

Related: https://github.com/sigstore/cosign/issues/784

#### Release Note
```release-note
The cosigned webhook now requires digest references to validate references.
```

PR #800, opened 2021-09-25
Add digest resolution to `cosigned`.

mattmoor:
This change introduces a mutating webhook to complement our validating webhook.

The validating webhook in #799 began rejecting tag reference because tags are mutable and can drift between validation and resolution by the kubelet.  This change introduces a mutating webhook that resolves tags to digests as resources are created, so that users aren't necessarily forced to provide digests, but we get the benefits of them nonetheless.

#### Ticket Link
Fixes: https://github.com/sigstore/cosign/issues/784

#### Release Note
```release-note
cosigned now includes a mutating webhook that resolves tags to digests
```

### DIFF
--- a/.github/workflows/kind-e2e-cosigned.yaml
+++ b/.github/workflows/kind-e2e-cosigned.yaml
@@ -64,6 +64,10 @@ jobs:
         sudo mv ko /usr/local/bin
         echo '::endgroup::'
 
+        echo '::group:: install yq'
+        go get github.com/mikefarah/yq/v4
+        echo '::endgroup::'
+
     - name: Check out code onto GOPATH
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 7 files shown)


mattmoor:
Marking this as WIP because of the remaining e2e TODO (w/ DO NOT SUBMIT), but I wanted to let the automation do a pass over things while I play with that 😁

mattmoor:
Okay, this should be RFAL.  I've updated the e2e test to check the digest resolution using a dry run on the resources.

mattmoor:
@dlorenc this is RFAL 😁

mattmoor:
Scratch that, the logs don't look right 🤔

mattmoor:
Ok, I fixed the script so it fails properly, and found the bug 🤦 

Turns out it doesn't matter if you resolve tags to digests if the digest still isn't signed!

I'm optimistic this will pass now 🤞

mattmoor:
💥 https://github.com/sigstore/cosign/pull/800/checks?check_run_id=3709807519#step:13:80

Alright, this is good to go 🎉

dlorenc (approved):
Nice!

Issue #807, opened 2021-09-27
Allow custom signature tag suffix via CLI parameter


mrjoelkamp:
**Description**
Add ability to customize the signature tag (prefix/suffix) before it gets pushed to the registry. 

When using [ECR Lifecycle Policies](https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html) you can filter image tags based on the tag prefix (i.e. `sha256-`). 

In my specific use case we have lifecycle policies for different environments (i.e. `develpment*`, `production*`). We keep `X` number of `develpment-*` and `production-*` tagged images. In order to persist the signatures associated with these container images, I need to be able to tag the signatures with a similar environment tag prefix. For example: `sha256.dev-[digest].sig`, to create a ECR Lifecycle policy for the tag prefix `sha256.dev` and keep X number of `deveopment` signatures and container images.

**Related Work**

Added `TagPrefix` and `TagSuffix` to `RegistryOpts` to allow customization of the signature tag where:
signature tag = `sha256[TagPrefix]-[Digest][TagSuffix]`

And the CLI flags `--signature-prefix` and `--signature-suffix` to set these options

PR #808, opened 2021-09-27
feat: add custom signature tag registry options

mrjoelkamp:
#### Summary

Added `TagPrefix` and `TagSuffix` to `TagOptions` and added `TagOptions` to `RegistryOptions` to pass custom tags to `ociremote.Options` to allow customization of the signature tag where:
signature tag = `[OptionalTagPrefix]sha256-[Digest].[TagExtension]`

And the CLI flags `--tag-prefix` and `--tag-suffix` to set these options 

#### Ticket Link

Fixes
https://github.com/sigstore/cosign/issues/807

cc: @n3wscott 

#### Release Note

```release-note
NONE
```


cmd/cosign/cli/options/flags.go
```diff
@@ -44,6 +44,8 @@ func NOf(args ...interface{}) int {
 // Deprecated: this will be deleted when the migration to cobra is finished.
 func ApplyRegistryFlags(regOpts *RegistryOptions, fs *flag.FlagSet) {
 	fs.BoolVar(&regOpts.AllowInsecure, "allow-insecure-registry", false, "whether to allow insecure connections to registries. Don't use this for anything but testing")
+	fs.StringVar(&regOpts.TagPrefix, "signature-prefix", "", "custom prefix to use for signature tag")
+	fs.StringVar(&regOpts.TagSuffix, "signature-suffix", "", "custom suffix to use for signature tag")
```
mattmoor, line 48: Hmm, these don't really feel like registry flags to me.

I think surfacing a way to customize suffix has come up a few times now, and this totally makes sense.  I think we should probably plumb through a similar change for attestations.

Especially if we leave the field in such a generic place, we should have the field name's specificity match the flag.  I could see `--suffix` setting `TagSuffix` or `--signature-suffix` setting `SignatureTagSuffix`, but as it is seems a bit odd.

mattmoor: cc @dekkagaijin @n3wscott  for thoughts on where this lives, and consistency with other flags.

cmd/cosign/cli/options/registry.go
```diff
@@ -36,10 +38,16 @@ var _ Interface = (*RegistryOptions)(nil)
...
+	cmd.Flags().StringVar(&o.TagSuffix, "signature-suffix", "", "custom suffix to use for signature tag")
 }
 
 func (o *RegistryOptions) ClientOpts(ctx context.Context) []ociremote.Option {
-	return []ociremote.Option{ociremote.WithRemoteOptions(o.GetRegistryClientOpts(ctx)...)}
+	if o.TagSuffix != "" {
+		return []ociremote.Option{ociremote.WithSignatureSuffix(o.TagSuffix), ociremote.WithSignaturePrefix(o.TagPrefix), ociremote.WithRemoteOptions(o.GetRegistryClientOpts(ctx)...)}
+	} else {
+		return []ociremote.Option{ociremote.WithSignaturePrefix(o.TagPrefix), ociremote.WithRemoteOptions(o.GetRegistryClientOpts(ctx)...)}
+	}
```
mattmoor, line 50: I guess I found it odd that this passes `TagPrefix` without a similar `""` check, which sort of bakes in the idea that `""` is the default.  Perhaps we should build this up?

pkg/oci/remote/options.go
```diff
@@ -78,6 +81,14 @@ func makeOptions(target name.Repository, opts ...Option) (*options, error) {
...
 }
 
+// WithSignaturePrefix is a functional option for overriding the default
+// signature tag prefix.
+func WithSignaturePrefix(prefix string) Option {
+	return func(o *options) {
+		o.SignaturePrefix = prefix
+	}
+}
+
```
mattmoor, line 91: A lot of the plumbing seems to be adding the prefix support, which I don't fully understand the need for, if folks can customize the suffix.  Seems like a lot of knobs, where just suffix would hopefully suffice.

cmd/cosign/cli/options/registry.go
mrjoelkamp: @mattmoor, I needed to conditionally check for `TagSuffix` because it would override `SignatureTagSuffix` to `""` instead of the default `.sig`. There wasn't a `TagPrefix` before so I defaulted it to `""` and don't need to perform the same check

pkg/oci/remote/options.go
mrjoelkamp: @mattmoor I started with suffix only, until I realized that ECR Lifecycle Policies only allow for [tag prefixes](https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lp_tag_prefix_list).

I added the prefix option to account for that and left in the suffix option in case someone wanted to use something different than `.sig`. I'd be okay will pulling the suffix code out but really need the prefix code at the moment.

mattmoor: `/me shakes fist at ECR`

I think I'd prefer we nail this for suffix (incl. stuff like `.att`) and then mirror everything for prefix.  Sorry if that's annoying.

cmd/cosign/cli/options/flags.go
mrjoelkamp: @mattmoor I agree, but the exiting suffix code was living in registry options so I extended it to include prefix and exposed the existing `TagSuffix` as a flag as well. `TagPrefix` is more important to me because of ECR Lifecycle Policy tag prefix filtering.

I'll push an update to make the `TagPrefix` signature, SBOM, and attestation agnostic.

pkg/oci/remote/options.go
mrjoelkamp: @mattmoor it sounds like this PR brought attention to a bigger code organization issue. I only exposed the existing `WithSignatureSuffix` to the CLI flags and added `WithPrefix` to the existing code base. Any chance we could merge this to get the feature in and then refactor how tags are formed in a separate PR?

samj1912:
Would it be better to introduce a gotemplate flag or something along those lines instead? It could be passed certain values like registry, digest, tag, image name etc and it would be left upto the user to have consistent templates during sign and verify?

dekkagaijin:
imho `[OptionalTagPrefix]sha256-[Digest].[TagExtension]` would be preferable, since it allows for "dumb"er implementation and keeps the conceptually atomic `sha256-[Digest]` in one piece.

We might consider allowing overriding `fn(image) -> attachment_tag` in the library, but for the CLI lets stick with `--attachment-tag-prefix`

mrjoelkamp:
> imho `[OptionalTagPrefix]sha256-[Digest].[TagExtension]` would be preferable, since it allows for "dumb"er implementation and keeps the conceptually atomic `sha256-[Digest]` in one piece.
> 
> We might consider allowing overriding `fn(image) -> attachment_tag` in the library, but for the CLI lets stick with `--attachment-tag-prefix`

@dekkagaijin I refactored a few things and I think that I have addressed most of the comments! How does it look?

cmd/cosign/cli/options/tags.go
```diff
@@ -0,0 +1,34 @@
...
+	TagSuffix string
+}
+
+var _ Interface = (*TagOptions)(nil)
+
+// AddFlags implements Interface
+func (o *TagOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.TagPrefix, "tag-prefix", "", "custom prefix to use for tags")
+	cmd.Flags().StringVar(&o.TagSuffix, "tag-suffix", "", "custom suffix to use for tags")
+}
```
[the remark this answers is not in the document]
dekkagaijin: @mattmoor wdty about standardizing on "attachment image tag suffix" == "attachment name"

mrjoelkamp:
> @mrjoelkamp If you resolve the docgen issue(`go run ./cmd/help/`), I'm happy to merge for now to save you more conflict grief

Awesome, thanks! Should be good to go now.

dekkagaijin (approved):

PR #817, opened 2021-09-28
Fix e2e postsubmit

dekkagaijin:
* Fixed regression in `copy`
* `-key` -> `--key`. Prevents the e2e script from throwing a ton of warnings

```release-note
NONE
```

### DIFF
--- a/cmd/cosign/cli/options/copy.go
+++ b/cmd/cosign/cli/options/copy.go
@@ -32,7 +32,7 @@ var _ Interface = (*CopyOptions)(nil)
 func (o *CopyOptions) AddFlags(cmd *cobra.Command) {
 	o.Registry.AddFlags(cmd)
 
-	cmd.Flags().BoolVar(&o.SignatureOnly, "sig-only", true,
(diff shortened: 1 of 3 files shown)


cmd/cosign/cli/options/copy.go
```diff
@@ -32,7 +32,7 @@ var _ Interface = (*CopyOptions)(nil)
 func (o *CopyOptions) AddFlags(cmd *cobra.Command) {
 	o.Registry.AddFlags(cmd)
 
-	cmd.Flags().BoolVar(&o.SignatureOnly, "sig-only", true,
+	cmd.Flags().BoolVar(&o.SignatureOnly, "sig-only", false,
```
mattmoor, line 35: Is this the regression?

mattmoor (approved):

cmd/cosign/cli/options/copy.go
dekkagaijin: yup, defaulted to false before migrating to cobra

PR #821, opened 2021-09-29
Make --cert-email actually do something

passcod:
#### Summary

Ref #622

The `--cert-email` option was never actually passed to verify's `CheckOpts`, so it never did anything.

#### Release Note

```release-note
Fixed the `--cert-email` option to actually work
```


cpanato (approved):
thanks! LGTM

PR #832, opened 2021-10-01
Fix `cosign attest` flags for `type`.

asraa:

#### Ticket Link

Fixes

#### Release Note

```release-note
* Bug fix: fixes --type flag for `cosign attest`
```

Previous behavior:
```
./cosign attest --predicate example.txt --key cosign.key --type custom gcr.io/asra-ali/busybox/demo
Error: signing gcr.io/asra-ali/busybox/demo: invalid predicate type: 
```

Now:
```
$ ./cosign attest --predicate example.txt --key cosign.key --type custom gcr.io/asra-ali/busybox/demo
Enter password for private key: Using payload from: example.txt
```

### DIFF
--- a/cmd/cosign/cli/options/predicate.go
+++ b/cmd/cosign/cli/options/predicate.go
@@ -32,6 +32,6 @@ func (o *PredicateOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.Path, "predicate", "",
 		"path to the predicate file.")
 
-	cmd.Flags().StringVar(&o.Path, "type", "custom",
+	cmd.Flags().StringVar(&o.Type, "type", "custom",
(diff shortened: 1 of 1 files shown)


cmd/cosign/cli/options/predicate.go
```diff
@@ -32,6 +32,6 @@ func (o *PredicateOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.Path, "predicate", "",
 		"path to the predicate file.")
 
-	cmd.Flags().StringVar(&o.Path, "type", "custom",
```
n3wscott, line 35: whooopps!! my bad, thanks for fixing!!

n3wscott (approved):

cmd/cosign/cli/options/predicate.go
asraa: happens :)

Issue #706, opened 2021-09-17
Recommendation: Move cosign to cobra. 


n3wscott:
**Description**

`ffcli` seems like an ok prototype tool, but it lacks several features that we could take advantage of, like tab support or rich input validation. Anyone opposed to using something more standard for interactive CLIs like Cobra?

I can make a follow up PR for what I am thinking using `sget`.

dlorenc:
+1. @ahmetb wanted to do this awhile ago too

n3wscott:
I created https://github.com/sigstore/cosign/pull/709 to show what I am thinking on moving to cobra in terms of shape. 3 main things to point out:

1. the `cmd/cli` code is to define the flags and cli interface.
2. the cli delegates to an instance object built up in the cli interpreter and is then executed via `Do(ctx)`
3. The code in `pkg/sget` holds all of the logic to do the `sget` task, and knows nothing of the cli. This allows the logic command to be vendored and used down stream without mixing the cli and the task it is doing.

n3wscott:
I am basing this off https://github.com/n3wscott/cli-base and https://github.com/kubernetes-sigs/zeitgeist/tree/master/buoy among other things.

cpanato:
+1

n3wscott:
There is a little wrinkle. `ffcli` supports non-POXIS standard single dash word flags and under POXIS, this is not allowed. See the fix required for `sget`: https://github.com/sigstore/cosign/pull/724.

This means we would have to change all of the flags for `cosign` to POXIS standard flags.

See https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

PR #835, opened 2021-10-01
find and fix examples where we are using the single dash style flags.

n3wscott:
#### Summary

find and fix examples where we are using the single dash style flags.

#### Ticket Link

Related to #706 

#### Release Note

```release-note
`cosign` now uses POXIS style flags.
```


dlorenc (approved):

imjasonh:
You might also want to update intro.gif: https://github.com/sigstore/cosign/blob/main/images/intro.gif

PR #839, opened 2021-10-04
Generate Markdown docs for cosign

imjasonh:
#### Summary

Now that https://github.com/sigstore/cosign/issues/706 is closed (👏 @n3wscott), we can use cobra's built-in doc generation to make Markdown docs available for `cosign`.

Example: https://github.com/imjasonh/cosign/blob/gendoc/doc/cosign.md

This makes it easier to reference CLI surfaces in docs and in conversation.

This also adds a CI check to ensure new surface changes are accompanied by doc regeneration.

Open Questions:
- Do we also want to generate docs for `sget`, which also lives in this repo? This might change where we want generated docs to live.

#### Ticket Link

n/a

#### Release Note

```release-note
cosign has generated Markdown docs available in the doc/ directory.
```


n3wscott (approved):
That is really awesome!!!

PR #851, opened 2021-10-06
verify: add support for rsapkcs15 keys

rchincha:
currently cosign only supports ecdsa keys for signature verification.

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>


dlorenc:
I think this is safe to add on verify, but I'm still not sure adding direct support for customizing algorithms on key-generation and signatures is a great idea.

dekkagaijin:
As Dan said, unless there's a concrete use-case that's being blocked by limiting support to ECDSA-SHA256 in `cosign` itself, we err on the side of limiting complexity (and choices).

rchincha:
> 
> 
> As Dan said, unless there's a concrete use-case that's being blocked by limiting support to ECDSA-SHA256 in `cosign` itself, we err on the side of limiting complexity (and therefore choices).

That the signing keys are RSA-2k? And cosign is not the one signing? [1]

[1] https://github.com/sigstore/cosign/blob/main/EXAMPLES.md

dekkagaijin:
The examples also mention the way to download signatures and verify with a separate tool, e.g.:

```shell
$ cosign download signature us-central1-docker.pkg.dev/dlorenc-vmtest2/test/taskrun > signatures.json
# There could be multiple signatures, let's pretend it's the last one.
# Extract the payload and signature, base64 decoding them.
$ cat signatures.json | tail -1 | jq -r .Payload | base64 -D > payload.txt
$ cat signatures.json | tail -1 | jq -r .Base64Signature | base64 -D > signature.txt
# Verify in openssl
$ openssl dgst -sha256 -verify pubkey.pem -signature signature.txt payload.txt
```

rchincha:
> 
> 
> The examples also mention the way to download signatures and verify with a separate tool, e.g.:
> 
> ```shell
> $ cosign download signature us-central1-docker.pkg.dev/dlorenc-vmtest2/test/taskrun > signatures.json
> # There could be multiple signatures, let's pretend it's the last one.
> # Extract the payload and signature, base64 decoding them.
> $ cat signatures.json | tail -1 | jq -r .Payload | base64 -D > payload.txt
> $ cat signatures.json | tail -1 | jq -r .Base64Signature | base64 -D > signature.txt
> # Verify in openssl
> $ openssl dgst -sha256 -verify pubkey.pem -signature signature.txt payload.txt
> ```

Sure, in which case the role cosign plays is rather limited - push and pull the signature layer. If so, that's fine, good to know.

dlorenc:
Sorry - yes this is fine in verify!

pkg/cosign/keys.go
```diff
@@ -105,6 +106,18 @@ func PemToECDSAKey(pemBytes []byte) (*ecdsa.PublicKey, error) {
 	return ecdsaPub, nil
 }
 
+func PemToRSAKey(pemBytes []byte) (*rsa.PublicKey, error) {
```
dlorenc, line 109: What do you think of merging these functions so we can do a type assertion switch instead of trying them in order as error fallbacks?

rchincha: Looking at top of sigstore/sigstore project for helper apis. Just thinking about this some more, to support a full verify list, more hash funcs may need to be checked against if done "automatically" without cmdline arg hints. For example, cosign assumes sha256, but sha512 is another possibility.

rchincha: Updated the PR using sigstore/sigstore helpers.
Caveat is we (continue to) assume SHA256.

Issue #844, opened 2021-10-04
[cosign/pkg] High-level, pluggable, composable, `cosign`ing and verification flows


dekkagaijin:
Right now `cosign` as a package is tightly coupled with its signing and verification implementation:

* _Registry-resident_ OCI containers as the signable object
* Generating `fulcio` certs
* Uploading keys to `rekor`

This leads to significant amounts of control flow being required to properly implement even [canonical signing/verifying operations](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign.go#L124-L246).

As a follow-up to @mattmoor's work to bury [OCI registry-specific implementation](https://github.com/sigstore/cosign/tree/main/pkg/oci) under a more generic API, we should also consider how we could present individual operations (e.g. `fulcio`, `rekor` upload/verification) as generic, pluggable parts of a high-level control flow.

dekkagaijin:
The imagined high-level API would be something like:

```
type SigningClient struct {
  steps []SigningOp
}

func (s *SigningClient) Sign(Signable) (Sig, error) {
  ...
}

func (s *SigningClient) Attest(Signable, Attestation) (Sig, error) {
  ...
}

type VerificationClient struct {
  steps []VerificationOp
}

func (v *VerificationClient) VerifySig(Verifiable) (error) {
  ...
}

func (v *VerificationClient) VerifyAttestation(Verifiable, Att) (error) {
  ...
}
```

or possibly, for the more functional people in the audience:

```
func  Sign(Signable, ops SignOp...) (Sig, error) {
  ...
}

func Attest(Signable, Attestation, ops SignOp...) (Sig, error) {
  ...
}
```

The inputs would be presented to steps in turn and results passed to subsequent steps. This way we can fully abstract away interactions with storage, eliminate complex control flow related to upload/verification with `fulcio` and `rekor`, and allow for consumers to easily insert their own logic (e.g. publish to PubSub)

mattmoor:
I think the key bit here is that we should decompose the CLI spaghetti monster around a few key interfaces:
```go
type Signer interface {
   Sign(Signable) (Sig, error)
   Attest(Signable, Attestation) (Sig, error)
}

type Verifier interface {
   ...
}
```

We could split out implementations of these interfaces into separate packages for:
1. Local public/private key
2. KMS's (including K8s secrets as faux-KMS)
3. Fulcio
4. Rekor (composed with one of the above)

[...]

PR #857, opened 2021-10-06
added `Attachment()` method to `oci.SignedEntity`

dekkagaijin:
This allows for users to easily fetch named File "attachments" associated with an image.

#### Ticket Link

https://github.com/sigstore/cosign/issues/844

#### Release Note

```release-note
[pkg/]: Added `Attachment()` method to `oci.SignedEntity` to retrieve arbitrary named "attachments"
```

### DIFF
--- a/pkg/oci/signed/image.go
+++ b/pkg/oci/signed/image.go
@@ -16,6 +16,8 @@
 package signed
 
 import (
+	"errors"
+
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 
 	"github.com/sigstore/cosign/pkg/oci"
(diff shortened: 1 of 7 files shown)


pkg/oci/signed/image.go
```diff
@@ -44,3 +46,8 @@ func (*image) Signatures() (oci.Signatures, error) {
 func (*image) Attestations() (oci.Signatures, error) {
 	return empty.Signatures(), nil
 }
+
+// Attestations implements oci.SignedImage
+func (*image) Attachment(name string) (oci.File, error) {
+	return nil, errors.New("unimplemented")
+}
```
mattmoor, line 53: It would be good to sketch out the `ociremote` forms of this, and use it in the read paths to validate this and see how the pieces will tie together.

We can follow up with things like `mutate.AttachFile` and `ociremote.WriteAttachment` for the "write" paths.

mattmoor: whoops looking in the wrong place.  This is `signed`, but `ociremote` is implemented.

mattmoor (approved):

Issue #866, opened 2021-10-10
`cosigned` should support verifying `Fulcio` signatures


mattmoor:
**Description**

Today the `cosigned` webhook requires folks to setup `k8s://cosign-system/verification-key` with the public key that should be used for verification.  The webhook passes the following `CheckOpts`:
https://github.com/sigstore/cosign/blob/278ad7d4063592b822ef58bf4a88abee7bb2eff3/pkg/cosign/kubernetes/webhook/validation.go#L61-L62

... which includes a `SigVerifier` based on the verification key, which sends us through this path in `Verify`:
https://github.com/sigstore/cosign/blob/278ad7d4063592b822ef58bf4a88abee7bb2eff3/pkg/cosign/verify.go#L143-L150

... but for Fulcio we want things to follow this path:
https://github.com/sigstore/cosign/blob/278ad7d4063592b822ef58bf4a88abee7bb2eff3/pkg/cosign/verify.go#L151-L184

mattmoor:
Here's sort of what I'm thinking:
1. Switch from using a secret for `verification-key` to using a `ConfigMap` (since it's a public key, when specified).
2. When the key isn't specified, verify against the Fulcio roots.

Over time, we can use this configmap to pass additional configuration, including:
 * Alternate Fulcio root certs (e.g. if a custom Fulcio root is in use),
 * A Rekor URL to verify transparency logs.

... and likely a bunch more (e.g. stuff like `CertEmail` in `CheckOpts`).

cc @dlorenc @dekkagaijin @n3wscott for thoughts

mattmoor:
I'm going to close this, and we can open a new issue for greater configurability.

PR #867, opened 2021-10-10
Allow `cosigned` to validate `Fulcio` signatures.

mattmoor:
This change makes it so that when `cosigned` is enabled on a namespace, and no public key is provided the webhook will verify things against the Fulcio root.

The basic idea here is that this roughly matches the CLI where omitting the key will verify things against the Fulcio root instead.

#### Ticket Link
Related: https://github.com/sigstore/cosign/issues/866

#### Release Note
```release-note
The cosigned webhook can now verify images signed by the Fulcio root, if no public key is present in the verification-key secret.
```


dlorenc (approved):

dlorenc:
I think we have some cleanup to do around expectations as we drop EXPERIMENTAL, but this SGTM.

PR #882, opened 2021-10-12
Clarify how to install sget

axelsimon:
Currently, sget is name dropped without install explanations :)

Also mention requirement for container registry and mention ttl.sh, a free OCI registry which requires no account, that can be used to easily test cosign.

Signed-off-by: axel simon <git@axelsimon.net>


dlorenc:
DCO seems stuck, manually merging since it's there.

Issue #880, opened 2021-10-11
Provide public accessors for each sub-command


mattmoor:
**Description**

Previously, you could invoke subcommands with something like: `cli.Sign().Exec(...)`,

Now, the code is constructed and added to the root commands via private methods:
https://github.com/sigstore/cosign/blob/f85fe3fb8e010f47f6a80185e1a9388e221a537f/cmd/cosign/cli/sign.go#L32

This might be okay, if these were thinner wrappers around the logic nested in Exec, but to pick on `Sign`, the wrapped `SignCmd` function takes 11 arguments, and this has changed since 1.2!

Given that the single use of `topLevel` (at least in `addSign`) is to call `topLevel.AddCommand(...)` these private accessors could likely just be eliminated and replaced with inline calls here:
https://github.com/sigstore/cosign/blob/a04f0601eccce41d02ca919ac830b650b8f1e4cc/cmd/cosign/cli/commands.go#L60-L82

to things like: `cmd.AddCommand(Sign())`.

Then I could continue to use `cli.Sign()` to get the (now) `*cobra.Command`.

cc @n3wscott @dekkagaijin

n3wscott:
Don't use the cli init, use the command that the cli sets up. There should always be a public struct or function that does the work. The cobra setup is just for cobra and private as planned. Don't reuse that, reuse what cobra is calling.

PR #883, opened 2021-10-12
Re-expose commands.

mattmoor:
I was trying to pull in the recent upstream updates to a downstream repo that was invoking the CLI via `cli.Sign().Exec(...)` and this regressed in the cobra migration.  In a perfect world, we could pull in the CLI functionality without needing to go through `cobra.Command` to do it, but we're not there yet.  The alternative to `cli.Sign().Exec(...)` is `cli.SignCmd(...)`, which takes 11 arguments and growing each release.  The alternative fo `cli.Attest().Exec(...)` is `cli.AttestCmd(...)` with 9 arguments.  etc, etc, etc...

What's more, the above functions are also colocated with their cobra wrappers so we also can't save on not vendoring cobra yet, so it's a mystery to me why we'd force folks to not use this interface.

#### Ticket Link
Fixes: https://github.com/sigstore/cosign/issues/880

#### Release Note

```release-note
NONE
```


mattmoor:
I also noticed that we were generating markdown docs without the `piv-tool` present, so this passes the needed tags to that command.

mattmoor:
touching things to kick DCO 🤞

PR #885, opened 2021-10-12
fix: show reasons of the rego validations

### LINKED ISSUE #884: enhance rego policy output
In https://github.com/sigstore/cosign/pull/872, the output on failing validation is

```
[policy.rego]
will be validating against Rego policies: [policy.rego]
There are 1 number of errors occurred during the validation:
- rego validation failed
```

It would be cool if the output could be influenced by the policy.rego code. To do that, we'd have to support a more complex result set, and have policies like

```rego
package signature

deny[msg] {
  input.Data != "bar"
  msg := sprintf("unexpected data: %v", [input.Data])
}

deny[msg] { 
  # timestamp logic
}

deny["Data missing"] { not input.Data }
```

The call to rego.eval would check the result set and include the information in the CLI output.

It's a pattern already used in conftest and gatekeeper, for other examples.

developer-guy:

#### Summary

enhance the output of the rego validation result

#### Ticket Link

Fixes #884 

#### Release Note

```release-note
fix: show reasons of the rego validations
```


cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -231,8 +229,8 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 
 			if len(regoPolicies) > 0 {
 				fmt.Fprintf(os.Stderr, "will be validating against Rego policies: %v\n", regoPolicies)
-				if err := rego.ValidateJSON(payload, regoPolicies); err != nil {
-					validationErrors = append(validationErrors, err)
+				if errs := rego.ValidateJSON(payload, regoPolicies); len(errs) > 0 {
+					validationErrors = append(validationErrors, errs...)
 				}
```
srenatus, line 234: [nit] Isn't this the same as 
```suggestion				
				validationErrors = append(validationErrors, rego.ValidateJSON(payload, regoPolicies)...)
```

(https://play.golang.org/p/UeieojxP1oz)

pkg/cosign/rego/rego.go
```diff
@@ -33,23 +34,31 @@ func ValidateJSON(jsonBody []byte, entrypoints []string) error {
...
 	if rs.Allowed() {
-		return nil
+		return errs
 	}
-	return errors.New("rego validation failed")
+
+	for _, result := range rs {
+		for _, expression := range result.Expressions {
+			for _, v := range expression.Value.([]interface{}) {
+				errs = append(errs, errors.Errorf("%s", v.(string)))
```
srenatus, line 59: We could be more cautious here:

```rego
deny[{}] = { true }
```
would make this panic

pkg/cosign/rego/rego.go
```diff
@@ -33,23 +34,31 @@ func ValidateJSON(jsonBody []byte, entrypoints []string) error {
...
-		return err
+		errs = append(errs, err)
 	}
 
 	var input interface{}
 	dec := json.NewDecoder(bytes.NewBuffer(jsonBody))
 	dec.UseNumber()
 	if err := dec.Decode(&input); err != nil {
-		return err
+		errs = append(errs, err)
```
srenatus, line 44: Should we not `return []error{err}` here? It's unlikely that the rego eval below does anything useful if we feed it bad input.

developer-guy: yes, totally agree with you, thank you so much.

pkg/cosign/rego/rego.go
developer-guy: what do you suggest I do? 🙏

developer-guy:
> Looks good -- the only thing I'd be concerned about is the naming of the `allow` rule:
> 
> ```rego
> allow["bad input"] {
>   # some logic
> }
> ```
> 
> just reads wrong, when you've been working with similar systems before.
> 
> I'd propose we flip the logic and have the rego eval query `data.signature.deny` or `data.signature.invalid` and _pass_ if the returned boolean is false; or if the returned set is empty.

yes, thank you so much @srenatus, with the latest changes it would be good to change the rule name from `allow` to `deny`. It makes more sense.

pkg/cosign/rego/rego.go
srenatus: something along those lines, maybe?
```go
if s, ok := v.(string); ok {
	errs = append(errs, fmt.Errorf(s))
} else {
	errs = append(errs, fmt.Errorf("%s": v))
}
```

developer-guy:
@srenatus I updated the code according to your comments. Maybe you should make a final review once you have time for it. Then @dlorenc can help us to merge this 🙏 😋🥳

srenatus (approved):
LGTM. Needs docs 😉

PR #854, opened 2021-10-06
feat(ci): Add GoFish support

dirien:
Signed-off-by: Engin Diri <engin.diri@mail.schwarz>

#### Summary

Hello cosign community and @cpanato,

I added GoFish support via goreleasers (@caarlos0). The only things we need before merging this PR:

A github repo called: fish-food  similar to https://github.com/goreleaser/fish-food and and ensure access to it during the release process.

```yaml
owner: sigstore
name: fish-food
```

#### Ticket Link

Fixes #855 

#### Release Note

NONE
```release-note

```

PR #893, opened 2021-10-14
fix(ci): do not need to set no_unique_dist_dir

caarlos0:
#### Summary

Its not really needed to set `no_unique_dist_dir`.
Also simplified the binary naming, with `no_unique_dist_dir: false`, goreleaser auto-creates an unique dir inside dist for each binary, and the archive config handles the "renaming" when uploading to github.

#### Ticket Link

as discussed in https://github.com/sigstore/cosign/pull/854#issuecomment-942841712

#### Release Note

```release-note

```

### DIFF
--- a/.github/workflows/validate-release.yml
+++ b/.github/workflows/validate-release.yml
@@ -55,9 +55,9 @@ jobs:
 
       - name: check binaries
         run: |
-          ./dist/cosign-linux-amd64 version
-          ./dist/cosigned-linux-amd64 --help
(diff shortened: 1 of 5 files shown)


caarlos0:
> IIRC I set `no_unique_dist_dir: true` and manually set the binary names because it simplified the logic to upload the release to GCS in the expected layout (i.e. flat) with the expected naming convention

oh, that makes sense... will close this then.

PR #894, opened 2021-10-14
Fix verify command line options

rgerganov:
Verify requires public key (not private)


hectorj2f (approved):

dlorenc:
Thank you @rgerganov!

PR #896, opened 2021-10-14
Pin crane dependency used in e2e tests

imjasonh:
#### Summary

`go install` without a version specified installs the latest semver release of the tool. This means that a compromise of the go-containerregistry repo could result in arbitrary code running in this workflow, e.g., when e2e tests invoke `crane ls`.

With a version specified, even if the repo moves the `v0.6.0` tag to another commit, the Go module proxy will serve the previous v0.6.0 code.

v0.6.0 is the latest release of crane: https://github.com/google/go-containerregistry/releases

Signed-off-by: Jason Hall <jasonhall@redhat.com>

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -37,6 +37,6 @@ jobs:
           service_account_key: ${{ secrets.GCP_CI_SERVICE_ACCOUNT }}
           export_default_credentials: true
       - run: |
-          go install github.com/google/go-containerregistry/cmd/crane
+          go install github.com/google/go-containerregistry/cmd/crane@v0.6.0
           gcloud auth configure-docker us-central1-docker.pkg.dev --quiet
           ./test/e2e_test_secrets.sh


.github/workflows/e2e_tests.yml
```diff
@@ -37,6 +37,6 @@ jobs:
           service_account_key: ${{ secrets.GCP_CI_SERVICE_ACCOUNT }}
           export_default_credentials: true
       - run: |
-          go install github.com/google/go-containerregistry/cmd/crane
+          go install github.com/google/go-containerregistry/cmd/crane@v0.6.0
```
naveensrinivasan, line 40: Why not pin by SHA go install github.com/google/go-containerregistry/cmd/crane@092caf0f50b9a09588b653e08480ec4077cb289f? I understand it is hard to know the version by SHA. But a comment with a version number would address that issue.

imjasonh: Practically speaking, there's no reason to prefer the SHA. Once the Go module proxy has seen a commit tagged as v0.6.0, it will cache and serve that content even if the v0.6.0 tag pointer moves to point to another commit.

So this is effectively the same as a SHA, and provides useful semantic information to readers.

PR #934, opened 2021-10-21
Verify a signature using secrets from a gitlab project

hectorj2f:
#### Summary

Add support to verify a signature using secrets from a gitlab project. Initially this feature is only available to Gitlab. Github does not support it to fetch the secret variables.

#### Ticket Link

Fixes

#### Release Note

```release-note
* Add support to verify artifacts using secrets from a Gitlab project
```


developer-guy (approved):
thank you so much @hectorj2f, LGTM !

PR #938, opened 2021-10-21
cmd/policy: ability to pass expire days

Dentrax:

#### Summary

Just noticed that we have the following [usage example](https://github.com/sigstore/cosign/blob/9e304d1486c0c8ecfc08ce3f775a88868d0a4356/cmd/cosign/cli/policy_init.go#L83):
```
# extract public key from private key to a specified out file.
  cosign policy init -ns <project_namespace> --maintainers {email_addresses} --threshold <int> --expires <int>(days)
```

I was passed `--expires 7` flag, and it seems does not work as expected: _(It adds [3 months by default](https://github.com/sigstore/cosign/blob/9e304d1486c0c8ecfc08ce3f775a88868d0a4356/pkg/cosign/tuf/policy.go#L82))_

My current date: `2021-10-21T20:52:12Z`

**Before**
```bash
$ cat ./o
{
    "signed": {
        "expires": "2022-01-21T20:54:44Z",
    }
}
```

**After**
```bash
$ cat ./o
{
    "signed": {
        "expires": "2021-10-28T20:58:53Z",
    }
}
```

#### Ticket Link

Fixes

#### Release Note

```release-note
* Ability to pass --expire for policy init command
```

cc: @asraa

### DIFF
--- a/cmd/cosign/cli/policy_init.go
+++ b/cmd/cosign/cli/policy_init.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
@@ -113,6 +114,10 @@ func initPolicy() *cobra.Command {
 			role.AddKeysWithThreshold(publicKeys, o.Threshold)
 			root.Roles["root"] = role
 			root.Namespace = o.ImageRef
+
+			if o.Expires > 0 {
+				root.Expires = time.Now().AddDate(0, 0, o.Expires).UTC().Round(time.Second)
+			}
 
 			policy, err := root.Marshal()
 			if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/cosign/cli/policy_init.go
```diff
@@ -114,6 +115,10 @@ func initPolicy() *cobra.Command {
 			root.Roles["root"] = role
 			root.Namespace = o.ImageRef
 
+			if o.Expires > 0 {
+				root.Expires = time.Now().Add(time.Duration(o.Expires) * 24 * time.Hour).UTC().Round(time.Second)
```
asraa, line 119: You can use `AddDate` which takes a time parameter instead of multiplying no `24 * time.Hour`. 

`time.Now().AddDate(0, 0, 21).UTC()`

asraa (changes requested):
Thanks for catching this! I think I just totally forgot about it!

PR #972, opened 2021-10-29
Add support for using k8schain under a flag.

mattmoor:
This will help avoid needing to configure credential helpers in environments supporting ambient authentication.

#### Release Note
```release-note
Add a --k8s-keychain option that enables cosign to support ambient registry credentials based on the "k8schain" library.
```

### DIFF
--- a/cmd/cosign/cli/options/flags.go
+++ b/cmd/cosign/cli/options/flags.go
@@ -16,11 +16,7 @@
 package options
 
 import (
-	"context"
 	"reflect"
-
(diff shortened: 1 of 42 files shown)


mattmoor:
Previously I would have to take the published images and run them through https://github.com/docker-credential-magic/docker-credential-magic, but with this it should just be a flag.

We could decide to detect we're running within k8s and do this by default, but I figured that would be more contentious.

n3wscott (approved):
Looks good to me!

PR #979, opened 2021-10-31
Invert upload flag to allow for not uploading attestation

luhring:
#### Summary

The `cosign attest` command has an `--upload` flag, designed to tell cosign to upload the generated attestation to the registry.

However, this bool value is set to `true` by default, which ends up meaning that **there's no way to instruct cosign _not to upload_** the generated attestation (IIUC).

This PR changes the flag to `--no-upload`, which inverts the logic and behaves as you'd expect. Without the flag, the attestation is uploaded. With the flag, the attestation is not uploaded, and instead, the attestation DSSE is sent to stdout.

**Note:** This is a breaking change for the CLI, since it removes an existing flag. The flag hadn't been affecting cosign's execution, so I'm not sure how many folks were using it. But regardless, I'm happy to approach this a different way to make the change _non-breaking_, just let me know!

#### Ticket Link

N/A

#### Release Note

```release-note
attest: replaces '--upload' flag with a '--no-upload' flag
```

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -70,7 +70,7 @@ func Attest() *cobra.Command {
 				OIDCClientSecret:         o.OIDC.ClientSecret,
 			}
 			for _, img := range args {
-				if err := attest.AttestCmd(cmd.Context(), ko, o.Registry, img, o.Cert, o.Upload, o.Predicate.Path, o.Force, o.Predicate.Type); err != nil {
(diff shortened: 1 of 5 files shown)


dlorenc:
> **Note:** This is a breaking change for the CLI, since it removes an existing flag. The flag hadn't been affecting cosign's execution, so I'm not sure how many folks were using it. But regardless, I'm happy to approach this a different way to make the change _non-breaking_, just let me know!

If the flag was broken I see no problem fixing it like this!

dlorenc (approved):

Issue #676, opened 2021-09-15
keyless verify-blob?


caarlos0:
**Question**

I am able to keyless sign a blog with:

```sh
COSIGN_EXPERIMENTAL=1 cosign sign-blob Desktop/file.pdf
```

But I can't verify it, as it asks for a `-key` or `-sk`.

Is it something planned to work in the future? 
If not, maybe we shouldn't allow keyless signing a blob? 
Maybe its just me doing something wrong? I don't see any mentions to blob in the [KEYLESS.md](https://github.com/sigstore/cosign/blob/main/KEYLESS.md) file... and seems like the full design doc is not public, so I'm not sure 🤔

dlorenc:
> Maybe its just me doing something wrong? I don't see any mentions to blob in the [KEYLESS.md](https://github.com/sigstore/cosign/blob/main/KEYLESS.md) file... and seems like the full design doc is not public, so I'm not sure 🤔

The design doc should be public! You just have to join sigstore-dev@googlegroups.com I think...

It looks like this is missing from the docs, but there's a "-cert" flag you need to pass to verify a blob using the "keyless" flow right now (when you initially sign an object, you get a certificate from Fulcio, unfortunately you need to remember that certificate somewhere today).

That cert can be looked up from Rekor if you only have the artifact, so we can automate/document this too.

DennisDenuto:
> That cert can be looked up from Rekor if you only have the artifact, so we can automate/document this too.

Hey @dlorenc is there any plan to automate looking up the cert using rekor when running `verify-blob`?

dlorenc:
I think we should do it! I don't have a plan to personally (just a bandwidth thing) but we can definitely help if someone wants to try and get it merged!

PR #991, opened 2021-11-02
Add ability for verify-blob to find signing cert in transparency log

DennisDenuto:
#### Summary

When a user has opted into keyless mode, and they have signed a blob, allow a user to verify a blob without having to specify a certificate.

questions / concerns:
- Currently, the first certificate found in the tlog (for the blob) is used when verifying. However, this might not be the correct cert to use when verifying. I'm not sure what the *best* ux around this should be. Perhaps a new flag should be introduced to help narrow down which cert is used to verify the blob.  (Allow the user to specify a subject / email address?) or perhaps cosign should loop through all the certificates it finds and attempt to verify the blob + sig against each one (this has performance implications)
- where is a good place to add tests?

#### Ticket Link
Aims to improve https://github.com/sigstore/cosign/issues/676#issuecomment-955102906

#### Release Notes

```release-note
When a user has opted into keyless mode, and they have signed a blob, allow a user to verify a blob without having to specify a certificate.
```

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -27,8 +27,8 @@ import (
 	"io"
 	"os"
 
+	"github.com/go-openapi/runtime"
 	"github.com/pkg/errors"
-
 	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/cmd/cosign/cli/sign"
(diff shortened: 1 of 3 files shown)


naveensrinivasan:
> #### Summary
> When a user has opted into keyless mode, and they have signed a blob, allow a user to verify a blob without having to specify a certificate.
> 
> questions / concerns:
> 
> * Currently, the first certificate found in the tlog (for the blob) is used when verifying. However, this might not be the correct cert to use when verifying. I'm not sure what the _best_ ux around this should be. Perhaps a new flag should be introduced to help narrow down which cert is used to verify the blob.  (Allow the user to specify a subject / email address?) or perhaps cosign should loop through all the certificates it finds and attempt to verify the blob + sig against each one (this has performance implications)
> * where is a good place to add tests?
> 
> #### Ticket Link
> Aims to improve [#676 (comment)](https://github.com/sigstore/cosign/issues/676#issuecomment-955102906)
> 
> #### Release Notes
> ```
> When a user has opted into keyless mode, and they have signed a blob, allow a user to verify a blob without having to specify a certificate.
> ```

We were looking for the same. Thanks

dlorenc:
This is awesome! Thanks so much!

dlorenc (approved):

naveensrinivasan:
@DennisDenuto  The `cosign` still looks for a cert `Error: verifying blob [go.mod]: exactly one of: key reference (--key), certificate (--cert) or hardware token (--sk) must be provided`. I pulled down the head and tried to verify this.

DennisDenuto:
@naveensrinivasan hmm did you set the experimental flag on? i.e `COSIGN_EXPERIMENTAL=1`

Issue #990, opened 2021-11-02
cosign doesn't provide an option to set timeout on rekor upload


naveensrinivasan:
I am trying to `cosign sign-blob` and run into a timeout issue 
```
use /home/runner/work/scorecard/scorecard/dist/scorecard-linux-arm64: Post \"https://rekor.sigstore.dev/api/v1/log/entries\": context deadline exceeded\n": exit status 1|

```
After some investigation, it looks like the code doesn't provide an option 
https://github.com/sigstore/cosign/blob/ae36ba5bd7fd839e6dceccfa2a27455572e8a3dc/pkg/cosign/tlog.go#L76

but `rekor` has that option https://github.com/sigstore/rekor/blob/0c2b37c15bbaf80a1c32d47b63956c5d499a6f2c/pkg/generated/client/entries/create_log_entry_parameters.go#L51

`func NewCreateLogEntryParamsWithTimeout(timeout time.Duration) *CreateLogEntryParams`

cpanato:
this is something we face before, but I'm not sure if we fix it @lukehinds did you make a PR for this fix?

naveensrinivasan:
If not I am open to fixing this. We need in Scorecard. Thanks

naveensrinivasan:
Fixed by https://github.com/sigstore/cosign/pull/1020

PR #1001, opened 2021-11-05
Included timeout option for uploading to Rekor

naveensrinivasan:
Included a `timeout` option to `sign-blob` because uploading a large binary to `rekor` was failing with timeout.

https://github.com/sigstore/cosign/issues/990


lukehinds:
Not sure I understand the logic here. rekor will timeout regardless of a client timing out, whereas with the client having a timeout you will never know the server has timed out in the first place. I would have thought it more useful for the client to know there has been timeout (and then take action accordingly), rather then bail early under the pretence there *may* timeout?

lukehinds:
interesting, I never knew we had rekor client timeouts. 

in regards to your proposal I am not of strong view here, so can let others chime in. There is no harm in having a timeout flag, I guess as long as users don't believe it will have influence on the servers timeout rules.

dlorenc (approved):

PR #1016, opened 2021-11-09
implement a new flag for sign-blob subcommand to save certificate to disk or stdout

shibumi:
#### Summary

With cosign 1.3.0 it is not possible to store the public certificate generated by the OIDC-issuer in a convenient way.
To store the public key locally on disk, the user has to  intercept the cosign stdout, parse it and store it.
This is inconvenient and error-prone.  We want to make signing and public key distribution as easy as possible, hence this PR introduces a new `pubkey-output` flag for the `sign-blob` sub-command.

The `pubkey-output` flag respects the `b64` flag and prints the certificate/key as file in a given file path.

#### Ticket Link

Fixes None

#### Release Note

```release-note
The cosign cli is now able to store the public certificate/key during sign-blob operations via the `pubkey-output` flag.
```

### DIFF
--- a/cmd/cosign/cli/options/signblob.go
+++ b/cmd/cosign/cli/options/signblob.go
@@ -22,16 +22,19 @@ import (
 )
 
 // SignBlobOptions is the top level wrapper for the sign-blob command.
+// The new output-certificate flag is only in use when COSIGN_EXPERIMENTAL is enabled
 type SignBlobOptions struct {
-	Key          string
-	Base64Output bool
-	Output       string // TODO: this should be the root output file arg.
-	SecurityKey  SecurityKeyOptions
-	Fulcio       FulcioOptions
-	Rekor        RekorOptions
-	OIDC         OIDCOptions
-	Registry     RegistryOptions
-	Timeout      time.Duration
+	Key               string
+	Base64Output      bool
+	Output            string // deprecated: TODO remove when the output flag is fully deprecated
+	OutputSignature   string // TODO: this should be the root output file arg.
+	OutputCertificate string // TODO: this should be the root output file arg.
+	SecurityKey       SecurityKeyOptions
+	Fulcio            FulcioOptions
+	Rekor             RekorOptions
+	OIDC              OIDCOptions
+	Registry          RegistryOptions
+	Timeout           time.Duration
 }
 
 var _ Interface = (*SignBlobOptions)(nil)
@@ -50,8 +53,14 @@ func (o *SignBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.Base64Output, "b64", true,
 		"whether to base64 encode the output")
 
-	cmd.Flags().StringVar(&o.Output, "output", "",
+	cmd.Flags().StringVar(&o.OutputSignature, "output-signature", "",
 		"write the signature to FILE")
+
+	// TODO: remove when output flag is fully deprecated
+	cmd.Flags().StringVar(&o.Output, "output", "", "write the signature to FILE")
+
+	cmd.Flags().StringVar(&o.OutputCertificate, "output-certificate", "",
+		"write the certificate to FILE")
 
 	cmd.Flags().DurationVar(&o.Timeout, "timeout", time.Second*30,
 		"HTTP Timeout defaults to 30 seconds")
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -52,9 +52,10 @@ type KeyOpts struct {
 }
 
 // nolint
-func SignBlobCmd(ctx context.Context, ko KeyOpts, regOpts options.RegistryOptions, payloadPath string, b64 bool, output string, timeout time.Duration) ([]byte, error) {
+func SignBlobCmd(ctx context.Context, ko KeyOpts, regOpts options.RegistryOptions, payloadPath string, b64 bool, outputSignature string, outputCertificate string, timeout time.Duration) ([]byte, error) {
 	var payload []byte
 	var err error
+	var rekorBytes []byte
 
 	if payloadPath == "-" {
 		payload, err = io.ReadAll(os.Stdin)
@@ -79,7 +80,6 @@ func SignBlobCmd(ctx context.Context, ko KeyOpts, regOpts options.RegistryOption
 
 	if options.EnableExperimental() {
 		// TODO: Refactor with sign.go
-		var rekorBytes []byte
 		if sv.Cert != nil {
 			fmt.Fprintf(os.Stderr, "signing with ephemeral certificate:\n%s\n", string(sv.Cert))
 			rekorBytes = sv.Cert
@@ -101,8 +101,8 @@ func SignBlobCmd(ctx context.Context, ko KeyOpts, regOpts options.RegistryOption
 		fmt.Fprintln(os.Stderr, "tlog entry created with index:", *entry.LogIndex)
 	}
 
-	if output != "" {
-		f, err := os.Create(output)
+	if outputSignature != "" {
+		f, err := os.Create(outputSignature)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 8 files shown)


cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -76,9 +76,9 @@ func SignBlobCmd(ctx context.Context, ko KeyOpts, regOpts options.RegistryOption
 		return nil, errors.Wrap(err, "signing blob")
 	}
 
+	var rekorBytes []byte
```
asraa, line 79: It seems like this will only pick up the public key when COSIGN_EXPERIMENTAL=1 is enabled. That mostly makes sense to me since otherwise you probably already have the key, but I would caveat in the doc string or somewhere else.

May also be worth adding for the sign subcommand too?

developer-guy:
we did the same in #1021

cmd/cosign/cli/options/signblob.go
```diff
@@ -49,4 +50,7 @@ func (o *SignBlobOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringVar(&o.Output, "output", "",
 		"write the signature to FILE")
+
+	cmd.Flags().StringVar(&o.PubKeyOutput, "pubkey-output", "",
```
Dentrax, line 54: What about changing this to `--output-cert`?

shibumi:
> we did the same in #1021

Yes :D just realized this. I like your PR more, but I am not sure if changing `output` to `output-signature` would break cosign's API stability.

cmd/cosign/cli/options/signblob.go
shibumi: @Dentrax I was a little bit confused about cert and public key... because this does not need to be a certificate it can also be a pubkey, IIRC.

@asraa are you fine with changing this to --output-cert instead of `-pubkey-output`?

Thoughts? @dlorenc

dlorenc: I think it should always be a cert for the keyless modes right?

developer-guy:
> > we did the same in #1021
> 
> Yes :D just realized this. I like your PR more, but I am not sure if changing `output` to `output-signature` would break cosign's API stability.

Thanks, @shibumi, yes we know, but it makes more sense to us instead of using just output, WDYT to make it consistent?

dlorenc:
> Thanks, @shibumi, yes we know, but it makes more sense to us instead of using just output, WDYT to make it consistent?

I think I'm fine with this breaking change. It's pretty minor and important for usability. If we need to we could always alias it for a release in cobra.

cmd/cosign/cli/options/signblob.go
asraa: Yeah, i think that's more clear.

cmd/cosign/cli/options/signblob.go
```diff
@@ -22,16 +22,18 @@ import (
 )
 
 // SignBlobOptions is the top level wrapper for the sign-blob command.
+// The new output-certificate flag is only in use when COSIGN_EXPERIMENTAL is enabled
```
shibumi, line 25: @asraa  is this docstring on the right position? I thought about adding it to the `SignBlob` function, but the SignBlob function has a `// nolint` on top.. I am not sure if I can have docstrings and nolint at the same time.

shibumi:
@developer-guy  @dlorenc done. Or do you prefer the short form for commands? Right now I chose `output-certificate` and `output-signature`. I see that @developer-guy  used `output-cert` and `output-sig`. Opinions?

Dentrax:
> @developer-guy @dlorenc done. Or do you prefer the short form for commands? Right now I chose `output-certificate` and `output-signature`. I see that @developer-guy used `output-cert` and `output-sig`. Opinions?

To make it more consistent with the `--signature` flag, I think it would better DX to use `--output-signature`, and eventually we should go with the `--output-certificate` if we decide to use long-form of `signature` instead of `sig`.

shibumi:
Good argument @Dentrax

developer-guy:
@shibumi would you mind adding I and friend @Dentrax as a collab to the project? Maybe we can add some code to this. 🤝🙋🏻‍♂️

shibumi:
@developer-guy do you mean editor permissions to this PR? I have no idea if this is possible in Github. I think you can fork my fork and then create a PR against my fork and I could merge it in my fork :D 

OR

You just drop me a patch file via mail.

What do you would like to change in this PR?

caarlos0 (approved):
nice!

Dentrax:
With the new release of [goreleaser v1.0.0](https://github.com/goreleaser/goreleaser/releases/tag/v1.0.0), I think we can merge this PR and plan to release on the upcoming cosign version. 🙏 Should we create `Release v1.4` issue additionally to keep track of the next milestone? @cpanato @dlorenc

developer-guy:
in addition to the comment above, with this, we will be able to do the following: 
```shell
signs:
  - cmd: cosign
    certificate: "${artifact}.cert"
    signature: "${artifact}.sig"
    args: ["sign-blob", "--oidc-issuer=https://token.actions.githubusercontent.com", "--output-signature=${signature}", "--output-certificate=${certificate}",  "${artifact}"]
```

For more detail: https://github.com/google/ko/pull/498/

dlorenc:
This is amazing!

caarlos0:
Once this is out, I'll write a post on how to use it with goreleaser on github actions as wel

dlorenc:
Looks like this just needs docs to be regenerated!

shibumi:
@dlorenc how do I regenerate the docs?

caarlos0:
BTW, this is a breaking change isnt it? As `--output` will not exist anymore... 

Maybe worth deprecating it first?

dlorenc:
> @dlorenc how do I regenerate the docs?

Looks like: `go run -tags pivkey,pkcs11key,cgo ./cmd/help/`



> BTW, this is a breaking change isnt it? As `--output` will not exist anymore...

Yeah, we probably should deprecate/alias first. I misunderstood and thought this flag only applied for experimental/keyless stuff, but it's used for normal blob signatures too. Great catch @caarlos0!

shibumi:
> Yeah, we probably should deprecate/alias first. I misunderstood and thought this flag only applied for experimental/keyless stuff, but it's used for normal blob signatures too. Great catch @caarlos0!

Sounds, good. Let me apply a new patch in the afternoon (that's in 2 hours in my time :D )

shibumi:
- [X] regenerated docs
- [X] deprecated output flag

New behavior:
```
❯ COSIGN_EXPERIMENTAL=1 ./cosign sign-blob --output=main.go.sig --output-certificate=cert.pub --oidc-issuer https://oauth2.sigstore.dev/auth main.go
WARNING: the '--output' flag is deprecated and will be removed in the future. Use '--output-signature
Using payload from: main.go
Generating ephemeral keys...
Retrieving signed certificate...
Your browser will now be opened to: ....
```

```
❯ COSIGN_EXPERIMENTAL=1 ./cosign sign-blob --output-signature=main.go.sig --output-certificate=cert.pub --oidc-issuer https://oauth2.sigstore.dev/auth main.go
Using payload from: main.go
Generating ephemeral keys...
Retrieving signed certificate...
Your browser will now be opened to: ...
```

developer-guy (approved):
LGTM, thanks @shibumi amazing work 👌🙋🏻‍♂️

shibumi:
@dlorenc The generate doc command does not seem to be sufficient.. I have generated the docs with it, but the pipeline still fails.

developer-guy:
@shibumi I think you should run `goimports -w ./...`.

developer-guy:
@dlorenc IMHO, this PR is ready for merge.

shibumi:
@developer-guy  thanks for your hint to goimports. I didn't know that exists :D

developer-guy:
> @developer-guy  thanks for your hint to goimports. I didn't know that exists :D 

I added two new targets to the makefile: fmt and checkfmt. You can use them now 🙋🏻‍♂️

caarlos0:
👀

PR #1030, opened 2021-11-11
Patch support attestation log search and bundle to payload hash check

houdini91:
Signed-off-by: houdini91 <mdstrauss91@gmail.com>

Hi guys.
My first PR to cosign hope i am not misunderstanding something critical.

## Summary
PR is a patch to the rekor validation flow when using attestation.
1) When bundle is not present searching for intoto rekor entries would fail because of search was using in to the wrong entry type (rekord vs intoto).
2) When verifying the attached rekor Bundle the certificate expire was checked and the validity of the bundle signiture.
I added a  hash test between the bundle content hash and actual payload hash.
I am not sure if the idea is to verify the bundle actual body is the responsibility of the policy.

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -137,20 +137,40 @@ func getTlogEntry(rekorClient *client.Rekor, uuid string) (*models.LogEntryAnon,
 	return nil, errors.New("empty response")
 }
 
+func proposedEntry(b64Sig string, payload, pubKey []byte) ([]models.ProposedEntry, error) {
+	var proposedEntry []models.ProposedEntry
+	signature, err := base64.StdEncoding.DecodeString(b64Sig)
+	if err != nil {
+		return nil, errors.Wrap(err, "decoding base64 signature")
+	}
+
+	if len(signature) == 0 {
+		te := intotoEntry(payload, pubKey)
+		entry := &models.Intoto{
+			APIVersion: swag.String(te.APIVersion()),
+			Spec:       te.IntotoObj,
+		}
+		proposedEntry = []models.ProposedEntry{entry}
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/cosign/verify.go
```diff
@@ -329,9 +331,70 @@ func VerifyBundle(sig oci.Signature) (bool, error) {
...
+	h := sha256.Sum256(payload)
+	payloadHash := hex.EncodeToString(h[:])
+
+	if alg != "sha256" || bundlehash != payloadHash {
+		return false, errors.Wrap(err, "matching bundle to payload")
+	}
 	return true, nil
 }
 
+func BundleHash(bundleBody, signature string, payload []byte) (string, string, error) {
```
dlorenc, line 354: nit: does this need to be public?

pkg/cosign/tlog.go
```diff
@@ -137,20 +137,40 @@ func getTlogEntry(rekorClient *client.Rekor, uuid string) (*models.LogEntryAnon,
...
 }
 
+func proposedEntry(b64Sig string, payload, pubKey []byte) ([]models.ProposedEntry, error) {
+	var proposedEntry []models.ProposedEntry
+	signature, err := base64.StdEncoding.DecodeString(b64Sig)
+	if err != nil {
+		return nil, errors.Wrap(err, "decoding base64 signature")
+	}
+
+	if len(signature) == 0 {
```
dlorenc, line 147: Thanks for tracking this down! I think I understand the logic, but would you mind leaving a comment here to explain it to future people?

dlorenc (commented):
A few small nits! Thanks for tracking this down!

dlorenc (approved):

dlorenc:
Thanks!!!

PR #1035, opened 2021-11-11
feat: validate whole statement not just predicate part

### LINKED ISSUE #1012: Change the verify attestation implementation
Right now when using the verify-attestation command, only the predicate is passed to the verification policy. This is nice, but has a few issues:

* It loses the predicate type itself. This is in the attestation header, but could be useful in policy.
* It makes the commands hard to use separately cosign verify-attestation with just a key will output the full attestation to stdout. This should be usable directly with "cue vet" and the same policy.

developer-guy:

#### Summary

validate whole statement not just predicate part
#### Ticket Link

Fixes #1012

#### Release Note

```release-note
feat: validate whole statement not just predicate part
```

render1636653473904


dlorenc (approved):
Thank you so much!

dlorenc:
We'll merge after we cut the release to avoid screwing up release notes :)

dekkagaijin (approved):
good stuff

Issue #1015, opened 2021-11-08
Cosign "save" and "load" commands


dlorenc:
We should add some commands to save an image and it's "attachments" to disk in some kind of bundle on disk, and then load that into a remote registry from disk. See this thread: https://twitter.com/sudo_bmitch/status/1457794784135761920

dlorenc:
cc @developer-guy this could be a fun one :)

priyawadhwa:
chatted with @developer-guy, i might give this a go! :)

developer-guy:
I assigned it to you @priyawadhwa. Thank you so much for asking it me first, that was very kind! Good luck with it. We can't wait to see the results! 🚀

dlorenc:
What's left here @priyawadhwa? This one would be great for the next release.

priyawadhwa:
hoping to have a PR open by tmrw! just have to hook up the commands to the `layout` package (currently seeing a bug with writing to a remote repo tho so just have to get to the bottom of that)

priyawadhwa:
it might be useful to add support for the layout on disk to `cosign verify` -- maybe something like `cosign verify -key [KEY] [path to image directory]`?

haydentherapper:
Chatted with @priyawadhwa about working on https://github.com/sigstore/cosign/issues/1015#issuecomment-975734697 to add support for verification from disk. I'll be taking a look at adding this!

PR #1040, opened 2021-11-12
Add layout package for writing and loading signatures from disk

priyawadhwa:
Added a `layout` package to `pkg/oci` which should make it easy to store images & associated signatures to disk! This is the first part of implementing https://github.com/sigstore/cosign/issues/1015

TODOs:
- [x] unit tests
- [x] moving the `sigLayer` implementation into a separate package since it's used by `remote` as well 

I'm still not sure if I implemented this the correct way, cc'ing @jonjohnsonjr and @mattmoor in case I'm totally on the wrong track 😅


pkg/oci/siglayer/siglayer.go
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package remote
+package siglayer
```
mattmoor, line 16: Can you make this package `pkg/oci/internal/signature`?

pkg/oci/siglayer/siglayer.go
```diff
@@ -28,12 +28,27 @@ import (
...
+)
+
 type sigLayer struct {
 	v1.Layer
-	img  *sigs
+	img  oci.Signatures
 	desc v1.Descriptor
 }
 
+func New(l v1.Layer, img oci.Signatures, desc v1.Descriptor) *sigLayer {
```
mattmoor, line 44:
```suggestion
func New(l v1.Layer, img oci.Signatures, desc v1.Descriptor) oci.Signature {
```

pkg/oci/siglayer/siglayer.go
```diff
@@ -28,12 +28,27 @@ import (
...
+	return &sigLayer{
+		Layer: l,
+		img:   img,
```
mattmoor, line 47: Poking around, the only use of `img` seems to be `LayerByDigest` and I wonder if this was before the `v1.Layer` was embedded...  I'd think we could just have `Payload` call `s.Layer.Compressed()` instead of looking it up by the descriptor digest and drop the `img` 🤔

pkg/oci/layout/image.go
```diff
@@ -0,0 +1,103 @@
...
+var _ oci.Signatures = (*sigs)(nil)
+
+// Get implements oci.Signatures
+func (s *sigs) Get() ([]oci.Signature, error) {
+	layers, err := s.Image.Layers()
+	if err != nil {
+		return nil, err
+	}
+	var signatures []oci.Signature
+	for _, l := range layers {
```
mattmoor, line 65: I'd rework the logic around this to fetch the `v1.Manifest`, walk the descriptors, and for each of the descriptors lookup the layer by digest (in the descriptor).  This is a weird use of `partial` and I'm not sure it WAI.

pkg/oci/layout/image.go
```diff
@@ -0,0 +1,103 @@
...
+// SignedImage provides access to a remote image reference, and its signatures.
+func SignedImage(path string) (oci.SignedImage, error) {
+	p, err := layout.FromPath(imagePath(path))
+	if err != nil {
+		return nil, err
+	}
+	img, err := p.Image(v1.Hash{})
+	if err != nil {
+		return nil, err
+	}
```
mattmoor, line 37: I'm less familiar with this path, so I'll have to do some reading of the signatures in here before I can weigh in.

cc @jonjohnsonjr  @imjasonh

pkg/oci/layout/write.go
```diff
@@ -0,0 +1,67 @@
...
+
+func signaturesPath(path string) string {
+	return filepath.Join(path, "signatures")
+}
+
+type pathFunc func(string) string
+
+func write(path string, pf pathFunc, img v1.Image) error {
+	p := pf(path)
+	// write empty image
```
mattmoor, line 57: index?

mattmoor: Can you elaborate on why this is necessary?

pkg/oci/layout/write.go
```diff
@@ -0,0 +1,67 @@
...
+	layoutPath, err := layout.Write(p, empty.Index)
+	if err != nil {
+		return err
+	}
+	// write image to disk
+	if err := layoutPath.AppendImage(img); err != nil {
```
mattmoor, line 63: Is the above because you need an empty layout index to append this image to?

pkg/oci/layout/write.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+	return nil
```
mattmoor, line 66:
```suggestion
	return layoutPath.AppendImage(img)
```

pkg/oci/layout/write.go
mattmoor: looks like it.

pkg/oci/layout/write.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func imagePath(path string) string {
+	return filepath.Join(path, "image")
+}
+
+func signaturesPath(path string) string {
+	return filepath.Join(path, "signatures")
+}
```
mattmoor, line 51: I'd stick with `sigs` for signatures (I'm a big fan of consistency / precedent, and the use of that suffix in the registry is why I'm leaning that way).

However, I think that we should generally thing about a set of options to customize these similar to `ociremote.WithSignatureSuffix`.  We can follow up with options stuff, I'm mostly thinking ahead.

pkg/oci/layout/image.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/sigstore/cosign/pkg/oci/siglayer"
+)
+
+// SignedImage provides access to a remote image reference, and its signatures.
+func SignedImage(path string) (oci.SignedImage, error) {
+	p, err := layout.FromPath(imagePath(path))
+	if err != nil {
+		return nil, err
+	}
+	img, err := p.Image(v1.Hash{})
```
mattmoor, line 34: You have this syntax a few places, I assume it is how singleton indices resolve the image?  Or maybe it accesses the "default" image similar to registries (linux/amd64)?  Worth a comment because if I don't know the next reader will likely struggle 😅

pkg/oci/layout/image.go
```diff
@@ -0,0 +1,103 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/sigstore/cosign/pkg/oci"
+	"github.com/sigstore/cosign/pkg/oci/siglayer"
+)
+
+// SignedImage provides access to a remote image reference, and its signatures.
+func SignedImage(path string) (oci.SignedImage, error) {
+	p, err := layout.FromPath(imagePath(path))
```
mattmoor, line 30: Is this ~always an index?  If so, I'm wondering if we should only expose `SignedIndex` 🤔  (or at least start there instead, and add "sugar" later)

mattmoor (commented):
This is definitely a good start.  Very cool (and unexpected!) to see 🤩

pkg/oci/layout/image.go
imjasonh: I'm not sure what I'm being asked to comment on by being mentioned in this review comment. `p.Image(v1.Hash{})` looks very weird to me though.

priyawadhwa: Yah, so if there's only a single image then `p.Image(v1.Hash{})` should work -- I ended up storing a single image at different subpaths in the directory so I think it should be ok (ref https://github.com/google/go-containerregistry/pull/710)

pkg/oci/layout/write.go
priyawadhwa: i think that's the case, @jonjohnsonjr do you know if there's a better way to do this?

pkg/oci/layout/write.go
priyawadhwa: sgtm, totally agree!

pkg/oci/layout/image.go
priyawadhwa: good call, i only knew about it from looking through ggcr PRs :) added in the comments.

pkg/oci/siglayer/siglayer.go
priyawadhwa: that makes sense to me. opened up https://github.com/sigstore/cosign/pull/1042 so we could see if the change makes sense separately!

pkg/oci/layout/image.go
jonjohnsonjr: The index is the entrypoint into this layout -- you could interpret the entire layout as a single index and push that to a registry, but I think it's more interesting to just use this index data structure as a list of what's contained within the directory. You can have a bunch of cosign-isms around annotations on the top-level `index.json` index that are meaningful to cosign without them leaking from this implementation.

pkg/oci/layout/image.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/sigstore/cosign/pkg/oci/internal/signature"
+)
+
+// SignedImage provides access to a remote image reference, and its signatures.
+func SignedImage(path string) (oci.SignedImage, error) {
+	p, err := layout.FromPath(imagePath(path))
+	if err != nil {
+		return nil, err
+	}
+	// there should only be one image in the index, so we can pass in empty hash to get it
```
jonjohnsonjr, line 33: This is ~fine, but I think aligning a little more closely with the OCI layout-isms would make things a bit nicer. E.g. if everything is reachable from `index.json`, I can just push it to a registry (with any tool that understands an OCI image layout) and have _every_ artifact be available in the registry.

pkg/oci/layout/image.go
jonjohnsonjr: It might be interesting to have an idea of a "snapshot" of a `SignedImage`, that is implemented as an index that points to every piece of relevant metadata?

priyawadhwa:
Hey @jonjohnsonjr & @mattmoor, I refactored a bit to store everything in a single index as recommended. The sample `index.json` looks like this right now, with a manifest for the image and a manifest for the signatures (distinguishable by annotation). Let me know what you think :) 

```json
{
   "schemaVersion": 2,
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 528,
         "digest": "sha256:69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a",
         "annotations": {
            "dev.cosignproject.cosign/image": "true"
         }
      },
      {
         "size": 501,
         "digest": "sha256:4c4dc104ee4357651d17a2e4f8faf2d64457acc7fff2d306312abf69c7bbe4c9",
         "annotations": {
            "dev.cosignproject.cosign/sigs": "true"
         }
      }
   ]
}
```

pkg/oci/layout/index.go
```diff
@@ -0,0 +1,113 @@
...
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+	"github.com/sigstore/cosign/pkg/oci"
+	"github.com/sigstore/cosign/pkg/oci/signed"
+)
+
+const (
+	imageAnnotation = "dev.cosignproject.cosign/image"
+	sigsAnnotation  = "dev.cosignproject.cosign/sigs"
```
mattmoor, line 29: @dlorenc What's standard here?  should this be `dev.sigstore.cosign`?  Are we documenting these anywhere?

pkg/oci/layout/index.go
```diff
@@ -0,0 +1,113 @@
...
+	for _, m := range manifest.Manifests {
+		if _, ok := m.Annotations[annotation]; ok {
+			img, err := i.Image(m.Digest)
+			if err != nil {
+				return nil, err
+			}
+			return img, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to find image")
```
mattmoor, line 101: nit: `errors.New` when no formatting directives

pkg/oci/layout/index.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, m := range manifest.Manifests {
+		if _, ok := m.Annotations[annotation]; ok {
+			img, err := i.Image(m.Digest)
+			if err != nil {
+				return nil, err
+			}
+			return img, nil
```
mattmoor, line 98:
```suggestion
			return i.Image(m.Digest)
```

pkg/oci/layout/index.go
dlorenc: It probably should be dev.sigstore but I apparently never switched them over from the first version, so dev.cosignproject is currently everywhere. We should open a separate issue to track/document/change all of these.

pkg/oci/layout/write.go
```diff
@@ -27,41 +25,29 @@ import (
...
+		return errors.Wrap(err, "appending signed image")
 	}
+	// write the signatures
 	sigs, err := si.Signatures()
 	if err != nil {
 		return errors.Wrap(err, "getting signatures")
 	}
-	if err := write(path, signaturesPath, sigs); err != nil {
-		return errors.Wrap(err, "writing signatures")
+	if err := appendImage(layoutPath, sigs, sigsAnnotation); err != nil {
```
mattmoor, line 42: If the resulting index were pushed, how would it run?

cc @jonjohnsonjr

jonjohnsonjr: Generally clients take the first entry that matches their criteria (platform).

This might be a reasonable knob to expose when pushing to a registry? By default, just push all the entries in `index.json`, but optionally include that `index.json` as a top-level index for everything related to the artifact. I don't know how this is intended to be used.

priyawadhwa: for `cosign load` I was planning on pushing everything in `index.json` for now and not including the `index.json` itself so this effectively is the same as running `cosign copy` -- i think it would be cool to optionally include pushing `index.json` as a feature for later (although running into some errors with it atm that would need to be addressed)

dlorenc (approved):
This looks close enough! Merging so we can keep going!

PR #1042, opened 2021-11-12
Remove img field from sigLayer

priyawadhwa:
We can use the already available Layer to get the payload, so we shouldn't need the `img` field. The only thing we lose is that we no longer call `img.LayerByDigest` so if the descriptor digest is invalid or doesn't match then we don't catch that anymore.

### DIFF
--- a/pkg/oci/remote/layer_test.go
+++ b/pkg/oci/remote/layer_test.go
@@ -22,12 +22,10 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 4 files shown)


pkg/oci/remote/layer_test.go
```diff
@@ -55,9 +53,7 @@ func TestSignature(t *testing.T) {
 	}{{
 		name: "just payload and signature",
 		l: &sigLayer{
-			img: &sigs{
-				Image: must(mutate.Append(ociempty.Signatures(), mutate.Addendum{Layer: layer})),
```
mattmoor, line 59: looks like `must` is now unused

dlorenc (approved):

PR #1043, opened 2021-11-12
split private signature and attestation verification fns

dekkagaijin:
```release-note
NONE
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -113,6 +113,15 @@ func verifyOCISignature(ctx context.Context, verifier signature.Verifier, sig oc
 	return verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader(payload), options.WithContext(ctx))
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cosign/verify.go
```diff
@@ -113,6 +113,15 @@ func verifyOCISignature(ctx context.Context, verifier signature.Verifier, sig oc
 	return verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader(payload), options.WithContext(ctx))
 }
 
+func verifyOCIAttestation(ctx context.Context, verifier signature.Verifier, att oci.Signature) error {
+	payload, err := att.Payload()
+	if err != nil {
+		return err
+	}
+	verifier = newReverseDSSEVerifier(verifier)
```
mattmoor, line 121: Do we even still need this?  Can it be simplified?

mattmoor (approved):

pkg/cosign/verify.go
dekkagaijin: it can be simplified, will continue in a subsequent PR

PR #1047, opened 2021-11-13
verify-blob: add URI to verify-blob output

### LINKED ISSUE #1046: More information about the certificate during verify operations
**Description**

Right now, cosign prints the following information when invoking `verify-blob`:
```
❯ COSIGN_EXPERIMENTAL=1  cosign verify-blob mnemonic-0.3.1.tar.gz --signature mnemonic-0.3.1.tar.gz.sig
Certificate is trusted by Fulcio Root CA
Email: []
Verified OK
tlog entry verified with uuid: "228527476f82c59641e27b8fb9b32f5fadbe47cf42c544ea97dc798490c4c14e" index: 853327
```

The Email entry is empty, because this signature got created via a workload identity.
The only way to validate  the workload identity is to retrieve the public certificate via rekor, storing it on disk and examining it with openssl:

```
❯ rekor-cli get --uuid 228527476f82c59641e27b8fb9b32f5fadbe47cf42c544ea97dc798490c4c14e --format json | jq -r '.Body.RekordObj.signature.publicKey.content' | base64 -d > pub.crt
❯ openssl x509 -noout -text -in pub.crt
```

This behavior is way to complicated. I suggest adding more information to the cosign output. The most interesting fields are these two, IMO:

```
            X509v3 Subject Alternative Name: critical
                URI:https://github.com/shibumi/mnemonic/.github/workflows/goreleaser.yml@refs/tags/v0.3.1
            1.3.6.1.4.1.57264.1.1:
                https://token.actions.githubusercontent.com
```

shibumi:
#### Summary

Adds an URI field to the verify-blob output.

Before this PR:
```
❯ COSIGN_EXPERIMENTAL=1  cosign verify-blob /tmp/mnemonic-0.3.1.tar.gz --signature /tmp/mnemonic-0.3.1.tar.gz.sig
Certificate is trusted by Fulcio Root CA
Email: []
Verified OK
tlog entry verified with uuid: "228527476f82c59641e27b8fb9b32f5fadbe47cf42c544ea97dc798490c4c14e" index: 853327
```

After this PR:

```
❯ COSIGN_EXPERIMENTAL=1  ./cosign verify-blob /tmp/mnemonic-0.3.1.tar.gz --signature /tmp/mnemonic-0.3.1.tar.gz.sig
Certificate is trusted by Fulcio Root CA
Email: []
URI: https://github.com/shibumi/mnemonic/.github/workflows/goreleaser.yml@refs/tags/v0.3.1
Issuer:  https://token.actions.githubusercontent.com
Verified OK
tlog entry verified with uuid: "228527476f82c59641e27b8fb9b32f5fadbe47cf42c544ea97dc798490c4c14e" index: 853327
```

#### Ticket Link

Fixes #1046 

#### Release Note

```release-note
verify-blob gained functionality for printing the URIs of the certificate if there are any.
```

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -174,6 +174,10 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, sigRef, blobRe
 		}
 		fmt.Fprintln(os.Stderr, "Certificate is trusted by Fulcio Root CA")
 		fmt.Fprintln(os.Stderr, "Email:", cert.EmailAddresses)
+		for _, uri := range cert.URIs {
+			fmt.Fprintf(os.Stderr, "URI: %s://%s%s\n", uri.Scheme, uri.Host, uri.Path)
+		}
+		fmt.Fprintln(os.Stderr, "Issuer: ", sigs.CertIssuerExtension(cert))
 	}
 	fmt.Fprintln(os.Stderr, "Verified OK")
 


shibumi:
@dlorenc  do you know how I get this information here?
```
            1.3.6.1.4.1.57264.1.1:
                https://token.actions.githubusercontent.com
```

Is it encoded as ASN1 struct in one of the other certificate fields?

shibumi:
This part works for printing the certificate extension:
```go
		for _, ext := range cert.Extensions {
			oidc := asn1.ObjectIdentifier{1,3,6,1,4,1,57264,1,1}
			if ext.Id.Equal(oidc) {
				fmt.Fprintln(os.Stderr, "1.3.6.1.4.1.57264.1.1: ", string(ext.Value))
			}
		}
```

But I have no idea if we might encounter more of these and if the certificate extension is always the same.
We might want to put that ObjectIdentifier as constant somewhere else.

shibumi:
I am not really convinced yet if printing the x509 extension/object identifier value is a good idea.
I assume that the current object identifier (the hard coded one) is only valid for Github, is this correct?

And throwing all values to the user might be more confusing than it actually helps. 

If the object identifier stays the same, it makes sense to include it. If not, we might just use the URIs?!

shibumi:
Further investigation later. Looks like this OID stays the same:

http://www.oid-info.com/cgi-bin/display?oid=1.3.6.1.4.1.57264&a=display

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -174,6 +175,15 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, sigRef, blobRe
 		}
 		fmt.Fprintln(os.Stderr, "Certificate is trusted by Fulcio Root CA")
 		fmt.Fprintln(os.Stderr, "Email:", cert.EmailAddresses)
+		for _, uri := range cert.URIs {
+			fmt.Fprintf(os.Stderr, "URI: %s://%s%s\n", uri.Scheme, uri.Host, uri.Path)
+		}
+		for _, ext := range cert.Extensions {
+			oidc := asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 1}
```
dlorenc, line 182: I think we might be able to reuse this: https://github.com/sigstore/cosign/blob/7066f122b8288e5621cf7e3dca600075bb70cd47/pkg/signature/keys.go#L219

shibumi: Done

shibumi:
New Output:
```
Certificate is trusted by Fulcio Root CA
Email: []
URI: https://github.com/shibumi/mnemonic/.github/workflows/goreleaser.yml@refs/tags/v0.3.1
Issuer:  https://token.actions.githubusercontent.com
Verified OK
tlog entry verified with uuid: "228527476f82c59641e27b8fb9b32f5fadbe47cf42c544ea97dc798490c4c14e" index: 853327
```

shibumi:
@dlorenc Would it make sense to give verify operations a format flag for machine-readable output? Or is this over the top?

dlorenc:
> Would it make sense to give verify operations a format flag for machine-readable output? Or is this over the top?

I think this makes sense, just a plain json output?

shibumi:
Yes, just a plain json output, but I think it makes sense to address this in another issue.

If we implement a format flag, we have to stay persistent with the JSON schema. People would rely on it, because it is machine readable.

I will create a separate issue for this.

shibumi:
Done: Separate Issue for JSON output: https://github.com/sigstore/cosign/issues/1048

dlorenc (approved):

PR #1052, opened 2021-11-15
Add PKCS11 tag in releaser and Makefile for Mac and Windows

garantir-km:
Signed-off-by: Kieran Miller <kieran.miller@garantir.io>

#### Summary
This PR adds the `pkcs11key` tag for both Mac and Windows releases. 
It also creates a new release `linux-pivkey-pkcs11key-amd64` which supports both PIV and PKCS11 keys, and adds its corresponding target `cosign-pivkey-pkcs11key` in the Makefile.

#### Release Note
```release-note
```


dlorenc:
Cc @cpanato can you take a look at this one?

cpanato:
do we need to generate two binaries one with `pivkey` and another with both `pivkey` and `pkcs11key`, any reason for that? or can we just build one binary with both

dlorenc:
> do we need to generate two binaries one with `pivkey` and another with both `pivkey` and `pkcs11key`, any reason for that? or can we just build one binary with both

I think it should be fine to do one for both.

garantir-km:
Does this look good to you ? @cpanato @dlorenc

cpanato (approved):
looks great to me
thanks for working on this
/lgtm

PR #1053, opened 2021-11-15
fix: reproducible builds

caarlos0:
#### Summary

On #1019 there are some discussion about the builds not being reproducible.

We could also ditch the ldflags and read from `debug.ReadBuildInfo()` only.

What this PR does is:

- enable go mod proxying, so all releases (except snapshots) are built from the gomod proxy version instead of local code

The other bits goreleaser needs for reproducible builds are already there, so this should be working assuming the same inputs are given.

More info:

- https://goreleaser.com/customization/build/#reproducible-builds
- https://goreleaser.com/customization/gomod/
- https://goreleaser.com/cookbooks/build-go-modules/

#### Ticket Link

- refs  #1019

#### Release Note

```release-note

```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -10,6 +10,9 @@ before:
   hooks:
   - go mod tidy
 
+gomod:
+  proxy: true
+
 builds:
 - id: linux
   binary: cosign-linux-{{ .Arch }}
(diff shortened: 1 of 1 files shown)


dlorenc:
We should be good to merge if you get the DCO bit signed!

caarlos0:
oh, no idea why it didn't signoff :|

dlorenc (approved):

cpanato:
thanks for that!

PR #1054, opened 2021-11-15
Update slsa-provenance predicate to v0.2

priyawadhwa:
Update the in-toto-golang dependency to generate the slsa-provenance predicate at version v0.2. This way chains can update to v0.2 as well without conflicting with the pinned version for in-toto-golang in cosign!


pkg/cosign/tuf/client.go
```diff
@@ -251,7 +251,7 @@ func GetTarget(ctx context.Context, name string, out client.Destination) error {
 	return getTargetHelper(name, out, c)
 }
 
-func getRootKeys(rootFileBytes []byte) ([]*data.Key, int, error) {
+func getRootKeys(rootFileBytes []byte) ([]*data.PublicKey, int, error) {
```
priyawadhwa, line 254: cc @asraa when I updated the intoto dependency it automatically updated `go-tuf` as well. Just wanted to make sure this looks correct to you?

asraa: resolved offline!

dlorenc (approved):

Issue #931, opened 2021-10-20
Consider separating Signature vs. Attestation at the Interface level.


vaikas:
**Description**
As I was ramping up on #930 and #919 I found it a bit confusing that both Signature and Attestation are treated as Signature, even though they actually behave differently. I tried to clarify in the documentation to hopefully help other newcomers (or maybe it was just me :) ) but that led me to think that given that since there are some codepaths where these are treated differently in the code if we might want to consider separating the Interface as well. Just a thought.

PR #1056, opened 2021-11-15
remove `reverseDSSEVerifier` in favor of using DSSE utilities directly

dekkagaijin:
More work on #844 and #931

```release-note
NONE
```


pkg/cosign/verify.go
```diff
@@ -34,6 +33,7 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/pkg/errors"
 
+	ssldsse "github.com/secure-systems-lab/go-securesystemslib/dsse"
```
mattmoor, line 36: Are we using this consistently everywhere?  I'm guessing "no" because it required changes to `go.mod`.  Want to make sure nothing else should be using this?

dekkagaijin: We're using it in `sigstore/dsse`. The go.mod changes were to update sigstore/sigstore to a version which exposed `dsse.VerifierAdapter`

mattmoor (approved):

Issue #594, opened 2021-08-30
cosigned: create a CRD to declare signed image policies per cluster


hectorj2f:
**Description**

The idea is to create a cluster-scoped CR to specify a signed image policy on their kubernetes cluster. This policy can be configured to verify signed images from specified patterns.

Based on our implementation of a similar admission controller, we expose a CR with several fields:

`publicKeys`: defines the list of valid public keys.
`registries`: defines the link between registries and defined keys where to apply these rules.
`unmatchedImagePolicy`:  specifies the behavior of the admission controller (Allow|Deny) whenever the policy is satisfied or not.

An example of a Policy:

```
kind: ClusterImagePolicy
metadata:
 name: image-policy
spec:
 unmatchedImagePolicy: <Allow/Deny>
 publicKeys:
 - name: cosign-key-1
   publicKey: <public-key>
 registries:
 - pattern: my.registry.org/library/*
   keys:
   - name: cosign-key-1
  ```

I cc @stormqueen1990 and @kkavitha @elfotografo007 as creators of this approach. So they can chime in here.

dlorenc:
For pattern - what glob matching semantics do we envision? Would it be simpler to limit to "prefix"? Or are we envisioning ** and * stuff for multiple nested "directories"?

hectorj2f:
Yes, we rely on `github.com/gobwas/glob`, so any pattern you could think it'll work here. Maybe we could limit it to a simpler use case for now.

hectorj2f:
@stormqueen1990 and @kkavitha @elfotografo007 Please, could you update the issue with the latest changes to the spec ?

hectorj2f:
It sounds good to me @DennyHoang. I suggest you open a PR with the proposed changes to show the benefits of using this CRD.

hectorj2f:
@mattmoor @mlieberman85 Our team created a PR https://github.com/sigstore/cosign/pull/1064 with a first approach, so we could start again the discussions.

JimBugwadia:
Looks like a number of the items discussed above and future considerations (match/exclude, HA, fail-closed vs open, external data, wildcarding, etc.) have been addressed for Kyverno along with other features like support for attestations.

Does it make sense to explore how to best reuse? 

If a stand-alone admission controller is desirable, perhaps we can consider a library approach like was done for Pod Security Admission, so we are not replicating work with minor variations.

@hectorj2f @dlorenc @mattmoor - would love to discuss thoughts around this.

[...]

PR #1064, opened 2021-11-16
cosigned: Add a policy CRD to verify image signatures

coyote240:
#### Summary
This pull request is a draft implementation of the CRD policy mentioned in this issue: https://github.com/sigstore/cosign/issues/594

This creates a CustomResourceDefinition that allows specifying:
1. An image name pattern to match container images
1. Signatures of the container images that match the pattern will be verified with associated keys
1. Resources with images not matching the pattern will be rejected (This workflow is up for discussion)

To run this:

Pre reqs:
1. `controller-gen` version `0.4.1`

Steps:
1. `make manifests` (This will generate the CRD yaml)
1. `make generate` (This will generate the deepcopy)
1. `make ko-local` (We added target to generate cosigned image)
1. `kubectl apply -f config/crd/bases` (This will create the CRD in the cluster)
1. Update the `config/webhook.yaml`. Set the public key in the `spec.keys.publicKey` field of the image-policy CRD.
1. `kubectl apply -f config/` (This will install webhook with necessary resources)

**Note**: The namespace still needs to be labeled to be monitored by the webhook

What needs to still be fleshed out:
1. Namespace/resource exclusion using the CRD
1. Way to specify the workflow if the images do not match any pattern (right now it gets rejected)
1. Allow multiple policies
1. `Keys` for the image pattern can be made optional if the signature needs to be verified using transparency logs

Please let us know your thoughts about this implementation

cc: @kkavitha @hectorj2f

#### Release Note
```release-note
1. Feature: Use an image policy CRD to verify signatures of container images

```


Makefile
```diff
@@ -92,6 +92,15 @@ cosign-pkcs11key: $(SRCS)
 cosigned: ## Build cosigned binary
 	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" -o $@ ./cmd/cosign/webhook
 
+
+.PHONY: manifests
```
hectorj2f, line 96: you might want to add `manifests`  to `cosigned` target.

config/webhook.yaml
```diff
@@ -40,10 +40,10 @@ spec:
       serviceAccountName: webhook
       containers:
       - name: webhook
+        imagePullPolicy: Always
```
mattmoor, line 43: This is more or less pointless with `ko` since you'll get a digest in the image fields.

config/webhook.yaml
```diff
@@ -87,11 +87,20 @@ spec:
       # high value that we respect whatever value it has configured for the lame duck grace period.
       terminationGracePeriodSeconds: 300
 ---
-apiVersion: v1
-kind: Secret
+apiVersion: sigstore.dev/v1alpha1
+kind: ClusterImagePolicy
```
mattmoor, line 91: Kubernetes it terrible about installing CRDs and immediately instantiating them.

IMO the default behavior should not require an instance of the CRD to be installed.

config/webhook.yaml
```diff
@@ -87,11 +87,20 @@ spec:
...
 metadata:
-  name: verification-key
-  namespace: cosign-system
-# stringData:
-#   cosign.pub: |
-#     <PEM encoded public key>
\ No newline at end of file
+  name: image-policy
```
mattmoor, line 93: I didn't realize this until below, but I was expecting the webhook to support N of these CRDs, not a singleton with a fixed name 😅

pkg/cosign/kubernetes/api/v1alpha1/imagepolicy_types.go
```diff
@@ -0,0 +1,98 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ClusterImagePolicySpec defines the desired state of ImagePolicy
+type ClusterImagePolicySpec struct {
```
mattmoor, line 24: To reiterate my request on the issue:
> I think we should probably start a doc enumerating key scenarios (perhaps ... can help provide a "voice of the user" to keep us honest?), and then sketch a more formal API spec where we spell out the fields and semantics. I'd be happy to collaborate if we want to get a group of interested folks spun up around this?

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -104,20 +108,23 @@ func (v *Validator) ValidateCronJob(ctx context.Context, c *duckv1.CronJob) *api
...
-	if err != nil {
-		return apis.ErrGeneric(err.Error(), apis.CurrentField)
-	}
-
-	keys, kerr := getKeys(ctx, s.Data)
-	if kerr != nil {
-		return kerr
+	var clusterPolicy = schema.GroupVersionResource{Group: "sigstore.dev", Version: "v1alpha1", Resource: "clusterimagepolicies"}
+	
+	imagePolicy, _ := v.dynamicClient.Resource(clusterPolicy).Get(ctx, "image-policy", metav1.GetOptions{})
```
mattmoor, line 119: This is an important error to not ignore

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -104,20 +108,23 @@ func (v *Validator) ValidateCronJob(ctx context.Context, c *duckv1.CronJob) *api
 }
 
 func (v *Validator) validatePodSpec(ctx context.Context, ps *corev1.PodSpec, opt k8schain.Options) (errs *apis.FieldError) {
+	// Read the CRD
```
mattmoor, line 111: What's immediately below this isn't reading the CRD.

pkg/cosign/kubernetes/webhook/validator.go
mattmoor: This also hits the API server on every webhook request, where the previous version hit the lister cache on requests.

mattmoor: I was also not expecting the use of a singleton here.  How are users expected to know this?  Are there other policy bits in Kubernetes that use a CRD for a global singleton configuration?

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -137,6 +144,27 @@ func (v *Validator) validatePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
 				).ViaFieldIndex(field, i))
 				continue
 			}
+			// Match the pattern
+			// if matched, use the respective key to verify
+			// Read the keys from the CRD from the namepattern that matches
+			keys := []*ecdsa.PublicKey{}
+			matched := false
+			for _, imagePattern := range imagePolicyType.Spec.Verification.Images {
+				imagePatternGlob := glob.MustCompile(imagePattern.NamePattern)
```
mattmoor, line 153: MustCompile panics when things fail, which is a very undesirable property of webhooks that validate pods and fail closed.  Is the validation checking that this is a valid regular expression to make this safe?

mattmoor (changes requested):
I'd really rather start from a doc outlining the schema and covering the semantics.  This is basically just using the CRD as a schematized ConfigMap, but I was expecting something closer to other K8s policy types.

config/webhook.yaml
coyote240: Agreed, 100%, though as a first pass, wanted to start here. I feel like there are questions re: resolution of duplicates (conflicts) across the _n_ policies that needs to be discussed.

pkg/cosign/kubernetes/api/v1alpha1/imagepolicy_types.go
kkavitha: Sorry, relatively new here! so I have a few questions. 
Do you have a drive where we can create/save a doc to detail this? I can start the doc. Whats the process to form an interest group? Do we just open this conversation up in the community meeting?

mattmoor: I think some sort of limited run API WG/task-force might be better than hashing it out in the bigger community meeting.  Maybe we can ask / socialize this at the next community call?  (cc @lukehinds @dlorenc )

lukehinds: add it to the agenda @mattmoor , sounds like a good idea!

lukehinds: @mattmoor et al,

had a chat with the TAC and recommend:

1. Set a date in the 2nd half of next week (or later) to meet on hangouts (to give enough time to invite others on next weeks community call)
2. Add sigstore-dev@googlegroups.com to the invite
3. You can repeat it for as long as you like.
4. Might be worth recording as well, I can add to the sigstore community YouTube channel

vaikas:
@coyote240 I think this was before the task force completed this work? Just wondering if this should still be kept open with the other ongoing work happening?

coyote240:
> @coyote240 I think this was before the task force completed this work? Just wondering if this should still be kept open with the other ongoing work happening?

Yes, this is ancient. We should certainly close it.

coyote240:
Closing, superseded by PR#1540

PR #1068, opened 2021-11-17
feat: support signature file in verify cmd

### LINKED ISSUE #1067: cosign verify not working with signature file
**Description**

We have an option to output signature to the standard output with the cosign sign command, but there is no way to use this signature with verify command.

Shouldn’t we accept this signature while using verify command? Because my signature is not uploaded to the registry, so, I can’t verify it with verify command, right?

you can reproduce what I’m saying by issuing the following commands on your terminal:

```shell
$ cosign generate-key-pair
$ cosign sign --upload=false --key cosign.key <yourdockerid>/alpine:latest > my.sig
$ cosign verify --key cosign.pub <yourdockerid>/alpine:latest
# it should fail, we should pass the signature to the verify command to make it success
```

cc: @Dentrax

developer-guy:
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>

#### Summary

Add ability to verify via signature reference in verify cmd

#### Ticket Link

Fixes #1067 

#### Release Note

```release-note
feat: support signature file in verify cmd
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -25,8 +25,12 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 	"time"
+
+	"github.com/sigstore/cosign/pkg/blob"
+	"github.com/sigstore/cosign/pkg/oci/static"
 
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/google/go-containerregistry/pkg/name"
@@ -43,6 +47,7 @@ import (
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/dsse"
 	"github.com/sigstore/sigstore/pkg/signature/options"
+	sigPayload "github.com/sigstore/sigstore/pkg/signature/payload"
 )
 
 // CheckOpts are the options for checking signatures.
@@ -67,6 +72,9 @@ type CheckOpts struct {
 	RootCerts *x509.CertPool
 	// CertEmail is the email expected for a certificate to be valid. The empty string means any certificate can be valid.
 	CertEmail string
+
+	// SignatureRef is the reference to the signature file
+	SignatureRef string
 }
 
 func getSignedEntity(signedImgRef name.Reference, regClientOpts []ociremote.Option) (oci.SignedEntity, v1.Hash, error) {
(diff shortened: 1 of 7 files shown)


pkg/cosign/verify.go
```diff
@@ -281,6 +337,11 @@ func VerifyImageSignatures(ctx context.Context, signedImgRef name.Reference, co
 	return checkedSignatures, bundleVerified, nil
 }
 
+func isb64(data []byte) bool {
```
naveensrinivasan, line 340: A suggestion. Why not inline this in the main function? This reduces the readability from jumping from one function to another. In go it is generally not a practice to have a small function with a simple check.

developer-guy:
kindly ping @dlorenc

dlorenc:
looks like this one got hit with a merge conflict. It LGTM otherwise.

developer-guy:
it should be okay now @dlorenc

pkg/cosign/verify.go
```diff
@@ -196,15 +213,54 @@ func VerifyImageSignatures(ctx context.Context, signedImgRef name.Reference, co
...
-		return nil, false, err
+
+	var sigs oci.Signatures
+	sigRef := co.SignatureRef
+	if sigRef == "" {
+		sigs, err = se.Signatures()
+		if err != nil {
+			return nil, false, err
+		}
+	} else {
```
hectorj2f, line 228: nit: I'd suggest moving this piece of code to a function to reduce the size and logic in this function.

pkg/cosign/verify.go
```diff
@@ -230,7 +286,7 @@ func VerifyImageSignatures(ctx context.Context, signedImgRef name.Reference, co
 					return err
 				}
 				if cert == nil {
-					return errors.New("no certificate found on signature")
+					return errors.New("no certificate found on sigRef")
```
hectorj2f, line 289: Suggestion: `signature` sounds correct to me, it refers to the command argument name, imho.

pkg/cosign/verify.go
hectorj2f: I agree with @naveensrinivasan, wrapping a function call inside of another function isn't ideal.

pkg/cosign/verify.go
developer-guy: thank you @hectorj2f, done.

pkg/cosign/verify.go
developer-guy: thank you @hectorj2f, done.

pkg/cosign/verify.go
developer-guy: thank you @hectorj2f, done.

hectorj2f (approved):
lgtm 👍🏻

Issue #946, opened 2021-10-24
Cosign with AWS KMS and AWS ECR doesn't work as expected for specific type of keys


beingamarnath:
### Description

Cosign has the support for AWS KMS as a means to provide keys for Signing and also AWS ECR is as well supported as a Container registry to have the signing information. 

So I tried to take the duo for testing and below are my findings .

P.S: I tried all the below setup also with Jfrog Artifactory as Container Registry and the results are the same.

AWS KMS 1 (alias/cosign-ec512):
-------------------------------
Key Type - Asymmetric
Origin: AWS_KMS
Key Spec: ECC_NIST_P521
Key Usage: Sign and Verify
Signing algorithms: ECDSA_SHA_512

Sign:
```
cosign sign --key awskms:///alias/cosign-ec512 <ACCOUNT>.dkr.ecr.<REGION>.amazonaws.com/<REPO>/<IMAGE_NAME>:ECDSA_SHA_512
```
Sign Output:
 ```
 Pushing signature to: <ACCOUNT>.dkr.ecr.<REGION>.amazonaws.com/<REPO>/<IMAGE_NAME>:sha256-6b444afd4dbdb32db7f5e31c7b70788680ff7e0876cf60e8acfbaa4b60740d15.sig
 ```
 Get Public Key:
 ```
 cosign public-key --key awskms:///alias/cosign-ec512 > kms-ec512.pub
 ```
 Verify:
 ```
cosign verify --key ./kms-ec512.pub <ACCOUNT>.dkr.ecr.<REGION>.amazonaws.com/<REPO>/<IMAGE_NAME>:ECDSA_SHA_512
 ```
 Verify Output:
 ```
error: no matching signatures:
failed to verify signature
 ```

AWS KMS 2 (alias/cosign-ec384):
-------------------------------
Key Type - Asymmetric
Origin: AWS_KMS
Key Spec: ECC_NIST_P384
Key Usage: Sign and Verify
Signing algorithms: ECDSA_SHA_384

Sign:
```
cosign sign --key awskms:///alias/cosign-ec384 <ACCOUNT>.dkr.ecr.<REGION>.amazonaws.com/<REPO>/<IMAGE_NAME>:ECDSA_SHA_384
```
Sign Output:
 ```
[...]

PR #1071, opened 2021-11-18
Add flag for manually specifying a hash algo when verifying

SgtCoDFish:
#### Summary

While, at the time of writing, it's not possible to manually specify the signature digest algorithm which should be used when signing an image, some KMS providers have key types which force a digest algorithm which isn't sha256; e.g. GCP KMS with RSA4096 / SHA512 keys.

Cosign will happily use these keys for signing and will infer the digest algorithm based on what the KMS provider mandates, leading to a situation where cosign generates signatures which it can't later verify.

This commit adds a new `--signature-digest-algorithm`  flag which allows the user to manually specify one of 4 commonly-used secure hashes when using `cosign verify`.

#### Notes

1. Normally I'd (of course) add an e2e test for this kind of thing, but I can't see an obvious way of generating, say, a SHA512 signature without using a cloud KMS currently, and unless I'm missing something it doesn't look trivial to add that. I think adding an e2e test would require adding `--signature-digest-algorithm` to `cosign sign`, which I think would be desirable but is a much bigger piece of work and maybe needs more discussion.
2. I'm not totally sold on  `--signature-digest-algorithm` as a flag name - it's a bit wordy. Happy to rename if someone has a better suggestion!

#### Testing

I've tested that this works correctly using a test image which I think _should_ be publicly available. Using this commit, I can successfully run:

```console
curl -sSL https://raw.githubusercontent.com/SgtCoDFish/ace/main/PUBKEY > /tmp/PUBKEY
$ ./cosign verify --key "/tmp/PUBKEY" --signature-digest-algorithm SHA512 ghcr.io/sgtcodfish/ace:0.5
```

#### Ticket Link

Related to #946, specifically [this comment](https://github.com/sigstore/cosign/issues/946#issuecomment-966519321). That issue is more broad, though, and there's probably more work to be done in the future for that issue to be completed.

#### Release Note

```release-note
- Added the `--signature-digest-algorithm` flag to `cosign verify`, allowing verification of container image signatures which were generated with a non-SHA256 signature algorithm
```

### DIFF
--- a/cmd/cosign/cli/options/signature_digest.go
+++ b/cmd/cosign/cli/options/signature_digest.go
@@ -1,0 +1,85 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package options
+
+import (
+	"crypto"
+	"fmt"
+	"sort"
+	"strings"
+
+	_ "crypto/sha256" // for sha224 + sha256
+	_ "crypto/sha512" // for sha384 + sha512
+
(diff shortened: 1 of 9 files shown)


dlorenc (approved):

dlorenc:
> 2\. I'm not totally sold on  `--signature-digest-algorithm` as a flag name - it's a bit wordy. Happy to rename if someone has a better suggestion!

I couldn't think of anything better...

SgtCoDFish:
Thanks very much, that was a really smooth process :grin:

PR #1081, opened 2021-11-19
Use hashed rekord type for tlog upload

asraa:

#### Summary

* Use a hashed rekord type for tlog upload to allow for signing large blobs with keyless flow.

**THIS IS A NOT A DRAFT ANYMORE BECAUSE...**
1. Prod rekor needs to be deployed with the new type **(DONE)**
2. The bigger issue is that if someone is using `COSIGN_EXPERIMENTAL` and uploading signed entries, let's say, with a KMS provider that uses a specific hash algorithm, then defaulting to the SHA256 I hard-coded in `rekorEntry` will be incorrect and will fail signature verification. The hash algorithm needs to be plumbed through, so I imagine that SignerVerifier would have to expose a new method to it's interface providing the hash function that it uses by default (doesn't look like cosign will ever use a `WithCryptoSignerOpts(otherHashFunc)` on its `SignMessage`s
- to get around this I guess I could specifically only have hashedRekord type being used for fulcio generated certs that we know will be using sha256? that'll at least help keyless sign-blobs. but i would rather not solve just a tiny part of the problem

#### Ticket Link

Fixes https://github.com/sigstore/rekor/issues/481

#### Release Note

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,13 +24,13 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/secure-systems-lab/go-securesystemslib v0.1.0
 	github.com/sigstore/fulcio v0.1.2-0.20210831152525-42f7422734bb
-	github.com/sigstore/rekor v0.3.1-0.20211117161348-09070aa96aef
+	github.com/sigstore/rekor v0.3.1-0.20211119021052-57b3fef494e0
 	github.com/sigstore/sigstore v1.0.2-0.20211115214857-534e133ebf9d
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/viper v1.9.0
 	github.com/stretchr/testify v1.7.0
 	github.com/tent/canonical-json-go v0.0.0-20130607151641-96e4ba3a7613
-	github.com/theupdateframework/go-tuf v0.0.0-20211115152232-a4f2dd6ea314
+	github.com/theupdateframework/go-tuf v0.0.0-20211006142131-1dc15a86c64d
 	github.com/xanzy/go-gitlab v0.52.0
 	golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -441,8 +441,10 @@ func VerifyBundle(ctx context.Context, sig oci.Signature) (bool, error) {
 func bundleHash(bundleBody, signature string) (string, string, error) {
 	var toto models.Intoto
 	var rekord models.Rekord
+	var hrekord models.Hashedrekord
 	var intotoObj models.IntotoV001Schema
 	var rekordObj models.RekordV001Schema
+	var hrekordObj models.HashedrekordV001Schema
 
 	bodyDecoded, err := base64.StdEncoding.DecodeString(bundleBody)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


dlorenc:
> 2\. The bigger issue is that if someone is using `COSIGN_EXPERIMENTAL` and uploading signed entries, let's say, with a KMS provider that uses a specific hash algorithm, then defaulting to the SHA256 I hard-coded in `rekorEntry` will be incorrect and will fail signature verification. The hash algorithm needs to be plumbed through, so I imagine that SignerVerifier would have to expose a new method to it's interface providing the hash function that it uses by default (doesn't look like cosign will ever use a `WithCryptoSignerOpts(otherHashFunc)` on its `SignMessage`s

The spec pretty much requires SHA256 right now. People can technically use other ones in a KMS, but I'd be fine fixing that later, maybe built on #1071

asraa:
> The spec pretty much requires SHA256 right now. 

Do you mean the spec for keyless signing? Or more generally. I'm just worried about someone doing KMS w/ sha512 + rekor upload

dlorenc:
> Do you mean the spec for keyless signing? Or more generally. I'm just worried about someone doing KMS w/ sha512 + rekor upload

The spec for all of cosign: https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md

asraa:
> The spec for all of cosign: https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md

doh right, thanks! yeah i suppose working out the issues for sign-blobs with KMS can come later

go.mod
```diff
@@ -55,3 +55,5 @@ require (
 	k8s.io/klog/v2 v2.20.0 // indirect
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b // indirect
 )
+
+replace github.com/sigstore/rekor => github.com/sigstore/rekor v0.3.1-0.20211119021052-57b3fef494e0
```
dlorenc, line 59: we can probably do this with a normal bump instead of the replace here.

dlorenc:
We should get this one in before v1.4.0 as well!

go.mod
asraa: I think we'll need a new release here in order to include hashedrekord type!

dlorenc: Ah I think you can do something like: `go get github.com/sigstore/rekor@cae63bd4e389b56e352c7ffebbb86d1c1437409b` to pin to an unreleased commit.

asraa: Argh thank you! sorry misinterpreted what you meant by normal bump!

asraa:
QQ about this: while this change happens, should cosign look for both the rekord proposed entry and, if it doesn't exist, the hashedrekord proposed entry? i think it's okay, because two sigs are always gonna be unique because of randomization, so one or the other should exist, not both. (I suppose it's possible for someone to hand craft a rekord and hashedrekord with the same exact signature, but then it shouldn't matter unless we accidentally serve a different form of the public key, e.g. the x509 cert vs the extracted public key that misses some client verification)

dlorenc:
> QQ about this: while this change happens, should cosign look for both the rekord proposed entry and, if it doesn't exist, the hashedrekord proposed entry? i think it's okay, because two sigs are always gonna be unique because of randomization, so one or the other should exist, not both. (I suppose it's possible for someone to hand craft a rekord and hashedrekord with the same exact signature, but then it shouldn't matter unless we accidentally serve a different form of the public key, e.g. the x509 cert vs the extracted public key that misses some client verification)

Just using the new format seems fine, most lookups go through the offline bundle verification anyway.

asraa:
Done! Updated the offline bundle verification to unmarshal either or. Tested both container/blob verification

https://gist.github.com/asraa/c2ef0bd3aacb7073b1e0736c0f9fd557

pkg/cosign/verify.go
```diff
@@ -469,20 +471,32 @@ func bundleHash(bundleBody, signature string) (string, string, error) {
 		return *intotoObj.Content.Hash.Algorithm, *intotoObj.Content.Hash.Value, nil
 	}
 
-	err = json.Unmarshal(bodyDecoded, &rekord)
+	if err := json.Unmarshal(bodyDecoded, &rekord); err == nil {
```
dlorenc, line 474: did you double check this actually fails? Sometimes unmarshal will succeed even if it's the wrong type. We might need to get a bit more clever here.

asraa: yep! can confirm that for a hashed rekord signature, it actually travels the second path and doesn't unmarshal to a rekord without an error

asraa: i just checked how it errors: errors out on "invalid kind value" (rekor checks to make sure that the type we want to unmarshal to is the kind in the JSON)

dlorenc: Awesome!

dlorenc (approved):

PR #1083, opened 2021-11-19
Drop `strfmt.Base64` from `pkg/oci`.

mattmoor:
I noticed that vendoring `pkg/oci` was pulling in stuff from `mongodb`, and pulling the thread it seems to come from this line.

`Base64` is an alias of `[]byte`, and changing it seems to break nothing, so great!

Signed-off-by: Matt Moore <mattmoor@chainguard.dev>

#### Release Note
```release-note
The SignedEntryTimestamp is now typed []byte, to get the previous behavior call strfmt.Base64(SignedEntryTimestamp)
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -403,7 +403,7 @@ func VerifyBundle(sig oci.Signature) (bool, error) {
 		return false, errors.Wrap(err, "pem to ecdsa")
 	}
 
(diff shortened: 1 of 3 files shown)


imjasonh (approved):

dlorenc (approved):

mattmoor:
Ok I lied, I was expecting build failures, so I passed `-run=^$` but there is one `""` -> `null` needed for this to not break.

mattmoor:
LOL we also had a place where we were casting to `[]byte()`!  🤣

PR #1091, opened 2021-11-22
Kubernetes: set secret immutable by default for version >= 1.21

### LINKED ISSUE #1090: provide an option to  enable creating immutable secrets while generating key pair with k8s scheme
**Description**

I think `immutability` is a critical part while working on multi-tenant systems like Kubernetes. At the time I wrote this, `cosign` supports generating key pairs, and storing them in Kubernetes secrets. Kubernetes introduces us `Immutable Secrets`[^1] with the v1.21.1 release, so, maybe users might want to use this option to enable storing key pairs in the immutable secrets.

cc: @Dentrax

[^1]: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable

Dentrax:
Fixes #1090

Co-authored-by: Erkan <erkan.zileli@trendyol.com>

### DIFF
--- a/pkg/cosign/kubernetes/client.go
+++ b/pkg/cosign/kubernetes/client.go
@@ -19,6 +19,7 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 
+	utilversion "k8s.io/apimachinery/pkg/util/version"
 	// Initialize all known client auth plugins
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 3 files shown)


Dentrax:
Not sure whether setting _immutable: true_ by default is good practice or not. Feel free to drop your feedback here, so we can add smth like `--immutable` flag to trigger the condition.

pkg/cosign/kubernetes/client.go
```diff
@@ -57,3 +58,16 @@ func Client() (kubernetes.Interface, error) {
 	}
 	return kubernetes.NewForConfig(config)
 }
+
+func CheckImmutableSecretSupported(client kubernetes.Interface) (bool, error) {
```
hectorj2f, line 62: nit: It is been used internally in the package, `func checkImm....`.

dlorenc: +1 here, otherwise LGTM!

dlorenc (approved):

PR #1092, opened 2021-11-22
Bump go-containerregistry to pickup the update to image-spec

dlorenc:
This is currently flagged by dependabot because of the CVE. It doesn't
actually matter here, bu we should try to silence the warning if we can.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


azure.key
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEzxfKqNeUnhJGkFKo8uE3GRoCmfK/
```
cpanato, line 2: this is expected?

dlorenc: Whoops, that was from debugging the azure kms bug. No idea how it ended up in this branch!

hectorj2f (approved):

PR #1095, opened 2021-11-22
feat: sign --output-certificate and verify --cert 

caarlos0:
#### Summary

`sign-blob` got a `--output-certificate` flag, so I thought it would be cool to do the same for `sign`.

Not sure if this makes sense, and this PR is very WIP just to see if it would work, but, if we decide to get this merged, it seems that we also need a `--cert` on the `verify` command.

Happy to hear your thoughts on this.

#### Ticket Link
n/a

#### Release Note

```release-note
added --output-certificate to sign and --cert to verify
```


dlorenc:
> Not sure if this makes sense, and this PR is very WIP just to see if it would work, but, if we decide to get this merged, it seems that we also need a `--cert` on the `verify` command.
> 
> Happy to hear your thoughts on this.

I think it's slightly less useful, but don't see an issue with it. Supporting `--cert` on `verify` is almost more compelling for orgs that sign with their own PKI - and it also motivates adding this to sign over here.

Go for it!

caarlos0:
I believe its a bit better now... not sure if we want to add the base64 thing to sign as well, as its not an option there today.

on that note: I wonder if its worth having it on other commands as well, and why is it the default... seems easy enough to `base64 foo > bar` if needed... but I'm probably just not aware of some usecase 🤔

cmd/cosign/cli/sign/sign.go
```diff
@@ -288,6 +288,19 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
...
+	if outputCertificate != "" {
+		rekorBytes, err := sv.Bytes(ctx)
+		if err != nil {
+			return err
+		}
+
+		if err := os.WriteFile(outputCertificate, rekorBytes, 0600); err != nil {
+			return err
+		}
+		// TODO: maybe accept a --b64 flag ass well?
```
mattmoor, line 301:
```suggestion
		// TODO: maybe accept a --b64 flag as well?
```

caarlos0: 😳

developer-guy (approved):
LGTM

PR #1098, opened 2021-11-24
continued sign refacc

dekkagaijin:
This continues to refactor signing operations re: #844 and #931

TODO:
* move over fulcio operations as well
* perform all signing operations via this new library
* perform all attestation operations via this new library
* for efficient reuse, expose the rekor client in tlog upload operations

```release-note
NONE
```

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -21,7 +21,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/rsa"
 	"crypto/x509"
-	"encoding/base64"
 	"encoding/pem"
 	"fmt"
 	"net/url"
@@ -209,11 +208,6 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
 		defer out.Close()
 	}
 
-	req := &icos.SigningRequest{
-		SignaturePayload: payload,
-		SignedEntity:     se,
-	}
-
 	var s icos.Signer
 	s = &icos.PayloadSigner{
 		PayloadSigner: sv,
@@ -229,31 +223,41 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
 			RekorURL: ko.RekorURL,
 		}
 	}
-	s = &icos.OCISignatureBuilder{
-		Inner: s,
-	}
-	s = &icos.OCISignatureAttacher{
-		DD:    dd,
-		Inner: s,
-	}
-	if upload {
-		s = &icos.RemoteSignerWrapper{
-			SignatureRepo: digest.Repository,
-			RegOpts:       regOpts,
 
-			Inner: s,
-		}
-	}
-
-	results, err := s.Sign(ctx, req)
+	ociSig, _, err := s.Sign(ctx, bytes.NewReader(payload))
 	if err != nil {
 		return err
 	}
 
-	b64sig := base64.StdEncoding.EncodeToString(results.Signature)
+	b64sig, err := ociSig.Base64Signature()
+	if err != nil {
+		return err
+	}
 	if _, err := out.Write([]byte(b64sig)); err != nil {
 		return errors.Wrap(err, "write signature to file")
 	}
+
+	if !upload {
+		return nil
+	}
+
+	// Attach the signature to the entity.
+	newSE, err := mutate.AttachSignatureToEntity(se, ociSig, mutate.WithDupeDetector(dd))
+	if err != nil {
+		return err
+	}
+
+	// Publish the signatures associated with this entity
+	walkOpts, err := regOpts.ClientOpts(ctx)
+	if err != nil {
+		return errors.Wrap(err, "constructing client options")
+	}
+
+	// Publish the signatures associated with this entity
+	if err := ociremote.WriteSignatures(digest.Repository, newSE, walkOpts...); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -38,6 +38,7 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	icos "github.com/sigstore/cosign/internal/pkg/cosign"
 	ifulcio "github.com/sigstore/cosign/internal/pkg/cosign/fulcio"
+	ipayload "github.com/sigstore/cosign/internal/pkg/cosign/payload"
 	irekor "github.com/sigstore/cosign/internal/pkg/cosign/rekor"
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/cosign/pkg/cosign/pivkey"
@@ -211,9 +212,7 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
 	}
 
 	var s icos.Signer
-	s = &icos.PayloadSigner{
-		PayloadSigner: sv,
-	}
+	s = ipayload.NewSigner(sv, nil, nil)
 	s = ifulcio.NewSigner(s, sv.Cert, sv.Chain)
 	if ShouldUploadToTlog(ctx, digest, force, ko.RekorURL) {
 		s = irekor.NewSigner(s, ko.RekorURL)
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -37,6 +37,8 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioverifier"
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	icos "github.com/sigstore/cosign/internal/pkg/cosign"
+	ifulcio "github.com/sigstore/cosign/internal/pkg/cosign/fulcio"
+	irekor "github.com/sigstore/cosign/internal/pkg/cosign/rekor"
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/cosign/pkg/cosign/pivkey"
 	"github.com/sigstore/cosign/pkg/cosign/pkcs11key"
@@ -212,16 +214,9 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
 	s = &icos.PayloadSigner{
 		PayloadSigner: sv,
 	}
-	s = &icos.FulcioSignerWrapper{
-		Cert:  sv.Cert,
-		Chain: sv.Chain,
-		Inner: s,
-	}
+	s = ifulcio.NewSigner(s, sv.Cert, sv.Chain)
 	if ShouldUploadToTlog(ctx, digest, force, ko.RekorURL) {
-		s = &icos.RekorSignerWrapper{
-			Inner:    s,
-			RekorURL: ko.RekorURL,
-		}
+		s = irekor.NewSigner(s, ko.RekorURL)
 	}
 
 	ociSig, _, err := s.Sign(ctx, bytes.NewReader(payload))
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -50,9 +50,6 @@ import (
 	providers "github.com/sigstore/cosign/pkg/providers/all"
 	sigs "github.com/sigstore/cosign/pkg/signature"
 	fulcPkgClient "github.com/sigstore/fulcio/pkg/client"
-	rekPkgClient "github.com/sigstore/rekor/pkg/client"
-	rekGenClient "github.com/sigstore/rekor/pkg/generated/client"
-	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	signatureoptions "github.com/sigstore/sigstore/pkg/signature/options"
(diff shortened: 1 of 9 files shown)


internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,188 @@
...
+type SigningResults struct {
+	SignedPayload []byte
+	Signature     []byte
+	Pub           crypto.PublicKey
+	Cert, Chain   []byte
+	Bundle        *oci.Bundle
+
+	OCISignature oci.Signature
+	SignedEntity oci.SignedEntity
+}
```
mattmoor, line 48: The SignedEntity is what's passed in, and virtually all of the stuff on top is accessible from the `oci.Signature`, so I think the result should just be `oci.Signature`.

internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Attach the signature to the entity.
+	newSE, err := mutate.AttachSignatureToEntity(results.SignedEntity, results.OCISignature, mutate.WithDupeDetector(sa.DD))
+	if err != nil {
+		return nil, err
+	}
+	results.SignedEntity = newSE
```
mattmoor, line 156: Don't conflate signing with mutating the signed entity to carry that signature.  The signer should *just* sign, let the caller worry about this.

internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,188 @@
...
+	pkOpts = append(pkOpts, ps.PublicKeyProviderOpts...)
+	pk, err := ps.PayloadSigner.PublicKey(pkOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &SigningResults{
+		Signature:     sig,
+		SignedPayload: req.SignaturePayload,
+		SignedEntity:  req.SignedEntity,
```
mattmoor, line 80: This seems to be the only use of `req.SignedEntity`.  What are our expectations of the Payload?

cmd/cosign/cli/sign/sign.go
```diff
@@ -431,14 +382,11 @@ func keylessSigner(ctx context.Context, ko KeyOpts) (*SignerVerifier, error) {
...
-			return nil, errors.Wrap(err, "getting key from Fulcio")
-		}
 	} else {
 		k, err = fulcioverifier.NewSigner(ctx, tok, ko.OIDCIssuer, ko.OIDCClientID, fClient)
-		if err != nil {
-			return nil, errors.Wrap(err, "getting key from Fulcio")
-		}
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "getting key from Fulcio")
```
mattmoor, line 389: This saves ~3 LoC and is less clear, I'd revert.

If you want to save 2 LoC without losing clarity, you can fold the call into the `if`:
```go
if k, err = fulcio.NewSigner(...); err != nil {
     return nil, errors.Wrap(err, "getting key from Fulcio")
}
```

mattmoor: Also, I would expect that ultimately the bonus logic in `fulcioverifier` just becomes yet-another composition of the `Signer` interface.

internal/pkg/cosign/sign.go
mattmoor: So it seems like this result struct is built up across multiple implementations that must be composed a particular way to work, which isn't really the goal of composition. 🤔 

Given the pieces I'm seeing I think the interface should just be:
```go
type Signer interface {
	Sign(ctx context.Context, payload []byte) (oci.Signature, error)
}
```

pkg/oci/remote/write.go
```diff
@@ -74,7 +74,7 @@ func WriteSignedImageIndexImages(ref name.Reference, sii oci.SignedImageIndex, o
 func WriteSignatures(repo name.Repository, se oci.SignedEntity, opts ...Option) error {
 	o := makeOptions(repo, opts...)
 
-	// Access the signature list to publish
+	// Access the signature list to published
```
mattmoor, line 77: this is an incorrect correction

dekkagaijin: ACK

internal/pkg/cosign/sign.go
dekkagaijin: roger

internal/pkg/cosign/sign.go
dekkagaijin: I was thinking a prior step might want to modify the `req.SignedEntity`, which is immutable, so I wanted to return the actual Entity in the results.

Our expectations for payload are to be able to upload it to Rekor, but we can get that from the `oci.SignedEntity`

cmd/cosign/cli/sign/sign.go
dekkagaijin: right, already have a TODO for it

internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// PayloadSigner implements `Signer`
+type PayloadSigner struct {
+	PayloadSigner         signature.Signer
+	PayloadSignerOpts     []signature.SignOption
+	PublicKeyProviderOpts []signature.PublicKeyOption
+}
+
+// Sign uses the PayloadSigner to sign the requested payload, then returns the signature, the public key associated with it, the signed payload
```
mattmoor, line 41: I'd move this onto `PayloadSigner` since most folks will see the interface comment via things like IDEs when looking at callsites to this method.

For interface impls, I always use: `// {function} implements {interface}`

internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/sigstore/cosign/pkg/oci/static"
+	"github.com/sigstore/sigstore/pkg/signature"
+	signatureoptions "github.com/sigstore/sigstore/pkg/signature/options"
+)
+
+type Signer interface {
+	Sign(ctx context.Context, payload io.Reader) (oci.Signature, crypto.PublicKey, error)
+}
+
+// PayloadSigner implements `Signer`
```
mattmoor, line 34: You can actually assert this (without allocation) using:
```go
var _ Signer = (*PayloadSigner)(nil)
```

I generally put this below struct definitions that I want implementing a particular interface.

internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,68 @@
...
+	"encoding/base64"
+	"io"
+
+	"github.com/sigstore/cosign/pkg/oci"
+	"github.com/sigstore/cosign/pkg/oci/static"
+	"github.com/sigstore/sigstore/pkg/signature"
+	signatureoptions "github.com/sigstore/sigstore/pkg/signature/options"
+)
+
+type Signer interface {
```
mattmoor, line 30: This is where you really want good comments documenting the expectations.

internal/pkg/cosign/rekor.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// RekorSignerWrapper implements `Signer`
+type RekorSignerWrapper struct {
+	Inner Signer
+
+	RekorURL string
+}
+
+// Sign calls a wrapped, inner signer then uploads either the Cert or Pub(licKey) of the results to Rekor, then adds the resulting `Bundle`
```
mattmoor, line 71: Same comment as below

mattmoor (commented):
I like the interface a lot more.  Left a few comments inline, but at a high level:
1. Can we shuffle the package structure around so that implementations generally live in their own packages with private types and public ctors in terms of the interface?
2. As I was building up `oci` I was trying to ensure fairly high coverage as I went, and it should now be fairly high.  Can we try to do the same here?

internal/pkg/cosign/sign.go
dekkagaijin: roger. The only thing not captured by `oci.Signature` was the signature's public key

internal/pkg/cosign/sign.go
```diff
@@ -0,0 +1,68 @@
...
+	Sign(ctx context.Context, payload io.Reader) (oci.Signature, crypto.PublicKey, error)
```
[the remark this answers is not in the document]
dekkagaijin: nope ¯\_(ツ)_/¯

It's not stored in the registry, so that makes sense.

internal/pkg/cosign/sign.go
dekkagaijin: ACK

internal/pkg/cosign/sign.go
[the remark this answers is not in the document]
mattmoor: What are callers consuming it to do today?

internal/pkg/cosign/sign.go
dekkagaijin: > I like the interface a lot more. Left a few comments inline, but at a high level:
> 
> 1. Can we shuffle the package structure around so that implementations generally live in their own packages with private types and public ctors in terms of the interface?

yed

> 2. As I was building up `oci` I was trying to ensure fairly high coverage as I went, and it should now be fairly high.  Can we try to do the same here?

Will do, but I'd like to have the package structure stable before I go implementing all the accoutrements for testing

internal/pkg/cosign/fulcio/fulcio.go
```diff
@@ -12,30 +12,31 @@
...
 	"github.com/sigstore/cosign/pkg/oci"
 	"github.com/sigstore/cosign/pkg/oci/static"
 )
 
-// FulcioSignerWrapper still needs to actually upload keys to Fulcio and receive
+// SignerWrapper still needs to actually upload keys to Fulcio and receive
 // the resulting `Cert` and `Chain`, which are added to the returned `oci.Signature`
-type FulcioSignerWrapper struct {
-	Inner Signer
+type SignerWrapper struct {
```
mattmoor, line 29: This should be private now.

internal/pkg/cosign/fulcio/fulcio.go
```diff
@@ -77,3 +78,12 @@ func (fs *FulcioSignerWrapper) Sign(ctx context.Context, payload io.Reader) (oci
 
 	return newSig, pub, nil
 }
+
+// NewSigner returns a *SignerWrapper which signs and uploads the given payload to Fulcio.
+func NewSigner(inner cosign.Signer, cert, chain []byte) *SignerWrapper {
```
mattmoor, line 83: return the interface

internal/pkg/cosign/fulcio/fulcio.go
```diff
@@ -77,3 +78,12 @@ func (fs *FulcioSignerWrapper) Sign(ctx context.Context, payload io.Reader) (oci
 
 	return newSig, pub, nil
 }
+
+// NewSigner returns a *SignerWrapper which signs and uploads the given payload to Fulcio.
```
mattmoor, line 82: This comment could use work :)

internal/pkg/cosign/fulcio/fulcio.go
```diff
@@ -12,30 +12,31 @@
...
 
-var _ Signer = (*FulcioSignerWrapper)(nil)
+var _ cosign.Signer = (*SignerWrapper)(nil)
 
-// Sign implements `Signer`
-func (fs *FulcioSignerWrapper) Sign(ctx context.Context, payload io.Reader) (oci.Signature, crypto.PublicKey, error) {
-	sig, pub, err := fs.Inner.Sign(ctx, payload)
+// Sign implements `cosign.Signer`
+func (fs *SignerWrapper) Sign(ctx context.Context, payload io.Reader) (oci.Signature, crypto.PublicKey, error) {
+	sig, pub, err := fs.inner.Sign(ctx, payload)
```
mattmoor, line 39: I don't really grok why this is wrapping *any* Signer instead of having a specific Signer implementation embedded here.  Can this truly be composed with other Signer implementations arbitrarily?  What other compositions make sense?

cmd/cosign/cli/sign/sign.go
```diff
@@ -212,16 +214,9 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
 	s = &icos.PayloadSigner{
```
mattmoor, line 214: I'd move the payload signer as well.

internal/pkg/cosign/fulcio/fulcio.go
dekkagaijin: not quite arbitrarily but we made the `Signer` abstraction for a reason. I think noting the requirements for our public instance in comments would be a legitimate way of documenting them (vs. using the type system)

internal/pkg/cosign/fulcio/fulcio.go
dekkagaijin: done

internal/pkg/cosign/fulcio/fulcio.go
dekkagaijin: done

internal/pkg/cosign/fulcio/fulcio.go
dekkagaijin: your FACE could use work

cmd/cosign/cli/sign/sign.go
dekkagaijin: roger

internal/pkg/cosign/sign.go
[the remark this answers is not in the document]
mattmoor: bump this comment?

cmd/cosign/cli/sign/sign.go
```diff
@@ -244,36 +210,33 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko KeyO
...
 	}
-	if _, err := out.Write([]byte(b64sig)); err != nil {
-		return errors.Wrap(err, "write signature to file")
-	}
 
-	if !upload {
-		return nil
+	var s icos.Signer
+	s = ipayload.NewSigner(sv, nil, nil)
+	s = ifulcio.NewSigner(s, sv.Cert, sv.Chain)
```
mattmoor, line 216: A few things about this line don't sit right with me:
1. Shouldn't it be predicated on some condition?
2. I still don't like the fact that the fulcio Signer wraps another random Signer where we have zero visibility into what it does.  I think the reality here is that the payload Signer is doing the fulcio signing, and the fulcio Signer is just decorating the resulting oci.Signature to match.  I think this is a fragile composition, and I think I'd rather see payload's Signer be a hidden implementation detail of a fulcio Signer than allow arbitrary composition here.

dekkagaijin: yes :p

"TODO:"
"* move over fulcio operations as well"

mattmoor (approved):
Let's land this and keep going :)

PR #1137, opened 2021-12-05
Add support for syft json type to cosign

### LINKED ISSUE #1136: Add support for syft SBOMs to cosign
**Description**

syft is a popular SBOM scanner/creator. It has it's own SBOM format that is used with Grype which is an open source vuln scanner. We should also support syft as a format apart from cyclonedx/spdx.

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

#### Summary

Add support for syft format in cosign attach sbom

#### Ticket Link

Fixes #1136 

#### Release Note

```release-note
Add support for syft format in cosign attach sbom
```

### DIFF
--- a/pkg/types/media.go
+++ b/pkg/types/media.go
@@ -17,7 +17,7 @@ package types
 
 const (
 	CycloneDXMediaType     = "application/vnd.cyclonedx"
-	SyftMediaType          = "application/vnd.syft"
+	SyftMediaType          = "application/vnd.syft+json"
 	SimpleSigningMediaType = "application/vnd.dev.cosign.simplesigning.v1+json"
 	SPDXMediaType          = "text/spdx"
 	WasmLayerMediaType     = "application/vnd.wasm.content.layer.v1+wasm"
--- a/pkg/types/media.go
+++ b/pkg/types/media.go
@@ -17,6 +17,7 @@ package types
 
 const (
 	CycloneDXMediaType     = "application/vnd.cyclonedx"
+	SyftMediaType          = "application/vnd.syft"
 	SimpleSigningMediaType = "application/vnd.dev.cosign.simplesigning.v1+json"
 	SPDXMediaType          = "text/spdx"
 	WasmLayerMediaType     = "application/vnd.wasm.content.layer.v1+wasm"
--- a/cmd/cosign/cli/options/attach.go
+++ b/cmd/cosign/cli/options/attach.go
@@ -61,7 +61,7 @@ func (o *AttachSBOMOptions) AddFlags(cmd *cobra.Command) {
 		"path to the sbom, or {-} for stdin")
 
 	cmd.Flags().StringVar(&o.SBOMType, "type", "spdx",
-		"type of sbom (spdx|cyclonedx)")
+		"type of sbom (spdx|cyclonedx|syft)")
 }
 
 func (o *AttachSBOMOptions) MediaType() (types.MediaType, error) {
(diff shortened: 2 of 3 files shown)


developer-guy:
wow, thanks @samj1912, this is amazing! I do even know Syft but did not realize that they use their own SBOM format.

PTAL @luhring

samj1912:
Not sure why the cosigned tests are failing but it looks unrelated to the PR

developer-guy:
> Not sure why the cosigned tests are failing but it looks unrelated to the PR

yes, seems something is broken, this is not related to your PR 🤝

dlorenc:
Looks great! Let's see if we can chase down the ci failure.

dlorenc:
Looks like the distroless image is not actually signed: https://github.com/GoogleContainerTools/distroless/issues/905

pkg/types/media.go
```diff
@@ -17,6 +17,7 @@ package types
 
 const (
 	CycloneDXMediaType     = "application/vnd.cyclonedx"
+	SyftMediaType          = "application/vnd.syft"
```
dlorenc, line 20: Actually for this we might need to check with @luhring - should this be something like `application/vnd.com.anchore.syft`?

luhring: Good question — we just recently registered `application/vnd.syft+json` with the IANA (https://www.iana.org/assignments/media-types/application/vnd.syft+json)

dlorenc: Perfect! @samj1912 could you add the +json part and then we should be good to merge!

samj1912: Sure, I didn't add it since we don't add the file extensions suffixes for cyclonedx or spdx either (spdx recently added application/spdx+json for their json documents). I was hoping to make a separate PR to add file extension suffixes for all of them. But can do it for syft specifically in this PR.

samj1912: Related https://github.com/spdx/spdx-spec/issues/577 and https://github.com/anchore/syft/issues/612

dlorenc: Ah ok. I'm not an expert here. There was actually some discussion about whether or not we were supposed to add it or not for the others and I can't remember the explanation.

samj1912: @dlorenc done - should be good to go. Will also try to update the others in another PR.

luhring: >There was actually some discussion about whether or not we were supposed to add it or not for the others

If we do find this, I'd be interested in learning more! I'm also not an expert on this.

From my understanding, it's important to have the `+json` part. This is because without it 1) it's no longer a registered media type, and 2) content interpretation becomes ambiguous to consumers of the data.

And I'd guess this applies equally to the already listed media types (e.g. `application/vnd.cyclonedx`) just as well, so I like your idea of a follow-up PR to address this, @samj1912!

dlorenc: Found it! It's long and confusing: https://github.com/sigstore/cosign/pull/439#discussion_r671834751

My understanding is that the suffixes are probably correct to add, but only if they were actually registered? Unless you're in the vendor sub-tree, where it might be ok, depending on which subtype you're using.

luhring: That's a great tl;dr. Thanks for the link!

Is this good to merge at this point?

dlorenc: Yep - sorry. There was a test flake I had to rerun.

PR #1147, opened 2021-12-06
Fix bug when saving and loading a manifest list

priyawadhwa:
Save and load were assuming we were only saving and loading SignedImages. This PR should add support for saving and loading SignedImageIndex's as well!

ref https://github.com/sigstore/cosign/issues/1015

```release-note
fix: saving and loading a signed image index to disk
```

cc @haydentherapper who is working on verification for this!

### DIFF
--- a/pkg/oci/layout/index.go
+++ b/pkg/oci/layout/index.go
@@ -25,10 +25,11 @@ import (
 )
 
 const (
-	kindAnnotation  = "kind"
-	imageAnnotation = "dev.cosignproject.cosign/image"
-	sigsAnnotation  = "dev.cosignproject.cosign/sigs"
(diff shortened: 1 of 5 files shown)


dekkagaijin (approved):

pkg/oci/layout/index.go
```diff
@@ -98,6 +99,9 @@ func (i *index) SignedImage(h v1.Hash) (oci.SignedImage, error) {
 	if err != nil {
 		return nil, err
 	}
+	if img == nil {
+		return nil, nil
+	}
```
dekkagaijin, line 104:
```suggestion
	if  img == nil || err != nil {
		return nil, err
	}
```

PR #1157, opened 2021-12-07
return error when rekor pub cannot be retrieved, fix file path construction

dekkagaijin:
This (probably) doesn't fix the error being seen by @cpanato during verification on windows, but it does surface the error and take care of some potential root causes

```release-note
NONE
```

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -43,14 +43,14 @@ import (
 // This is the rekor public key target name
 var rekorTargetStr = `rekor.pub`
 
-func GetRekorPub(ctx context.Context) string {
+// GetRekorPub retrieves the rekor public key from the embedded or cached TUF root. If expired, makes a
+// network call to retrieve the updated target.
+func GetRekorPub(ctx context.Context) ([]byte, error) {
 	buf := tuf.ByteDestination{Buffer: &bytes.Buffer{}}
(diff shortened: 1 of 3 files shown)


mattmoor (approved):
Any way we can get Windows testing with actions?

pkg/cosign/tlog.go
```diff
@@ -43,14 +43,14 @@ import (
...
-func GetRekorPub(ctx context.Context) string {
+// GetRekorPub retrieves the rekor public key from the embedded or cached TUF root. If expired, makes a
+// network call to retrieve the updated target.
+func GetRekorPub(ctx context.Context) ([]byte, error) {
 	buf := tuf.ByteDestination{Buffer: &bytes.Buffer{}}
-	// Retrieves the rekor public key from the embedded or cached TUF root. If expired, makes a
-	// network call to retrieve the updated target.
 	if err := tuf.GetTarget(ctx, rekorTargetStr, &buf); err != nil {
-		panic("error retrieving rekor public key")
+		return nil, err
```
asraa, line 51: this might be a local cache being old/invalid -- maybe we can say "please remove local cache" (this should be a const str in the tuf code DefaultCache or something)

dekkagaijin: PRs welcome :D

If there's a place where we can definitively say "this local cache is old/invalid and we can't fix it," lets do it

Issue #1153, opened 2021-12-07
Cosign crashes when password input is requested from terminal in Windows


gjirm:
On Windows when password should be read from terminal, application immediately crashes with the following error:

```
Enter password for private key:
Error: The handle is invalid.
main.go:46: error during command execution: The handle is invalid.
```
For example when running these commands:

```
cosign generate-key-pair
cosign sign --key cosign.key ...
```

I think it is related to the bug #862 I previously reported.

### My environment

- Windows 11 (21H2 - build 22000.318), Windows 10 (21H2 - build 19044.1348)
- cosign v1.4.0
- Window Terminal  v1.11.2921.0
  - native cmd.exe
  - PowerShell 7.2.0

dekkagaijin:
@gjirm I wasn't able to reproduce the issue. Would you be able to add a test case to https://github.com/sigstore/cosign/blob/main/test/e2e_test.ps1 that provokes the bug?

gjirm:
> @gjirm I wasn't able to reproduce the issue. Would you be able to add a test case to https://github.com/sigstore/cosign/blob/main/test/e2e_test.ps1 that provokes the bug?

@dekkagaijin Unfortunately, I don't know how to add a test case, because it only happens to me if I run the command without entering a password (via a environment variable or via a pipe). See example below for `cosign generate-key-pair`:

![cosign_bug_1153](https://user-images.githubusercontent.com/32864045/145793598-b490904e-69eb-41f8-9727-b858dc88401a.gif)

jowko:
I also have similar issue. I use Windows 10 Pro. I downloaded v1.5.1 version of cosign.
When I run `cosign generate-key-pair` in PowerShell/cmd i get such thing:
```
Enter password for private key: Error: The handle is invalid.
main.go:46: error during command execution: The handle is invalid.
```
When I run the same command in git bash, then bash hangs and I need to kill this command:
{code}
$ cosign generate-key-pair -d

{code}
Providing password via environment variable resolves this issue.

PR #1158, opened 2021-12-07
expand CI testing to Windows and OSX, fix issues uncovered

dekkagaijin:
I expanded our Windows CI test coverage with the goal of root causing https://github.com/sigstore/cosign/issues/1153. Still haven't done that, yet, but I figure I should get this in

Mitigates https://github.com/sigstore/cosign/issues/1160 -related failures in testing

```release-note
bugfix: `cosign` Windows releases panicking under certain circumstances
```

### DIFF
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -55,7 +55,7 @@ var rootClient *client.Client
 var rootClientMu = &sync.Mutex{}
 
 func GetEmbeddedRoot() ([]byte, error) {
-	return root.ReadFile(filepath.Join("repository", "root.json"))
+	return root.ReadFile("repository/root.json")
 }
 
 func CosignCachedRoot() string {
@@ -133,7 +133,7 @@ func RootClient(ctx context.Context, remote client.RemoteStore, altRoot []byte) 
 		if os.IsNotExist(err) && altRoot == nil {
 			// Cache does not exist, check if the embedded metadata is currently valid.
 			// TODO(asraa): Need a better way to check if local metadata is verified at this stage.
-			timestamp, err := root.ReadFile(filepath.Join("repository", "timestamp.json"))
+			timestamp, err := root.ReadFile("repository/timestamp.json")
 			if err != nil {
 				return nil, errors.Wrap(err, "reading local timestamp")
 			}
@@ -142,12 +142,12 @@ func RootClient(ctx context.Context, remote client.RemoteStore, altRoot []byte) 
 				if err := local.SetMeta("timestamp.json", timestamp); err != nil {
 					return nil, errors.Wrap(err, "setting local meta")
 				}
-				for _, metadata := range []string{"root.json", "targets.json", "snapshot.json"} {
-					msg, err := root.ReadFile(filepath.Join("repository", metadata))
+				for _, mdFilename := range []string{"root.json", "targets.json", "snapshot.json"} {
+					msg, err := root.ReadFile("repository/" + mdFilename)
 					if err != nil {
 						return nil, errors.Wrap(err, "reading local root")
 					}
-					if err := local.SetMeta(metadata, msg); err != nil {
+					if err := local.SetMeta(mdFilename, msg); err != nil {
 						return nil, errors.Wrap(err, "setting local meta")
 					}
 				}
@@ -185,7 +185,7 @@ func RootClient(ctx context.Context, remote client.RemoteStore, altRoot []byte) 
 				if altRoot != nil {
 					trustedRoot = altRoot
 				} else {
-					trustedRoot, err = root.ReadFile(filepath.Join("repository", "root.json"))
+					trustedRoot, err = root.ReadFile("repository/root.json")
 					if err != nil {
 						return nil, errors.Wrap(err, "reading embedded trusted root")
 					}
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/cosign/tuf/client_test.go
```diff
@@ -127,10 +128,14 @@ func TestValidMetadata(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error")
 	}
+	if cl, isCloser := local.(io.Closer); isCloser {
+		// TODO: this is a hack to free the file descriptors, need to patch `tuf_leveldbstore.FileLocalStore` to return a io.Closer
```
mattmoor, line 132: Open an issue to track?

pkg/cosign/tuf/client.go
```diff
@@ -247,7 +258,7 @@ func GetTarget(ctx context.Context, name string, out client.Destination) error {
 
 	// Retrieves the target and writes to out. This may make a network call and cache if
 	// the embedded or cached root is invalid (e.g. expired).
-	return getTargetHelper(name, out, c)
+	return getTargetHelper(name, out, c, false)
```
mattmoor, line 261: I don't understand why this path doesn't require coherence, is there someone else we can tag in?

cc @dlorenc

pkg/cosign/tuf/client.go
```diff
@@ -97,11 +95,11 @@ func (b *ByteDestination) Delete() error {
...
 // Retrieves a local target, either from the cached root or the embedded metadata.
 func getLocalTarget(name string) (fs.File, error) {
-	if _, err := os.Stat(CosignCachedTargets()); !os.IsNotExist(err) {
+	if f, err := os.Open(filepath.Join(CosignCachedTargets(), name)); err == nil {
 		// Return local cached target
-		return os.Open(path.Join(CosignCachedTargets(), name))
+		return f, nil
 	}
-	return root.Open(path.Join("repository", "targets", name))
+	return root.Open("repository/targets/" + name)
```
mattmoor, line 102: why doesn't this need `filepath.Join`? 🤔

pkg/cosign/tuf/client.go
```diff
@@ -57,7 +55,7 @@ var rootClient *client.Client
 var rootClientMu = &sync.Mutex{}
 
 func GetEmbeddedRoot() ([]byte, error) {
-	return root.ReadFile(filepath.Join("repository", "root.json"))
+	return root.ReadFile("repository/root.json")
```
mattmoor, line 58: Why not `filepathJoin`?

pkg/cosign/tuf/client.go
```diff
@@ -44,8 +43,7 @@ const (
 	DefaultRemoteRoot = "sigstore-tuf-root"
 )
 
-//go:embed repository/*.json
-//go:embed repository/targets/*.pem repository/targets/*.pub
```
mattmoor, line 48: What happened to these?

pkg/cosign/tuf/client.go
dekkagaijin: the `embed.FS` it creates uses unix-like path separators, even on Windows /shrug

pkg/cosign/tuf/client.go
dekkagaijin: that'd be an option, but behavior depends on whether or not we're reading from the embedded FS or real FS, so it'd make more sense to create an interface

pkg/cosign/tuf/client.go
dekkagaijin: embedding a directory copies everything recursively, except for files beginning with `.` or `_`. We ended up missing a bunch of embedded files due to a misunderstanding about the behavior of `*` in the directive

pkg/cosign/tuf/client.go
mattmoor: Woof, can you add comments to that effect?

pkg/cosign/tuf/client.go
dekkagaijin: This helper is doing a bit too much, IMHO. 

In general we shouldn't be re-syncing with the remote TUF repo unless absolutely necessarily, and certainly not on every single read of the local cache

pkg/cosign/tuf/client_test.go
dekkagaijin: https://github.com/sigstore/cosign/issues/1160

pkg/cosign/tuf/client.go
dekkagaijin: done (at the helper fn)

pkg/cosign/tuf/client.go
dekkagaijin: created private global functions, PTAL

pkg/cosign/tuf/client_test.go
dekkagaijin: also added link in a commit

cpanato (commented):
lgtm

pkg/cosign/tuf/client.go
dlorenc: I'd grab @asraa

mattmoor: I'd make this `false /* requiresCoherence */` and add a comment explaining why this path doesn't require it.

pkg/cosign/tuf/client.go
```diff
@@ -125,91 +132,101 @@ func isExpiredMetadata(metadata []byte) bool {
...
-		path := filepath.Join(CosignCachedRoot(), "tuf.db")
-		_, err := os.Open(path)
-		if os.IsNotExist(err) && altRoot == nil {
-			// Cache does not exist, check if the embedded metadata is currently valid.
-			// TODO(asraa): Need a better way to check if local metadata is verified at this stage.
-			timestamp, err := root.ReadFile(filepath.Join("repository", "timestamp.json"))
-			if err != nil {
-				return nil, errors.Wrap(err, "reading local timestamp")
+
+	if rootClient != nil {
```
mattmoor: The whitespace changes from flipping this (while I agree!) make this unnecessarily hard to review.  I'd flip this back and make that change separately (and IMO go straight to `sync.Once`)

dekkagaijin: done

mattmoor (approved):

pkg/cosign/tuf/client.go
asraa: I'm doing some clean-up here, just FYI. We should be requiring coherence between the target we have the metadata we have locally. That's what this does here. We can disable the remote check, but we should issue a warning that the current `targets.json` and the `target` you have don't line up (this probably only happens if you have a stale cache, or something else wonky happened).

PR #1159, opened 2021-12-07
Add option to verify local image

haydentherapper:
This builds on the previous work to support saving an image and its signature for airgapped environments. This adds a flag, `--local-image`, to verify a signed image from disk.

Example:

```
cosign save us-west1-docker.pkg.dev/project/docker-repo/image:tag1 --dir .
cosign verify --key cosign.pub --local-image . --verbose
```

With `--verbose`, we see no network calls.

Ref #1015

#### Release Note
```release-note
Add `cosign verify --local-image path` for verifying signed images from disk
```

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -58,6 +58,7 @@ type VerifyCommand struct {
 	Annotations   sigs.AnnotationsMap
 	SignatureRef  string
 	HashAlgorithm crypto.Hash
+	LocalImage    bool
 }
 
 // Exec runs the verification command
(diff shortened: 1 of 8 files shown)


haydentherapper:
cc @priyawadhwa

haydentherapper:
cc @dlorenc - FYI, this should allow for offline verification too

dlorenc:
Amazing!!!

cc @jonjohnsonjr

cmd/cosign/cli/verify/verify.go
```diff
@@ -138,22 +139,31 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-		ref, err := name.ParseReference(img)
-		if err != nil {
-			return errors.Wrap(err, "parsing reference")
-		}
-		ref, err = sign.GetAttachedImageRef(ref, c.Attachment, ociremoteOpts...)
-		if err != nil {
-			return errors.Wrapf(err, "resolving attachment type %s for image %s", c.Attachment, img)
-		}
+		if c.LocalImage {
+			verified, bundleVerified, err := cosign.VerifyLocalImageSignatures(ctx, img, co)
```
jonjohnsonjr, line 143: A single directory can contain multiple images -- I'd consider making `LocalImage` a string that points to the path instead of a bool and use `img` to address the image within that path.

The UX here gets a little funky -- for a directory with a single image, you probably want to just default to it and not require users to pass it as an arg as well.

@priyawadhwa might have some ideas here.

priyawadhwa: i think we [overwrite](https://github.com/sigstore/cosign/blob/main/pkg/oci/layout/write.go#L29) the directory every time we run `cosign save` so only a single image should be allowed at the moment -- eventually if we add support for multiple images in a directory then we'll need to figure out what the UX would look like

test/e2e_test.go
```diff
@@ -81,13 +81,30 @@ var verify = func(keyRef, imageRef string, checkClaims bool, annotations map[str
 		Annotations:   sigs.AnnotationsMap{Annotations: annotations},
 		Attachment:    attachment,
 		HashAlgorithm: crypto.SHA256,
+		// LocalImage: true,
```
priyawadhwa, line 84: nit: leftover debugging?

priyawadhwa (commented):
nice, this looks great!

test/e2e_test.go
haydentherapper: Removed

cmd/cosign/cli/verify/verify.go
haydentherapper: Thanks for pointing this out! I agree that since we don't support persisting multiple images in a single directory, this shouldn't be an issue for now. 

For the future, I like the approach of specifying the image along with the path by flag.

PR #1168, opened 2021-12-08
feat: vuln attest support

### LINKED ISSUE #442: Determine attestation format for vuln scans
**Description**

From a slack conversation with @developer-guy @Dentrax and @joshuagl, capturing it here so I don't forget!

We've already discussed support for signing/verifying/storing `attestations`, using the In-Toto model: https://github.com/in-toto/attestation, and @developer-guy brought up a common use case around storing vulnerability scan reports for an image in the registry.

Storing scans is slightly problematic in that the scan results are *timely*, a scan with no vulnerabilities a month ago doesn't mean that the image has no know vulnerabilities today. Rather than storing scan results directly, we can convert them into an `attestation`, by adding a timestamp and a signer.

Rather than saying "this has no vulnerabilties", the statement becomes "this had no vulnerabilities as of $TIME, as seen by $SIGNER". Then the $SIGNER signs this, making it a formal attestation.

Policy engines can then rely on these, and policies can be written in the form of "there must be a clean scan report within the last X days".

I think to implement all of this in cosign, we would do something like:

* define a new [provenance predicate](https://github.com/in-toto/attestation/tree/main/spec) either here in cosign or in the in-toto repo to convey a timestamped scan report.
* add some cosign commands:

`cosign sign-attestation -f <attestation>`
`cosign attach attestation -f <attestation> `

We can still use all of the KMS/key/ephemeral support for signing - but these are slightly different from the rest of what cosign works with, as the attestation itself is signed, not the OCI image wrapper. The scan report would be generated and timestamped, then cosign would sign the file locally (similar to sign-blob). Then the signed blob can get attached to an image.

Does this all make sense?

developer-guy:
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>

#### Summary

Users might upload vulnerability scan results in form of attestation within the OCI registry

#### Ticket Link

Fixes #442 

#### Release Note

```release-note
feat: vuln attest support
```

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -32,12 +32,46 @@ import (
 const (
 	// CosignCustomProvenanceV01 specifies the type of the Predicate.
 	CosignCustomProvenanceV01 = "cosign.sigstore.dev/attestation/v1"
+
+	// CosignVulnProvenanceV01 specifies the type of VulnerabilityScan Predicate
+	CosignVulnProvenanceV01 = "cosign.sigstore.dev/attestation/vuln/v1"
 )
 
 // CosignPredicate specifies the format of the Custom Predicate.
 type CosignPredicate struct {
 	Data      interface{}
 	Timestamp string
+}
+
+// VulnPredicate specifies the format of the Vulnerability Scan Predicate
+type CosignVulnPredicate struct {
+	Invocation Invocation `json:"invocation"`
+	Scanners   []Scanner  `json:"scanners"`
+	Metadata   Metadata   `json:"metadata"`
+}
+
+type Invocation struct {
+	Parameters interface{} `json:"parameters"`
+	URI        string      `json:"uri"`
+	EventID    string      `json:"event_id"`
+	BuilderID  string      `json:"builder.id"`
+}
+
+type DB struct {
+	URI     string `json:"uri"`
+	Version string `json:"version"`
+}
+
+type Scanner struct {
+	URI     string                 `json:"uri"`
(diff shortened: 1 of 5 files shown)


dlorenc:
Looks good, the lint errors should be easy!

pkg/cosign/attestation/attestation.go
```diff
@@ -40,6 +43,37 @@ type CosignPredicate struct {
...
 }
 
+// VulnPredicate specifies the format of the Vulnerability Scan Predicate
+type CosignVulnPredicate struct {
+	Invocation Invocation `json:"invocation"`
+	Scanners   []Scanner  `json:"scanners"`
+	Metadata   Metadata   `json:"metadata"`
+}
+
+type Invocation struct {
```
dlorenc, line 53: We should probably also add a README in here under specs/ that describes all of these fields in detail and how they should be used.

developer-guy: Exactly! This is the next that we are going to do🤝

developer-guy:
hey @dlorenc, I know spec documentation is not perfect but we tried to do our best, can you please review it, thanks in advance 🤝

developer-guy:
kindly ping @dlorenc

dlorenc:
> kindly ping @dlorenc

Sorry missed this because of the WIP in the title. Is it ready for a merge?

developer-guy:
I think yes, but it'd be nice if you can review it one more time I guess.

dlorenc (approved):

dlorenc:
LGTM!

PR #1170, opened 2021-12-08
fix deprecation warning from setup-gcloud

dekkagaijin:
Fixes warnings e.g. https://github.com/sigstore/cosign/actions/runs/1555909393

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -32,11 +32,14 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.17.x'
+      - id: auth
+        uses: google-github-actions/auth@v0.4.2
+        with:
+          credentials_json: ${{ secrets.GCP_CI_SERVICE_ACCOUNT }}
       - name: Set up Cloud SDK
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e_tests.yml
```diff
@@ -32,11 +32,14 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.17.x'
+      - id: auth
+        uses: google-github-actions/auth@v0.4.2
+        with:
+          credentials_json: ${{ secrets.GCP_CI_SERVICE_ACCOUNT }}
```
mattmoor, line 38: Can we file an issue to replace this with token federation?

mattmoor (approved):

.github/workflows/e2e_tests.yml
dekkagaijin: yes, but that requires additional setup. I can add an issue

dekkagaijin:
sigh... looks like the setup action dropped support for base64 input

bobcallaway:
I think this is fixed via workload identity in #1330

PR #1174, opened 2021-12-08
Add option to verify attestations from local image

haydentherapper:
This is very similar to a previous PR (#1159) that added support for verifying signatures from a local image.

Ref #1015 

Example:

```
...generate attestation...
cosign save us-west1-docker.pkg.dev/project/docker-repo/image:tag1 --dir .
cosign verify-attestation --key cosign.pub --local-image . --verbose
```

With `--verbose`, we see no network calls.

#### Release Note

```release-note
Add `cosign verify-attestation --local-image path` for verifying signed images with attestations from disk
```

### DIFF
--- a/cmd/cosign/cli/options/verify.go
+++ b/cmd/cosign/cli/options/verify.go
@@ -70,7 +70,7 @@ func (o *VerifyOptions) AddFlags(cmd *cobra.Command) {
 		"signature content or path or remote URL")
 
 	cmd.Flags().BoolVar(&o.LocalImage, "local-image", false,
-		"whether the path to the image is a local directory")
+		"whether the specified image is a path to an image saved locally via 'cosign save'")
 }
 
 // VerifyAttestationOptions is the top level wrapper for the `verify attestation` command.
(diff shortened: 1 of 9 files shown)


haydentherapper:
cc @priyawadhwa

cmd/cosign/cli/options/verify.go
```diff
@@ -108,6 +109,9 @@ func (o *VerifyAttestationOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringVarP(&o.Output, "output", "o", "json",
 		"output format for the signing image information (json|text)")
+
+	cmd.Flags().BoolVar(&o.LocalImage, "local-image", false,
+		"whether the path to the image is a local directory")
```
priyawadhwa, line 114: wdyt of clarifying that the image has to be saved with cosign, something like

```
"path to an image saved locally via `cosign save`"
```

priyawadhwa (approved):
just one nit, lgtm!

cmd/cosign/cli/options/verify.go
priyawadhwa: this would match the description below as well! we might want to update the description for the flag in `cosign verify` as well

haydentherapper: Thanks, updated!

Issue #1179, opened 2021-12-09
"cosign initialize" doesn't work


rgerganov:
I am trying to run `cosign` with a local TUF root. My understanding is that I should use the `initialize` command like this:
```
cosign initialize --root /local/tuf/repo/root.json
```
However this fails with:
```
Error: initializing root client: updating from remote TUF repository: updating tuf metadata: tuf: failed to decode timestamp.json: expired at 2021-12-02 15:41:56 +0000 UTC
main.go:46: error during command execution: initializing root client: updating from remote TUF repository: updating tuf metadata: tuf: failed to decode timestamp.json: expired at 2021-12-02 15:41:56 +0000 UTC
```
Running without `--root`:
```
cosign initialize    
Error: updating local metadata and targets: updating tuf metadata: tuf: failed to decode timestamp.json: expired at 2021-12-02 15:41:56 +0000 UTC
main.go:46: error during command execution: updating local metadata and targets: updating tuf metadata: tuf: failed to decode timestamp.json: expired at 2021-12-02 15:41:56 +0000 UTC
```

It appears it always fetch stuff from GCS which has an expired timestamp. I know there is a `--mirror` flag which can override the default GCS location but I don't want to put my stuff of GCS.

priyawadhwa:
cc @asraa might know more about this?

asraa:
Fixed now -- cosigns updated with the latest root, but GCS isn't synced. I'll make a fix PR to auto-update it and tag you

rgerganov:
But why do we need GCS in the first place? Why it's not possible to use local TUF root?

asraa:
> But why do we need GCS in the first place? Why it's not possible to use local TUF root?

`cosign initialize` is a manual fetch. You don't have to run it at all, you use the local TUF root during a cosign verification. `cosign initialize` is just our system of allowing users to manually updating/downloading in case they want to verify with a remote.

rgerganov:
I am not sure I understand, so I will describe what I am trying to achieve.
I have a local instance of Fulcio and CT log. When I experiment with keyless signing, I am getting an error which says that SCT signature received from Fulcio cannot be verified. This is because cosign is using `ctfe.pub` from its embedded TUF repo. I want to change the TUF repo and provide my own `ctfe.pub`. I thought the way to do this is with `cosign initialize`. Is this correct?

asraa:
Ah, sorry, didn't realize you wanted your entire own root!

[...]

PR #1185, opened 2021-12-10
Allow fetching TUF root from HTTP

rgerganov:
This patch allows fetching TUF root from HTTP in addition to GCS. If the
specified mirror is an HTTP link, then HTTPRemoteStore will be used.
Otherwise it defaults to GCS.

### DIFF
--- a/cmd/cosign/cli/initialize/init.go
+++ b/cmd/cosign/cli/initialize/init.go
@@ -18,9 +18,11 @@ package initialize
 import (
 	"context"
 	_ "embed" // To enable the `go:embed` directive.
+	"net/url"
 
 	"github.com/sigstore/cosign/pkg/blob"
 	"github.com/sigstore/cosign/pkg/cosign/tuf"
+	"github.com/theupdateframework/go-tuf/client"
 )
 
 func DoInitialize(ctx context.Context, root, mirror string, threshold int) error {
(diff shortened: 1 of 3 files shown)


hectorj2f (commented):
@rad Could you add a test ? So we can validate both use cases.

hectorj2f:
This PR is related to: https://github.com/sigstore/cosign/issues/1179

hectorj2f (commented):
@rgerganov You are also missing to update the command description to be able to use a different mirror in `cmd/cosign/cli/initialize.go`:
`SigStore remote TUF repository is pulled from the GCS mirror at sigstore-tuf-root.`

cmd/cosign/cli/initialize/init.go
```diff
@@ -35,7 +37,12 @@ func DoInitialize(ctx context.Context, root, mirror string, threshold int) error
 	}
 
 	// Initialize the remote repository.
-	remote, err := tuf.GcsRemoteStore(ctx, mirror, nil, nil)
+	var remote client.RemoteStore
+	if strings.HasPrefix(mirror, "http://") || strings.HasPrefix(mirror, "https://") {
```
hectorj2f, line 41: Idea: Perhaps you could use  `u, err := url.Parse(mirror)` to validate the mirror points to a valid URL, or simply check whether the mirror is `sigstore-tuf-root`. 

cc @asraa

rgerganov:
> @rgerganov You are also missing to update the command description to be able to use a different mirror in `cmd/cosign/cli/initialize.go`: `SigStore remote TUF repository is pulled from the GCS mirror at sigstore-tuf-root.`

This description is about the default behavior which is not changed (i.e. pulling from GCS)

cmd/cosign/cli/initialize/init.go
asraa: +1 preference to parse the URL to check if it's valid.

I would caution against hardcoding a check on whether it's `sigstore-tuf-root` though.

asraa (commented):
Thank you!

cmd/cosign/cli/initialize/init.go
rgerganov: Done

asraa (approved):

PR #1189, opened 2021-12-10
create KeylessSigner

dekkagaijin:
#### Ticket Link
#844

#### Release Note
```release-note
NONE
```

### DIFF
--- a/internal/pkg/cosign/keyless/signer.go
+++ b/internal/pkg/cosign/keyless/signer.go
@@ -1,0 +1,77 @@
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keyless
+
+import (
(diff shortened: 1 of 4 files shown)


internal/pkg/cosign/keyless/signer.go
```diff
@@ -0,0 +1,77 @@
...
+
+// NewSigner generates a new private signing key and returns a `cosign.Signer` which creates signatures with it.
+func NewSigner() (icosign.Signer, error) {
+	priv, err := cosign.GeneratePrivateKey()
+	if err != nil {
+		return nil, errors.Wrap(err, "generating cert")
+	}
+	s, err := signature.LoadECDSASignerVerifier(priv, crypto.SHA256)
+	if err != nil {
+		return nil, err
```
mattmoor, line 72: We should consistently wrap errors, I think.

internal/pkg/cosign/keyless/signer.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return ociSig, pub, err
+}
+
+// NewSigner generates a new private signing key and returns a `cosign.Signer` which creates signatures with it.
+func NewSigner() (icosign.Signer, error) {
+	priv, err := cosign.GeneratePrivateKey()
+	if err != nil {
+		return nil, errors.Wrap(err, "generating cert")
+	}
```
mattmoor, line 69: Generally when we say "keyless" we mean the full Fulcio flow with the ephemeral certificate chained into the root CA (and optionally logged to Rekor).  Perhaps to avoid creating confusion over the term "keyless" (since it is only part) we should call this "ephemeral"?

internal/pkg/cosign/keyless/signer.go
dekkagaijin: done

mattmoor (approved):

Issue #1195, opened 2021-12-13
The `upload` flag of `cosign initialize` is not used


rgerganov:
`cosign initialize` has an `upload` flag which is supposed to set a threshold of root key signers and has default value 3. Currently this value is not used in any way because we always take the root threshold of the local repo and use this number to update from the remote one.
I propose to change the name of the flag from `upload` to `threshold` and use it to override the root threshold when updating from the remote TUF repo. I believe this was the original intent of this flag.

rgerganov:
After discussion with @asraa we decided to simply delete this flag (PR #1201)

PR #1196, opened 2021-12-13
Allow setting root threshold for "cosign initialize"

rgerganov:
There is "upload" flag which does nothing. This patch renames the flag
to "threshold" and make it override the threshold of root key signers
when updating from the remote TUF repo.

Closes: #1195

### DIFF
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -325,7 +325,7 @@ func downloadRemoteTarget(name string, c *client.Client, out client.Destination)
 
 // Instantiates the global TUF client. Uses the embedded (by default trusted) root in cosign
 // unless a custom root is provided. This will always perform a remote call to update.
-func Init(ctx context.Context, altRootBytes []byte, remote client.RemoteStore, threshold int) error {
+func Init(ctx context.Context, altRootBytes []byte, remote client.RemoteStore) error {
 	rootClient, err := RootClient(ctx, remote, altRootBytes)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


dlorenc:
Cc @asraa

pkg/cosign/tuf/client.go
```diff
@@ -336,12 +336,12 @@ func Init(ctx context.Context, altRootBytes []byte, remote client.RemoteStore, t
 			return err
 		}
 	}
-	rootKeys, rootThreshold, err := getRootKeys(altRootBytes)
```
asraa, line 339: hey! so i'm thinking maybe we want to just remove the flag entirely -- right now it was getting the threshold from the current trusted root rather than a threshold out of band like in the flag. This was only because `Init` reuqires a list of root keys and a threshold, but since that's much harder for a user to get (you need to self-parse the root.json) i had just added the whole trusted root as the flag (and then parse for you). WDYT?

otherwise if you would like to keep the flag in case you want more signers than the specified threshold in the root.json, i would say we default to the threshold in the return value here.

asraa (commented):
Thanks! Just a comment here.

pkg/cosign/tuf/client.go
rgerganov: I was also thinking to just delete the flag but wasn't sure if there is a use case for it. Let's keep it simple and delete it. I will update the patch.

rgerganov:
I will close this PR and open a new one

PR #1199, opened 2021-12-14
create `mutate` functions for `oci.Signature`

dekkagaijin:
#### Ticket Link
#844

#### Release Note
```release-note
NONE
```

### DIFF
--- a/pkg/oci/mutate/signature.go
+++ b/pkg/oci/mutate/signature.go
@@ -124,70 +124,68 @@ func (sw *sigWrapper) Size() (int64, error) {
 	return sw.wrapped.Size()
 }
 
-// SignatureAnnotations returns a new `oci.Signature` based on the provided one
-func SignatureAnnotations(sig oci.Signature, newAnnotations map[string]string) (oci.Signature, error) {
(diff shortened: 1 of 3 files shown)


pkg/oci/mutate/signature.go
```diff
@@ -0,0 +1,193 @@
...
+	return sw.wrapped.Uncompressed()
+}
+
+// Size implements v1.Layer
+func (sw *sigWrapper) Size() (int64, error) {
+	return sw.wrapped.Size()
+}
+
+// SignatureAnnotations returns a new `oci.Signature` based on the provided one
+func SignatureAnnotations(sig oci.Signature, newAnnotations map[string]string) (oci.Signature, error) {
```
mattmoor, line 128: I wonder if we should expose `mutate.Signature(sig, opts...)` with functional options for fiddling with the individual properties.  This would let us batch a set of alterations together and feels less verbose 🤔

dekkagaijin: Thought about that; we'd have to either expose the static.Options or create new ones here. I'd be OK with having a common options folder

PR #1201, opened 2021-12-14
Remove the "upload" flag for "cosign initialize"

rgerganov:
The "upload" flag is not used anywhere and it is not really needed. When
we update from the remote TUF repo, we expect the same number of root
signatures (or more) which is a sensible default.

Closes: #1195

Signed-off-by: Radoslav Gerganov <rgerganov@vmware.com>


hectorj2f (approved):
lgtm

PR #1203, opened 2021-12-14
[WIP] Implements new policy init iteration

lukehinds:
Implement a policy init with no signed body and remove unneeded tuf'isms

Users will now instead sign the whole policy and attach the sig
to the cosign signature annotation,rather than the signature body
of the policy

This also introduces a `previous_root` key which will allow
clients that consume policy (such as sget), to walk back a linear
proof of previous signers.

Signed-off-by: Luke Hinds <lhinds@redhat.com>

cc @asraa @imjasonh @jyotsna-penumaka

### DIFF
--- a/cmd/cosign/cli/policy_init.go
+++ b/cmd/cosign/cli/policy_init.go
@@ -18,7 +18,6 @@ package cli
 import (
 	"bytes"
 	"context"
-	// "encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -42,7 +41,6 @@ import (
 	cremote "github.com/sigstore/cosign/pkg/cosign/remote"
 	"github.com/sigstore/cosign/pkg/cosign/tuf"
 	"github.com/sigstore/cosign/pkg/sget"
-	// sigs "github.com/sigstore/cosign/pkg/signature"
 	signatureoptions "github.com/sigstore/sigstore/pkg/signature/options"
 	"github.com/spf13/cobra"
 )
--- a/cmd/cosign/cli/policy_init.go
+++ b/cmd/cosign/cli/policy_init.go
@@ -18,7 +18,7 @@ package cli
 import (
 	"bytes"
 	"context"
-	"encoding/base64"
+	// "encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -42,7 +42,7 @@ import (
 	cremote "github.com/sigstore/cosign/pkg/cosign/remote"
 	"github.com/sigstore/cosign/pkg/cosign/tuf"
 	"github.com/sigstore/cosign/pkg/sget"
-	sigs "github.com/sigstore/cosign/pkg/signature"
+	// sigs "github.com/sigstore/cosign/pkg/signature"
 	signatureoptions "github.com/sigstore/sigstore/pkg/signature/options"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


lukehinds:
still a stray sig to deal with. will get that

cmd/cosign/cli/policy_init.go
```diff
@@ -116,6 +114,7 @@ func initPolicy() *cobra.Command {
 			role.AddKeysWithThreshold(publicKeys, o.Threshold)
 			root.Roles["root"] = role
 			root.Namespace = o.ImageRef
+			root.PreviousRoot = "null"
```
imjasonh, line 117: Can we just have it be an empty string, or omitted entirely? I'd rather not have the `"null"` string mean something special if we can avoid it.

pkg/cosign/tuf/policy.go
```diff
@@ -66,15 +58,14 @@ func (k *Key) ContainsID(id string) bool {
...
 type Root struct {
-	Type        string           `json:"_type"`
-	SpecVersion string           `json:"spec_version"`
-	Version     int              `json:"version"`
-	Expires     time.Time        `json:"expires"`
-	Keys        map[string]*Key  `json:"keys"`
-	Roles       map[string]*Role `json:"roles"`
-	Namespace   string           `json:"namespace"`
-
-	ConsistentSnapshot bool `json:"consistent_snapshot"`
```
imjasonh, line 77: What's `ConsistentSnapshot`? Is it okay that we're removing it here?

pkg/cosign/tuf/signer.go
```diff
@@ -36,10 +36,9 @@ type FulcioKeyVal struct {
 func FulcioVerificationKey(email string, issuer string) *Key {
 	keyValBytes, _ := json.Marshal(FulcioKeyVal{Identity: email, Issuer: issuer})
 	return &Key{
-		Type:       KeyTypeFulcio,
-		Scheme:     KeySchemeFulcio,
-		Algorithms: KeyAlgorithms,
```
imjasonh, line 41: What's `KeyAlgorithms`? Is it okay that we're removing it here?

imjasonh (commented):
I still need to look at this closer, just some initial comments.

pkg/cosign/tuf/policy.go
asraa: Yeah, it's TUF language for versioning the names of the policy, which we aren't doing yet anyway (e.g. naming the root policies 1.root, 2.root, 3.root).... unless we do want to do that now that I think about it...

pkg/cosign/tuf/signer.go
asraa: That supplies what hash is used to create the key identifier (the if in  id : keystuff). i think we can remove and fix on sha256, since it's not a security concern, we just need a unique id

asraa (commented):
Thanks! I think the `signPolicy` code can almost be totally obliterated btw -- no longer need to re-upload the root. Should look something like a wrapper to `SignCmd`

pkg/cosign/tuf/policy.go
```diff
@@ -31,22 +31,14 @@ import (
...
+type Policy struct {
+	Policy json.RawMessage `json:"policy"`
 }
 
 type Key struct {
-	Type       string          `json:"keytype"`
-	Scheme     string          `json:"scheme"`
-	Algorithms []string        `json:"keyid_hash_algorithms,omitempty"`
-	Value      json.RawMessage `json:"keyval"`
+	Type   string          `json:"keytype"`
```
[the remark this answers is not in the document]
lukehinds: I think it makes sense for PreviousRoot (changing null to an empty string), I don't believe we would want namespace etc to be optional. Which ones did you have in mind?

asraa:
Reviving this, are we aware of https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md, and how should we comply to this?

lukehinds:
I don't think this is going anywhere now and no intention to keep working on it myself, so will close. If anyone wants to pick up this work, feel free to grab what code is useful to you.

Issue #1207, opened 2021-12-14
Kind e2e tests are broken


dekkagaijin:
e.g. https://github.com/sigstore/cosign/runs/4526494006?check_suite_focus=true

PR #1208, opened 2021-12-14
Fall back to in-memory cache when local TUF root cannot be instantiated

dekkagaijin:
#### Ticket Link
Fixes https://github.com/sigstore/cosign/issues/1207

#### Release Note
```release-note
fixed: panics like `panic ... creating root cert pool: retrieving trusted root; local cache may be corrupt: creating cached local store: mkdir $HOME/.sigstore: read-only file system`
```

### DIFF
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -171,12 +171,10 @@ func initGlobalRootClient(ctx context.Context, remote client.RemoteStore, altRoo
 		}
 	}
 
-	local, err := tuf_leveldbstore.FileLocalStore(localCacheDBPath)
-	if err != nil {
-		if !errors.Is(err, fs.ErrPermission) {
-			return nil, errors.Wrap(err, "creating cached local store")
-		}
-		local = client.MemoryLocalStore()
+	local := client.MemoryLocalStore()
+	if localDB, err := tuf_leveldbstore.FileLocalStore(localCacheDBPath); err == nil {
+		// TODO: log errors
+		local = localDB
 	}
 
 	// We may need to download latest metadata and targets if the cache is un-initialized or expired.
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -173,7 +173,10 @@ func initGlobalRootClient(ctx context.Context, remote client.RemoteStore, altRoo
 
 	local, err := tuf_leveldbstore.FileLocalStore(localCacheDBPath)
 	if err != nil {
-		return nil, errors.Wrap(err, "creating cached local store")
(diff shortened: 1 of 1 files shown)


dlorenc:
Any idea why this is failing in e2e?

asraa:
i'm not 100% sure if this is needed? I'm going to do a bit of a code cleanup in this whole code and make sure that we can (a) do a non-write mode and fetch inmemory (b) do a non-write and non-network mode and fallback to old targets and (c) obv hold onto the normal flow if we can write and and fetch

dekkagaijin:
> Any idea why this is failing in e2e?

We're still looking for `target` metadata files (in addition to local root cache) on disc, which still blows up when we don't have read permissions. The "fix" in this PR is incomplete

> i'm not 100% sure if this is needed? I'm going to do a bit of a code cleanup in this whole code

SGTM

Itxaka:
This also fixes an issue when launching several cosign in parallel as they fight to open the same `TUF_ROOT` files and fail with `resource temporarily unavailable`

we worked around this by giving each cosign its own temporal `TUF_ROOT` :)

hectorj2f:
We are facing this same issue.

dlorenc:
I think @asraa had a fix ready but then ran into an issue: https://twitter.com/AsraEntr0py/status/1473641163320180736

@asraa do you want someone else to pick this one up?

hectorj2f:
@dlorenc It is failing because this branch is missing the changes here https://github.com/sigstore/cosign/pull/1209. That is why the cosigned test is failing. I could get a PR rebased (https://github.com/hectorj2f/cosign/tree/read-only) with @dekkagaijin changes and latest version of main to merge these changes without bothering @dekkagaijin.

hectorj2f:
@dlorenc @asraa let me know if you want me to pick it up.

asraa:
@hectorj2f @ltxaka -- not sure if this usecase works for you but since then @dlorenc added an option to SIGSTORE_NO_CACHE that stores all the targets in-memory. Does that work for you? Alterntatively, does setting the TUF_ROOT env variable suffice like you say?

(sorry for the delay! my personal laptop had also gotten stolen so i was pretty device free. i'm back now!)

Itxaka:
For us, just setting the TUF_ROOT was enough, but our use case was several sigstore binaries being launched in parallel and fighting for those files so it may not cover other paths in which this appeared..

For SIGSTORE_NO_CACHE I would need to try, but I'm pretty sure it would also fix it, no file written means no fighting to open the same file, so it would work.

dlorenc:
I think this one can be closed now!

hectorj2f:
@asraa I apologize for missing your comment. Yes, it works.

PR #1216, opened 2021-12-15
feat: implement cosign download attestation

venafi-iw:
#### Summary
This PR provides support for downloading in-toto attestation(s) from a supplied container image.

#### Release Note

* Add CLI option to download attestations from a supplied container image:
  `cosign download attestation <IMAGE>`


dlorenc:
This makes sense to me - at first I was thinking it could be dangerous to download these without verification but the attestation is pretty clearly a signed object that requires unpacking/verification before use.

I'm +1 here, but want to let this sit for a bit to see if others agree.

dlorenc (approved):

dlorenc:
OK, seems good to merge!

PR #1221, opened 2021-12-15
create `DSSEAttestor` interface, `payload.DSSEAttestor` implementation

dekkagaijin:
This is the `Attestor` equivalent of the previous `Signer` refactoring, plus nips and tucks here and there

TODO:

* implement the rest of the attestors
* move over `fulcio` operations into their respective `Signer` and `DSSEAttestor` implementations, stop using the fulcio `signature.Signer`
* perform all signing and attestation operations via new `/pkg` libraries

#### Ticket Link
#844 #931

#### Release Note
```release-note
NONE
```


internal/pkg/cosign/payload/attestor.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
+	"github.com/sigstore/cosign/internal/pkg/cosign"
+	"github.com/sigstore/cosign/pkg/oci"
+	"github.com/sigstore/cosign/pkg/oci/static"
+	"github.com/sigstore/cosign/pkg/types"
+	"github.com/sigstore/sigstore/pkg/signature"
+)
+
+type payloadAttestor struct {
+	payloadSigner
```
mattmoor, line 34: Don't inline this.  Inlining this means that this now publicly implements the Signer interface, which we don't want (it is an internal detail).

internal/pkg/cosign/payload/signer.go
```diff
@@ -38,40 +38,57 @@ var _ cosign.Signer = (*payloadSigner)(nil)
...
 	}
 
-	pkOpts := []signature.PublicKeyOption{signatureoptions.WithContext(ctx)}
-	pkOpts = append(pkOpts, ps.publicKeyProviderOpts...)
-	pk, err := ps.payloadSigner.PublicKey(pkOpts...)
+	return ociSig, pk, nil
+}
+
+func (ps *payloadSigner) signPayload(ctx context.Context, payload io.Reader) (payloadBytes, sig []byte, pk crypto.PublicKey, err error) {
+	payloadBytes, err = io.ReadAll(payload)
```
mattmoor, line 56: There are a lot of paths where we `io.ReadAll` and `bytes.NewBuffer(someBytes)` around this signature.

I'm like 95% certain that if you kill this return value and forced the caller to read it themselves and pass a `bytes.NewReader` that the overall change would be a bunch simpler and clearer.

4 return values is a lot and this is sorta starting to have a code smell of 1 function doing too many things (as evidenced by its signature).  Already a better name would be `readAndSignPayload`, but I think we should drop the return value, which kills the `readAnd` bit.

internal/pkg/cosign/payload/attestor.go
```diff
@@ -0,0 +1,86 @@
...
+	pae := dsse.PAE(pa.payloadType, p)
+
+	pb, sig, pk, err := pa.signPayload(ctx, bytes.NewReader(pae))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	envelope := dsse.Envelope{
+		PayloadType: pa.payloadType,
+		Payload:     base64.StdEncoding.EncodeToString(pb),
```
mattmoor, line 56: `pb` is the same as `pae`

internal/pkg/cosign/payload/signer.go
```diff
@@ -38,40 +38,57 @@ var _ cosign.Signer = (*payloadSigner)(nil)
 
 // Sign implements `Signer`
 func (ps *payloadSigner) Sign(ctx context.Context, payload io.Reader) (oci.Signature, crypto.PublicKey, error) {
-	payloadBytes, err := io.ReadAll(payload)
+	payloadBytes, sig, pk, err := ps.signPayload(ctx, payload)
```
mattmoor, line 41: I'd preface this with `payloadBytes, err := io.ReadAll(payload)` and pass `bytes.NewBuffer(payloadBytes)` to `newSigner`

internal/pkg/cosign/payload/signer.go
```diff
@@ -38,40 +38,57 @@ var _ cosign.Signer = (*payloadSigner)(nil)
...
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	b64sig := base64.StdEncoding.EncodeToString(sig)
-	ociSig, err := static.NewSignature(payloadBytes, b64sig)
+	sOpts := []signature.SignOption{signatureoptions.WithContext(ctx)}
+	sOpts = append(sOpts, ps.payloadSignerOpts...)
+	sig, err = ps.payloadSigner.SignMessage(bytes.NewReader(payloadBytes), sOpts...)
```
mattmoor, line 63: On the attest path (where we have bigger payloads) we read the payload into the DSSE, wrapped it into a `bytes.Buffer` to pass in here, then we `io.ReadAll()` above only to wrap it in another `bytes.Buffer` here 🙃

internal/pkg/cosign/payload/signer.go
```diff
@@ -38,40 +38,57 @@ var _ cosign.Signer = (*payloadSigner)(nil)
...
 		payloadSigner:         s,
 		payloadSignerOpts:     sOpts,
 		publicKeyProviderOpts: pkOpts,
 	}
 }
+
+// NewSigner returns a `cosign.Signer` which uses the given `signature.Signer` to sign requested payloads.
+func NewSigner(s signature.Signer,
+	sOpts []signature.SignOption,
+	pkOpts []signature.PublicKeyOption) cosign.Signer {
```
mattmoor, line 91: A trick for passing multiple kinds of options via variadic args is to take `...interface{}` and then walk the list with a type switch to place each option into an appropriately typed list.

You lose a bit of static type checking, but any tested callsite will get dynamically checked.

dekkagaijin: I hate it but will direct my rage at Go typing

mattmoor: Maybe next they will add union types :)

internal/pkg/cosign/payload/attestor.go
dekkagaijin: ACK

internal/pkg/cosign/payload/signer.go
dekkagaijin: you mean `signPayload`?
done

internal/pkg/cosign/payload/signer.go
dekkagaijin: c'est la vie

internal/pkg/cosign/payload/signer.go
dekkagaijin: Receiving `io.Reader`s for signatures and payloads is an unexpectedly popular request, but I'd be OK supporting an interface which takes byte arrays

internal/pkg/cosign/payload/attestor.go
dekkagaijin: done

internal/pkg/cosign/payload/signer.go
mattmoor: I was actually suggesting we keep the `io.Reader`, but shuffling things so there are fewer layers of `bytes.NewBuffer` by reorienting expectations of who reads what.  That said, the only meaningful change here is in private stuff, so we can freely change things later :)

internal/pkg/cosign/payload/signer.go
mattmoor: yes thanks

mattmoor (approved):

PR #1226, opened 2021-12-16
add `fulcio.DSSEAttestor`

dekkagaijin:
This is just to create a `Attestor` equivalent for the eventual upstream splitting of `Signing` and `Attesting`.

#### Ticket Link
#844 #931

#### Release Note
```release-note
NONE
```

### DIFF
--- a/internal/pkg/cosign/fulcio/attestor.go
+++ b/internal/pkg/cosign/fulcio/attestor.go
@@ -1,0 +1,60 @@
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


internal/pkg/cosign/fulcio/attestor.go
```diff
@@ -0,0 +1,61 @@
...
+var _ cosign.DSSEAttestor = (*fulcioAttestor)(nil)
+
+// Attest implements `cosign.DSSEAttestor`
+func (fa *fulcioAttestor) DSSEAttest(ctx context.Context, payload io.Reader) (oci.Signature, crypto.PublicKey, error) {
+	att, pub, err := fa.inner.DSSEAttest(ctx, payload)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO(dekkagaijin): move the fulcio SignerVerififer logic here
```
mattmoor, line 44: Can we just do this?

I feel like we keep rehashing the argument about `fulcio` being a wrapper 😉

dekkagaijin: I would, but doing that involves changing _everything_

mattmoor: Only if you *use* it right?  We talked about forking chunks of logic over here and then replacing the spaghetti monster with the composable blocks 😇

dekkagaijin: fair, I'll migrate things over for the Attestor

Issue #1229, opened 2021-12-16
Do not require multiple Fulcio certs in the TUF root


rgerganov:
`cosign` requires both `fulcio.crt.pem` and `fulcio_v1.crt.pem` in the TUF root which doesn't make sense when using local TUF. `fulcio_v1.crt.pem` was added in the embedded TUF in order to support Fulcio v1 but it shouldn't be required when users initialize `cosign` with their own TUF repo.

I will submit a patch which will make `fulcio_v1.crt.pem` optional.

/cc @asraa

asraa:
I suppose this would also be fixed by https://github.com/sigstore/cosign/issues/1224

PR #1230, opened 2021-12-16
Do not require multiple Fulcio certs in the TUF root

rgerganov:
cosign requires both fulcio.crt.pem and fulcio_v1.crt.pem in the TUF
root which doesn't make sense when using local TUF. fulcio_v1.crt.pem
was added in the embedded TUF in order to support Fulcio v1 but it
shouldn't be required when users initialize cosign with their own TUF
repo.

Closes: #1229

Signed-off-by: Radoslav Gerganov <rgerganov@vmware.com>

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
+++ b/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"sync"
 
-	"github.com/pkg/errors"
 	"github.com/sigstore/cosign/pkg/cosign/tuf"
 )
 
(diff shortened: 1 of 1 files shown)


hectorj2f (commented):
Lgtm

cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
```diff
@@ -64,11 +64,16 @@ func initRoots() *x509.CertPool {
 		// Retrieve from the embedded or cached TUF root. If expired, a network
 		// call is made to update the root.
 		ctx := context.Background() // TODO: pass in context?
-		for _, fulcioTarget := range []string{fulcioTargetStr, fulcioV1TargetStr} {
-			buf := tuf.ByteDestination{Buffer: &bytes.Buffer{}}
-			if err := tuf.GetTarget(ctx, fulcioTarget, &buf); err != nil {
-				panic(errors.Wrap(err, "creating root cert pool"))
-			}
+		buf := tuf.ByteDestination{Buffer: &bytes.Buffer{}}
```
asraa, line 67: WDYT about keeping the previous code and use a bool to panic if one of the targets wasn't added?
(as in gracefully fail, just fail when none of the errors were nil)

asraa (commented):
Thank you so so much for these changes!

I think long term we might want some kind of general discovery like look for targets named `fulcio.*`, and as long as one of them is successfully added, don't panic.

cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
rgerganov: make sense, I will update the patch

dlorenc (approved):

PR #1241, opened 2021-12-22
Add support for other public key types for SCT verification, allow override for testing.

### LINKED ISSUE #1240: Support other Public keys than ecdsa when verifying SCT
**Description**

For verifying the SCT cosign seems to assume that public keys are always going to be Elliptical. I also verified that if I spin up an instance of CT Log here:
https://github.com/google/certificate-transparency-go/blob/master/trillian/docs/ManualDeployment.md#key-generation

With RSA keys, then cosign is unable to verify the SCT produced by the CT Log.

The assumption seems to be here:
https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go#L57

And with some debugging magic enabled, it fails:
```
getting signer: getting key from Fulcio: verifying SCT: cannot verify RSA signature with *ecdsa.PublicKey key
```

So, I was planning to add other public key types supported for verification.

vaikas:
#### Summary

Fixes #1240 

Also when using, say a testing CT log (where I found the above issue), it would be nice to be able to specify a different public key to use for SCT validation, so basically the public key for the CT Log. Other option of course would be to just use the insecure bit, but I reckon this is nicer and allows for more flexibility.
Introduce `CT_LOG_PUBLIC_KEY_FILE` environmental variable that allows one to override (with nice warnings to boot) where the public key used for validating the SCT.

#### Ticket Link

Fixes

#### Release Note

```release-note
Introduce an env variable `CT_LOG_PUBLIC_KEY_FILE` that allows one to use a different public key for validating SCT from the CT Log.
```


cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
```diff
@@ -37,6 +40,10 @@ import (
 // This is the CT log public key target name
 var ctPublicKeyStr = `ctfe.pub`
 
+// Setting this env variable will over ride what is used to validate
+// the SCT coming back from Fulcio.
+const altCTLogPublicKeyLocation = "CT_LOG_PUBLIC_KEY_FILE"
```
dlorenc, line 45: What do you think about prefixing this with `SIGSTORE_`? We have some env vars prefixed with `COSIGN_` for cosign-specific things, and then`SIGSTORE_ROOT_FILE` and `SIGSTORE_NO_CACHE` for the TUF root itself.

dlorenc (approved):

Issue #1242, opened 2021-12-22
[sign] output log misrepresents target repo for signatures


Itxaka:
**Description**

When pushing signatures to a different repository than the image we are signing there is a message in the terminal that seems to tell us that we are pushing to the same repo where the image resides.

This doesn't take into account the `COSIGN_REPOSITORY` value for the output which can cause confusion in users (I got confused and kind of terrified of pushing to a different repo!)

PR #1243, opened 2021-12-22
Log the proper remote repo for the signatures on verify

Itxaka:
#### Summary

Currently we print the image repo being signed as if we were pushing the
signatures to that same repo, which could lead to users being surprised.

Instead try to get the override and print that instead

#### Ticket Link

Resolves: #1242

#### Release Note

```release-note
NONE
```


dlorenc (approved):

Itxaka:
woops sorry, I thought the `make golangci-lint` would run the linter, seems like it only installs it.

PR #1245, opened 2021-12-22
feat: resolve --cert from URL

caarlos0:
#### Summary

`verify` and `verify-blob`'s `--cert` param should accept an URL as well, so we can verify without previously downloading files, e.g.:

```sh
cosign verify-blob \
  --cert https://github.com/goreleaser/supply-chain-example/releases/download/v1.0.0/checksums.txt.pem \
  --signature https://github.com/goreleaser/supply-chain-example/releases/download/v1.0.0/checksums.txt.sig \
  https://github.com/goreleaser/supply-chain-example/releases/download/v1.0.0/checksums.txt
```

#### Ticket Link

https://github.com/gythialy/golang-cross/pull/67/files#r774120479

#### Release Note

```release-note
verify and verify-blob resolve --cert from URL
```

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -32,6 +32,7 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
 	"github.com/sigstore/cosign/cmd/cosign/cli/sign"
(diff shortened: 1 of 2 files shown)


developer-guy (approved):
LGTM

dlorenc:
This is safe IMO - the cert itself is signed by fulcio and must be in a transparency log, so the location of it can be untrusted.

PR #1248, opened 2021-12-23
Fix semantic bug in DSSE specification.

dlorenc:
The 'payloadType' field in a DSSE envelope is always supposed to be set
to the in-toto media type. We were incorrectly placing the inner predicateType
here.

The DSSE library probably shouldn't even expose this as a parameter given that it's
a MUST in the spec.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -138,7 +138,7 @@ func AttestCmd(ctx context.Context, ko sign.KeyOpts, regOpts options.RegistryOpt
 		return errors.Wrap(err, "getting signer")
 	}
 	defer sv.Close()
-	wrapped := dsse.WrapSigner(sv, predicateURI)
+	wrapped := dsse.WrapSigner(sv, types.IntotoPayloadType)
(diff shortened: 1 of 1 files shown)


dlorenc:
cc @mlieberman85 for a second instance of that bug we found earlier :)

dlorenc:
The spec itself is here: https://github.com/in-toto/attestation/tree/main/spec#envelope

loosebazooka (approved):
stop working dan it's the holidays. :snowman: :pie:

loosebazooka:
>The DSSE library probably shouldn't even expose this as a parameter given that it's
a MUST in the spec.

yeah that's weird

dlorenc:
NVM I got the layering wrong - the DSSE implementation is fine.

In DSS, you can set any payloadType: https://github.com/secure-systems-lab/dsse/blob/master/protocol.md

The In-ToTo Attestation Spec is the one that requires that payloadType: https://github.com/in-toto/attestation/tree/main/spec#envelope

So fixing it here makes sense, given it's a library for generating in-toto attestations.

PR #1249, opened 2021-12-24
Fix semantic bugs in attestation verifification.

dlorenc:
DSSE requires that the payload types match, but the Attestation
spec requires that the payload type be the specific in-toto one. We
don't actually do anything with the content here other than dump it
in plain text, but semantically we're calling these attestations in the
CLI so we should enforce that part of the specification as well.

#### Ticket Link

Fixes

#### Release Note

```release-note
BREAKING: In-Toto Attestations generated in older versions of cosign will not verify in newer versions. 
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/sigstore/cosign/pkg/blob"
 	"github.com/sigstore/cosign/pkg/oci/static"
+	"github.com/sigstore/cosign/pkg/types"
 
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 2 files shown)


dlorenc:
Similar to #1248 but reverse.

pkg/cosign/verify.go
```diff
@@ -106,18 +107,25 @@ func verifyOCISignature(ctx context.Context, verifier signature.Verifier, sig oc
...
+func verifyOCIAttestation(_ context.Context, verifier signature.Verifier, att payloader) error {
 	payload, err := att.Payload()
 	if err != nil {
 		return err
 	}
 
 	env := ssldsse.Envelope{}
 	if err := json.Unmarshal(payload, &env); err != nil {
-		return nil
+		return err
```
loosebazooka, line 123: wild

loosebazooka (approved):

PR #1253, opened 2021-12-26
feat: support attach attestation

### LINKED ISSUE #1202: consider attaching attestations to the OCI registry but verify its signature first
**Description**

As we support attaching SBOMs and signatures by using the `cosign attach` command, why don't we add attaching attestation files support as an addition to that? The only option people have is that `cosign attest,` but it also signs the attestation and attaches it to the OCI Registry. What I want to do is that only store the attestation files in OCI registries.

cc: @dlorenc @hectorj2f

developer-guy:

#### Summary

To test this, run the following commands:

```shell
$ crane copy alpine <registry>/<username>/alpine
$ echo "foo" > bar
$ cosign attest --key cosign.key --predicate bar <registry>/<username>/alpine
$ cosign download attestation <registry>/<username>/alpine > signedpayload.json
$ cosign attach attestation --key cosign.pub --attestation signedpayload.json <registry>/<username>/alpine
```

#### Ticket Link

Fixes #1202 

#### Release Note

```release-note
feat: support attach attestation
```

### DIFF
--- a/cmd/cosign/cli/attach/attestation.go
+++ b/cmd/cosign/cli/attach/attestation.go
@@ -1,0 +1,88 @@
+// Copyright 2021 The Sigstore Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attach
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+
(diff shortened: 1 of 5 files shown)


cmd/cosign/cli/attach/attestation.go
```diff
@@ -0,0 +1,135 @@
...
+			if err != nil {
+				return errors.Wrap(err, "creating Rekor client")
+			}
+			co.RekorClient = rekorClient
+		}
+		co.RootCerts = fulcio.GetRoots()
+	}
+
+	// Keys are optional!
+	if keyRef != "" {
```
dlorenc, line 61: Sorry! I think I wasn't clear earlier. We don't actually need to verify the signature itself (that seems a bit overkill), just verify that the attestation object is well-formed and contains a signature.

My worry was someone doing something like: `cosign attach attestation --attestation=README.md <image>` and assuming it would sign the README.md file automatically.

developer-guy: ah okay, now I got it, thank you, it should be fixed right now. 🤝

dlorenc (approved):

dlorenc:
An e2e test for this one would be great, but we can do that on a follow-up.

PR #1255, opened 2021-12-27
Moved certificate output before checking for upload during signing

sabre1041:
Moved logic to print certificate during signing earlier on otherwise logic is not reached when `--upload=false`


dlorenc (approved):

dlorenc:
GitHub is struggling today.

sabre1041:
@sabre1041 github must have the holiday hangover :)

PR #1265, opened 2022-01-02
Add attest blob functionality

mlieberman85:
This mostly copies the functionality of image attestation and blob
signing.

Signed-off-by: Michael <mlieberman85@gmail.com>

#### Summary
This will allow users to attest local blobs similar to attesting images, following a similar pattern to signing blobs/images. Currently the code allows for attesting local blobs only. It probably wouldn't take much effort to include just a hash to attest as not all blobs will be available locally for checksumming. This also supports printing attestation locally or uploading it to Rekor.

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -1,0 +1,153 @@
+package attest
+
+import (
+	"bytes"
+	"context"
+	"crypto"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sigstore/cosign/cmd/cosign/cli/options"
+	"github.com/sigstore/cosign/cmd/cosign/cli/sign"
+	"github.com/sigstore/cosign/pkg/cosign"
+	"github.com/sigstore/cosign/pkg/cosign/attestation"
+	"github.com/sigstore/cosign/pkg/oci/static"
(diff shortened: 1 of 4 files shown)


mlieberman85:
This is still very much a WIP. Need to add some abstractions to deduplicate some of the code between attest and attestblob.

dlorenc (commented):
This looks like the right shape overall, thanks!

mlieberman85:
Ughh I'll fix this one minute

mlieberman85:
I have some additional stuff I've changed/added. I'm also going to include signing of the hash directly.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -0,0 +1,158 @@
...
+	if artifactHash == "" {
+		digest, _, err := signature.ComputeDigestForSigning(bytes.NewReader(artifact), crypto.SHA256, []crypto.Hash{crypto.SHA256, crypto.SHA384})
+		if err != nil {
+			return err
+		}
+		hexDigest = strings.ToLower(hex.EncodeToString(digest))
+	} else {
+		hexDigest = artifactHash
+	}
+	wrapped := dsse.WrapSigner(sv, predicateURI)
```
priyawadhwa, line 97: This should be:

```
wrapped := dsse.WrapSigner(sv, types.IntotoPayloadType)
```

Rekor search expects that the `PayloadType` is of type `Intoto` when searching through `IntotoRecord` object types:

https://github.com/sigstore/rekor/blob/main/pkg/types/intoto/v0.0.1/entry.go#L82

mlieberman85: Thanks!

PR #1268, opened 2022-01-04
One-to-one mapping of invocation to scan result

### LINKED ISSUE #1262: Consider one-to-one mapping of an invocation to scan result
The recent addition of the [`COSIGN_VULN_ATTESTATION_SPEC`](https://github.com/sigstore/cosign/blob/3dd690ea18cd65adbcea2e9ae9686cbb4b2ee428/specs/COSIGN_VULN_ATTESTATION_SPEC.md) supports storing multiple scan results originating from a single invocation. I think it would be more intuitive if we encouraged folks to map a single invocation to a single scan result.

This would not only make it easier to construct the payload, but easier to write and evaluate policy against as well. I think we should discourage concatenating the outputs from multiple scans in a single invocation. Doing so would be inconsistent with what `invocation` is intended to represent as far as I can tell.

Constructing separate vulnerability attestations for different types of scanners makes it easier to write a process that evaluates one or more policies against them. For example, you would download a single, named vulnerability attestation and have some guarantee (or at least a reasonable expectation) that it will contain a result set of a particular schema.

Slack thread for reference: https://sigstore.slack.com/archives/C01PZKDL4DP/p1640878335099100

jdolitsky:
In COSIGN_VULN_ATTESTATION_SPEC, modify the "scanners" field to be a single object "scanner" (vs. an array) so that image scans are tied to a single invocation.

Resolves #1262

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -46,7 +46,7 @@ type CosignPredicate struct {
 // VulnPredicate specifies the format of the Vulnerability Scan Predicate
 type CosignVulnPredicate struct {
 	Invocation Invocation `json:"invocation"`
-	Scanners   []Scanner  `json:"scanners"`
+	Scanner    Scanner    `json:"scanner"`
 	Metadata   Metadata   `json:"metadata"`
 }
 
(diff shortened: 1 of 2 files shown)


dlorenc:
Welcome!

dlorenc (approved):

Issue #1273, opened 2022-01-05
Log verification for expired targets


haydentherapper:
**Overview**

This is a tracking issue for supporting verification for expired/rotated targets. @asraa and I will be working on this.

Currently, cosign assumes the latest TUF metadata can be used to validate signatures. As the Fulcio CA certificate will expire, we will have to rotate that target at some point. This will cause cosign to not be able to validate the signatures that chain up to the expired CA certificate, since the TUF metadata will contain a different certificate. 

We propose bundling a pointer to the metadata used when generating the signature. We will do this by including the snapshot or timestamp JSON in the signature bundle. Cosign will use this to find versioned TUF metadata. 

[Design doc](https://docs.google.com/document/d/1jIE44Fw7CVr0dLLwK9pYgNCIkzdCCeIUVuwgr0HRa8E/edit#)

**Tasks**

For cosign:

- [ ] Add current snapshot/timestamp.json to an entry’s bundle
- [ ] Add verification by pulling the referenced versioned x.targets.json and x.root.json
- [ ] Cache previous x.targets.json files and expired Fulcio certificates

There are a few tasks for the [Sigstore TUF repo](https://github.com/sigstore/root-signing/).

- [ ] Enable consistent snapshots in Sigstore root
- [ ] Add versioned targets.json and complete list of targets to Sigstore’s TUF root (and synced to GCS bucket)
- [ ] Generate snapshot/timestamp more frequently

asraa:
Some more thoughts to make the experience of expired target verification along with BYO TUF root:
- [ ] we may want to change the sigstore TUF root to pull in directories of targets rather than hard-coded target names (e.g. add all targets in a path `fulcio/**` to the trusted root pool rather than pull in a hardcoded target name `fulcio_v1.crt.pem`). this would reduce friction between any target version names and BYO TUF people

haydentherapper:
As I was implementing verification using the TUF timestamp, I ran into an issue. As a reminder, the verification flow is roughly:
1. Read the timestamp from an annotation on the OCI image
2. Using the timestamp, fetch the versioned snapshot
3. Using the snapshot, fetch the versioned root (N) and targets
4. Verify the root by walking the chain from 1.root.json to N.root.json
5. Verify the timestamp, snapshot and targets
6. Use the old targets to verify the OCI image

[...]

PR #1274, opened 2022-01-05
Update signature spec with TUF timestamp

haydentherapper:
The timestamp will be used to find a versioned TUF target metadata.
This allows targets to be rotated while still being able to validate
old entries that were signed by previous targets.

Ref #1273

#### Summary
Cosign currently can only verify signatures from current TUF metadata. This will cause signature verification to fail for old entries once the target, such as the Fulcio certificate, is rotated. Including the timestamp, along with additional work to persist versioned TUF targets, will allow cosign to find the previous target used to generate the signature.

I will implement this feature to be backwards compatible. If an entry does not include the timestamp, then cosign will use the latest TUF metadata.

#1273 contains additional context.

#### Release Note
```release-note
* Updated signature spec with TUF timestamp
```

### DIFF
--- a/specs/SIGNATURE_SPEC.md
+++ b/specs/SIGNATURE_SPEC.md
@@ -100,6 +100,7 @@ Gyp4apdU7AXEwysEQIb034aPrTlpmxh90SnTZFs2DHOvCjCPPAmoWfuQUwPhSPRb
 ```
 
 * `bundle` string
+
   This OPTIONAL property contains a JSON formatted `bundle` type, which can be used for offline verification.
   Example `bundle`:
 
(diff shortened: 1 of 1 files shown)


haydentherapper:
cc @asraa

asraa (commented):
Should this be signed? If not, would just putting the version number of the timestamp be equivalent?

Also -- should we 100% go for doing the timestamp lookup vs holding old fulcio / rekor certs into the current targets metadata to simplify logic?

asraa:
> Should this be signed? If not, would just putting the version number of the timestamp be equivalent?

JK I am wrong, this is not equivalent, the full timestamp contains the signatures on the timestamp.json

haydentherapper:
> Should this be signed? If not, would just putting the version number of the timestamp be equivalent?

Yep, since timestamp.json is signed, we can use the versioned root role to verify it and avoid attacks that trick a client into using a different version of the TUF metadata (for example, maybe one of the targets was compromised) to verify the signature.

> Also -- should we 100% go for doing the timestamp lookup vs holding old fulcio / rekor certs into the current targets metadata to simplify logic?

I'd prefer to avoid growing targets indefinitely. As discussed offline, this also makes it easier for those who want offline TUF metadata for verifying old signatures where the TUF metadata is no longer hosted. \
One benefit of keeping all old data in targets is if we want to mark a target as untrusted, we simply remove it from the list. However, this can be solved separately, as a signed list of untrusted targets.

dlorenc:
I like this approach! Nice work!

Should we merge this as is or wait until the implementation is ready?

bobcallaway (approved):
lgtm

asraa (approved):

haydentherapper:
Let's merge now, I'm working on the implementation, it should be ready soon.

PR #1280, opened 2022-01-07
Fix a few bugs in cosign initialize

haydentherapper:
* In getRoot, the metadata is always stored at the top level,
  not under repository/. This resulted in initialize always falling
  back to the embedded root.
* In Initialize, download all metadata and targets. This should
  avoid a disk write on verify.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -138,11 +138,12 @@ func New(ctx context.Context, remote client.RemoteStore, cacheRoot string) (*TUF
 }
 
 func getRoot(meta map[string]json.RawMessage) (json.RawMessage, error) {
-	trustedRoot, ok := meta[filepath.Join("repository", "root.json")]
+	trustedRoot, ok := meta["root.json"]
 	if ok {
 		return trustedRoot, nil
 	}
-	trustedRoot, err := embeddedRootRepo.ReadFile("repository/root.json")
+	// On first initialize, there will be no root in the TUF DB, so read from embedded.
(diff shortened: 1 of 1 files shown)


haydentherapper:
cc @asraa @dlorenc

haydentherapper:
Failing on just windows, gotta fix that

dlorenc:
> Failing on just windows, gotta fix that

Maybe useful- files seem to silently get converted to /r/n line endings when accessed on windows, even from the embedded data  I had to jump through some hoops to convert back before checking signatures.

asraa (approved):

pkg/cosign/tuf/client.go
```diff
@@ -212,7 +216,7 @@ func localStore(cacheRoot string) (client.LocalStore, error) {
 func embeddedLocalStore() (client.LocalStore, error) {
 	local := client.MemoryLocalStore()
 	for _, mdFilename := range []string{"root.json", "targets.json", "snapshot.json", "timestamp.json"} {
-		b, err := embeddedRootRepo.ReadFile(path.Join("repository", mdFilename))
+		b, err := embeddedRootRepo.ReadFile(filepath.Join("repository", mdFilename))
```
asraa, line 219: maybe it's this? https://github.com/golang/go/issues/45230

haydentherapper: Yep, that was it! We have to use `path`.

dlorenc:
Thanks! Nicely done.

Issue #950, opened 2021-10-25
support verification of the signature with multiple public keys


developer-guy:
**Description**

The idea is actually came from here[^1]. We can support verification of the signature with multiple public keys, I couldn't think over the design much yet but, at the end of the day it might look like the following:

```shell
$ cosign verify --key awskms:///1234abcd-12ab-34cd-56ef-1234567890ab --key hashivault://testkey --key k8s://default/mysecret  <IMAGE>
```

[^1]: https://github.com/kyverno/kyverno/issues/2583

cc: @dentrax @dlorenc @JimBugwadia @hectorj2f

JimBugwadia:
Currently `cosign.VerifySignatures` takes a `signature.Verifier` and loops through all signatures and invokes the verifier.  Can this be refactored to push multiple signatures to a new abstraction that performs the logic to verify the entire set against policies or other declarations? 

Something like this:

```golang

type Verifier interface {
	 VerifySignatures(signatures []oci.Signature, opts ...signature.VerifyOption) error
}

type MultiKeyVerifier struct {
	// the number of verifiers that must be applied (defaults to 1)
	// - all: min = len(verifiers)
	// - any: min = 1
	// - count (e.g. 2/5): min = count
	min int
	// the list of keys to attempt
	verifiers []Verifier
}

func (mv *MultiKeyVerifier) VerifySignatures(signatures []oci.Signature, opts ...signature.VerifyOption) error {
	// todo - verify signatures based on signature sets
	return nil
}

type CueVerifier struct {
  // todo - declare or construct a Cue policy
}

func (mv *CueVerifier ) VerifySignatures(signatures []oci.Signature, opts ...signature.VerifyOption) error {
	// todo - verify signatures based on Cue policy
        return nil
}

```

We can then support different verifiers e.g. CueVerifier, MultiKeyVerifier, SingleKeyVerifier, etc. that perform the necessary checks and pass or fail the entire signature set or attestation set.

Dentrax:
Kind ping here 🙏

houdini91:
Issue/PR may be relevant.
[https://github.com/secure-systems-lab/go-securesystemslib/issues/4](url)
https://github.com/secure-systems-lab/go-securesystemslib/pull/7

JimBugwadia:
Hi @houdini91 @Dentrax  - is the plan to use the `MultiEnvelopeSigner` in `cosign.VerifySignatures`?

developer-guy:
[...]

PR #1287, opened 2022-01-07
Fix a couple bugs in cert verification for blobs

priyawadhwa:
1. We weren't checking the expiry of the cert against the time the entry was added to rekor
2. We weren't verifying the cert passed in by `--cert`

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
 	"flag"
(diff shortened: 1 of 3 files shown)


cmd/cosign/cli/verify/verify.go
```diff
@@ -132,7 +133,12 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 			return errors.Wrap(err, "initializing piv token verifier")
 		}
 	case certRef != "":
-		pubKey, err = loadCertFromFileOrURL(c.CertRef)
+		certs, err := loadCertFromFileOrURL(c.CertRef)
+		if err != nil {
+			return err
+		}
+		cert := certs[0]
```
haydentherapper, line 140: In loadCertFromFileOrURL, should we enforce that only one cert was provided in the reference? Alternatively, outside of the scope of this PR, should we support attempting verification with multiple trusted certs/keys?

pkg/cosign/verify.go
```diff
@@ -537,7 +537,7 @@ func verifyImageAttestations(ctx context.Context, atts oci.Signatures, h v1.Hash
 	return checkedAttestations, bundleVerified, nil
 }
 
-func checkExpiry(cert *x509.Certificate, it time.Time) error {
+func CheckExpiry(cert *x509.Certificate, it time.Time) error {
```
haydentherapper, line 540: Can you add a comment since this is a public function?

cmd/cosign/cli/verify/verify.go
priyawadhwa: i think for now we only use one cert, so we could enforce that. i can make that change.

ref https://github.com/sigstore/cosign/issues/950 for discussion about verification w/ multiple keys!

dlorenc (approved):

Issue #1298, opened 2022-01-11
Support for Signing String Literals


medenzon:
**Description**

Adding support for signing string literals. This is similarly to sign-blob, except it allows for a string to be passed in instead of being read from a file.

Example:

```shell
$ cosign sign-string --key cosign.key "hello world"
Enter password for private key: MEUCIB2yVuhvuwKLH4AiodNzvxVOvsKVygyYXZpa8R7kUHQiAiEAuuBTC2rJF3DkbT+LH4sZUMkO17LWpvYC/7bhdOG1Hy4=
```

```shell
$ cosign verify-string --key cosign.pub --signature MEUCIB2yVuhvuwKLH4AiodNzvxVOvsKVygyYXZpa8R7kUHQiAiEAuuBTC2rJF3DkbT+LH4sZUMkO17LWpvYC/7bhdOG1Hy4= "hello world"
Verified OK
```

medenzon:
Please see pull request https://github.com/sigstore/cosign/pull/1299

dlorenc:
Closing as the attached PR was closed!

PR #1299, opened 2022-01-11
added string literal signing functionality

medenzon:
#### Summary

Adding support for signing string literals. This is similar to blob signing, except it allows for a string to be passed in instead of being read from a file.

```shell
$ cosign sign-string --key cosign.key "hello world"
Enter password for private key: MEUCIB2yVuhvuwKLH4AiodNzvxVOvsKVygyYXZpa8R7kUHQiAiEAuuBTC2rJF3DkbT+LH4sZUMkO17LWpvYC/7bhdOG1Hy4=
```

```shell
$ cosign verify-string --key cosign.pub --signature MEUCIB2yVuhvuwKLH4AiodNzvxVOvsKVygyYXZpa8R7kUHQiAiEAuuBTC2rJF3DkbT+LH4sZUMkO17LWpvYC/7bhdOG1Hy4= "hello world"
Verified OK
```

#### Ticket Link

https://github.com/sigstore/cosign/issues/1298

Fixes

#### Release Note

```release-note
* Add functionality to sign string literals, using `cosign sign-string`
```

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -46,3 +46,15 @@ The following checks were performed on each of these signatures:
   - Any certificates were verified against the Fulcio roots.
 {"critical":{"identity":{"docker-reference":"us.gcr.io/dlorenc-vmtest2/demo"},"image":{"docker-manifest-digest":"sha256:124e1fdee94fe5c5f902bc94da2d6e2fea243934c74e76c2368acdc8d3ac7155"},"type":"cosign container image signature"},"optional":null}
 ```
+
+## Sign Strings
+
+```shell
+$ cosign sign-string --key cosign.key "hello world"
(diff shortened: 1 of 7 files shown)


dlorenc:
Thanks for the PR!

Could you explain the use-case a bit? Is there a reason something like sign-blob with a process substitution wouldn't work?

`cosign sign-blob <(echo -n string)`

medenzon:
@dlorenc Thats' my bad, I didnt' know it could be used like that. Thanks for posting the example, it does fit my use case. I'll close the PR, thanks for reviewing.

medenzon:
@dlorenc sorry for the delay -- we tried doing this but it writes to a temp file and doesn't handle as a string literal. do you have any insight into how we can do this without my PR?

Issue #549, opened 2021-08-18
support for other private key type


shimish2:
**Question**
Does cosign supports signing images using key that does not have its type "ENCRYPTED COSIGN PRIVATE KEY".

dlorenc:
What type of key are you trying to use?

Cosign supports several KMS systems which can work with a handful of different keys types. For local, on disk keys, we only support the PEM type you mentioned, because there is no other single widely used format that we'd recommend.

Cosign can also generate payloads that can be signed by other tools, and then upload those same signatures to a registry. This can be used to integrate with other tools and formats not natively supported in cosign.

shimish2:
I am using normal PKI key generated using openssl.

dlorenc:
Ah, thanks. Does it begin with BEGIN PRIVATE KEY or BEGIN ENCRYPTED PRIVATE KEY ?

I think we could add a cosign import key command to generate a cosign key from an openssl one. The encryption on the openssl keys isn't great, so I don't like the idea of encouraging the usage of them directly in cosign sign, but an import for compatibility makes sense.

dlorenc:
Would you be interested in contributing this? I could help out and walk you through it if you're interested!

shimish2:
It begins with BEGIN RSA PRIVATE KEY, I would be happy to contribute this.

dlorenc:
Great! Let me know if you need any pointers to get started. I'll be available on slack tomorrow!

dlorenc:
Import was done, so this is complete!

PR #1300, opened 2022-01-11
Add support for importing PKCS#8 private keys, and add validation

haydentherapper:
#### Summary

This adds support for PKCS#8 encoded private keys, which means
we can also import ED25519 keys. I've added tests for PKCS#8
RSA and ECDSA keys too.

I also added some validation of keys before importing. For RSA,
we will require that the key size be between 2048 and 4096.
For ECDSA keys, we will only disallow NIST P-224, since Cosign
generates P-256 by default. Other curves are not supported
by Go's crypto library.

#### Ticket Link

#549

#### Release Note

```release-note
* Added support for importing PKCS#8 private keys
* Added validation of imported private keys
```

### DIFF
--- a/pkg/cosign/keys.go
+++ b/pkg/cosign/keys.go
@@ -18,6 +18,7 @@ package cosign
 import (
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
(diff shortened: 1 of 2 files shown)


haydentherapper:
cc @venafi-iw

haydentherapper:
It's up for debate if we want to have an upper bound on the size of the RSA key, but larger than 4096 gets a bit unwieldy, requiring more storage space and some verification libraries can't handle large RSA keys.

dlorenc:
This looks awesome!

dlorenc (approved):

PR #1306, opened 2022-01-13
Add --bundle flag to sign-blob and verify-blob

### LINKED ISSUE #1193: Specify and implement on-disk "verification bundle" format
Right now we stash a bunch of information for verifying signatures (fixed-key and keyless) in OCI images. This includes fields like:

* the signature itself
* the signed payload
* the certificate (and chain)
* the transparency log index/uuid/inclusion proof
* a signed timestamp

Full info is here: https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md

This works great for OCI stuff (`cosign sign`) but doesn't really work for blobs. Right now users that sign blobs have to keep track of all of these things on their own.

We should try to standardize a single "bundle" for all of this stuff on disk. It could be as simple as just the JSON info we stick into OCI, but there might be better options. JSON is a necessary prereq for working with OCI so relying on it there isn't a big deal, but for blobs maybe we could avoid a JSON dependency with something simpler.

The overall flow could be something like:

`cosign sign-blob <blob> > bundle.sig`

`cosign verify-blob <blob> --bundle bundle.sig`

Where this would contain enough information for a keyless verification and offline transparency log lookup.

cc @luhring

priyawadhwa:
Add in `--bundle` flag to `cosign sign-blob` and `cosign verify-blob`. The bundle stores everything required to verify a blob in a file, so now you can do something like this:

```
cosign sign-blob --bundle bundle.sig <blob> 

cosign verify-blob --bundle bundle.sig <blob> 
```

The bundle is basically just JSON with the Signature, Cert, and RekorBundle included.

Some thoughts I had, which might be future work:
- For now I put everything behind the `--bundle` flag, but later we might want `sign-blob` to print out the entire bundle instead of just the signature by default
- Right now other flags don't effect the output of `--bundle`, but later we might want certain flags to work together (e.g. `--b64`)
- Since we already have the concept of a "rekor bundle" maybe we should call this something else?

fixes #1193 

```release-note
Add --bundle flag to sign-blob and verify-blob
```


haydentherapper:
For https://github.com/sigstore/cosign/pull/1294/, we're going to include a TUF timestamp in the OCI signature. Will this automatically pick this up, or will we need to append the timestamp to the bundle?

dlorenc:
> * For now I put everything behind the `--bundle` flag, but later we might want `sign-blob` to print out the entire bundle instead of just the signature by default

I'd be fine adding it by default now - this is only relevant for the `COSIGN_EXPERIMENTAL sign-blob` workflow right?

priyawadhwa:
@haydentherapper I think it'll need to be added in, I can update that!

@dlorenc it should work for all workflows right now (for the regular workflow it'll just store the signature in the file and nothing else). Should we only do it for experimental right now?

PR #1308, opened 2022-01-13
Add flag to verify OIDC issuer in certificate 

haydentherapper:
#### Summary

This PR adds support for --cert-oidc-issuer to verify the OIDC issuer of a Fulcio certificate. In combination with --cert-email, users should be able to verify and pin the expected identity of the Fulcio certificate.

Also added certificate generation test utilities. These might be more useful in sigstore/sigstore.

#### Ticket Link

Ref #556

#### Release Note

```release-note
* Added `--cert-oidc-issuer` to verify the OIDC issuer of a Fulcio certificate
```


dlorenc:
Nice!

@mattmoor FYI for the policy task force

mattmoor:
Awesome.  Looks like it needs a rebase

cmd/cosign/cli/options/certificate.go
```diff
@@ -0,0 +1,40 @@
...
+
+var _ Interface = (*RekorOptions)(nil)
+
+// AddFlags implements Interface
+func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.Cert, "cert", "",
+		"path to the public certificate")
+
+	cmd.Flags().StringVar(&o.CertEmail, "cert-email", "",
+		"the email expected in a valid Fulcio certificate")
```
mattmoor, line 36: We should consider changing this since only a handful of things use email, and the rest use URIs.

Doesn't have to be here (probably shouldn't be), but something related that's been bothering me.

cmd/cosign/cli/options/certificate.go
```diff
@@ -0,0 +1,40 @@
...
+
+	cmd.Flags().StringVar(&o.CertOidcIssuer, "cert-oidc-issuer", "",
+		"the OIDC issuer expected in a valid Fulcio certificate, e.g. https://accounts.google.com or https://oauth2.sigstore.dev/auth")
```
mattmoor, line 39: I'd probably bias towards Github as an example here because I believe it's disproportionately the biggest issuer in our logs now.

mattmoor (commented):
This is great to see 👍

cmd/cosign/cli/options/certificate.go
mattmoor: A lot of that info is in extensions already (I think you added it!), and I think this is where it may make more sense to pass some payload with all the data to something like Cue/Rego for policy may be more manageable.

haydentherapper: Left a comment in the bug, I think a cert subject flag would be great to handle both email and URIs.

Given that there is ongoing work for policies (correct?), I agree that we should probably avoid adding complex policy checks to flags, and keep flag-based checks simple.

cmd/cosign/cli/options/certificate.go
haydentherapper: Good to know! Changed the first to `https://token.actions.githubusercontent.com`

haydentherapper:
Rebased and updated with flag doc change. I also removed `x509.ExtKeyUsage(x509.KeyUsageDigitalSignature)` from the list of EKUs used during cert verification, because digital signature is not an extended key usage.

dlorenc (approved):

Issue #1289, opened 2022-01-07
TUF: Support option to reference mirror during verification/signing


haydentherapper:
**Description**

Currently, `cosign initialize` is the only way to update BYO TUF metadata. If the TUF metadata is out of date on verification/signing, then cosign will fetch the updated TUF metadata from the [hosted GCS bucket](https://github.com/sigstore/cosign/blob/main/pkg/cosign/tuf/client.go#L61).

One option would be to move `mirror` and `root` to global flags, so that these could be used with any `cosign` command. The TUF client would need to be refactored to look for these flags when updating TUF metadata.

#1288 improves UX for this issue, so that the flag values don't have to be passed with each invocation.

cc @asraa

asraa:
Just adding some more context:

> then cosign will fetch the updated TUF metadata from the hosted GCS bucket.

This is because of the default fallback. It's kind of questionable on whether we want to fallback gracefully or fail when your cache is expired or corrupt

haydentherapper:
Adding support for this will also allow cosign to support BYO TUF with no disk writes (`SIGSTORE_NO_CACHE`), since the TUF client would use the mirror to fetch TUF metadata/targets, and then those would be persisted in-memory.

rgerganov:
It'd be also nice if cosign is keeping the mirror URL which is passed on `cosign initialize` and tries to update from it when the local cache has expired.

Issue #1293, opened 2022-01-10
TUF: Initializing cosign with expired TUF metadata


asraa:
**Description**

Just requesting better error handling here... if you run `cosign initialize` with a TUF repo that has expired metadata (e.g. the timestamp expires), then you'll panic on a following `cosign verify` since none of the fulcio roots could be used. We need better error handling, 

(1) Early error handling: warn a user when initializing a TUF repo with expired metadata
(2) This problem could happen in the interim as well, initialize valid, then the cached metadata expires. There's no current good fallback to the embedded root or warning that the cache is expired. I think both (1) and a cache warning is needed. I'll save a fallback discussion for https://github.com/sigstore/cosign/issues/1289

Issue #1224, opened 2021-12-15
TUF: Support fallback on trusted Sigstore target for custom roots.


asraa:
**Description**

* This came up when I was testing and demo-ing a BYO TUF root. Currently, we require you sign off on all targets that are needed for verification, including Rekor pubkey, Fulcio root CA, CTFE key. If a project wants to only host their own rekor instance, they will still want to fallback on Sigstore's trusted Fulcio instance.

haydentherapper:
Could we also call this support for multiple TUF roots?

I think we'd want a mode where there's never fallback, if someone never wants to trust the hosted infrastructure. If someone wanted to use the fallback, they could initialize cosign with their custom root and Sigstore root. 

On verification, we'd have to update all expired caches, unless there's some mapping between a signature and its root.

rgerganov:
> I think we'd want a mode where there's never fallback, if someone never wants to trust the hosted infrastructure. If someone wanted to use the fallback, they could initialize cosign with their custom root and Sigstore root.
> 

This is possible now. You can create your own TUF root and then initialize cosign like this:
`cosign initialize --mirror http://localhost:8080/repository --root /tmp/tuf/repository/root.json`

haydentherapper:
That would be for hosting a separate root, correct? I was suggesting we add support for multiple roots, and we don't treat the Sigstore root differently than any other root (besides maybe a special flag to initialize the cosign root). For example, supporting something like:

`cosign initialize --root /tmp/tuf/repository/root.json<delimiter>http://localhost:8080/repository --root /other/root.json<delimiter>http://website:8080/repository --root cosign-root-and-mirror

asraa:
> I was suggesting we add support for multiple roots,

Why would you want multiple TUF roots? You want to combine support for verifying against different roots without having to run cosign initialize between verifications?

haydentherapper:
Yep, exactly. Say you want to trust what's signed by the Sigstore root and also some private root too.

PR #1315, opened 2022-01-13
Cache the location of the remote repository when running cosign initialize

asraa:

#### Summary

Caches the remote repository location so that updates will pull from the cached remote instead of defaulting to the sigstore GCS bucket. This allows clients to automatically pull updates from their custom root when their root is expired and avoids a local/remote mismatch.

#### Ticket Link

This fixes https://github.com/sigstore/cosign/issues/1289 per @rgerganov and https://github.com/sigstore/cosign/issues/1293

#### Release Note

```release-note

```

### DIFF
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -77,16 +77,20 @@ func NewFromEnv(ctx context.Context) (*TUF, error) {
 	// We need to update our tufdb.
 	trustedRoot, err := getRoot(trustedMeta)
 	if err != nil {
+		t.local.Close()
 		return nil, errors.Wrap(err, "getting trusted root")
 	}
 	rootKeys, rootThreshold, err := getRootKeys(trustedRoot)
 	if err != nil {
+		t.local.Close()
 		return nil, errors.Wrap(err, "bad trusted root")
 	}
 	if err := t.client.Init(rootKeys, rootThreshold); err != nil {
+		t.local.Close()
 		return nil, errors.Wrap(err, "unable to initialize client, local cache may be corrupt")
 	}
 	if err := t.updateMetadataAndDownloadTargets(); err != nil {
+		t.local.Close()
 		return nil, errors.Wrap(err, "updating local metadata and targets")
 	}
 
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -442,10 +442,13 @@ func newTuf(ctx context.Context) (*TUF, error) {
 	// default remote root.
 	mirror := DefaultRemoteRoot
 	b, err := os.ReadFile(cachedRemote(rootCacheDir()))
-	remoteInfo := remoteCache{}
-	if err := json.Unmarshal(b, &remoteInfo); err == nil {
-		mirror = remoteInfo.Mirror
+	if err == nil {
+		remoteInfo := remoteCache{}
+		if err := json.Unmarshal(b, &remoteInfo); err == nil {
+			mirror = remoteInfo.Mirror
+		}
 	}
+
 	remote, err := remoteFromMirror(ctx, mirror)
 	if err != nil {
 		return nil, err
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -50,6 +50,12 @@ type TUF struct {
 	remote  client.RemoteStore
 }
 
+// RemoteCache contains information to cache on the location of the remote
+// repository.
+type remoteCache struct {
+	Mirror string `json:"mirror"`
+}
+
 // Close closes the local TUF store. Should only be called once per client.
 func (t *TUF) Close() error {
 	return t.local.Close()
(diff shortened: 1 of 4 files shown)


pkg/cosign/tuf/client_test.go
```diff
@@ -51,12 +51,7 @@ func TestNewFromEnv(t *testing.T) {
 	tuf.Close()
 	checkTargetsAndMeta(t, tuf)
 
-	// Now let's explicitly make a root.
-	remote, err := GcsRemoteStore(ctx, DefaultRemoteRoot, nil, nil)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := Initialize(remote, nil); err != nil {
+	if err := Initialize(ctx, DefaultRemoteRoot, nil); err != nil {
```
haydentherapper, line 54: Can we update the test to check that we've written remote.txt?

pkg/cosign/tuf/client.go
```diff
@@ -327,6 +297,10 @@ func rootCacheDir() string {
 	return rootDir
 }
 
+func cachedRemote(cacheRoot string) string {
+	return filepath.Join(cacheRoot, "remote.txt")
```
haydentherapper, line 301: Do you think we should write serialized structured data to this instead of just a string? Then we can easily append to it in the future.

asraa: Yeah, I was kind of wondering whether we need something like a JSON blob here. I didn't want to overcomplicate it though. The only thing I could think of was multiple mirrors? WDYT?

haydentherapper: Yea, that would enable support for multiple roots of trust like I had mentioned here - https://github.com/sigstore/cosign/issues/1224#issuecomment-998161822

I think making the change to JSON with support for multiple mirrors (even if we just read only the first mirror for now, with a TODO for support for multiple mirrors in another PR) sounds good.

pkg/cosign/tuf/client_test.go
asraa: yeah! actually gonna add a new test of custom root, that would help some later on tests anyway

asraa: added a test that actually uses the cached remote.txt location to pull more metadata!

asraa:
addressed the JSON comment and added a test! @haydentherapper

pkg/cosign/tuf/client.go
```diff
@@ -114,21 +81,22 @@ func New(ctx context.Context, remote client.RemoteStore, cacheRoot string) (*TUF
 	}
 
 	// We need to update our tufdb.
-	// Warning: If a local cache already exists, you may get a local/remote mismatch
-	// since the default remote may not match the remote repository configured during
-	// a cosign initialize.
 	trustedRoot, err := getRoot(trustedMeta)
 	if err != nil {
+		t.local.Close()
```
haydentherapper, line 86: nit: You can call `t.Close()`, which currently just wraps `t.local.Close()`, but maybe we'll need to extend it at some point

pkg/cosign/tuf/client.go
```diff
@@ -46,6 +47,13 @@ type TUF struct {
 	client  *client.Client
 	targets targetImpl
 	local   client.LocalStore
+	remote  client.RemoteStore
+}
+
+// RemoteCache contains information to cache on the location of the remote
+// repository.
+type remoteCache struct {
+	Mirror string `json:"mirror"`
```
haydentherapper, line 56: To avoid exporting, should we use lowercase `mirror`? Since `remoteCache` isn't exported, we shouldn't need to export the struct variable

pkg/cosign/tuf/client_test.go
```diff
@@ -132,6 +138,68 @@ func TestCache(t *testing.T) {
...
+	if _, err = NewFromEnv(ctx); err == nil {
+		t.Errorf("expected expired timestamp from the remote")
+	}
+	expCleanup()
+
+	// Update remote targets, issue a timestamp v2.
+	updateTufRepo(t, td, r, "foo1")
+
+	// Force expiration on the first timestamp.
+	expCleanup = forceExpirationVersion(1, false)
```
haydentherapper, line 190: To clarify, why do we need to force the expiration again for version 1, after it's been done on line 180?

haydentherapper (approved):
Looks great! The tests look fantastic!

pkg/cosign/tuf/client_test.go
asraa: There's two places where we're forcing verification to fail -- once in cosign when we check specifically for the timestamp, and once internally in go-tuf. The first call disables both so that internal go-tuf actually fails when calling `Update()` (otherwise `NewFromEnv` will enventually pass -- fails on our check, then passes with the Update). 

In the second call, we want to make cosign think it needs to fetch an update because the timestamp v1 is expired, but let Update fetch all the latest data from remote without forcing failure.

AS I WRITE THIS: I realize I should just split this up to to only cleanup the internal go-tuf force failure. Thanks :)

pkg/cosign/tuf/client.go
asraa: If it's not exported json won't serialize it :/

haydentherapper: Ah that's unfortunate, oh well

haydentherapper (approved):
LGTM!

dlorenc (approved):

Issue #838, opened 2021-10-04
Support GitHub and GitLab services for generate-key-pair


Dentrax:
**Description**

Consider the following example:
```
$ cosign generate-key-pair -kms github://repository/project
$ cosign generate-key-pair -kms gitlab://repository/project
```

What these providers going to do is that simply calling `secret` API to _create_ or _update_ the generated _private key_ in the secret variables. Something like: `[UPSERT] /repos/:owner/:repo/actions/secrets` into `$COSIGN_PRIVATE_KEY` variable.

GitHub: https://docs.github.com/en/rest/reference/actions#secrets
GitLab: https://docs.gitlab.com/ee/api/project_level_variables.html

**P.S:** I am not so sure if we should call `GitHub` and `GitLab` as a KMS provider. What about `-git` flag?

^ @developer-guy @erkanzileli 🤗

dlorenc:
I like it!

hectorj2f:
+1 😄

PR #1317, opened 2022-01-14
fix: add public key encryption to github secrets

gcbmykola:
Current cosign github:// URI reference key generation feature
is not using the repository public key for secrets encryption.
The key reference is only added for Github secrets API request,
the actual secret is provisioned in clear text, is not encrypted
before it reaches the GitHub.

As a result, when the secret is decrypted when used in a codespace
(for example by Github Action) latter fails silently, and the
resulting secret value is empty.

Fix use libsodium sealed box to encrypt secrets before they reach
GitHub as per API guidelines.

Manual test:
Before fix
1) $cosign generate-key-pair github://[OWNER]/[PROJECT_NAME]
2) Run action that uses COSIGN_PASSWORD, COSIGN_PRIVATE_KEY secrets
Actual result
Secret sizes are equal to 0

After fix
1) $cosign generate-key-pair github://[OWNER]/[PROJECT_NAME]
2) Run action that uses COSIGN_PASSWORD, COSIGN_PRIVATE_KEY secrets
Actual result
Secret sizes are as provisioned by cosign

Fixes: #838

Signed-off-by: Mykola Kondratenko <mykola.kondratenko@qlik.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ fmt: ## Format all go files
 	goimports -w $(GOFILES)
 
 cosign: $(SRCS)
-	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" -o $@ ./cmd/cosign
+	CGO_ENABLED=1 go build -trimpath -ldflags "$(LDFLAGS)" -o $@ ./cmd/cosign
 
 cosign-pivkey-pkcs11key: $(SRCS)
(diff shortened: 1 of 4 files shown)


dlorenc:
cc @developer-guy can you take a look here?

developer-guy:
Not yet but it is on my list 😍 but I like the idea 👏👍

PR #1320, opened 2022-01-14
Fix TestSignBlobBundle

priyawadhwa:
There must have been a race condition between merging https://github.com/sigstore/cosign/pull/1306 and an extra arg being added to `VerifyBlobCmd` -- this should fix the failing test.


bobcallaway (approved):
lgtm

PR #1325, opened 2022-01-15
Enable CodeQL runs on PRs and weekly

### LINKED ISSUE #1324: enable CodeQL on all PRs, not just code on main
Tracking issue for:
- [ ] https://github.com/sigstore/cosign/security/code-scanning/67

bobcallaway:
Fixes #1324

Signed-off-by: Bob Callaway <bob.callaway@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```


dlorenc:
So we need it on PR runs as well as the cron? It takes an incredibly long time and almost never fails so I've been trying to leave it for post-submit only.

I think we've only ever had one failure here and it was a false positive.

.github/workflows/codeql-analysis.yml
```diff
@@ -34,10 +38,31 @@ jobs:
...
+          ~/go/pkg/mod
+          ~/.cache/go-build
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+
+    - name: Setup Go Version
+      uses: actions/setup-go@424fc82d43fa5a37540bae62709ddcc23d9520d4 #v2.1.5
+      with:
+        go-version: '1.16.x'
```
cpanato, line 64: why 1.16 and not 1.17?

cpanato:
> So we need it on PR runs as well as the cron? It takes an incredibly long time and almost never fails so I've been trying to leave it for post-submit only.
> 
> I think we've only ever had one failure here and it was a false positive.

i agree, a post-job is good and this does not hurt much us

.github/workflows/codeql-analysis.yml
cpanato: @bobcallaway we are using go1.18, can you update to that?

cpanato (changes requested):

.github/workflows/codeql-analysis.yml
cpanato: now maybe 1.19 :)

hectorj2f:
I've seen PR checks to slow down a lot, we need to measure how much time, it will take. Otherwise I'll suggest to run them post push.

PR #568, opened 2021-08-24
update Go to 1.17.0

dekkagaijin:
Signed-off-by: Jake Sanders <jsand@google.com>

PR #1340, opened 2022-01-19
Bump recommended Go development version in README

znewman01:
#### Summary

[#568](https://github.com/sigstore/cosign/pull/568) bumps the Go version for tests to 1.17. Now they break for 1.16:

```sh
$ TMPDIR=$(mktemp -d)
$ cd $TMPDIR
$ git clone https://github.com/sigstore/cosign 2> /dev/null
$ cd cosign
$ for go in go_1_16 go_1_17; do
>   nix-shell \
>     -p ${go} \
>     -p gnumake \
>     --command 'make test' \
>     > /dev/null 2>&1
>   if [ $? -eq 0 ]; then
>     echo ${go} good
>   else
>     echo ${go} bad
>   fi
> done
go_1_16 bad
go_1_17 good
```

[#1252](https://github.com/sigstore/cosign/pull/1252) adds a call to `testing.T.Setenv`, added in [1.17](https://go-review.googlesource.com/c/go/+/326790/3/api/go1.17.txt), which caused this breakage.

(We're still testing *builds* on 1.16, so that's fine.)

#### Ticket Link

Fixes N/A

#### Release Note

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ For Linux and macOS binaries see the [GitHub release assets](https://github.com/
 
 ## Developer Installation
 
-If you have Go 1.16+, you can setup a development environment:
+If you have Go 1.17+, you can setup a development environment:
 
     $ git clone https://github.com/sigstore/cosign
     $ cd cosign


README.md
```diff
@@ -30,7 +30,7 @@ For Linux and macOS binaries see the [GitHub release assets](https://github.com/
 
 ## Developer Installation
 
-If you have Go 1.16+, you can setup a development environment:
+If you have Go 1.17+, you can setup a development environment:
```
cpanato, line 33: if we bump here in the readme, we should change the go module as well

cc @mattmoor @dlorenc

znewman01: I'm not an expert on Go modules but AFAICT the [Go version in `go.mod`](https://go.dev/ref/mod#go-mod-file-go) has 2 effects:

1. Restricts language features introduced after that version.
2. Changes the behavior of the `go` command. In particular, for the 1.16->1.17 transition the consequential changes:
    - starts including explicit `requires` in `go.mod` for transitive dependencies
    - vendoring acts a little different (starts recording dependencies' Go versions from `go.mod`, omits `go.mod` and `go.sum` for vendored dependencies)

Because as of right now we still want to support 1.16, my initial thought is that (1) is a good thing for development and I'm inclined not to change it. Of course, will defer to you all :)

dlorenc: We chatted about bumping the other one to 1.17 soon, this one seems fine as is and we can do the other one later.

dlorenc (approved):

Issue #1321, opened 2022-01-14
Add DSSE support to verify-attestation


rgreinho:
## Description

As described in [chains issue #329](https://github.com/tektoncd/chains/issues/329), when the signature(s) get stored as a taskrun annotation, it is represented using the [dsse envelope format](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md). For example:

```json
{
  "payloadType": "application/vnd.in-toto+json",
[...]

PR #1355, opened 2022-01-25
Update verify-blob to support DSSEs

rgreinho:
Adds the ability to verify DSSEs from the `verify-blob` command.

Fixes sigstore/cosign#1321

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -23,6 +23,7 @@ import (
 	_ "crypto/sha256" // for `crypto.SHA256`
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -30,6 +31,7 @@ import (
 
 	"github.com/go-openapi/runtime"
 	"github.com/pkg/errors"
+	ssldsse "github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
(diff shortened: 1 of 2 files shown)


dlorenc:
cc @priyawadhwa

rgreinho:
Using the signature and the envelope I put in the [issue](https://github.com/sigstore/cosign/issues/1321#issuecomment-1018678455), I used it like that:

```
$ ./cosign verify-blob --key cosign.pub --signature dsse.sig dsse.sig
Verified OK
```

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -162,9 +166,28 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
 		}
 	}
 
-	// verify the signature
-	if err := verifier.VerifySignature(bytes.NewReader([]byte(sig)), bytes.NewReader(blobBytes)); err != nil {
-		return err
+	// Is DSSE?
+	DSSEenvelope := ssldsse.Envelope{}
+	if err := json.Unmarshal(blobBytes, &DSSEenvelope); err == nil {
```
priyawadhwa, line 171: Could we move this into a separate function and add a unit test for it?

priyawadhwa (commented):
Thanks for adding this @rgreinho ! Just added one comment about unit testing :)

cmd/cosign/cli/verify/verify_blob.go
rgreinho: Ok, I found some help in the sigstore package, no the code is shorter and cleaner.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -162,6 +167,11 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
 		}
 	}
 
+	// Use the DSSE verifier if the payload is a DSSE with the In-Toto format.
+	if isIntotoDSSE, _ := isIntotoDSSE(blobBytes); isIntotoDSSE {
```
priyawadhwa, line 171: let's either check the error here and return it, or change this function to only return a `bool` if the error should be ignored :)

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -93,3 +95,45 @@ func TestSignaturesBundle(t *testing.T) {
...
+	}
+}
+
+func TestIsIntotoDSSEWithBytes(t *testing.T) {
+	tts := []struct {
+		envelope     []byte
+		isIntotoDSSE bool
+	}{
+		{
+			envelope:     []byte("This is no valid"),
```
priyawadhwa, line 128: could you add a test with a regular signature in here just to make sure it still fails?

cmd/cosign/cli/verify/verify_blob.go
rgreinho: Copy that! Addressed ✅

cmd/cosign/cli/verify/verify_blob_test.go
rgreinho: Oki doki. I think I did what you wanted, my apologies if I misunderstood.

priyawadhwa (approved):
LGTM, thanks @rgreinho !

PR #1039, opened 2021-11-12
feat: replace option for same attestation

developer-guy:
Signed-off-by: Furkan <furkan.turkal@trendyol.com>

feat: replace option for same attestation

Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

#### Summary

replace option for same attestation

#### Ticket Link

Fixes #923 

#### Release Note

```release-note
feat: replace option for same attestation
```

Huge thanks tho @imjasonh !! 

cc: @JimBugwadia @dlorenc @cpanato 

ezgif com-gif-maker(5)

Issue #923, opened 2021-10-19
option to replace attestations


JimBugwadia:
**Description**

Provide a way to replace (vs. append) attestations.

Currently the same attestation type can be appended multiple times. A proposal is to support a `--replace` which would change the behavior to replace an attestation of the same type instead of appending it to the existing list of attestations.

Slack discussion at: https://sigstore.slack.com/archives/C01PZKDL4DP/p1633463916305100.

JimBugwadia:
Thanks, @dlorenc! Please assign to me and I will work on a PR.

developer-guy:
amazing thought @JimBugwadia, we (w/@dentrax @erkanzileli) are available for a while, so, we can work on this issue if you let us do that 🤩

JimBugwadia:
@developer-guy  please proceed, thanks!

PR #1366, opened 2022-01-26
Fix comparison in replace option for attestation

bburky:
#### Summary
Currently the replace option compares `r.predicateURI == payloadData["payloadType"]` when evaluating each signature:
https://github.com/sigstore/cosign/blob/21e6b806143652331b77c29c62a131fb33a71ae0/pkg/cosign/remote/remote.go#L142

This appears to be comparing the payload's `predicateType` (e.g a URI `https://example.com/predicate` or `cosign.sigstore.dev/attestation/v1` for custom) to the signature's `payloadType` (which is always(?) `application/vnd.in-toto+json`). This appears to be a bug unless I'm misunderstanding something.

Fix this by making the replace option replace all attestations which match the current predicate type. I think this is the intended behavior?

The command line output has been updated to clarify the actions taken, clearly listing all the attestations retained and deleted:
```
$ cosign attest --type https://example.com/predicate --key "$key" --predicate predicate.json --replace "localhost:5000/$image"
Not replacing attestation predicate: https://example.com/aaaa
Replacing attestation predicate: https://example.com/predicate
Replacing attestation predicate: https://example.com/predicate
```

This feature was initially requested in #923 and implemented in #1039.

#### Ticket Link

No ticket.

#### Release Note

```release-note
* Fixed predicate type comparison in replace option of attestation
```

### DIFF
--- a/pkg/cosign/remote/remote.go
+++ b/pkg/cosign/remote/remote.go
@@ -137,30 +137,29 @@ func (r *ro) Replace(signatures oci.Signatures, o oci.Signature) (oci.Signatures
 			return nil, fmt.Errorf("unmarshal payload data: %w", err)
 		}
 
-		var decodedPayload []byte
-		if val, ok := signaturePayload["payload"]; ok {
-			decodedPayload, err = base64.StdEncoding.DecodeString(val.(string))
-			if err != nil {
-				return nil, fmt.Errorf("could not decode 'payload': %w", err)
-			}
-		} else {
-			return nil, fmt.Errorf("could not find 'payload' in payload data")
+		val, ok := signaturePayload["payload"]
+		if !ok {
+		   return nil, fmt.Errorf("could not find 'payload' in payload data")
+		}
+		decodedPayload, err := base64.StdEncoding.DecodeString(val.(string))
+		if err != nil {
+			return nil, fmt.Errorf("could not decode 'payload': %w", err)
 		}
 
 		var payloadData map[string]interface{}
 		if err := json.Unmarshal(decodedPayload, &payloadData); err != nil {
 			return nil, fmt.Errorf("unmarshal payloadData: %w", err)
 		}
-		if val, ok := payloadData["predicateType"]; ok {
-			if r.predicateURI == val {
-				fmt.Fprintln(os.Stderr, "Replacing attestation predicate:", r.predicateURI)
-				continue
-			} else {
-				fmt.Fprintln(os.Stderr, "Not replacing attestation predicate:", val)
-				sigsCopy = append(sigsCopy, s)
-			}
-		} else {
+		val, ok = payloadData["predicateType"]
+		if !ok {
 			return nil, fmt.Errorf("could not find 'predicateType' in payload data")
+		}
+		if r.predicateURI == val {
+			fmt.Fprintln(os.Stderr, "Replacing attestation predicate:", r.predicateURI)
+			continue
+		} else {
+			fmt.Fprintln(os.Stderr, "Not replacing attestation predicate:", val)
+			sigsCopy = append(sigsCopy, s)
 		}
 	}
 
--- a/pkg/cosign/remote/remote.go
+++ b/pkg/cosign/remote/remote.go
@@ -139,7 +139,7 @@ func (r *ro) Replace(signatures oci.Signatures, o oci.Signature) (oci.Signatures
 
 		val, ok := signaturePayload["payload"]
 		if !ok {
-		   return nil, fmt.Errorf("could not find 'payload' in payload data")
+			return nil, fmt.Errorf("could not find 'payload' in payload data")
 		}
 		decodedPayload, err := base64.StdEncoding.DecodeString(val.(string))
 		if err != nil {
--- a/pkg/cosign/remote/remote.go
+++ b/pkg/cosign/remote/remote.go
@@ -127,24 +127,40 @@ func (r *ro) Replace(signatures oci.Signatures, o oci.Signature) (oci.Signatures
 	}
 
 	for _, s := range sigs {
-		var payloadData map[string]interface{}
-
+		var signaturePayload map[string]interface{}
 		p, err := s.Payload()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


bburky:
I think this broke with #1248 which changed the DSSE envelope to always be the in-toto media type.

priyawadhwa (commented):
Thanks for fixing this @bburky! Just left a couple nits. It might also be useful to test the `--replace` flag so that this type of bug can't be reintroduced.

I think modifying the existing test would probably work:

https://github.com/sigstore/cosign/blob/d2781b81e9d0bb6149aec4dd99bffee9fe197fc8/test/e2e_test.go#L204

You could try attest a second time with the `--replace` flag and make sure the # of attestations on the image is still 1 (if replacement fails, then there should be an additional attestation, on which the test would fail)

lmk if you run into any issues!

pkg/cosign/remote/remote.go
```diff
@@ -127,24 +127,40 @@ func (r *ro) Replace(signatures oci.Signatures, o oci.Signature) (oci.Signatures
...
+		err = json.Unmarshal(p, &signaturePayload)
 		if err != nil {
 			return nil, fmt.Errorf("unmarshal payload data: %w", err)
 		}
 
-		if r.predicateURI == payloadData["payloadType"] {
-			fmt.Fprintln(os.Stderr, "Attestation already present, not adding new one.")
-			continue
+		var decodedPayload []byte
+		if val, ok := signaturePayload["payload"]; ok {
```
priyawadhwa, line 141: nit: refactoring this a little would be more readable


```go
val, ok := signaturePayload["payload"]
if !ok {
   return nil, fmt.Errorf("could not find 'payload' in payload data")
}
decodedPayload, err = base64.StdEncoding.DecodeString(val.(string))
if err != nil {
	return nil, fmt.Errorf("could not decode 'payload': %w", err)
}


```

pkg/cosign/remote/remote.go
```diff
@@ -127,24 +127,40 @@ func (r *ro) Replace(signatures oci.Signatures, o oci.Signature) (oci.Signatures
...
+			}
+		} else {
+			return nil, fmt.Errorf("could not find 'payload' in payload data")
+		}
+
+		var payloadData map[string]interface{}
+		if err := json.Unmarshal(decodedPayload, &payloadData); err != nil {
+			return nil, fmt.Errorf("unmarshal payloadData: %w", err)
+		}
+		if val, ok := payloadData["predicateType"]; ok {
```
priyawadhwa, line 154: this could be refactored similarly to the above comment!

bburky: `json.Unmarshal()` returns the data by reference, this isn't possible. The other refactoring looks fine though.

bburky:
@priyawadhwa I can't easily add the test to `TestAttestVerify()`. The high level command interfaces of `VerifyAttestationCommand` or `VerifyCommand` only give you an `error` result. I cannot easily get any other output from it unless I redirect stdout and parse JSON. I can do that, but I wanted to see if you had a simpler design in mind. Is there a lower level function I can easily invoke to retrieve an array of attestations?

bburky:
I added an e2e test for the the replace option using `cosign.FetchAttestationsForReference()` to download the attestations.

_Somehow_ this is only failing on [`random.Image()`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/random#Image) generated containers. Somehow this causes _only_ the replace option to not upload any new attestations. It silently fails.

I used skopeo to copy real images to a local registry when testing, the replace option works perfectly fine on them. I can reproduce the buggy behavior on the command line with `--replace` if I upload a `random.Image()` to a local registry, it's not unique to the unit tests.

I honestly have no idea what's wrong. This bug predates my changes, it occurs with the v1.5.0 release `cosign` binary on any image from `random.Image()` when using `--replace`.

test/e2e_test.go
```diff
@@ -254,6 +254,29 @@ func TestAttestVerify(t *testing.T) {
 
 	// Look for a specific annotation
 	mustErr(verify(pubKeyPath, imgName, true, map[string]interface{}{"foo": "bar"}, ""), t)
+
+	// Attest again with replace=true, replacing the previous attestation
+	must(attest.AttestCmd(ctx, ko, options.RegistryOptions{}, imgName, "", false, slsaAttestationPath, false,
+		"slsaprovenance", true, ftime.Duration(30*time.Second)), t)
```
priyawadhwa, line 260: After attesting the first time with `replace=true`, let's just make sure there's 1 attestation. that should be enough for this test right now!

priyawadhwa (commented):
Thanks for trying to add the test @bburky ! I left a comment for simplifying the test for now, & if you wouldn't mind maybe you could open an issue for the bug you found.

test/e2e_test.go
bburky: Ok, I removed it. The second test did allow demonstrating the bug I mentioned though. Currently this test is incorrectly passing because replace=true is never working on `random.Image()` images.  See #1378

If you're fine with this test incorrectly passing and tracking the bug in #1378, then I think we're fine.

priyawadhwa: Sorry to change my mind on this, but that's a good point! Maybe we leave your test in and just skip it for now, and then we can link it in the issue. Whoever picks up the issue already has the test built in :)

bburky: Done.

bburky: I'm not sure if there needs to be anything done to skip the test? Let me know.

Please test this PR manually on the command line considering there's only sort of a usable test.

priyawadhwa: Actually let's copy the test into `TestAttestationReplace`, and then you can throw in a `t.Skip()` on the first line for now.

dlorenc:
Looks like DCO got dropped somehow, you'll need to add that back.

bburky:
I think everything should be ready now.

priyawadhwa (approved):
thanks @bburky !

PR #1394, opened 2022-02-02
[WIP] proposal for new oidc and fulcio abstractions

dekkagaijin:
Re: https://github.com/sigstore/cosign/issues/844
PTAL @mattmoor

### DIFF
--- a/internal/pkg/cosign/fulcio/attestor.go
+++ b/internal/pkg/cosign/fulcio/attestor.go
@@ -1,0 +1,101 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fulcio
+
+import (
+	"bytes"
+	"context"
+	"crypto"
+	"encoding/base64"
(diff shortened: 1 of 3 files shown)


internal/pkg/cosign/fulcio/attestor.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return att, pk, nil
+}
+
+// NewDSSEAttestor returns a `cosign.DSSEAttestor`
+func NewDSSEAttestor(payloadType string,
+	s signature.Signer, oidp oidc.Provider) cosign.DSSEAttestor {
```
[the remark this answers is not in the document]
dekkagaijin: so that one can use a non-ephemeral key, or an ephemeral key generated elsewhere, and associate it with an OID

internal/pkg/cosign/fulcio/client.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// *NOTE* This is a higher-level client than exists in `sigstore/fulcio`, should probably live there
+
+// Client is a Fulcio client.
+type Client struct {
+	APIClient fulcioapi.Client
+}
+
+// GetCert retrieves a Fulcio certificate which associates the `Signer`'s public key with the ID provided
```
mattmoor, line 36: It would be tremendously helpful if the comments were actually useful in describing what you were trying to accomplish with the higher level client.

It seems like your goal is to capture some of the challenge semantics (e.g. proving we have the private key by signing the subject of the IDToken), but TBQH the objectives of this as a higher level client aren't clear to me at all, which makes it sort of hard to judge good or bad.

dekkagaijin: fair enough. The higher level client I was thinking of uses "real" types, e.g. `x509.Certificate` instead of serialized `[]byte`s, and performs the boilerplate operations to get the ID token, sign the subject, and retrieve a signing cert from Fulcio.

PR #1413, opened 2022-02-07
use the upstream kubernetes version lib and ldflags

n3wscott:
#### Summary

See  https://github.com/kubernetes-sigs/release-utils/pull/28

This leverages the upstream k8s release-utils version strings and output, so we can drop it locally.

#### Ticket Link

Relates to https://github.com/kubernetes-sigs/release-utils/issues/23

#### Release Note
```release-note
NONE
```


cpanato:
> cc @cpanato, i like the code deletion but it deletes his really cool ASCII logo!

:( true

but this is not adding another external dependency to the project? 

besides the fact we loose the starwars logo lol
maybe we can add an improvement in the lib that generate the ASCII :)

n3wscott:
It supports fonts that you can load if you really want that font back.

cpanato:
that is fine, we don't need to be that picky, maybe I can update that when are bored lol

cpanato (approved):

PR #1423, opened 2022-02-08
Fetch verification targets by TUF custom metadata

### LINKED ISSUE #1342: TUF: Use custom metadata to identify TUF target usage
**Description**

Currently, we hardcode TUF targets by name, e.g. `fulcio.crt.pem`. As we rotate logs year by year, we will want to annotate custom metadata about these targets. This will also help users make their own TUF targets without relying on fixed file names.

We should probably include:
* the component the target is used for: (fulcio, rekor, ctfe)
* the dates the target will be valid for or some usage information

c/f this comment https://github.com/sigstore/cosign/issues/1273#issuecomment-1006116223

cc @haydentherapper

haydentherapper:
#### Summary

Continuing #1273, this adds support for fetching a list of TUF targets by custom metadata. This will be used for verification using expired/rotated targets. All targets will be persisted in the TUF target metadata. Only one will be marked as `active` while the others are marked as `expired`.

The custom metadata on a target looks like looks like:
```
      "fulcio.crt.pem": {
        "custom": {
          "sigstore": {
            "status": "Active",
            "usage": "Fulcio"
          }
        },
        "hashes": {
          "sha512": "f2e33a6dc208cee1f51d33bbea675ab0f0ced269617497985f9a0680689ee7073e4b6f8fef64c91bda590d30c129b3070dddce824c05bc165ac9802f0705cab6"
        },
        "length": 740
      },
```

Possible statuses are `active` or `expired`, and possible usages are `fulcio`, `rekor`, and `ctfe`. `unknown` is also supported for each as a default value.

To test this, I:
* Used the current TUF targets to sign and verify. These TUF targets do not contain custom metadata, so this tested the fallback.
* Initialized a TUF repository with custom metadata for each target (based on [the script](https://gist.github.com/asraa/947f1a38afd03af57c7b71d893c36af0) from the BYO TUF blog post). I changed the filenames too so no fallback would happen. Initialized Cosign with the custom TUF metadata. Successfully signed and verified. 

Next steps:
* Use the `Status` of the custom metadata to inform users when verifying with an expired target
* Remove the OCI timestamp code
* After TUF Root v3 is published and bundled with Cosign, we can remove the code that falls back to targets by filename.

#### Ticket Link

Fixes #1342

Ref #1273

#### Release Note

```release-note
TUF targets can now be referenced by custom metadata instead of filename

Multiple verification keys can be used to verify Fulcio certificates, Rekor entries, or Fulcio CT log SCTs
```

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
+++ b/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
@@ -51,26 +51,27 @@ const altCTLogPublicKeyLocation = "SIGSTORE_CT_LOG_PUBLIC_KEY_FILE"
 // The SCT is a `Signed Certificate Timestamp`, which promises that
 // the certificate issued by Fulcio was also added to the public CT log within
 // some defined time period
-func verifySCT(certPEM, rawSCT []byte) error {
+func verifySCT(certPEM, rawSCT []byte, ctx context.Context) error {
 	var pubKeys []crypto.PublicKey
 	rootEnv := os.Getenv(altCTLogPublicKeyLocation)
 	if rootEnv == "" {
-		ctx := context.TODO()
-		tuf, err := tuf.NewFromEnv(ctx)
+		tufClient, err := tuf.NewFromEnv(ctx)
 		if err != nil {
 			return err
 		}
-		defer tuf.Close()
-		ctPub, err := tuf.GetTarget(ctPublicKeyStr)
+		defer tufClient.Close()
+
+		targets, err := tufClient.GetTargetsByMeta(tuf.CTFE, []string{ctPublicKeyStr})
 		if err != nil {
 			return err
 		}
-		// Is there a reason why this must be ECDSA key?
-		pubKey, err := cosign.PemToECDSAKey(ctPub)
-		if err != nil {
-			return errors.Wrap(err, "converting Public CT to ECDSAKey")
+		for _, t := range targets {
+			ctPub, err := cosign.PemToECDSAKey(t.Target)
+			if err != nil {
+				return errors.Wrap(err, "converting Public CT to ECDSAKey")
+			}
+			pubKeys = append(pubKeys, ctPub)
 		}
-		pubKeys = append(pubKeys, pubKey)
 	} else {
 		fmt.Fprintf(os.Stderr, "**Warning** Using a non-standard public key for verifying SCT: %s\n", rootEnv)
 		raw, err := os.ReadFile(rootEnv)
@@ -82,6 +83,9 @@ func verifySCT(certPEM, rawSCT []byte) error {
 			return errors.Wrap(err, "error parsing alternate public key from the file")
 		}
 		pubKeys = append(pubKeys, pubKey)
+	}
+	if len(pubKeys) == 0 {
+		return errors.New("none of the CTFE keys have been found")
 	}
 	cert, err := x509util.CertificateFromPEM(certPEM)
 	if err != nil {
@@ -110,7 +114,7 @@ func NewSigner(ctx context.Context, idToken, oidcIssuer, oidcClientID, oidcClien
 	}
 
 	// verify the sct
-	if err := verifySCT(fs.Cert, fs.SCT); err != nil {
+	if err := verifySCT(fs.Cert, fs.SCT, ctx); err != nil {
 		return nil, errors.Wrap(err, "verifying SCT")
 	}
 	fmt.Fprintln(os.Stderr, "Successfully verified SCT...")
--- a/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
+++ b/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
@@ -51,7 +51,7 @@ const altCTLogPublicKeyLocation = "SIGSTORE_CT_LOG_PUBLIC_KEY_FILE"
 // The SCT is a `Signed Certificate Timestamp`, which promises that
 // the certificate issued by Fulcio was also added to the public CT log within
 // some defined time period
-func verifySCT(certPEM, rawSCT []byte, ctx context.Context) error {
+func verifySCT(ctx context.Context, certPEM, rawSCT []byte) error {
 	var pubKeys []crypto.PublicKey
 	rootEnv := os.Getenv(altCTLogPublicKeyLocation)
 	if rootEnv == "" {
@@ -114,7 +114,7 @@ func NewSigner(ctx context.Context, idToken, oidcIssuer, oidcClientID, oidcClien
 	}
 
 	// verify the sct
-	if err := verifySCT(fs.Cert, fs.SCT, ctx); err != nil {
+	if err := verifySCT(ctx, fs.Cert, fs.SCT); err != nil {
 		return nil, errors.Wrap(err, "verifying SCT")
 	}
 	fmt.Fprintln(os.Stderr, "Successfully verified SCT...")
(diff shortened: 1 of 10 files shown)


haydentherapper:
cc @asraa @dlorenc @bobcallaway

haydentherapper:
Looking at failing test now, this doesn't fail locally (sorting enums for comparing two arrays)

dlorenc:
These test failures look related to the PR, any idea?

haydentherapper:
Reran tests, the unit tests succeeded (I saw the flaky "timestamp.json" off by two bytes), but the KinD e2e tests are failing. Investigating now

haydentherapper:
It appears the tests collided with an ongoing distroless release. It pulled the latest image before signing was complete. No more errors!

dlorenc:
Cc @znewman01 could you take a look?

znewman01:
> could you take a look?

Yup, sometime in the next couple hours.

asraa (changes requested):

pkg/cosign/tuf/client.go
```diff
@@ -247,6 +261,44 @@ func (t *TUF) GetTarget(name string) ([]byte, error) {
 	return targetBytes, nil
 }
 
+// Get target files by a custom usage metadata tag. If there are no files found,
+// use the fallback target names to fetch the targets by name.
+func (t *TUF) GetTargetsByMeta(usage UsageKind, fallbacks []string) ([]TargetFile, error) {
+	targets, err := t.client.Targets()
```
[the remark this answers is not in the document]
haydentherapper: Yep, this is a limitation of the underlying library. Definitely think a `DelegatedTargets` method would be valuable.

pkg/cosign/tuf/client.go
```diff
@@ -247,6 +261,47 @@ func (t *TUF) GetTarget(name string) ([]byte, error) {
...
+		var scm sigstoreCustomMetadata
+		err := json.Unmarshal(*targetMeta.Custom, &scm)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "**Warning** Custom metadata not configured properly for target %s, skipping target\n", name)
+			continue
+		}
+		if scm.Sigstore.Usage == usage {
+			target, err := t.GetTarget(name)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "**Warning** Unable to fetch target %s, skipping target\n", name)
```
asraa, line 286: This feels like a legit error: `GetTargets` gives the string but it fails download? 
Or are you anticipating that maybe there's a difference between delegated targets and better safe than sorry?

pkg/cosign/tuf/client.go
```diff
@@ -247,6 +261,47 @@ func (t *TUF) GetTarget(name string) ([]byte, error) {
...
+			}
+		}
+	}
+	if len(matchedTargets) == 0 {
+		for _, fallback := range fallbacks {
+			target, err := t.GetTarget(fallback)
+			if err != nil {
+				return nil, errors.Wrap(err, "error getting target")
+			}
+			matchedTargets = append(matchedTargets, TargetFile{Target: target, Status: Active})
```
asraa, line 299: should we make this an unknown status?

asraa (changes requested):

pkg/cosign/tuf/client.go
haydentherapper: Yea, I went back and forth on this. I think it's fair to say that by this point, an error would be caused by something out of the user's control and therefore we should fail.
FWIW, we'll remove fallbacks after root v3, so we can make both of these cases errors then.

pkg/cosign/tuf/client.go
haydentherapper: I wanted this to be Active because we trust the fallbacks as active now. Once root v3 is bundled with cosign, we can remove this entirely. Does that sound good?

pkg/cosign/tuf/client_test.go
```diff
@@ -256,6 +256,17 @@ func TestGetTargetsByMeta(t *testing.T) {
 	if targets[0].Status != Active {
 		t.Fatalf("target without custom metadata not active, got: %v", targets[0].Status)
 	}
+	// Fetch multiple targets with no custom metadata.
+	targets, err = tufObj.GetTargetsByMeta(UnknownUsage, []string{"fooNoCustom.txt", "fooNoCustomOther.txt"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(targets) != 2 {
+		t.Fatalf("expected one target without custom metadata, got %d targets", len(targets))
```
znewman01, line 265: s/1/2/

znewman01 (approved):
LGTM

cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
```diff
@@ -51,26 +51,27 @@ const altCTLogPublicKeyLocation = "SIGSTORE_CT_LOG_PUBLIC_KEY_FILE"
...
 		}
-		defer tuf.Close()
-		ctPub, err := tuf.GetTarget(ctPublicKeyStr)
+		defer tufClient.Close()
+
+		targets, err := tufClient.GetTargetsByMeta(tuf.CTFE, []string{ctPublicKeyStr})
 		if err != nil {
 			return err
 		}
-		// Is there a reason why this must be ECDSA key?
```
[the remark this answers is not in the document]
haydentherapper: The code makes assumptions about all of the TUF targets. We should clearly document these assumptions in cosign/root-signing. We could add support for other keys if there was a need, and it might come up with BYO-TUF. I'd encourage others to use ECDSA, primarily because you can have much smaller keys and get the same security strength.

pkg/cosign/tuf/client_test.go
```diff
@@ -206,6 +209,82 @@ func TestCustomRoot(t *testing.T) {
 	tufObj.Close()
 }
 
+func TestGetTargetsByMeta(t *testing.T) {
+	ctx := context.Background()
+	// Create a remote repository.
+	td := t.TempDir()
+	remote, _ := newTufCustomRepo(t, td, "foo")
```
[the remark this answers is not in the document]
haydentherapper: Added a comment on `newTufCustomRepo` for the repo layout.

pkg/cosign/tuf/client_test.go
```diff
@@ -206,6 +209,82 @@ func TestCustomRoot(t *testing.T) {
...
+	}
+	rootBytes, ok := meta["root.json"]
+	if !ok {
+		t.Error(err)
+	}
+	if err := Initialize(ctx, s.URL, rootBytes); err != nil {
+		t.Error(err)
+	}
+	if l := dirLen(t, tufRoot); l == 0 {
+		t.Errorf("expected filesystem writes, got %d entries", l)
```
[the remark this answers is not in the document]
haydentherapper: It's a check to make sure Initialize wrote something to the `tufRoot` directory.

pkg/cosign/tuf/client_test.go
haydentherapper: Thanks!

haydentherapper (commented):
Thanks for the review!

cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
```diff
@@ -51,26 +51,27 @@ const altCTLogPublicKeyLocation = "SIGSTORE_CT_LOG_PUBLIC_KEY_FILE"
 // The SCT is a `Signed Certificate Timestamp`, which promises that
 // the certificate issued by Fulcio was also added to the public CT log within
 // some defined time period
-func verifySCT(certPEM, rawSCT []byte) error {
+func verifySCT(certPEM, rawSCT []byte, ctx context.Context) error {
```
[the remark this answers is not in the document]
haydentherapper: This was likely from an outdated commit, it was fixed in one of the recent commits.

pkg/cosign/tuf/client.go
asraa: sg!

asraa (approved):

PR #1428, opened 2022-02-08
Remove TUF timestamp from OCI signature bundle

haydentherapper:
As described in #1273, this solution does not work
because the TUF root is not included in the snapshot.
Removing unused code.

Confirmed that verifying images with a timestamp
annotation still works. Confimed that signing and
verifying works locally too.

#### Ticket Link

Fixes

#### Release Note

```release-note
Removed unused TUF timestamp from OCI signature
```


haydentherapper:
How to pad your github stats - Add a feature and remove it a week later

haydentherapper:
Rollback of #1274, #1316, and #1294, leaving in tests for signers that were added as part of the PRs.

dlorenc (approved):

PR #1429, opened 2022-02-09
Add docs on API stability and deprecation table

priyawadhwa:
Adding these in preparation for GA. I'm not sure if we have features we're in the process of deprecating, but I can add them in if we do!

These follow the policy described [here](https://docs.sigstore.dev/api-stability).

Ref https://github.com/sigstore/sigstore-website/issues/119

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -1,0 +1,10 @@
+# Deprecations
+
+This doc lists deprecated features in `cosign`.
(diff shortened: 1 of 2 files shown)


dlorenc (approved):

dlorenc:
I think you can be a bit more generous! Most of the key based stuff is GA today. I think it's documented in the readme.

priyawadhwa:
Thanks @dlorenc ! Added in those details :)

dlorenc (approved):

PR #1432, opened 2022-02-09
Improve log lines to match with implementation

marcofranssen:
#### Summary

Updates the printed info shown when using `generate-keypair` for Github and Gitlab providers.

#### Ticket Link

Fixes N.A.

#### Release Note

```release-note
Improve generate-keypair info printing
```

### DIFF
--- a/pkg/cosign/git/github/github.go
+++ b/pkg/cosign/git/github/github.go
@@ -82,7 +82,7 @@ func (g *Gh) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
 
 	passwordSecretEnvResp, err := client.Actions.CreateOrUpdateRepoSecret(ctx, owner, repo, passwordSecretEnv)
 	if err != nil {
-		return errors.Wrap(err, "could not create \"COSIGN_PASSWORD\" environment variable")
+		return errors.Wrap(err, "could not create \"COSIGN_PASSWORD\" github actions secret")
 	}
 
 	if passwordSecretEnvResp.StatusCode < 200 && passwordSecretEnvResp.StatusCode >= 300 {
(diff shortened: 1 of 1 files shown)


pkg/cosign/git/github/github.go
```diff
@@ -82,15 +82,15 @@ func (g *Gh) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
...
+		return errors.Wrap(err, "could not create \"COSIGN_PASSWORD\" github actions secret")
 	}
 
 	if passwordSecretEnvResp.StatusCode < 200 && passwordSecretEnvResp.StatusCode >= 300 {
 		bodyBytes, _ := io.ReadAll(passwordSecretEnvResp.Body)
 		return fmt.Errorf("%s", bodyBytes)
 	}
 
-	fmt.Fprintln(os.Stderr, "Private key written to COSIGN_PASSWORD environment variable")
+	fmt.Fprintln(os.Stdout, "Password written to COSIGN_PASSWORD github actions secret")
```
dlorenc, line 93: Seems like the rest of the passphrase commands print to stderr, any reason for the switch here?

marcofranssen: Because what we are printing are not errors. Now you mention it, I'm thinking what would be the reason these are printed to Stderr?

dlorenc: Typically logging statements go to stderr, while things intended to be machine-readable or passed to other commands via pipes go to stdout. I don't think there's a clear answer for this command, I just think we should keep it consistent.

marcofranssen: Agree, let me revert that change for now.

dlorenc (approved):

PR #1436, opened 2022-02-09
generate and upload release yaml

### LINKED ISSUE #1414: Publish release YAMLs for cosigned
This would make it easier to install directly from CI without needing helm.

cc @n3wscott @mattmoor

k4leung4:
#### Summary

Add workflow that triggers on published release to generate a yaml and upload it to the release for the triggered tag.
The workflow can also be triggered manually that requires the tag to be specified.

#### Ticket Link

Fixes #1414

#### Release Note

```release-note
NONE
```

Signed-off-by: Kenny Leung <kleung@chainguard.dev>

### DIFF
--- a/.github/workflows/publish-release-yaml.yaml
+++ b/.github/workflows/publish-release-yaml.yaml
@@ -38,43 +38,43 @@ jobs:
       id-token: write
       contents: read  
 
-  env:
-    KO_DOCKER_REPO: 'gcr.io/projectsigstore'
-    KOCACHE: '~/ko'
-    REPO: 'sigstore/cosign'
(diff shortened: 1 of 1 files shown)


dlorenc:
That was fast!

cpanato (commented):
this will not remove the binaries we push when doing the release? i guess it will just append the new files, but just to make sure :)

cpanato (commented):
can this be part of the cloudbuild release job or do we want to run it as action?

k4leung4:
Marking this PR obsolete as it makes more sense to move this into cloudbuild.

Opened PR #1453.

PR #1457, opened 2022-02-15
Adjust the format of the document.

kaitoii11:
#### Summary
Adjust the format of the completion  document for better visibility.

### DIFF
--- a/doc/cosign_completion.md
+++ b/doc/cosign_completion.md
@@ -5,30 +5,38 @@ Generate completion script
 ### Synopsis
 
 To load completions:
+
 Bash:
+````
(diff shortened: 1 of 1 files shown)


priyawadhwa:
Hey @kaitoii11 thanks for fixing this. This doc is auto-generated via `make docgen` so unit tests will fail if it is edited manually. These changes would have to happen in the description for the command [here](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/completion.go#L29-L51) & then you could run`make docgen` to update the do, but that would make output in the terminal kinda ugly.

Unfortunately I think that leaving this as is might be the best option for usability.

PR #1464, opened 2022-02-15
fix(sign): refactor unsupported provider log

### LINKED ISSUE #1461: error occurred: no provider found for that key reference, will try to load key from disk
**Description**

An unexpected error is printed when using a signing key locally on disk.

Steps to reproduce

```shell
$ go install github.com/sigstore/cosign/cmd/cosign@v1.5.1
$ echo "squash" > bug.txt
$ export COSIGN_PASSWORD=test1234
$ cosign generate-keypair
$ cosign sign-blob --key ./cosign.key bug.txt
Using payload from: bug.txt
an error occurred: no provider found for that key reference, will try to load key from disk...
MEUCIBJwGMXOpIigeLPn56K8LfR8om1iS9svRv4zUCS1DsjpAiEAm4d5AenYmRZThahd2jItQlHVPFa1opsGHY+lLXaF+70=
```

**Expected**

No error to show up on the CLI.

```shell
$ cosign sign-blob --key ./cosign.key bug.txt
Using payload from: bug.txt
MEUCIBJwGMXOpIigeLPn56K8LfR8om1iS9svRv4zUCS1DsjpAiEAm4d5AenYmRZThahd2jItQlHVPFa1opsGHY+lLXaF+70=
```

Dentrax:

#### Summary
In case unexpected _provider_ given, following error will be thrown:
```
./cosign sign-blob --key foo://cosign.key bug.txt
Using payload from: bug.txt
Error: signing bug.txt: reading key: unsupported provider given
main.go:46: error during command execution: signing bug.txt: reading key: unsupported provider given
```

But couldn't get why there are 2 same logs 

#### Ticket Link

Fixes #1461

#### Release Note

```release-note
Fixes an issue where the sign command is throwing an unsupported provider log without checking whether it actually is.
```


dlorenc (approved):

pkg/signature/keys.go
```diff
@@ -145,14 +145,17 @@ func SignerVerifierFromKeyRef(ctx context.Context, keyRef string, pf cosign.Pass
 		}
 
 		return cosign.LoadPrivateKey([]byte(pk), []byte(pass))
+	default:
```
dlorenc, line 148: Actually, would this make sense as just a case?

```golang
case strings.Contains(keyRef, "://"):
  return nil, fmt.Errorf("unsupported provider given")
```

dlorenc (approved):

PR #1467, opened 2022-02-15
Verify tlog entries using the Rekor public keys from TUF

haydentherapper:
Typical verification used VerifyBundle, which called
GetRekorKeys, which fetches keys from the TUF repo.
If the bundle was not present or for a specific error
when a duplicate log entry was present, then the tlog
entry would be verified using a public key fetched from
Rekor's API. This key was not verified using TUF metadata.

This change simply removes the API call and uses Rekor
public keys from the TUF repo.

Tested locally by not including the Rekor bundle in
the OCI signature, which will hit the code path to
fetch the entry from the log.

#### Ticket Link

Fixes

#### Release Note

```release-note
Switched verifying non-bundled transparency log entries to using keys from the TUF repository
```

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -34,7 +34,6 @@ import (
 
 	"github.com/sigstore/rekor/pkg/generated/client"
 	"github.com/sigstore/rekor/pkg/generated/client/entries"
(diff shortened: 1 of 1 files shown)


haydentherapper:
cc @asraa

dlorenc (approved):

asraa (approved):

PR #1468, opened 2022-02-15
Print message when verifying with old TUF targets

haydentherapper:
#### Summary

This adds console messages when the TUF metadata
used for Rekor or the CTFE key is marked as expired.

I haven't added a log message for Fulcio yet. The way
that certificates are verified is different. Instead of
multiple verifications where we can easily determine
which key successfully verified an object, the x509
library uses a CertPool and returns a valid chain.
I'll need to plumb through the TUF information.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added information when old TUF targets are used to verify signatures
```

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
+++ b/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
@@ -52,7 +52,7 @@ const altCTLogPublicKeyLocation = "SIGSTORE_CT_LOG_PUBLIC_KEY_FILE"
 // the certificate issued by Fulcio was also added to the public CT log within
 // some defined time period
 func verifySCT(ctx context.Context, certPEM, rawSCT []byte) error {
-	var pubKeys []crypto.PublicKey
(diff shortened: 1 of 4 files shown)


haydentherapper:
Note that this includes the commit from #1467 too. Let me know if you have any feedback on the UX of this. I don't want this to be a scary warning, but I do think users should still be notified.

cc @asraa @dlorenc

dlorenc:
The UX here is fine, is there any way we could make this work for machine readable output too? Maybe include the TUF root we verified against?

haydentherapper:
I'll take a look at piping some information to stdout. Is there any existing format for machine readable output? Just print JSON?

dlorenc:
> I'll take a look at piping some information to stdout. Is there any existing format for machine readable output? Just print JSON?

This can come in a separate PR - but yeah it's typically just when we do --format=json. The basic idea is just that if we feel information is important enough to display to a person, it probably makes sense to make it parseable in a script too.

asraa (approved):

PR #1470, opened 2022-02-16
tests: `/bin/bash` -> `/usr/bin/env bash`

znewman01:
#### Summary
This uses `bash` from the user's `$PATH`, which is (usually) preferable to hardcoding a particular `bash` location: https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

Selfishly, without this change these scripts don't work on NixOS.

#### Ticket Link

Fixes N/A

#### Release Note

```release-note
NONE
```

### DIFF
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2021 The Sigstore Authors.
 #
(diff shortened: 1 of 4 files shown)


znewman01:
Filed https://github.com/sigstore/cosign/issues/1471 for the flaky CI

PR #1490, opened 2022-02-21
Pick up some of the shared workflows

mattmoor:
This picks up a shared definition of `DO NOT SUBMIT` to avoid divergent shell code.

This also picks up our KinD diagnostic dumping logic from downstream, which we've been using to debug assorted KinD-based failures.

I have a few more workflows on my list that we can share this way, and I'll likely keep adding them here.

@cpanato Not sure what the best way of getting this to use the dependabot form is, so PTAL and LMK how best to adjust this.  We haven't cut a release yet, but we can once we've gotten a bunch more of our workflows shared this way.

#### Ticket Link

N/A

#### Release Note

```release-note
NONE
```


dlorenc (approved):

mattmoor:
I think we should probably snap a `v1` before we merge this (and/or pin to a commit), but I wanted to find out from @cpanato the best way to align with what we do here.

.github/workflows/donotsubmit.yaml
```diff
@@ -17,39 +17,4 @@ jobs:
...
-          xargs grep -n "DO NOT SUBMIT" |
-          reviewdog -efm="%f:%l:%m" \
-                -name="DO NOT SUBMIT" \
-                -reporter="github-pr-check" \
-                -filter-mode="added" \
-                -fail-on-error="true" \
-                -level="error"
-
-          echo '::endgroup::'
+        uses: chainguard-dev/actions/donotsubmit@main
```
cpanato, line 20: can we use a git hash instead of main?

cpanato:
@mattmoor, we need to add the following in the dependabot.yml definition

```yaml
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "daily"
```

then it will check the actions/jobs and update them according

.github/workflows/donotsubmit.yaml
mattmoor: Done

mattmoor:
Added the dependabot config as well.

.github/workflows/whitespace.yaml
```diff
@@ -13,84 +13,11 @@ jobs:
...
-              echo $x:$((1 + $(wc -l $x | tr -s ' ' | cut -d' ' -f 1))): Missing newline
-            fi
-          done |
-          reviewdog -efm="%f:%l: %m" \
-                -name="EOF Newline" \
-                -reporter="github-pr-check" \
-                -filter-mode="added" \
-                -fail-on-error="true" \
-                -level="error"
-          echo '::endgroup::'
```
cpanato, line 96: nit/non blocking: new line

cpanato (approved):
cool!

.github/workflows/whitespace.yaml
mattmoor: This was before 🤣  👍

.github/dependabot.yml
```diff
@@ -20,3 +20,7 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"
```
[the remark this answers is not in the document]
mattmoor: done

.github/workflows/whitespace.yaml
cpanato: 🙈

cpanato (approved):

PR #1514, opened 2022-02-24
Update the k8s codegen library and actually re-run update-codegen.

dlorenc:
Replaces #1508 

#### Ticket Link

Fixes

#### Release Note

```release-note

```


mattmoor:
We are on K8s v0.22 libs for now.  We can pin it with replace if it'll squelch `dependabot`?

PR #1540, opened 2022-03-02
Add fields in policy CRD

### LINKED ISSUE #1417: feature: create image policy types and CRDs
Associated doc: https://docs.google.com/document/d/1gBLEOOHWOmvHVsoJbgGU74GdwA6CGxMRp3MAeEB50l4/edit#

Based upon discovery in the referenced doc, create necessary types and CRD to support future work on Image Policy for cosigned admission controller.

Outcome:
* Necessary data types have been defined
* Validation and default values have been established

@mattmoor @coyote240 @hectorj2f 

See also: #1418 #1419

kkavitha:
#### Summary
This PR adds fields to the policy CRD
Additional doc: https://docs.google.com/document/d/1gBLEOOHWOmvHVsoJbgGU74GdwA6CGxMRp3MAeEB50l4/edit#heading=h.kvgjd3pmidrj
#### Ticket Link
Fixes #1417 

cc: @mattmoor @hectorj2f @vaikas


pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,79 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+type Authority struct {
+	Key     KeyRef     `json:"key"`
+	Keyless KeylessRef `json:"keyless"`
+	Sources []Source   `json:"source"`
+	CTLog   TLog       `json:"ctlog"`
+}
+
+// This references a public verification key stored in
+// a secret in the cosign-system namespace.
+type KeyRef struct {
```
vaikas, line 70: Consider adding a comment stating that exactly one must be specified.

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,79 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+	SecretRef *v1.SecretReference `json:"secretref"`
```
vaikas, line 71: because it's one of, I think all these should have:
// +optional

and add the json omitempty tag?

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,79 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+
+// Source specifies the location of the signature
+type Source struct {
+	OCI string `json:"oci"`
+}
+
+// TLog specifies the URL to a transparency log that holds
+// the signature and public key information
+type TLog struct {
+	URL string `json:"url"`
```
vaikas, line 86: I would prefer if the URLs were:
import this:
```
	"knative.dev/pkg/apis"
```

And then here instead of string, use:
	URL *apis.URL `json:"url,omitempty"`

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,79 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+}
+
+// KeylessRef contains location of the validating certificate and the identities
+// against which to verify. KeylessRef will contain either the URL to the verifying
+// certificate, or it will contain the certificate data inline or in a secret.
+type KeylessRef struct {
+	URL        string             `json:"url"`
```
vaikas, line 93: ditto here for the URL
	URL *apis.URL `json:"url,omitempty"`

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,79 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+// The authorities block defines the rules for discovering and
+// validating signatures.  Signatures are
+// cryptographically verified using one of the "key" or "keyless"
+// fields.
+// When multiple authorities are specified, any of them may be used
+// to source the valid signature we are looking for to admit an
+// image.
+
+type Authority struct {
+	Key     KeyRef     `json:"key"`
```
vaikas, line 62: nit, some of these are optional, so they should be tagged with:
// +optional

omitempty json tag

And they should be pointers.

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,79 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+
+// Identity may contain the issue and/or the subject found in the transparency log.
+// Either field supports a pattern glob.
+type Identity struct {
+	Issuer  string `json:"issuer"`
+	Subject string `json:"subject"`
+}
+
+// CAKey contains inline public-key data
+type CAKey struct {
```
vaikas, line 107: I'm wondering if we could add SecretRef here instead of name, my understanding was that it was meant to be used for airgaps, where the CA was on the cluster.
Also, if we then add KMS into this struct, it looks identical to KeyRef, so in there we can just inline it. And we can get rid of L96?

vaikas (commented):
Awesomesauce! Few comments about the datastructures, thanks for splitting out only the API changes.

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
vaikas: And now that I think about it, we don't even need to inline it in KeyRef if we think it won't be needing futureproofing. If we are confident (which I'm not :) ), then we could replace L62 with CAKey. I think however it still would make sense to inline into KeyRef, in case there are other things that show up only for KeyRef.

coyote240: Really good point re: SecretRef, putting a change in for that now.

Rather than replacing the KeyRef in Authority with CAKey, we're thinking we can replace CAKey type with KeyRef type.

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
coyote240: +1

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
vaikas: Yeah, or that :) 👍

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,81 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+	// +optional
+	CTLog *TLog `json:"ctlog,omitempty"`
+}
+
+// This references a public verification key stored in
+// a secret in the cosign-system namespace.
+// A KeyRef must specify only one of SecretRef, Data or KMS
+type KeyRef struct {
+	// +optional
+	SecretRef *v1.SecretReference `json:"secretref,omitempty"`
```
vaikas, line 77: This does allow for the namespace to be specified. I personally like this flexibility, and we can validate if a particular instance does not allow for per-namespace secrets. Just jotting it down, because in the api working group we had talked about the problem of having to manage secrets in many namespaces and decided that the way around that would be restrict that they live in a single namespace to reduce management headaches. This allows us to do both 👍

vaikas, line 77: Nit: json tag should be secretRef as per the doc and well, by convention :)

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,81 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+	// Data contains the inline public key
+	// +optional
+	Data string `json:"data,omitempty"`
+	// KMS contains the KMS url of the public key
+	// +optional
+	KMS string `json:"kms	,omitempty"`
```
vaikas, line 83: this looks a bit wonky?
kms,omitempty
instead?

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -49,9 +36,81 @@ func (*ClusterImagePolicy) GetGroupVersionKind() schema.GroupVersionKind {
...
+}
+
+// KeylessRef contains location of the validating certificate and the identities
+// against which to verify. KeylessRef will contain either the URL to the verifying
+// certificate, or it will contain the certificate data inline or in a secret.
+type KeylessRef struct {
+	// +optional
+	URL *apis.URL `json:"url,omitempty"`
+	// +optional
+	Identities *[]Identity `json:"identities,omitempty"`
```
vaikas, line 104: hm. I don't this needs to be a pointer. Because the array should have a nil value.

vaikas (commented):
Woot! Couple of nits. Thanks for the quick turnaround!

vaikas:
@kkavitha looks like there's a few style issues also, just a heads up.

vaikas (approved):
Thank you so much for getting this done. Let's get it in so we can start sharding the remaining work. We can always change things if during implementation we find some oopses here.

hectorj2f:
❤️  Awesome work!

PR #1548, opened 2022-03-03
1417 policy validations

kkavitha:
#### Summary
Create a validation webhook for policy CRD and their tests
Fixes: #1417 

cc: @hectorj2f @mattmoor @vaikas


vaikas:
I'm just curious why the test run needed to be approved since it says first time contributor, but we got the other PR merged yesterday, so I don't understand why it blocked you again? 🤔

kkavitha:
It didn't need approval to run when I was the author. Looks like Adam has to contribute a PR under his name to get the permission. The merged PR was under my name.

config/501-policy-webhook-configurations.yaml
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
```
vaikas, line 17: nit: I don't think we need this, or L34 below.

test/e2e_test_policy_crd.sh
```diff
@@ -0,0 +1,270 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+echo '::group:: invalid policy: both glob and regex'
+
+cat > policy.yaml <<EOF
```
vaikas, line 20: I wonder if we might want to create something like:
./test/testadata/success
./test/testdata/failure

directories and then create files there with names like:
./test/testdata/failure/both-glob-and-regex.yaml
./test/testdata/failure/key-with-multiple-properties.yaml

And then the both-glob-and-regex.yaml would look like:
```
# Invalid policy: both glob and regex
apiVersion: cosigned.sigstore.dev/v1alpha1
kind: ClusterImagePolicy
metadata:
  name: image-policy
spec:
  images:
  - glob: image**
    regex: image.*
    authorities:
    - key:
        data: "---somedata---"
```

Then we could have a for loop for both directories and for failures they should fail, and for successes they
should be created. For success case, then we could also delete after created, so that we don't have to deal
with conflicting names.

Just thinking that having those files might make it easier then to add new tests in the future as well as have some example configurations that could also serve as documentation. They could also be used for testing the validations for the fields. Just a thought.

config/501-policy-webhook-configurations.yaml
hectorj2f: you're right, we don't need it.

test/e2e_test_policy_crd.sh
hectorj2f: If we go with the proposed filenames under testdata, I'd suggest to add a `cosigned` prefix to know where they are been used.

hectorj2f (commented):
@kkavitha I checked the code, the `e2e_test_cosigned.sh` are failing to reject invalid pods because the cluster image policy has not been installed. We'll need to install so the tests are working again,

test/e2e_test_policy_crd.sh
kkavitha: That makes sense. I will push them to new files.

config/501-policy-webhook-configurations.yaml
kkavitha: Cool, will remove it!

test/e2e_test_policy_crd.sh
```diff
@@ -0,0 +1,109 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+echo '::group:: invalid policy: both glob and regex'
```
vaikas, line 19: Awesomesauce! So, if we don't hand code these files, but use a for loop here:
```
for i in `ls ./test/testdata/cosigned/invalid/`
do
  if kubectl create -f ./test/testdata/cosigned/invalid/$i ; then
  then
    echo $i failed when it should not have
    exit 1
  else $i rejected as expected
done
```

Or something like that. Then if new files are added to .../invalid/ directory, then we don't have to modify this file at all.

What do you think?

kkavitha: Yeah that sounds good. Will make the change!

test/e2e_test_policy_crd.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+echo '::group:: Invalid policy tests:'
```
mattmoor, line 19: I think this is missing its `::endgroup::`

test/testdata/cosigned/valid/valid-policy.yaml
```diff
@@ -0,0 +1,46 @@
...
+            namespace: some-namespacemak
+    - keyless:
+        identities:
+        - issuer: "issue-details"
+          subject: "subject-details"
+    - keyless:
+        identities:
+        - issuer: "issue-details1"
+    - key:
+        data: "---some-key---"
```
vaikas, line 40: I created this:
https://github.com/sigstore/cosign/issues/1581
to track tighter validations.
But let's get this in, it's a great starting point!

vaikas (approved):
Thanks!

PR #1556, opened 2022-03-06
oci/signature: expose descriptor reference to interface

Dentrax:

#### Ticket Link

Fixes

#### Release Note

```release-note
Ability to expose v1.Descriptor reference in sigLayer{}
```

### DIFF
--- a/pkg/oci/internal/signature/layer.go
+++ b/pkg/oci/internal/signature/layer.go
@@ -50,6 +50,11 @@ func New(l v1.Layer, desc v1.Descriptor) oci.Signature {
 
 var _ oci.Signature = (*sigLayer)(nil)
 
+// Descriptor holds a reference from the manifest to one of its constituent elements.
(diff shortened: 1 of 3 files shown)


Dentrax:
The thing I want to do here is that accessing _owner manifest_'s descriptor in `sigLayer` struct. By doing so, I can get the _OCI image manifest_'s config.

But couldn't figure out how can I implement this in `staticLayer{}`, since we don't have information about `v1.Image` during initialization:

```go
func NewSignature(payload []byte, b64sig string, opts ...Option) (oci.Signature, error)
```

We have to pass `v1.Image` to get _config_. I could create a new option, something like:
```go
func WithImage(image v1.Image) Option {
	return func(o *options) {
		o.image = image
	}
}
```

Which will turn into:
```go
manifest, _ := s.Image.Manifest()
descriptor := manifest.Config()
```

I'm not sure if this is right way to handle this.

PR #1580, opened 2022-03-09
Support deletion of ClusterImagePolicy

### LINKED ISSUE #1572: ClusterImagePolicy needs to remove policies from CM on deletions.
After #1562 goes in as a follow on, we need to add support for when CIP is deleted that we then remove the entry from the config--image-policies ConfigMap.

This requires implementing FinalizeKind method.

vaikas:
By removing them properly from the CM.

Fixes #1572 

#### Ticket Link

Fixes

#### Release Note

```release-note

```


mattmoor (approved):
🤩

PR #1595, opened 2022-03-11
Add ability to inline secrets from SecretRef to configmap.

### LINKED ISSUE #1573: Resolve Secrets from CIP into ConfigMap
As a follow on to #1562 we need to read the Secrets (which are really just public keys in this use case) and add them into the config-image-policies ConfigMap.

There's couple of questions about this. If a Secret has multiple keys in it, should we cat them all into a single `data` entry in the fields. Related,is it meant that this field can hold multiple keys or just a single key?

https://github.com/sigstore/cosign/blob/main/pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go#L94

vaikas:
Use tracker to keep track of changes to secrets, for example
if a secret doesn't exist initially, once it shows up. Tested
obvs with UT but also tested on a real cluster that the tracker
is keeping track of changes to secrets.

Fix #1573

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
+++ b/pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
@@ -16,16 +16,15 @@ package clusterimagepolicy
 
 import (
 	"context"
-	"errors"
+	"fmt"
 
 	"github.com/sigstore/cosign/pkg/apis/config"
-	"k8s.io/apimachinery/pkg/types"
-
 	"github.com/sigstore/cosign/pkg/apis/cosigned/v1alpha1"
 	clusterimagepolicyreconciler "github.com/sigstore/cosign/pkg/client/injection/reconciler/cosigned/v1alpha1/clusterimagepolicy"
 	"github.com/sigstore/cosign/pkg/reconciler/clusterimagepolicy/resources"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
```diff
@@ -116,18 +116,60 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, cip *v1alpha1.ClusterImag
...
+		return fmt.Errorf("failed to track changes to secret %q : %w", keyref.SecretRef.Name, err)
+	}
+	secret, err := r.secretlister.Secrets(system.Namespace()).Get(keyref.SecretRef.Name)
+	if err != nil {
+		return err
+	}
+	if len(secret.Data) == 0 {
+		return fmt.Errorf("secret %q contains no data", keyref.SecretRef.Name)
+	}
+	if len(secret.Data) > 1 {
```
hectorj2f, line 166: what about validating that the data field contains a `publicKey` field like here https://github.com/sigstore/cosign/pull/1595/files#diff-f95503af2405ad769f56a5e3ec662f2d57380441e636ba81ffe33b93e00ea4ebR437 ?

vaikas: I'll do that in a followup?
https://github.com/sigstore/cosign/issues/1596

hectorj2f (approved):

PR #1605, opened 2022-03-14
Ensure entry is removed from CM on secret error.

### LINKED ISSUE #1601: Remove entries from CM if secret goes missing
If a secret exists, and an entry with the Secret inlined has been created in the CM, but subsequently the secret goes missing, we should remove the entry from the CM since the entry is no longer valid.

vaikas:

#### Summary
If there's an entry in the Configmap, say a secret existed and a valid entry was made, but then secret was deleted or modified, remove an existing entry.

Fix #1601 

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cpanato (approved):

hectorj2f:
@vaikas thanks. My concern was between steps 1 and 2. I am used to set an ownership reference to the secret, so whenever someone attempts to delete it should get an error that is used by our other resource.

vaikas:
@hectorj2f yeah, that is usually how we'd do it if we'd be the one controlling the resource. We could add an ownerref but that feels a bit wonky since we don't really own that in this case. But, even if we did that, we would still have to deal with the case where the secret contains invalid data.

hectorj2f (approved):

PR #1615, opened 2022-03-16
Use latest knative/pkg's configmap informer

tcnghia:
#### Summary
Use the latest version of `knative/pkg` to pull in the dynamic configmap informer (https://github.com/knative/pkg/pull/2466)

#### Release Note

```release-note
NONE
```


pkg/reconciler/clusterimagepolicy/controller.go
```diff
@@ -47,19 +48,19 @@ func NewController(
...
-	nsSecretInformer := nsinformerfactory.Get(ctx).Core().V1().Secrets()
-	nsConfigMapInformer := nsinformerfactory.Get(ctx).Core().V1().ConfigMaps()
+	secretInformer := secretinformer.Get(ctx)
+	configMapInformer := cminformer.Get(ctx)
 
 	// Start the informers we got from the SharedInformerFactory above because
 	// injection doesn't do that for us since we're injecting the Factory and
 	// not the informers.
-	if err := controller.StartInformers(ctx.Done(), nsSecretInformer.Informer(), nsConfigMapInformer.Informer()); err != nil {
+	if err := controller.StartInformers(ctx.Done(), secretInformer.Informer(), configMapInformer.Informer()); err != nil {
```
vaikas, line 57: I know this is just a draft, but jotting this down.
Now that both of these are coming from the informers, instead of the factory, I think we do not need to manually start them.
https://github.com/knative/pkg/blob/main/injection/clients/namespacedkube/informers/core/v1/configmap/configmap.go#L36

For example, above, registers cm informer, so it should be started automagically by the sharedwithmain.

tcnghia: Done

pkg/reconciler/clusterimagepolicy/controller_test.go
```diff
@@ -18,16 +18,21 @@ import (
...
+	fakesecret "knative.dev/pkg/injection/clients/namespacedkube/informers/core/v1/secret/fake"
 	_ "knative.dev/pkg/injection/clients/namespacedkube/informers/factory/fake"
 )
 
 func TestNew(t *testing.T) {
-	ctx, _ := SetupFakeContext(t)
+	ctx, _ := rtesting.SetupFakeContext(t)
+
+	fakesecret.Get(ctx)
+	fakecm.Get(ctx)
```
mattmoor, line 35: You can just use `_ "..."` above to link these instead of adding these calls.

mattmoor (approved):

pkg/reconciler/clusterimagepolicy/controller_test.go
tcnghia: Done

hectorj2f (approved):

cpanato (approved):

mattmoor:
going to close and reopen to try and kick DCO...

mattmoor (approved):

PR #1623, opened 2022-03-16
Validate authority keys

### LINKED ISSUE #1419: feature: integrate policy CRD into admission webhook
Use the created policy/configmaps created in #1417 and #1418 in the webhook to enforce policy

@mattmoor @coyote240 @hectorj2f

coyote240:
Resolves #1419 

#### Summary

Begin integration of clusterimagepolicy into cosigned webhook.

@dennyhoang


pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -138,7 +139,30 @@ func (v *Validator) validatePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
 				continue
 			}
 
-			if err := valid(ctx, ref, keys, ociremote.WithRemoteOptions(remote.WithAuthFromKeychain(kc))); err != nil {
+			tempKeys := keys
+			config := config.FromContext(ctx)
+			if config != nil {
+				authorities, err := config.ImagePolicyConfig.GetAuthorities(ref.Name())
```
DennyHoang, line 145: Question: Wondering how I could structure a unit-test to mock out `ctx` such that `config.FromContext(ctx)` returns a config with mocked data for `authorities[].key.data`. Currently, I tried to use `context.WithValue(ctx, config.CfgKey{}, <ImagePolicyConfig populated struct>)` and this did not work (I made `cfgKey{}` public in `store.go` to temporarily try this).

@dlorenc @vaikas

DennyHoang: Yup, pulling out the logic into a function and going to test that specifically. I will be passing in a mocked config now so I can isolate the testing to the logic there. 

As for testing the integration of that function, I'll look into that further afterwards.

vaikas:
Fantastic! Thanks for doing this. We also need to add validation here (since we didn't support it at the time, I made it disallowed).
https://github.com/sigstore/cosign/blob/main/pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation.go#L61

Then we should probably in the validation function see if the regex compiles, wdyt? I think this can then we caught earlier, during the creation of the resource.
https://github.com/sigstore/cosign/pull/1623/files#diff-5cd49ec5dc4f031cb339593e719593fd2d06d1ee635e62360629fd048a9f0610R104

If we use something like this during the validation?
https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/regexp/regexp.go;l=134

And then maybe add a valid regex in here (or create a new file with just regex?):
https://github.com/sigstore/cosign/blob/main/test/testdata/cosigned/valid/valid-policy.yaml

And maybe add an invalid regex file here as well?
https://github.com/sigstore/cosign/tree/main/test/testdata/cosigned/invalid

pkg/cosign/kubernetes/webhook/validator.go
vaikas: I think using this should work for this:
https://github.com/sigstore/cosign/blob/main/pkg/apis/config/store.go#L53

DennyHoang: Oh my, how did I not see that... Thanks, will give it a try!

vaikas (approved):
Thanks for doing this, and thanks for cleaning up the tests! I'm fine with this going on if you want to do the errors in the followup, but please create an issue to track that.

pkg/cosign/kubernetes/webhook/validation.go
```diff
@@ -121,6 +106,27 @@ func getKeys(ctx context.Context, cfg map[string][]byte) ([]*ecdsa.PublicKey, *a
...
+	pems := parsePems([]byte(pubKey))
+	for _, p := range pems {
+		key, err := x509.ParsePKIXPublicKey(p.Bytes)
+		if err != nil {
+			errs = append(errs, err)
+		} else {
+			keys = append(keys, key.(*ecdsa.PublicKey))
+		}
+	}
+	if keys == nil {
```
hectorj2f, line 124: Suggestion: you may want to do `len(keys) == 0` instead.

pkg/cosign/kubernetes/webhook/validator.go
hectorj2f: It looks good to me.

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -153,6 +161,29 @@ func (v *Validator) validatePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
 	return errs
 }
 
+func getAuthorityKeys(ctx context.Context, ref name.Reference, config *config.Config) []*ecdsa.PublicKey {
```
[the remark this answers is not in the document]
DennyHoang: Updated to surface errs and it is currently logged as a warning. 
Tracking issue to actually enforce the errors in: 
https://github.com/sigstore/cosign/issues/1642

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -153,6 +166,26 @@ func (v *Validator) validatePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
 	return errs
 }
 
+func getAuthorityKeys(ctx context.Context, ref name.Reference, config *config.Config) (keys []*ecdsa.PublicKey, errs *apis.FieldError) {
```
vaikas, line 169: I'm not sure we want to make these as apis.FieldError. We usually use those as validation time errors because they are user facing directly. I'd just been thinking that we'll use err for the return value, not apis.FieldError. <time passes...> But then as I started thinking about it some more, the are user facing at least in some cases because they will be returned to the user when they try to deploy something.

vaikas (commented):
Do you want to switch from Draft? I think it's in a shape where we can get it in and continue working on it, wdyt?

pkg/cosign/kubernetes/webhook/validator.go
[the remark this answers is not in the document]
vaikas: odd, this was a comment I made in the earlier version. Not sure why it's showing up as a comment now. Resolving as you made this change :)

vaikas:
Also I've been working on adding the Keyless stuff into this, so if we get this merged, we can keep working on respective pieces :)

pkg/cosign/kubernetes/webhook/validator.go
DennyHoang: I also felt similar that it should have been `errors` initially. I followed using `apis.ErrGeneric` because `getAuthorityKeys/parseAuthorityKeys` followed the same logic from `getKeys`. `getKeys` returned errors using `apis.ErrGeneric`. I just tried to surface the lowest level error up and made it a common type. Should I change it such that getAuthorityKeys returns errors only and then have `validatePodSpec` determine how to interpret generic `error` such as into apis.FieldError? 

Or is this something that can be changed later when we actually start erroring out more concretely? The change seems trivial enough either way you want it.

hectorj2f: Question: why do we even need to parse the authority keys here ? The function call comes from validatePodSpec, i assume the parsing of the authorities has been done when creating the image policy config, rather when validating the pod spec.

DennyHoang:
@vaikas I can't update the PR from Draft since "Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can mark a draft pull request as ready for review."

coyote240:
Set to "Ready for Review"

pkg/cosign/kubernetes/webhook/validator.go
DennyHoang: When we get the `ImagePolicyConfig`, it contains `Policies.[]ClusterImagePolicySpec.[]Authority`. The returned `Authority` contains `KeyRef.Data` which is a string. `parseAuthorityKeys` takes that string and parses it into `ecdsa.PublicKey`.

hectorj2f (approved):

DennyHoang:
@hectorj2f @vaikas 
Can you please re-run the workflows? I fixed the lint issues and updated e2e test script to delete the applied CRD between runs as the second run of the `kubectl create` was erroring from the previous successful run.

dlorenc:
> Can you please re-run the workflows? I fixed the lint issues and updated e2e test script to delete the applied CRD between runs as the second run of the `kubectl create` was erroring from the previous successful run.

Rerunning!

pkg/cosign/kubernetes/webhook/validator.go
vaikas: Well, sorry I was thinking that we could be defining the signature as error, I think at least (too lazy to look up as I'm on the phone) we could still be returning apierrors as they are errors but not bleed that into the signature of the function.

dlorenc:
reran it one more time, looked like a GCP flake.

hectorj2f:
Yes, it was a GCP flake.

Issue #1554, opened 2022-03-04
Allow specifying CA chain when not using Fulcio


bburky:
When using a local PEM key file, a PKCS#11 key, or a KMS key a user should be able to specify `--cert` and something like `--chain` to specify the [`dev.sigstore.cosign/certificate`](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#certificate) and [`dev.sigstore.cosign/chain`](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#chain).

The following already works to specify the certificate, and correctly adds it to the cosign signature:
```
cosign sign --key awskms:///... --cert cert.pem example.com/foo
```

However, `--cert` is silently ignored in the following command (the certificate from the token is used instead of the provided one): (this is a separate bug really, should I create a new issue for it?)
```
cosign sign --key pkcs11:... --cert cert.pem example.com/foo
```

It is not possible with either of these to specify the CA chain. If my certificate is signed by existing PKI, I would like to include the intermediate certificates and root in the signature.

Two possible implementations would be:
1. Add a new argument like `--chain` which includes a chain of certificates.
1. Allow the `--cert` file to include both the leaf cert, and a chain of intermediates.
   * One issue with this is that a user may want to use the public key on the PKCS#11 token and provide a chain to include in the signature. This option would force them to extract the public key from their token to include in this file. If the chain is supplied separately with `--chain` they do not need to.

Providing the certificate chain is incompatible with using Fulcio, and should return an error.

dlorenc:
+1 to allowing this somehow. I've been meaning to think more about it forever but haven't had the time. Thanks for opening it!

bburky:
Also, `cosign verify` should support verification. I think the go api already should "just work" if you specify the roots correctly.

Something like:
```
# possibly somewhat confusing:
cosign verify --key ca-bundle.pem
# or a new flag, something like this which more closely matches the go api:
cosign verify --roots ca-bundle.pem
```

bburky:
@dlorenc  Do you have a preference for new flags like `--chain` or putting a concatenated CA chain in the `--cert` PEM file?

dlorenc:
I'd probably defer to @haydentherapper on that!

bburky:
[...]

PR #1656, opened 2022-03-24
Add certificate chain flag for signing

haydentherapper:
This allows users to pass their own certificate chain to include in the
OCI signature. The chain is checked for validity using the provided
certificate.

Also refactored the check for matching public keys using a method from
sigstore/sigstore, comparing the certificate's key with the provided
key. Also added this check when extracting the PKCS11 certificate.

Certificate chains must be PEM-encoded. I changed the text of the
certificate flag to also specify a preference for PEM encoding, but
didn't remove the code that handles DER encoding for backwards
compatibility.

#### Ticket Link

Ref #1554

#### Release Note

```release-note
Added cert-chain flag for including certificate chain in OCI signature on signing
```


haydentherapper:
cc @bburky @dlorenc

haydentherapper:
@dlorenc Are you familiar with `./hack/update-codegen.sh`? I ran it, but it had a few errors, and ended up generating a bunch of code files in `vendors/` rather than updating the licenses.

haydentherapper:
Figured out the issue, looks like some part of the script needs a license binary to be installed via `go install` instead of `go get`.

bburky:
I tested with a couple configurations:
```
# The following works great and adds the chain annotation:
cosign sign --key "awskms:///..." --cert cert.pem  --cert-chain chain.pem "localhost:5000/$image"  
# The following does _not_ work. The certificate is automatically extracted from the smartcard, but the provided chain is not added as an annotation:
cosign sign --key "pkcs11:..."  --cert-chain chain.pem "localhost:5000/$image"  
```

In general, I've noticed the (optional) `--cert` flag doesn't seem to work for `--key "pkcs11:..."` and I think this is the same issue. The cert provided from the token is used unconditionally, the cert on the command line is not used. It should probably be an error to pass `--cert`, or it should be used instead of the one on the token.

Perhaps the `cosign sign --key "pkcs11:..."  --cert-chain chain.pem` behavior is related? It seems that the provided cert chain is ignored.

haydentherapper:
Thanks for testing this out. Currently, with a pkcs11 key, it attempts to fetch a certificate from the token and then returns early. You're correct, it's currently ignoring the cert and cert chain flags. I just need to change it to not exit early.

dlorenc (approved):

PR #1657, opened 2022-03-25
Use syscall.Stdin for input handle. Fixes #1153

### LINKED ISSUE #1153: Cosign crashes when password input is requested from terminal in Windows

mdp:
#### Summary

Cosign currently passes in the int '0' for the Stdin handle when reading from term. This unfortunately doesn't work on Windows, and should actually use `syscall.Stdin` to get the platform specific Stdin handle.

#### Ticket Link

Fixes #1153 

#### Release Note

* Bugfix to handle input on Windows platforms without crashing.


dlorenc:
Thanks for the fix!

mdp:
Sorry about spamming the CI/PR, let me figure out this lint issue on my own fork and I'll push when it's ready (it has to do with unconvert detecting an uneccessary conversion of 'int' on *nix but it's a uintptr on windows)

mdp:
This should be working now and rebased onto the latest main branch. I "nolint'd" the int conversion, but the alternative is to build a platform specific ReadPassword to handle the int vs uintptr divergence. Happy to quickly build that out if you think it's the right direction.

dlorenc (approved):

Issue #1105, opened 2021-11-25
random callback uri port


tuananh:
the url generated is localhost:xxxxx . however the port is random each time so it doesn't match with what i have register in oidc idp (dex) and i get this error when the url opened in browser

```
Unregistered redirect_uri ("http://localhost:56947/auth/callback").
```

is there anyway to fix the port?

tuananh:
@dlorenc i omit it so i assume it's the public dev one? the client i have registered in dex looks like below. 

```yaml
staticClients:
      - id: sigstore
        secret: __CHANGE_ME__
        name: 'sigstore'
        # Where the app will be running.
        redirectURIs:
        - 'http://localhost:56921/auth/callback'
        # - 'http://127.0.0.1:5555/callback'
```

as you can see, the redirect uri has port in it but everytime i run the below command, the port changed so i got `unregistered redirect_uri`

```
COSIGN_EXPERIMENTAL=1 cosign sign \
    --oidc-issuer "http://<my-dex-instance-url>" \
    README.md
```

feelepxyz:
👋  I [ran into this](https://github.com/lukehinds/sigstore-the-hard-way/issues/11) when trying to get [sigstore-the-hard-way](https://github.com/lukehinds/sigstore-the-hard-way) up and running. 

I think [this change in cosign is relevant](https://github.com/sigstore/sigstore/pull/120), it changed the local server to bind to a random available port, to handle it already being taken on the developers machine.

I'm not yet aware of a way to handle a random port range in dex. There's this fix proposed in dex but looks inactive: https://github.com/dexidp/dex/pull/1783

tuananh:
@dlorenc can you advise? is there any known workaround for this?

hectorj2f:
The main issue is that certain OIDC providers do not allow to have a redirect URI to a random port (e.g. auth0). In. our code, we set a redirect listener where the port is randomly picked by the OS. As a consequence, those OIDC providers return a `redirect_uri` mismatch because the port changes on every call. 

@dlorenc @tuananh A solution would be to expose a flag to set the listener of the redirect to a specific port or a redirect_uri itself. I could get some changes to fix it.

hectorj2f:
The code is  in: https://github.com/sigstore/sigstore/blob/main/pkg/oauthflow/interactive.go#L148, and even if the [RFC8252](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) mentions 
```
[...]

PR #1658, opened 2022-03-25
add oidc-callback-port flag

tuananh:
Signed-off-by: Tuan Anh Tran <anhtt17@techcombank.com.vn>

Require this PR in sigstore https://github.com/sigstore/sigstore/pull/347

#### Summary

Add `--oidc-callback-port` flag to set callback port to a specific port rather than a random port. this is due to some provider do callback url whitelisting.

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1105

#### Release Note

```release-note

```


hectorj2f (commented):
@tuananh Thanks for the PR 👏🏻 . 
But I am not sure if defining the callback port is the optimal choice. I was thinking about exposing a flag to set the **RedirectURI** itself `oauth2.Config` and pass it as part of the existing https://cs.opensource.google/go/x/oauth2/+/6242fa91:oauth2.go;l=56 properties. I can think of many other situations where `localhost` should not be used.

PR #1659, opened 2022-03-25
Add support for certificate chain to verify certificate

haydentherapper:
This adds a flag to provide a chain of CA certificates to verify a
certificate provided by flag. Callers should include a chain from the
parent of the certificate to the root.

While it'd be ideal to force the root to be specified out of band, by
TUF, that code is currently intertwined with expectations around Fulcio
and Rekor usage.

#### Ticket Link

Ref https://github.com/sigstore/cosign/issues/1554

#### Release Note

```release-note
Added certificate chain flag for verifying a certificate provided by flag
```


hectorj2f (approved):

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -121,6 +123,24 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		if err != nil {
 			return errors.Wrap(err, "loading certificate from reference")
 		}
+		// Verify certificate with chain
+		// First intermediate at chain[0], root at chain[n-1]
```
dlorenc, line 127: should we factor this logic out into something shared? I think we probably have this in a few other places too...

haydentherapper: Good suggestion! I added a function that takes the chain and calls `ValidateAndUnpackCert`. This has the added benefit of also verifying the OIDC subject/issuer flags too.

dlorenc: Nice!

bburky:
I'm putting my verification notes and comments here, let me know if you'd rather all the comments in one place.

Is the following expected to work? Sign with a full chain, and verify with only the root? I was unable to get this to work. I verified the chain using `openssl verify -CAfile root.pem -untrusted intermediate.pem cert.pem`.
```
./cosign sign  "localhost:5000/$image"  --key "pkcs11:$key" --cert-chain full-chain.pem
# The following returns `error during command execution: x509: certificate signed by unknown authority`
./cosign verify  "localhost:5000/$image"  --cert cert.pem --cert-chain root.pem
```

Some follow up operations could make sense to add longer term:
```
# Don't pin the specific cert, just check that it was signed by the specified CA
$ cosign verify "$image" --cert-chain chain.pem
# Check the Email SAN of the leaf certificate:
$ cosign verify "$image" --cert-chain chain.pem  --cert-email user@example.com

# The only current operation possible with --cert-chain requires --cert too:
$ cosign verify "$image" --cert-chain chain.pem --cert cert.pem

# Additionally, the current error message here is misleading, `--cert` is also accepted here:
$ cosign verify "$image"
Error: exactly one of: key reference (--key), or hardware token (--sk) must be provided
main.go:46: error during command execution: exactly one of: key reference (--key), or hardware token (--sk) must be provided
```

bburky:
I'm actually not sure if `--cert-chain` makes sense as an argument for `cosign verify`. Would a `--roots` flag make more sense actually? The verifier user wants to ask "is this signature made by a root I trust?" and should not need to provide the specific intermediates (unless the signature is missing a chain annotation).

A user may want to provide a _root bundle_, not a specific chain for verification. This isn't possible with a `--cert-chain` flag. Please see my comment on the above Kyverno issue too, it's along the same lines.

haydentherapper:
> Is the following expected to work? Sign with a full chain, and verify with only the root? I was unable to get this to work. I verified the chain using `openssl verify -CAfile root.pem -untrusted intermediate.pem cert.pem`.

No, it would not work without the intermediate. The verification library needs the intermediate CA in order to build a chain from leaf to root. For OpenSSL, `-untrusted` is the same as the `Intermediates` certificate pool in Golang. If you were to drop `-untrusted`, verification should fail. 

> `Don't pin the specific cert, just check that it was signed by the specified CA`
> `$ cosign verify "$image" --cert-chain chain.pem`

See the last part of this comment.

> `Check the Email SAN of the leaf certificate:`
> `$ cosign verify "$image" --cert-chain chain.pem  --cert-email user@example.com`

This actually should work! It would require the `--cert` flag currently though.

`cosign verify --cert cert.txt --cert-chain .sigstore/root/targets/fulcio_v1.crt.pem --cert-email "<my-email>" --cert-oidc-issuer "<oidc-issuer>" us-west1-docker.pkg.dev/project/repo/image:tag`

> `# Additionally, the current error message here is misleading, `--cert` is also accepted here:`

Good find, will update this!

> Would a `--roots` flag make more sense actually?

This gets back to an initial remark I made about not supporting roots specified from the command line. I agree that it'd be reasonable to support a root bundle (and a bundle of intermediates for path finding too). I'd prefer to encourage users to use TUF for this however. This provides an out-of-band mechanism for forming the trust bundle, and that trust bundle can be verified by the TUF framework (Users can provide their own TUF metadata via `cosign initialize`). An alternative is creating a verification policy, something like what Cosigned is working on.

My concern with providing a certificate chain without pinning the certificate means a client could be coerced into using a chain that issued a certificate from an untrusted identity. The same issue arises with a trust bundle. By pinning the certificate and/or the identity of the certificate, this risk is reduced (the assumption being it should be easier to confirm a trusted identity rather than a trusted chain).

bburky:
Ok thanks, it sounds like validation is working as intended then. This does support verifying "is this image signed by this specific cert with this specific full chain".

I look forward to seeing how other tooling can take advantage of these signatures and provide flexible verification policy.

PR #1662, opened 2022-03-25
client: improve tlog verification by comparing leaf hash, minor refactor

asraa:

#### Summary

This does two things
* refactor tlog verification so that multiple rekor clients aren't created in the verify-blob flow, and calls to rekor are minimized
* compare the leaf hash with the entry body similar to https://github.com/sigstore/rekor/pull/469

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -295,7 +295,7 @@ func verifyRekorEntry(ctx context.Context, ko sign.KeyOpts, e *models.LogEntryAn
 	}
 
 	// if we have a cert, we should check expiry
-	// The IntegratedTime verified in VerifyTlog (in FindTlogEntry)
+	// The IntegratedTime verified in VerifyTlog
 	return cosign.CheckExpiry(cert, time.Unix(*e.IntegratedTime, 0))
 }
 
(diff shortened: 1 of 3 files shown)


cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -253,33 +264,38 @@ func verifyRekorEntry(ctx context.Context, ko sign.KeyOpts, pubKey signature.Ver
...
+				return err
+			}
+		}
+		uuid, e, _, err = cosign.FindTlogEntry(ctx, rekorClient, b64sig, blobBytes, pubBytes)
 		if err != nil {
 			return err
 		}
 	}
-	uuid, index, err := cosign.FindTlogEntry(ctx, rekorClient, b64sig, blobBytes, pubBytes)
+	verifiedEntry, err := cosign.VerifyTLogEntry(ctx, rekorClient, e, uuid)
```
asraa, line 288: FYI it was very unclear that FindTlogEntry performed the merkle verification as well. I rescoped this so FindTlogEntry only performs the lookup by proposed entry (which also reduces the extra call to findtlogentry when we already found it in the keyless flow)

PR #1672, opened 2022-03-28
Fix handling of policy in verify-attestation

lcarva:
#### Summary

The first commit fixes the sanity checks during policy processing. Previously, the code was comparing a predicate URI against a payload type. This is never expected to match.

The code was modifying to extract the actual predicate type from within the payload so the verification can be done successfully.

The second commit addresses multiple issues when applying a rego policy against the payload of an attestation: 

1. If `data.signature.deny` evaluated to `true`, the policy verification would pass. This is obviously unexpected. The code now looks for `data.signature.allow` instead, and expects it to be `true`.
2. If a query result returned an undefined result, the policy verification would pass. The code now explicitly checks for this
condition and ensures that if `ResultSet.Allowed()` returns `false` the policy verification also fails.
3. Improve error messages to assist user in defining correct variable name and type.
4. Add unit tests to validate behavior and prevent breaking changes in the future.

#### Release Note

```release-note
* Modified rego policy evaluation in the `verify-attestation` sub-command to query for the variable `data.signature.allow`.
* Fixed rego policy evaluation to no longer pass when query result is undefined.
* Fixed rego policy evaluation to no longer pass when `data.signature.deny` is `true`.
* Fixed predicate type sanity checks during policy evaluation in the `verify-attestation` sub-command.
```

### DIFF
--- a/pkg/cosign/rego/rego_test.go
+++ b/pkg/cosign/rego/rego_test.go
@@ -1,0 +1,100 @@
+//
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rego
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+const simpleJSONBody = `{
+	"_type": "https://in-toto.io/Statement/v0.1",
+	"predicateType": "https://slsa.dev/provenance/v0.2"
+}`
+
+func TestValidationJSON(t *testing.T) {
+	cases := []struct {
+		name     string
+		jsonBody string
+		policy   string
+		pass     bool
+		errors   []string
+	}{
+		{
+			name:     "passing policy",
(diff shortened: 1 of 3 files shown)


pkg/cosign/rego/rego_test.go
```diff
@@ -0,0 +1,97 @@
...
+			`,
+			pass:   false,
+			errors: []string{"expression value, false, is not true"},
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			policyFileName := filepath.Join(t.TempDir(), "policy.rego")
+			if err := os.WriteFile(policyFileName, []byte(tt.policy), 0644); err != nil {
```
lcarva, line 79: On Windows, `os.WriteFile` doesn't return an error (good), but when `rego.ValidateJSON` tries to read the file, Windows complains it's not there:
```
CreateFile \Users\RUNNER~1\AppData\Local\Temp\TestValidationJSONpassing_policy1204380751\001\policy.rego: The system cannot find the path specified.
```

*Even though the Windows call is `CreateFile`, I believe it's actually used for both reading and writing files.

There are other parts of the code that follow this same pattern and the tests seem to work just fine. I got my hands on a windows machine and will try to reproduce this locally.

lcarva: I cannot reproduce this locally with golang 1.17.8 (same used in tests) on Windows 11. I could disable these new tests for Windows but that is not an ideal solution.

@dlorenc, any suggestions?

lcarva: This is an issue on rego itself: https://github.com/open-policy-agent/opa/issues/4521

Adjusted the tests to take this into account.

vaikas:
Thanks @lcarva for doing this! I'm going to merge this so I can start adding the e2e tests in a followup.

PR #1673, opened 2022-03-28
Find all valid entries in verify-blob

priyawadhwa:
At the moment we only check the most recent entry, and fail if verification fails.
Instead, we should look at all matching entries to see if any of them verify.

I also tried to add a test that will run in github actions and leverage github OIDC. The test passes locally, but it'll probably take a few iterations for this to actually pass in github actions since i'm not sure how to test it otherwise.

cc @znewman01 thanks for your feedback on this!

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,44 +162,69 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
...
+	if err := verifyRekorEntry(ctx, ko, verifier, cert, b64sig, blobBytes); err != nil {
+		return err
+	}
+
+	fmt.Fprintln(os.Stderr, "Verified OK")
+	return nil
+}
 
-		tlogEntry, err := cosign.GetTlogEntry(ctx, rClient, uuids[0])
+func verifyUUIDs(ctx context.Context, ko sign.KeyOpts, rClient *client.Rekor, certEmail, certOidcIssuer, sig, b64sig string, uuids []string, blobBytes []byte) error {
```
znewman01, line 187: nit: I wouldn't say that this is "verifying UUIDs," more like it's "verifying signatures by UUID"

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,44 +162,69 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
...
+			continue
+		}
+		validUUID = true
 	}
-
-	// verify the rekor entry
-	if err := verifyRekorEntry(ctx, ko, verifier, cert, b64sig, blobBytes); err != nil {
-		return err
+	if !validUUID {
+		return errors.New("could not find a valid tlog entry for provided blob")
```
znewman01, line 226: I can imagine this being really difficult to debug. Maybe: "could not find....blob. found X, none verified successfully"

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,44 +162,69 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
...
+		return err
+	}
+
+	fmt.Fprintln(os.Stderr, "Verified OK")
+	return nil
+}
 
-		tlogEntry, err := cosign.GetTlogEntry(ctx, rClient, uuids[0])
+func verifyUUIDs(ctx context.Context, ko sign.KeyOpts, rClient *client.Rekor, certEmail, certOidcIssuer, sig, b64sig string, uuids []string, blobBytes []byte) error {
+	var validUUID bool
```
znewman01, line 188: same nit as the function name

znewman01:
LGTM pending tests

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,44 +162,69 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
...
-	}
 
-	// verify the signature
-	if err := verifier.VerifySignature(bytes.NewReader([]byte(sig)), bytes.NewReader(blobBytes)); err != nil {
-		return err
+		// verify the rekor entry
+		if err := verifyRekorEntry(ctx, ko, verifier, cert, b64sig, blobBytes); err != nil {
+			continue
+		}
+		validUUID = true
```
znewman01, line 223: Hmm, looks like we're not doing anything with most of the certs that we check, in which case you probably could stop after the first one.

Longer term I think we want to make sure we print them all or at least record them somehow.

priyawadhwa: Right now all valid entries will be printed in `verifyRekorEntry` which we call for each UUID -- 

https://github.com/sigstore/cosign/blob/ae7f873dd97a60f80a60896c7a2b72c0361a2d4f/cmd/cosign/cli/verify/verify_blob.go#L289

so all valid entries are getting printed to stderr. i can change this to only return one as we were doing before if you think that's better!

sample output:
```
cosign verify-blob --signature sig ./blob
tlog entry verified with uuid: "06b1f4d42d88247a4db98127e47c31ed3cc404024951542c0e98757c965ea7e2" index: 1819736
Verified OK
```

znewman01: Ohhhhhh right, missed that. I was imagining like `cosign verify` where we print out a JSON blob to STDOUT, where you'd want to return a list of all the valid entries for the CLI front-end to handle.

I think keep it for now!

dlorenc (approved):

haydentherapper:
cc @asraa fyi since you had taken a look at this

asraa:
Oh, thanks Hayden! Yep, that was the TODO added here after adding some verification/refactor https://github.com/sigstore/cosign/pull/1662

I'll make sure my changes apply over this

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,44 +162,74 @@ func VerifyBlobCmd(ctx context.Context, ko sign.KeyOpts, certRef, certEmail, cer
...
+		if err := verifier.VerifySignature(bytes.NewReader([]byte(sig)), bytes.NewReader(blobBytes)); err != nil {
+			continue
 		}
-	}
 
-	// Use the DSSE verifier if the payload is a DSSE with the In-Toto format.
-	if isIntotoDSSE(blobBytes) {
-		verifier = dsse.WrapVerifier(verifier)
+		// verify the rekor entry
+		if err := verifyRekorEntry(ctx, ko, verifier, cert, b64sig, blobBytes); err != nil {
```
asraa, line 220: FYI inside this we make another extraneous call to Rekor to get the TLOG entry UUID (https://github.com/sigstore/cosign/blob/e2017af9d4ed79c3e15d944eb19a818905af47c2/cmd/cosign/cli/verify/verify_blob.go#L326) and perform tlog verification (check inclusion proof etc) inside this. I had this refactored/separated out https://github.com/sigstore/cosign/pull/1662 so I'll rebase and make it even more clean here!

priyawadhwa: thanks @asraa !

PR #1680, opened 2022-03-29
Support in-toto payloadType in verify-attestation

### LINKED ISSUE #1679: Not supporting in-toto DSS payloadType in verify-attestation
**Description**

I've created an in-toto attestation and uploaded it to a image registry. When I tried to use `cosign verify-attestation --policy my.cue ...` the policy written in CUE was not asserted. When debugging it seems to have been ignored because of the `payloadType` check[1] against known policies, and `application/vnd.in-toto+json` not being defined/handled there.

As far as I understand the `application/vnd.in-toto+json` should be the correct DSSE payload type for the in-toto attestation and this should be allowed to assert against a supplied policy.

This is the image reference of the image and attestation:

```sh
$ docker pull zregvart/blank:latest # image
$ docker pull zregvart/blank:sha256-76addd9802ebcdb49bd3beac67d1b6c4ff2514bf867bb02f2b3605f633ba7f3b.att # attestation
```

Given this CUE policy:
```cue
predicate: passed: false
```

Verify attestation should fail, but it doesn't
```sh
$ cosign verify-attestation docker.io/zregvart/blank:latest \
  --key cosign.pub \
  --policy policy-test.cue
Verification for docker.io/zregvart/blank:latest --
{"payloadType":"application/vnd.in-toto+json","payload":"eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjAuMSIsInByZWRpY2F0ZVR5cGUiOiJjb3NpZ24uc2lnc3RvcmUuZGV2L2F0dGVzdGF0aW9uL3YxIiwic3ViamVjdCI6W3sibmFtZSI6ImluZGV4LmRvY2tlci5pby96cmVndmFydC9ibGFuayIsImRpZ2VzdCI6eyJzaGEyNTYiOiI3NmFkZGQ5ODAyZWJjZGI0OWJkM2JlYWM2N2QxYjZjNGZmMjUxNGJmODY3YmIwMmYyYjM2MDVmNjMzYmE3ZjNiIn19XSwicHJlZGljYXRlIjp7IkRhdGEiOiJ7XG4gIFwiX3R5cGVcIjogXCJodHRwczovL2luLXRvdG8uaW8vU3RhdGVtZW50L3YwLjFcIixcbiAgXCJwcmVkaWNhdGVUeXBlXCI6IFwiaHR0cHM6Ly9leGFtcGxlLmNvbS9UZXN0UmVzdWx0L3YxXCIsXG4gIFwic3ViamVjdFwiOiBbXG4gICAge1xuICAgICAgXCJuYW1lXCI6IFwiZG9ja2VyLmlvL3pyZWd2YXJ0L2JsYW5rXCIsXG4gICAgICBcImRpZ2VzdFwiOiB7XG4gICAgICAgIFwic2hhMjU2XCI6IFwiNzZhZGRkOTgwMmViY2RiNDliZDNiZWFjNjdkMWI2YzRmZjI1MTRiZjg2N2JiMDJmMmIzNjA1ZjYzM2JhN2YzYlwiXG4gICAgICB9XG4gICAgfVxuICBdLFxuICBcInByZWRpY2F0ZVwiOiB7XG4gICAgXCJwYXNzZWRcIjogdHJ1ZVxuICB9XG59XG4iLCJUaW1lc3RhbXAiOiIyMDIyLTAzLTI5VDEyOjA0OjA1WiJ9fQ==","signatures":[{"keyid":"","sig":"MEUCIQDIJIRqkvxWyZWczeFwfBYRh2PffJHVQmq+IRszfKqo4QIgLRRSWZ9zghJ6l45OMcdqYYt41snTwe2eBGKmxoCLvl8="}]}
```

The attestation was created using cosign:

```sh
$ cosign attest --key cosign.key --predicate predicate.json docker.io/zregvart/blank:latest
```

[1] https://github.com/sigstore/cosign/blob/ba50ee05c87701f532c63ec6d0fd0d7f3c03d34d/cmd/cosign/cli/verify/verify_attestation.go#L204

zregvart:
#### Summary
Adds support for the in-toto (`application/vnd.in-toto+json`) DSSE payload type in verify-attestation subcommand.

For example, verifying that against a CUE policy:

```sh
$ cosign verify-attestation docker.io/zregvart/blank:latest \
  --key cosign.pub \
  --policy policy-test.cue \
  --type intoto
```

where `policy-test.cue` is:

```cue
predicate: passed: true
```

Signed-off-by: Zoran Regvart <zoran@regvart.com>

#### Ticket Link

Fixes #1679

#### Release Note

```release-note
Support for `--policy intoto` in `cosign verify-attestation`
```

### DIFF
--- a/cmd/cosign/cli/options/predicate.go
+++ b/cmd/cosign/cli/options/predicate.go
@@ -33,6 +33,7 @@ const (
 	PredicateSPDX   = "spdx"
 	PredicateLink   = "link"
 	PredicateVuln   = "vuln"
+	PredicateInToto = "intoto"
 )
 
 // PredicateTypeMap is the mapping between the predicate `type` option to predicate URI.
@@ -42,6 +43,7 @@ var PredicateTypeMap = map[string]string{
 	PredicateSPDX:   in_toto.PredicateSPDX,
 	PredicateLink:   in_toto.PredicateLinkV1,
 	PredicateVuln:   attestation.CosignVulnProvenanceV01,
+	PredicateInToto: in_toto.PayloadType,
 }
 
 // PredicateOptions is the wrapper for predicate related options.
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -259,6 +259,15 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 				if err != nil {
 					return fmt.Errorf("error when generating SPDXStatement: %w", err)
 				}
+			case options.PredicateInToto:
+				var statement in_toto.Statement
+				if err := json.Unmarshal(decodedPayload, &statement); err != nil {
+					return fmt.Errorf("unmarshal in_toto.Statement: %w", err)
+				}
+				payload, err = json.Marshal(statement)
+				if err != nil {
+					return fmt.Errorf("error when generating in_toto.Statement: %w", err)
+				}
 			}
 
 			if len(cuePolicies) > 0 {


cmd/cosign/cli/options/predicate.go
```diff
@@ -42,6 +43,7 @@ var PredicateTypeMap = map[string]string{
 	PredicateSPDX:   in_toto.PredicateSPDX,
 	PredicateLink:   in_toto.PredicateLinkV1,
 	PredicateVuln:   attestation.CosignVulnProvenanceV01,
+	PredicateInToto: in_toto.PayloadType,
```
lcarva, line 46: I don't think this is correct. A payload is not the same thing as a predicate. IMO, the "sanity checks" in `verify_attestation.go` are wrong and will never pass. See the first commit in https://github.com/sigstore/cosign/pull/1672.

zregvart: Yes I think you are right, there seems to be an underlying semantic discrepancy here (like in #1249 and #1248), this was a quick first pass at getting the functionality I needed. The code here doesn't seem to be used outside of the verify-attestation code path, so it seemed safe to modify even if it is patchy.

priyawadhwa: we could check if an attestation is in DSSE format (using smth like [this](https://github.com/sigstore/cosign/blob/ba50ee05c87701f532c63ec6d0fd0d7f3c03d34d/cmd/cosign/cli/verify/verify_blob.go#L418)) & then go through a different flow for verification since it doesn't clearly map to the current predicate checks?

zregvart: @priyawadhwa that makes sense I can rework this PR if this is the way to go
@lcarva I tried running with the changes from #1672 and sadly that doesn't help for CUE case as well, if I run with a CUE policy that should deem the attestation invalid (i.e. change from `true` to `false` above) the verification still passes.

zregvart:
I'm closing this in favor of #1685 that'll add tests and I think the changes in #1672 can be based on that as well.

Issue #1642, opened 2022-03-21
[cosigned] Enforce authority key errors 


DennyHoang:
**Description**

With the implementation of https://github.com/sigstore/cosign/pull/1623, we currently only warn on errors and do not actually error out when there are issues with getting and using authority keys so that it does not introduce breaking behavior.

When we enforce the new clusterimagepolicy resource and the authority keys, we should update [these lines](https://github.com/sigstore/cosign/pull/1623/files#diff-506dbced92e7841cdd78058e329145cd4c554d7b7a61837db031cb128e9d73dfR148) to add to `errs` instead of a Warn log.

Tracking that we should enforce fail on errors once full migration to using clusterimagepolicy is done. 

cc: @hectorj2f @vaikas

hectorj2f:
@DennyHoang I am not sure. Why don't we attempt to parse the content of  KeyRef.Data into a ecdsa.PublicKey when creating/update the ClusterImagePolicy resource ? That way we won't need to parse it when validating every single PodSpec.

DennyHoang:
Closing. 
Issue is no longer a problem as errors are propagated up now.

PR #1681, opened 2022-03-29
Early parse pubkey

DennyHoang:
Addressing https://github.com/sigstore/cosign/issues/1642#issuecomment-1079069545

Summary
Move authority key ecdsa public key parsing to reconciler. 
Also store the public key into the configmap during initial parsing so that we do not redo the parsing for each time the key data was used. 

Later improvement, we could possibly remove Authority.Key.Data from configmap. But doing incremental changes first.


vaikas:
By adding them to the ClusterImagePolicy CRD they now become part of the exposed API. I'm not sure we want to do that, as in do we want folks to be able to specify public keys like this?

If you are (and I think you are?) the goal is to simplify the 'internal' as in the ConfigMap representation between the reconciler and the webhook, I think that makes sense and as you point out in a follow on work comment, we could simplify some of the fields in the internal struct (no need for secret ref for example, since that has been resolved). However, I do wonder if the 'external' and 'internal' start to diverge if we should create a different internal representation that's only used between the reconciler/webhook only and it's not exposed to the end users since it's an implementation detail.

So, I guess my preference would be to not expose this in the external API, but instead define the internal API :) Does that make sense?

DennyHoang:
> So, I guess my preference would be to not expose this in the external API, but instead define the internal API :) Does that make sense?

@vaikas Ye, makes sense. Just not 100% atm on how to implement an "internal" API portion and how to separate the existing external API. I did it in the reconciler initially to just put it as close to the existing logic for when secrets were handled. I was initially considering of possibly storing the publicKey in `store.go:Config` alongside `ImagePolicyConfig` such that during `ToContext` or instantiation of the `Config`, I could parse and store the keys there and then.

How would it be best to approach "separating" internal and external API?

vaikas:
I would probably add a new struct with all the json tags so it serializes/unserializes cleanly just like we have for the existing clusterimagepolicy.Spec.Authority and strip out the pieces that the internal does not need. You could add that struct in the say pkg/reconciler/cosigned or something like that. Then here:
https://github.com/sigstore/cosign/blob/main/pkg/reconciler/clusterimagepolicy/resources/configmap.go#L52
Where we serialize the existing CIP, you'd use the internal representation instead of marshalling the entry into the configmap.
Probably right after this:
https://github.com/sigstore/cosign/blob/main/pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go#L57
You could convert from CIP to internal and be pretty good to go?

And here it gets patched (so, replaces the CIP entry, now with the internal one instead of the external one).
https://github.com/sigstore/cosign/blob/main/pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go#L92

And then of course the store would need to use the internal vs. external api.

Does any of that make sense? Top level idea is just to define a struct that we can serialize/unserialize and replace places where we use CIP to use that instead.

DennyHoang:
Makes sense, I will give that a try tomorrow. The only notable thing I was wondering was if I would still encounter the codegen issue that is currently failing. 

```
# github.com/sigstore/cosign/pkg/apis/cosigned/v1alpha1
pkg/apis/cosigned/v1alpha1/zz_generated.deepcopy.go:201:10: (*in).DeepCopyInto undefined (type *ecdsa.PublicKey has no field or method DeepCopyInto)
```

I see a `deepcopy` when creating a patch at the moment. So if i switch out the CIP entry to the internal one, it would still have deepcopy. Thus lead to that undefined deepcopy that I am not sure how to best resolve with my current understanding of codegen.

vaikas:
I don't think we'd need a deepcopy (I might be forgetting something) for the internal strruct. The reason why the deepcopy is needed for the external version is because it's being modified in places so we do not want to modify the informers copy of the CIP that's handed to us => we make a copy with Deepcopy.  Well, and it has to be for codegen purposes, but we shouldn't need codegen. Just a struct that we can marshal/unmarshal into string that we then slap into the configmap as a string.

So, I think we'd have (very high level).
 * Construct the CIP we want (for example, resolve the secrets, KMS, etc. into inlined data) in the ReconcileKind. So we'll deepcopy and modify 'our' copy of the CIP and get it ready to be used internally.
 * From the CIP above, construct the 'internal' data struct
 * Marshal the internal datastruct into a string and update the Configmap entry with CIP.Name=>'marshalled string'
 * Now, to create the patch, you're right we use deepcopy, but that deepcopy is on the ConfigMap, and since our entries there are just serialized 'internal' structs, and just strings, that will not be a problem.

So, there should be no for deepcopy, since only operating on strings, but as I said mayhaps I am forgetting something.

pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
```diff
@@ -119,7 +133,64 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, cip *v1alpha1.ClusterImag
...
+	if err != nil {
+		if !apierrs.IsNotFound(err) {
+			logging.FromContext(ctx).Errorf("Failed to get configmap: %v", err)
+		}
+	} else if err := r.removeCIPEntry(ctx, existing, cipName); err != nil {
+		logging.FromContext(ctx).Errorf("Failed to get configmap: %v", err)
+	}
+}
+
+func parseAuthorityKeys(ctx context.Context, pubKey string) ([]*ecdsa.PublicKey, error) {
```
hectorj2f, line 167: nit: this funcion does more than parsing, i'd rename it to convertAuthorityKeys

pkg/reconciler/cosigned/clusterimagepolicy_internal.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2022 The Sigstore Authors.
```
hectorj2f, line 1: Suggestion: If this version is internal, it could go under `pkg/api/cosigned` where the rest of version are located. I prefer that rather than in the reconciler.

DennyHoang: Agreed, I just pushed this the moment I got it working to get early reviews before more potential changes

DennyHoang: Hmm, for now, I moved it to `pkg/api/cosigned/private`
I couldn't use `pkg/api/cosigned/internal`
Reason explained here I believe:
https://stackoverflow.com/questions/59342373/use-of-internal-package-not-allowed

pkg/apis/config/image_policies.go
```diff
@@ -41,13 +41,13 @@ type ImagePolicyConfig struct {
 	// in the Secret. So it seemed unnecessary to mirror those all over.
 	// Key in the map is the name of the ClusterImagePolicy where the Policy
 	// was received from.
-	Policies map[string]v1alpha1.ClusterImagePolicySpec
+	Policies map[string]internalcip.ClusterImagePolicySpec
```
vaikas: You can now probably remove my TODO since you did it :) Thanks!

pkg/reconciler/cosigned/clusterimagepolicy_internal.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// ClusterImagePolicy defines the images that go through verification
+// and the authorities used for verification.
+// This is the internal representation of the external v1alpha1.ClusterImagePolicy.
+// KeyRef does not store secretRefs in internal representation.
+// KeyRef does store parsed publicKeys from Data in internal representation.
+
+type ClusterImagePolicy struct {
+	metav1.TypeMeta   `json:",inline"`
```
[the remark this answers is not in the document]
DennyHoang: Yup, sounds good. With removing the `TypeMeta` and `ObjectMeta`, just had to change some method signatures to accept `cipName` string as the `ClusterImagePolicy` struct no longer had that inherited field

internal/pkg/apis/cosigned/clusterimagepolicy_types.go
```diff
@@ -0,0 +1,65 @@
...
+	// +optional
+	CTLog *v1alpha1.TLog `json:"ctlog,omitempty"`
+}
+
+// This references a public verification key stored in
+// a secret in the cosign-system namespace.
+type KeyRef struct {
+	// Data contains the inline public key
+	// +optional
+	Data string `json:"data,omitempty"`
```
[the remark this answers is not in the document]
hectorj2f: If we don't need it, let's remove it.

pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
```diff
@@ -121,15 +133,14 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, cip *v1alpha1.ClusterImag
...
+	return r.removeCIPEntry(ctx, existing, cip.Name)
 }
 
 // convertPublicKeys will go through the CIP and try to convert key data
 // to ecdsa.PublicKey. Makes a copy of the CIP before modifying it and
 // returns the copy
-func (r *Reconciler) convertKeyData(ctx context.Context, cip *v1alpha1.ClusterImagePolicy) (*v1alpha1.ClusterImagePolicy, error) {
-	ret := cip.DeepCopy()
-	for _, authority := range ret.Spec.Authorities {
+func (r *Reconciler) convertKeyData(ctx context.Context, cip *internalcip.ClusterImagePolicy) (*internalcip.ClusterImagePolicy, error) {
```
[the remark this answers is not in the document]
DennyHoang: Yup, updated

pkg/apis/config/image_policies.go
DennyHoang: Awesome! removed

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -268,27 +268,22 @@ func ValidatePolicy(ctx context.Context, ref name.Reference, kc authn.Keychain,
...
-					continue
-				} else {
-					if len(sps) > 0 {
-						logging.FromContext(ctx).Debugf("validated signature for %s, got %d signatures", len(sps))
-						return sps, nil
-					}
-					logging.FromContext(ctx).Errorf("no validSignatures found for %s", ref.Name())
-					authorityErrors = append(authorityErrors, fmt.Errorf("no valid signatures found for %s", ref.Name()))
+				if len(sps) > 0 {
+					logging.FromContext(ctx).Debugf("validated signature for %s, got %d signatures", len(sps))
```
hectorj2f, line 282: i believe you're missing one argument here.

DennyHoang: Oh, there is another instance below as well, ill add it there as well

internal/pkg/apis/cosigned/clusterimagepolicy_types.go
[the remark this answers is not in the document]
DennyHoang: Discussed an approach with @hectorj2f , going to try to make that change.
- dont use unmarshal/marshal
- have a Convert function that does the conversion between external to internal CIP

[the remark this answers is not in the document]
vaikas: I'm fine with actually checking this in, and improving the conversion function in the follow on PRs, since:
 * That's all going to be internal to the reconciler
 * API won't change (Data will be dropped, is not used).
So, up to you if you want to add to this PR, or just merge this as is and then work on the conversion in a follow up?

vaikas (approved):

vaikas:
@hectorj2f @DennyHoang do you want to merge this as is and do the conversion differently / drop the Data field in a follow on PR?

DennyHoang:
Oops, gotta fix some tests after the rebase to upstream

PR #1685, opened 2022-03-30
Add e2e test for attest / verify-attestation

vaikas:

#### Summary
Add an e2e test for validating creation of
an attestation and then validating it against a cue policy that will work and one that
doesn't work. Now, This is the example that we use in our documentation, and it does
now pass. I reckon with these tests we can add few more of these test cases with
different payload types and keep validating them by adding them in the
test/testdata/policies. Just a beginning.

#### Ticket Link

Fixes

#### Release Note

```release-note
```

### DIFF
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -1,0 +1,194 @@
+# Copyright 2022 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


cpanato (approved):
that is cool, thanks

dlorenc (approved):

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -195,22 +195,6 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
-			if !ok {
-				return fmt.Errorf("invalid predicate type: %s", c.PredicateType)
-			}
-
-			// sanity checks
-			if val, ok := payloadData["payloadType"]; ok {
-				// we need to check only given type from the cli flag
-				// so we are skipping other types
-				if predicateURI != val {
-					continue
```
lcarva, line 208: I think removing this check altogether causes the `--type` parameter to be completely ignored and the policy to be applied to all attestations found. My reading of the code is that this is not intended. I tried a different take in resolving this issue in the first commit of https://github.com/sigstore/cosign/pull/1672.

PR #1694, opened 2022-03-31
verify: remove extra calls to rekor for verify and verify-blob

asraa:

#### Summary

Removes extra calls to Rekor during verification:
1) When verifying a blob, we retrieve the entry UUIDs (by an artifact hash search), and then call `verifyRekorEntry`, which then takes the (cert, pubkey, artifact) and makes a call to rekor to get the entry. Instead, we pass the entry since we already have it.

2) Distangles VerifyTLogEntry from inside of FindTlogEntry. Users can call this directly. This means that we don't need to `FindTlogEntry` to verify it if we already have it, which saves the call from (1).

3) In both `verify` and `verify-blob`, we call `verifyRekorEntry`, which has ANOTHER call to `GetTlogEntry` after `FindTlogEntry` to get the actual entry, we already got it! Refactored `FindTlogEntry` to give you back the... entry.

4) Move the UUID verification into functions that are asking for an entry by UUID (e.g. when you get or find an entry)

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -23,6 +23,7 @@ import (
 	_ "crypto/sha256" // for `crypto.SHA256`
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


dlorenc (approved):

asraa:
cc @rohankh532 this makes `VerifyTLogEntry` public, see the verify-blob commands for usage

priyawadhwa (approved):
nice!

Issue #1717, opened 2022-04-07
cosigned: fail to parse configMap content with the policies


hectorj2f:
**Description**

Cosigned is failing to parse the content of the ConfigMap with all the public keys with the following error:

```
[...]

PR #1719, opened 2022-04-07
[Cosigned] Fix publicKey unmarshal

DennyHoang:

#### Summary
Unmarshaling PublicKeys causes an error:
```
{"severity":"ERROR","timestamp":"2022-04-06T14:33:44.2084731Z","logger":"cosigned.config-store","caller":"configmap/store.go:148","message":"Error updating image-policies config \"config-image-policies\": \"failed to parse the entry <ENTRY> : math/big: cannot unmarshal \\\"2.7708706494268807e+76\\\" into a *big.Int\"","commit":"9008111","stacktrace":"<STACKTRACE>"}
```

#### Ticket Link
Fixes https://github.com/sigstore/cosign/issues/1717

#### Notes
I am aware I am temporarily duplicating the function for parsing PEMs into the internal clusterimagepolicy_types file.
I have another branch that will address this alongside consolidating a lot of the clusterimagepolicy logic into the types file.

cc: @hectorj2f @vaikas @coyote240

### DIFF
--- a/.github/workflows/kind-cluster-image-policy.yaml
+++ b/.github/workflows/kind-cluster-image-policy.yaml
@@ -120,9 +120,9 @@ jobs:
         echo Created image $demoimage2
         popd
 
-    - name: Deploy ClusterImagePolicy
+    - name: Deploy ClusterImagePolicy With Keyless Signing
       run: |
-        kubectl apply -f ./test/testdata/cosigned/e2e/cip.yaml
+        kubectl apply -f ./test/testdata/cosigned/e2e/cip-keyless.yaml
 
     - name: Sign demoimage with cosign
       run: |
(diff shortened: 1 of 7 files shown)


vaikas:
To make sure that we do not regress in the future, would you mind adding a test with public keys (existing only tests signing with Fulcio so this was not caught by the existing tests).
https://github.com/sigstore/cosign/blob/main/.github/workflows/kind-cluster-image-policy.yaml#L127

hectorj2f (approved):

.github/workflows/kind-cluster-image-policy.yaml
```diff
@@ -149,7 +149,49 @@ jobs:
...
+    - name: Generate New Signing Key
+      run: |
+        COSIGN_PASSWORD="" ./cosign generate-key-pair
+
+    - name: Deploy ClusterImagePolicy With Key Signing
+      run: |
+        yq '. | .spec.authorities[0].key.data |= load_str("cosign.pub")' ./test/testdata/cosigned/e2e/cip-key.yaml | \
+          kubectl apply -f -
+
+    - name: Sign demoimage with cosign-test key
```
vaikas, line 169: Awesomesauce! Thanks for doing this. I wonder if we should / could add one more test above this before signing.
Create the CIP with the key, this means that all CIP must be met for the image to be admitted.
Check that the image is rejected. This is now a neat test because it also tests the AND condition of the CIP. It passed above, we added a CIP that should now be rejected. 
Then sign.

So, I guess right before this step, add a test that roughly makes sure that demoimage is rejected.

Something like this:
```
      - name: Verify with two CIP one not signed with public key
      run: |
        if kubectl create -n demo-key-signing job demo --image=${{ env.demoimage }} ; then
          echo Failed to block unsigned Job creation!
          exit 1
```
What do you think?

DennyHoang: Added! Thanks for suggestion!

vaikas (approved):
Awesomesauce! Thanks!

PR #1726, opened 2022-04-08
[cosigned] The webhook name is now configurable via --webhook-name flag

vpnachev:
#### Summary
The webhook name is now configurable via --webhook-name flag for cosigned. This way, cosigned can be deployed multiple times in the same cluster with different webhook configurations, e.g. namespace and/or object selectors. Also, each webhook can be served by different instance of the cosigned using different public keys.

#### Ticket Link

Fixes

#### Release Note

```release-note
The name of the cosigned webhook can be now configured via the flag `--webhook-name`, the default value is `cosigned.sigstore.dev`.
```

### DIFF
--- a/cmd/cosign/webhook/main.go
+++ b/cmd/cosign/webhook/main.go
@@ -47,7 +47,7 @@ var secretName = flag.String("secret-name", "", "The name of the secret in the w
 // webhookName holds the name of the validating webhook to set up with the
 // types we are watching.  If this changes, you must also change:
 //    ./config/500-webhook-configuration.yaml
-const webhookName = "cosigned.sigstore.dev"
+var webhookName = flag.String("webhook-name", "cosigned.sigstore.dev", "The name of the webhook.")
 
 func main() {
 	opts := webhook.Options{
@@ -92,7 +92,7 @@ func NewValidatingAdmissionController(ctx context.Context, cmw configmap.Watcher
 
 	return validation.NewAdmissionController(ctx,
 		// Name of the resource webhook.
-		webhookName,
+		*webhookName,
 
 		// The path on which to serve the webhook.
 		"/validations",
@@ -123,7 +123,7 @@ func NewMutatingAdmissionController(ctx context.Context, cmw configmap.Watcher) 
 
 	return defaulting.NewAdmissionController(ctx,
 		// Name of the resource webhook.
-		webhookName,
+		*webhookName,
 
 		// The path on which to serve the webhook.
 		"/mutations",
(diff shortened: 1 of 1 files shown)


dlorenc (approved):

cmd/cosign/webhook/main.go
```diff
@@ -92,7 +92,7 @@ func NewValidatingAdmissionController(ctx context.Context, cmw configmap.Watcher
 
 	return validation.NewAdmissionController(ctx,
 		// Name of the resource webhook.
-		webhookName,
+		*webhookName,
```
hectorj2f, line 95: Note: If you change this, you will to change on the scripts too.

vpnachev: Do you mean the helm [templates](https://github.com/sigstore/helm-charts/blob/main/charts/cosigned/templates/webhook/webhook_mutating.yaml) and [config files](https://github.com/sigstore/cosign/blob/main/config/500-webhook-configuration.yaml)? If yes, I rely on this comment https://github.com/sigstore/cosign/blob/56af7f3f51556cb2374d6556ad08e5a2b1d764a2/cmd/cosign/webhook/main.go#L47-L49 to get the attention to whomever tries to change it. Do you suggest to enhance the help string of the flag?

hectorj2f: Yes, we'll need to update the comment in that piece of code accordingly. 
Also we'll need to add that field in the values.yaml of the helm-chart, so everything gets changed based on the value of the field. I could take care of that.

vpnachev: I have updated the comment and the flag doc with 62e602886e8101ae2768e1395662cc8de10e26ad. 
Looking in the webhook initialization in [1] and [2] there is a lot of configurations that can be injected, but also skipped in different conditions, or even fail when URL is used instead of in-cluster service.  I doubt I can describe all of them with just few words, and I don't think here is the right place for detailed documentation. 
If you have better suggestion/wording, please propose it.  

[1] https://github.com/knative/pkg/blob/e325df66cb51d341b5d1142fa42f75539dfe290a/webhook/resourcesemantics/validation/reconcile_config.go#L99 
[2] https://github.com/knative/pkg/blob/e325df66cb51d341b5d1142fa42f75539dfe290a/webhook/resourcesemantics/defaulting/defaulting.go#L170

hectorj2f: We'll get some documentation for cosigned new changes soon.

dlorenc:
@hectorj2f is this ready to merge?

hectorj2f:
@dlorenc Yes.
@vpnachev Could you add the flag and respective changes to the webhook configuration if the flag is set in sigstore/helm-charts/cosigned ?

vpnachev:
Yes, I can open a PR in the next days about it.

PR #1727, opened 2022-04-08
Make public all types required to use ValidatePolicy

jdolitsky:
Unable to use `ValidatePolicy` without access to the `Authority` type (and subtypes)


vaikas:
Sweet! Now we can reuse the validation of policies. Need to just expose the Convert from External to Webhook.

hectorj2f:
@jdolitsky @vaikas Question: Why do we want to make an internal-only resource type available ? Why don't we use the external type v1alpha1 ?

vaikas:
Great question! The motivation for the CIP work was two fold:
 1. User facing way to configure policies (aka, the ClusterImagePolicy CRD)
 2. Expose a SDK for other tooling (for example Kyverno, or even cosign the CLI command, etc.).

For the second point, the ValidatePolicy method was exposed, but the argument takes the 'normalized' variant, that has everything that can be resolved into public keys so that the method doesn't need access to secrets, etc. So we needed to expose the struct that the method takes so that it can be reused, we also need to expose a method that takes a CIP CRD and converts to the normalized variant.
Does that make sense @hectorj2f

hectorj2f:
@vaikas Thanks for the comment. It makes sense.

PR #1736, opened 2022-04-11
[Cosigned] Convert functions for webhookCIP from v1alpha1

DennyHoang:

#### Summary
Continual improvement to webhookCIP
- Less reliance on marshal/unmarshal to allow for more customization and streamlined conversion between CIP and the webhookCIP representation
- KeylessRef.CACert now uses the webhookCIP representation of KeyRef
- ~~Logic for inlinedData from secretRef and KMS moved from reconciler file to the `clusterimagepolicy_types` for consolidation and removing bulk in reconciler file~~
- No new tests were added at the moment because the `clusterimagepolicy_test` already had decent coverage for all the inlining data etc

### DIFF
--- a/pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
+++ b/pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
@@ -21,6 +21,8 @@ import (
 	"encoding/pem"
 
 	"github.com/sigstore/cosign/pkg/apis/cosigned/v1alpha1"
+
+	"knative.dev/pkg/apis"
 )
 
 // ClusterImagePolicy defines the images that go through verification
@@ -37,7 +39,7 @@ type Authority struct {
 	// +optional
 	Key *KeyRef `json:"key,omitempty"`
 	// +optional
-	Keyless *v1alpha1.KeylessRef `json:"keyless,omitempty"`
+	Keyless *KeylessRef `json:"keyless,omitempty"`
 	// +optional
 	Sources []v1alpha1.Source `json:"source,omitempty"`
 	// +optional
@@ -50,13 +52,19 @@ type KeyRef struct {
 	// Data contains the inline public key
 	// +optional
 	Data string `json:"data,omitempty"`
-	// KMS contains the KMS url of the public key
-	// +optional
-	KMS string `json:"kms,omitempty"`
 	// PublicKeys are not marshalled because JSON unmarshalling
 	// errors for *big.Int
 	// +optional
 	PublicKeys []*ecdsa.PublicKey `json:"-"`
+}
+
+type KeylessRef struct {
+	// +optional
+	URL *apis.URL `json:"url,omitempty"`
+	// +optional
+	Identities []v1alpha1.Identity `json:"identities,omitempty"`
+	// +optional
+	CACert *KeyRef `json:"ca-cert,omitempty"`
 }
 
 // UnmarshalJSON populates the PublicKeys using Data because
@@ -73,7 +81,7 @@ func (k *KeyRef) UnmarshalJSON(data []byte) error {
 	k.Data = ret["data"]
 
 	if ret["data"] != "" {
-		publicKeys, err = convertKeyDataToPublicKeys(ret["data"])
+		publicKeys, err = ConvertKeyDataToPublicKeys(ret["data"])
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -15,12 +15,25 @@
...
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
+	"fmt"
+	"strings"
 
 	"github.com/sigstore/cosign/pkg/apis/cosigned/v1alpha1"
+	"github.com/sigstore/cosign/pkg/apis/utils"
```
hectorj2f, line 28: nit: separate the groups of imports as groups (sigstore ones, external, golang packages).

pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -37,7 +50,7 @@ type Authority struct {
 	// +optional
 	Key *KeyRef `json:"key,omitempty"`
 	// +optional
-	Keyless *v1alpha1.KeylessRef `json:"keyless,omitempty"`
```
hectorj2f, line 40: suggestion: I am considering whether we should create again all the resources again. As a consequence, we should create the type for ImagePattern, Sources and Identities here to be consistent.

hectorj2f: @vaikas wdyt?

hectorj2f (commented):
lgtm, so far.

pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
vaikas: Consistency is good indeed :) So, I'd be down for recreating them here, or always use one from the CRD.

DennyHoang: I kept the other ones as imports from the CRD because we arent currently manipulating it at the moment and it is a direct 1:1 mapping. 
We actually manipulate keyRefs so that is why Authorities and key[less]Refs were re-implemented as required.

When the other properties diverge, I feel that updating it then makes more sense to keep potential updates in one place as Ville mentions until that diverging.

pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -84,9 +106,103 @@ func (k *KeyRef) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-func convertKeyDataToPublicKeys(pubKey string) ([]*ecdsa.PublicKey, error) {
+func ConvertClusterImagePolicyV1alpha1ToWebhook(ctx context.Context, in *v1alpha1.ClusterImagePolicy, rtracker tracker.Interface, secretLister corev1listers.SecretLister) (*ClusterImagePolicy, error) {
```
vaikas, line 109: I'd rather not have Tracker / SecretLister exposed here. I'm envisioning using these functions for example, from Cosign the CLI for various reasons, so having the Tracker exposed or secretlister doesn't make sense.
I would rather have this be a three pass operation from the Reconciler:
 1. Resolve the inline references
 2. Convert to Webhook.

Right now I can't do the second one from CLI (or maybe from other users of this library) because I don't have secretlisters or trackers. So, I'd rather have a Convert function that _only_ converts to inline webhook data with no side effects.

So, I guess I'd rather have a method in Reconciler that handles the CIP->CIP-With-References-Inlined as a method that can take trackers / listers as necessary.
Then, the second step would just take the CIP-With-References-Inlined and this has no side effects at all, and is re-useable from other places as a standalone method that will then allow those places to call the ValidatePolicy.

Thoughts?

hectorj2f: If we see those functions been used somewhere else, then lets keep them where they can be reused.

hectorj2f: > , I guess I'd rather have a method in Reconciler that handles the CIP->CIP-With-References-Inlined as a method that can take trackers / listers as necessary.

I like this option.

pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -84,9 +92,59 @@ func (k *KeyRef) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-func convertKeyDataToPublicKeys(pubKey string) ([]*ecdsa.PublicKey, error) {
-	keys := []*ecdsa.PublicKey{}
+func ConvertClusterImagePolicyV1alpha1ToWebhook(in *v1alpha1.ClusterImagePolicy) *ClusterImagePolicy {
+	copyIn := in.DeepCopy()
```
vaikas, line 96: nit, I think this is not stricly necessary because we don't modify, but since we use the images from L105 as inputs, can't think right now if that will make a difference.

vaikas (approved):
Thanks for cleaning up the marshal/nonmarshal stuff woot woot.

PR #1748, opened 2022-04-12
Use bundle log ID to find verification key

haydentherapper:
Instead of iterating over all log verification keys, we use the log ID
in the bundle to find the correct key. This avoids iterating over all
trusted keys.

Note that the log ID is not signed. If the offline bundle is manipulated
so that the log ID does not resolve to a valid key, this will not result
in a denial of service attack, because cosign will fall back to fetching
the entry from the online log.

The equivalent change for CT entry verification has been made in another
PR.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


haydentherapper:
cc @asraa

dlorenc (approved):

PR #1750, opened 2022-04-13
test: create fake TUF test root and create test SETs for verification

asraa:
#### Summary

* Adds a fake test TUF root set up that can be used to configure test rekor and fulcio signing keys
* Completes a fake SET test using a test TUF root

cc @haydentherapper 

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/cosign/tuf/testutils.go
+++ b/pkg/cosign/tuf/testutils.go
@@ -33,7 +33,7 @@ import (
 )
 
 type TestSigstoreRoot struct {
-	Rekor             signature.SignerVerifier
+	Rekor             signature.Verifier
 	FulcioCertificate *x509.Certificate
 	// TODO: Include a CTFE key if/when cosign verifies SCT.
 }
--- a/pkg/cosign/tuf/testutils.go
+++ b/pkg/cosign/tuf/testutils.go
@@ -1,0 +1,126 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tuf
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	"github.com/sigstore/sigstore/pkg/signature"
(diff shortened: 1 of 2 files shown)


pkg/cosign/tuf/testutils.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := r.Commit(); err != nil {
+		t.Error(err)
+	}
+	// Serve remote repository.
+	s := httptest.NewServer(http.FileServer(http.Dir(filepath.Join(td, "repository"))))
+	defer s.Close()
+
+	// Initialize with custom root.
+	tufRoot := t.TempDir()
+	t.Setenv("TUF_ROOT", tufRoot)
```
haydentherapper, line 113: Do we need to unset the environment variable via a defer, or does the test automatically clean it up?

pkg/cosign/tuf/testutils.go
```diff
@@ -0,0 +1,126 @@
...
+		t.Error(err)
+	}
+	if err := r.Timestamp(); err != nil {
+		t.Error(err)
+	}
+	if err := r.Commit(); err != nil {
+		t.Error(err)
+	}
+	// Serve remote repository.
+	s := httptest.NewServer(http.FileServer(http.Dir(filepath.Join(td, "repository"))))
```
haydentherapper, line 108: nit: Kinda outside the scope of this PR cause I think it'd require a more significant refactoring, but it'd be nice if there was a way to initialize the TUF metadata without an outbound network call. I don't love having to stand up HTTP servers in unit tests.

pkg/cosign/tuf/testutils.go
```diff
@@ -0,0 +1,126 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/theupdateframework/go-tuf"
+)
+
+type TestSigstoreRoot struct {
```
haydentherapper, line 35: nit: Should we store the private key of the certificate too? I think we should either have the TestRoot struct only hold the public keys (so a `signature.Verifier`), or also hold the private keys so it can be passed around.

haydentherapper (approved):
Great work, this is excellent! I'll add the tests for the CT log once this is merged!

pkg/cosign/tuf/testutils.go
asraa: Good question! t.Setenv will already clean it up when the test is complete. https://pkg.go.dev/testing#B.Setenv

pkg/cosign/tuf/testutils.go
asraa: FWIW it's a local server, but I agree -- t we also wanted to keep a remote cache URL/bucket information, so pulling out the remote as a parameter to `Initialize` wouldn't let the TufClient have that info. 

However, we can just support local filesystem roots as "mirrors". Tests can provide the path to the root.  

https://github.com/sigstore/cosign/issues/1753

Will do as a follow up!

pkg/cosign/tuf/testutils.go
asraa: Since they're tests meant to create fake signed data, they'll already have access to private key material, so I'm indifferent on this. For consistency I'll make it the public root material. If we find it easier to keep the private material we can change that.

dlorenc (approved):

Issue #1651, opened 2022-03-23
Implement CIP.Source 


vaikas:
This is breaking apart a tracking task for wiring / using the Sources in Authority section for ClusterImagePolicy.

DennyHoang:
Going to take a crack at this

hectorj2f:
@DennyHoang Nice!

PR #1754, opened 2022-04-13
Cosigned validate against remote sig src

DennyHoang:

#### Summary
- Implementing Authority.Sources to allow verification against signatures in different registry locations
- Also noticed a bug where we are not creating remoteOptions with the keychain and therefore could not verify against private repositories
- RemoteOpts follow the same unmarshalJSON override as publicKeys. This is due to the remoteOption type not being supported by json [un]marshal
- RemoteOpts are parsed and stored as a part of the CIP config to reduce parsing within the validation loop

#### Ticket Link
Fixes https://github.com/sigstore/cosign/issues/1651

#### Release Note
```release-note
* Specifying remote source for signatures in cosigned
```

### DIFF
--- a/pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation_test.go
+++ b/pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation_test.go
@@ -397,7 +397,57 @@ func TestAuthoritiesValidation(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:      "Should pass when source oci is present",
+			expectErr: false,
+			policy: ClusterImagePolicy{
+				Spec: ClusterImagePolicySpec{
+					Images: []ImagePattern{{Regex: ".*"}},
+					Authorities: []Authority{
+						{
+							Key:     &KeyRef{KMS: "kms://key/path"},
+							Sources: []Source{{OCI: "registry.example.com"}},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:        "Should fail when source oci is empty",
+			expectErr:   true,
+			errorString: "missing field(s): spec.authorities[0].source.oci",
+			policy: ClusterImagePolicy{
+				Spec: ClusterImagePolicySpec{
+					Images: []ImagePattern{{Regex: ".*"}},
+					Authorities: []Authority{
+						{
+							Key:     &KeyRef{KMS: "kms://key/path"},
+							Sources: []Source{{OCI: ""}},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:      "Should pass with multiple source oci is present",
+			expectErr: false,
+			policy: ClusterImagePolicy{
+				Spec: ClusterImagePolicySpec{
+					Images: []ImagePattern{{Regex: ".*"}},
+					Authorities: []Authority{
+						{
+							Key: &KeyRef{KMS: "kms://key/path"},
+							Sources: []Source{
+								{OCI: "registry1"},
+								{OCI: "registry2"},
+							},
+						},
+					},
+				},
+			},
+		},
 	}
+
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			err := test.policy.Validate(context.TODO())
--- a/pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
+++ b/pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
@@ -20,7 +20,10 @@ import (
 	"encoding/json"
 	"encoding/pem"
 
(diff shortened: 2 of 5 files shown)


pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation_test.go
```diff
@@ -397,6 +397,61 @@ func TestAuthoritiesValidation(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:        "Should fail when source is empty",
+			expectErr:   true,
+			errorString: "missing field(s): spec.authorities[0].source.oci",
```
hectorj2f, line 442: Umm, but sources could be empty. It could be optionally used.

hectorj2f: we should only used if present.

pkg/cosign/kubernetes/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -44,6 +47,9 @@ type Authority struct {
 	Sources []v1alpha1.Source `json:"source,omitempty"`
 	// +optional
 	CTLog *v1alpha1.TLog `json:"ctlog,omitempty"`
+	// RemoteOpts are not marshalled because they are an unsupported type
+	// +optional
+	RemoteOpts []remote.Option `json:"-"`
```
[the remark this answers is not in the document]
DennyHoang: We keep the sources as a string in `authority.source[].oci`
RemoteOpts are populated during the unmarshal that occurs when `store.go` creates the clusterImagePolicy object from the configmap. We use an overridden `func (a *Authority) UnmarshalJSON(data []byte) error` to then populate the RemoteOpts at that time.

So summary
- Reconciler->CIP-resolved->webhook->configmap will store source[].oci as string
- configmap->webhook in store.go is where we do the pre-processing for remoteOpts

.github/workflows/kind-cluster-image-policy.yaml
```diff
@@ -160,10 +160,12 @@ jobs:
     - name: Generate New Signing Key
       run: |
         COSIGN_PASSWORD="" ./cosign generate-key-pair
+        mv cosign.key cosign-key.key
```
hectorj2f, line 163: @DennyHoang  Is this intentional ?

DennyHoang: Yes, for now, I was going to run `generate-key-pair` multiple times in the workflow file. 
So I move the cosign.key and pub after creation.

There is a lower step where I generate for the remote signing key

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation_test.go
DennyHoang: tl;dr: Test removed.

When I tried to set yaml with
```yaml
spec:
  images:
    - glob: "*"
  authorities:
    - key:
        data: ...
      source:
      - oci: "index.docker.io/dhoang679/nginx-signatures"
      -
```

It responds with:
```
error: error validating "tmp/cip.yaml": error validating data: ValidationError(ClusterImagePolicy.spec.authorities[0].source): unknown object type "nil" in ClusterImagePolicy.spec.authorities[0].source[1]; if you choose to ignore these errors, turn validation off with --validate=false
```

However, this error cant be simulated by unit tests. It did not even enter the validation functions as far as I can tell.

Maybe I misunderstand the `json:"source,omitempty"` as it does not omit when it is an null object from the yaml...

I tried changing the `[]Source` to use a pointer representation `[]*Source` but the unit test with `[]&Source{ nil }` object within did not trigger the same validation error as setting it through yaml.

So went with removing this test for now.

.github/workflows/kind-cluster-image-policy.yaml
hectorj2f: Umm, okay. But I don't see cosign.pub been used later in this script.

DennyHoang: Oh, all the cosign.pub were renamed as either "cosign-remote.pub" or "cosign-key.pub" to match their key counterpart. 

Maybe the name for "cosign-key" file pairing was a bad choice. I could be more descriptive with "cosign-colocated-signing.key" and for the remote one, "cosign-remote-signing.key" and their pub respectively.

One key designated for having signature colocated with image, and the second key for signing with signatures in remote location.

DennyHoang: resolving as I did update those names as mentioned

hectorj2f (approved):

vaikas (approved):
Awesomesauce! Thanks for doing!

PR #1759, opened 2022-04-14
Implement identities, fix bug in webhook validation.

vaikas:

#### Summary
* Implement ClusterImagePolicy for Keyless for identities, so you can restrict which issuer/subject needs to be matched.
* Also fixed the webhook validator logic that actually would prevent being able to specify both the URL as well as identities in keyless, even though it's a valid configuration as per the design spec.
* Add e2e tests for them.

#### Ticket Link

Fixes

#### Release Note

```release-note
Add support for specifying identities (issuer/subject matching) in ClusterImagePolicy
Fix bug that would now allow creation of CIP with identities
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -27,9 +27,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
+	"github.com/sigstore/cosign/pkg/apis/cosigned/v1alpha1"
 	cbundle "github.com/sigstore/cosign/pkg/cosign/bundle"
 	"github.com/sigstore/cosign/pkg/cosign/tuf"
 
@@ -84,6 +86,11 @@ type CheckOpts struct {
 
 	// SignatureRef is the reference to the signature file
 	SignatureRef string
+
+	// Identities is an array of Identity (Subject, Issuer) matchers that have
+	// to be met for the signature to ve valid.
+	// Supercedes CertEmail / CertOidcIssuer
+	Identities []v1alpha1.Identity
 }
 
 func getSignedEntity(signedImgRef name.Reference, regClientOpts []ociremote.Option) (oci.SignedEntity, v1.Hash, error) {
@@ -143,7 +150,7 @@ func verifyOCIAttestation(_ context.Context, verifier signature.Verifier, att pa
 }
 
 // ValidateAndUnpackCert creates a Verifier from a certificate. Veries that the certificate
-// chains up to a trusted root. Optionally verifies the subject of the certificate.
+// chains up to a trusted root. Optionally verifies the subject and issuer of the certificate.
 func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Verifier, error) {
 	verifier, err := signature.LoadVerifier(cert.PublicKey, crypto.SHA256)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/cosign/verify.go
```diff
@@ -84,6 +86,11 @@ type CheckOpts struct {
 
 	// SignatureRef is the reference to the signature file
 	SignatureRef string
+
+	// Identities is an array of Identity (Subject, Issuer) matchers that have
+	// to be met for the signature to ve valid.
+	// Supercedes CertEmail / CertOidcIssuer
```
mattmoor, line 92: Can we change those to pass a `[]v1alpha1.Identity`?

pkg/cosign/verify.go
```diff
@@ -167,23 +174,87 @@ func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Ver
...
+	// matches, call that good, otherwise, return an error.
+	if len(co.Identities) > 0 {
+		for _, identity := range co.Identities {
+			issuerMatches := false
+			// Check the issuer first
+			fmt.Fprintf(os.Stderr, "Checking identity: %+v", identity)
+			if identity.Issuer != "" {
+				issuer := getIssuer(cert)
+				if regex, err := regexp.Compile(identity.Issuer); err != nil {
+					return nil, fmt.Errorf("malformed issuer in identity: %s : %w", identity.Issuer, err)
```
mattmoor, line 191: We should check in the webhook that these are valid regex

pkg/cosign/verify.go
```diff
@@ -167,23 +174,87 @@ func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Ver
...
+				issuerMatches = true
+			}
+
+			// Then the subject
+			subjectMatches := false
+			if identity.Subject != "" {
+				regex, err := regexp.Compile(identity.Subject)
+				if err != nil {
+					return nil, fmt.Errorf("malformed subject in identity: %s : %w", identity.Subject, err)
+				}
```
mattmoor, line 206: same

pkg/cosign/verify.go
vaikas: Thought about it, but did not want to change them because they would match other things than email and that would be a change in behaviour. So, I'd rather not do that as part of this PR.

pkg/cosign/verify.go
vaikas: Yeah, I'll do that in a follow up PR and add tests for them, ok?

pkg/cosign/verify.go
vaikas: same same :)

mattmoor (approved):

pkg/cosign/verify.go
```diff
@@ -84,6 +86,11 @@ type CheckOpts struct {
 
 	// SignatureRef is the reference to the signature file
 	SignatureRef string
+
+	// Identities is an array of Identity (Subject, Issuer) matchers that have
+	// to be met for the signature to ve valid.
+	// Supercedes CertEmail / CertOidcIssuer
+	Identities []v1alpha1.Identity
```
mtrmac, line 93: This causes every user of `pkg/cosign` to depend on k8s; updating to a recent Cosign version thus adds 1.157 _million_ lines of code.

Please consider some other API design.

vaikas: zomg, I'm so sorry about that, didn't realize that:
https://github.com/sigstore/cosign/pull/1790

@mtrmac

PR #1760, opened 2022-04-14
Bump Fulcio client lib and use LegacyClient

nsmith5:
####  Summary

With @bobcallaway's recent work the client in Fulcio library was renamed to LegacyClient. This change bumps the Fulcio package version and uses that new LegacyClient name. Once GRPC is rolled out in production we should move off of this client as it uses the v1 API's


mattmoor (commented):
I think `.fulcioverifier.go.swp` is unintended?

nsmith5:
Yup, removed and need a mod tidy as well

mattmoor (dismissed):

nsmith5:
Hmm I see the GRPC refactor did pull in some banned dependencies (trillian + grpc middleware) that work against our slimming down the cosign binary....

Should we make a `client` library upstream and not use the `api` package so we can slim down the client library as much as possible?

haydentherapper:
I removed `"github.com/google/certificate-transparency-go"` in the embedded SCT PR, we need to depend on it.

nsmith5:
Ah ok, yeah I don't think we can run away from depending on grpc either now so these changes make sense then..

haydentherapper:
I wonder if the issue is the same thing I was running into in https://github.com/sigstore/cosign/pull/1731/, where one of the Trillian packages defines a flag that conflicts with what's defined in cosignd.

haydentherapper:
This is why in https://github.com/sigstore/cosign/pull/1731/ I copied in a file, so that when I used that package, I was only pulling in the bare minimum.

nsmith5:
closing in favour of #1762

PR #1762, opened 2022-04-14
Switch to calling gRPC-based fulcio v2 APIs 

bobcallaway:
Waiting for fulcio gRPC to be enabled in prod

@nsmith5


nsmith5:
Hmm it looks like you might be hitting similar code-gen issues I ran into. You'll need to run `hack/update-codegen.sh` I think and make want this patch for that file: https://github.com/sigstore/cosign/pull/1760/files#diff-149dfe7bb29d1191dceae3a52915e750e64b7f87257a5fb309c29d3056e2a95dL61-L65

Even after updating the codegen for cosigned I hit some CI issues... no sure what is up there

nsmith5:
If we're close to getting GRPC running in prod I'll close my PR in favour of yours  as well

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -411,15 +416,26 @@ func (v *Validator) resolvePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
 	resolveContainers(ps.Containers)
 }
 
-func getFulcioCert(u *apis.URL) (*x509.CertPool, error) {
-	fClient := api.NewClient(u.URL())
-	rootCertResponse, err := fClient.RootCert()
+func getFulcioCert(fulcioURL *apis.URL) (*x509.CertPool, error) {
+	var opts []grpc.DialOption
+	if fulcioURL.Scheme == "https" {
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS13})))
```
haydentherapper, line 422: nit: Is TLS 1.3 prevalent enough, or should we default to TLS 1.2?

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -131,17 +140,23 @@ func NewSigner(ctx context.Context, idToken, oidcIssuer, oidcClientID, oidcClien
 	default:
 		flow = FlowNormal
 	}
-	Resp, err := GetCert(ctx, priv, idToken, flow, oidcIssuer, oidcClientID, oidcClientSecret, oidcRedirectURL, fClient) // TODO, use the chain.
+	resp, err := GetCert(ctx, priv, idToken, flow, oidcIssuer, oidcClientID, oidcClientSecret, oidcRedirectURL, fClient) // TODO, use the chain.
```
haydentherapper, line 143: nit: Chain is used!

Use the chain Luke

nsmith5:
@bobcallaway is it worth resurrecting my other PR just to unblock bumping the Fulcio library here? Or are we pretty close to being able to switch over to GRPC with this work?

bobcallaway:
i'm almost done with a rebase on this; where do we stand on getting the v0.4.0 fulcio code into prod?

dlorenc:
> i'm almost done with a rebase on this; where do we stand on getting the v0.4.0 fulcio code into prod?

Should be next week. I just need to cut over the IPs in DNS.

dlorenc:
cc @k4leung4

haydentherapper:
We're discussing a rollout plan - Will ping this PR once Fulcio's been updated

haydentherapper:
@bobcallaway .4 is out! This should be good to go

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -161,11 +180,28 @@ func GetIntermediates() *x509.CertPool {
...
-	fClient := api.NewClient(fulcioServer, api.WithUserAgent(clioptions.UserAgent()))
-	return fClient, nil
+	host := fulcioServer.Host
+	switch fulcioServer.Scheme {
+	case "https":
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS12})))
+		if fulcioServer.Port() == "" {
+			host = fmt.Sprintf("%s:443", host)
+		}
+	default:
```
haydentherapper, line 196: Why do we have the option for an insecure mode? For local testing?

bobcallaway:
yes, that was the intent.

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -411,15 +416,34 @@ func (v *Validator) resolvePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
 	resolveContainers(ps.Containers)
 }
 
-func getFulcioCert(u *apis.URL) (*x509.CertPool, error) {
-	fClient := api.NewClient(u.URL())
-	rootCertResponse, err := fClient.RootCert()
+func getFulcioCert(fulcioURL *apis.URL) (*x509.CertPool, error) {
```
haydentherapper, line 419: Should we be providing this client in Fulcio to avoid duplication? I had hoped setting up a gRPC connection required very little boilerplate, but given that we need scheme/port detection, it might be good to centralize it.

fyi @wlynch, who would need the same in gitsign

haydentherapper (changes requested):

bobcallaway:
looks like scaffolding only exposes the HTTP port for fulcio, not the grpc port which is the root issue with this test failure.

@haydentherapper

haydentherapper:
@vaikas fyi, https://github.com/sigstore/cosign/pull/1762#issuecomment-1258831932

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -202,11 +222,28 @@ func GetIntermediates() (*x509.CertPool, error) {
...
-	fClient := api.NewClient(fulcioServer, api.WithUserAgent(options.UserAgent()))
-	return fClient, nil
+	host := fulcioServer.Host
+	switch fulcioServer.Scheme {
+	case "https":
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS12})))
+		if fulcioServer.Port() == "" {
+			host = fmt.Sprintf("%s:443", host)
+		}
+	default:
```
haydentherapper, line 238: There's an open issue to add an insecure flag - https://github.com/sigstore/cosign/issues/2290

We should only default to a TLS-less connection with that flag set. Revising the above comment, if we match on port and or scheme:

* scheme=https => TLS
* port=443 => TLS
* insecure flag set => no TLS

We could just ignore port 80, and in testing it's unlikely you'd be using port 80 anyways, probably 8000+

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -202,11 +222,28 @@ func GetIntermediates() (*x509.CertPool, error) {
...
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
asraa, line 239: TODO: add a retry policy?

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -73,19 +77,28 @@ func getCertForOauthID(priv *ecdsa.PrivateKey, fc api.Client, connector oidcConn
...
 
-	cr := api.CertificateRequest{
-		PublicKey: api.Key{
-			Algorithm: "ecdsa",
-			Content:   pubBytes,
+	cscr := &fulciopb.CreateSigningCertificateRequest{
+		Key: &fulciopb.CreateSigningCertificateRequest_PublicKeyRequest{
+			PublicKeyRequest: &fulciopb.PublicKeyRequest{
+				PublicKey: &fulciopb.PublicKey{
+					Algorithm: fulciopb.PublicKeyAlgorithm_ECDSA,
```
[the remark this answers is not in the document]
bobcallaway: line 71 has `pubPEM, err := cryptoutils.MarshalPublicKeyToPEM(priv.Public())` which is off of an `*ecdsa.PrivateKey` so this is accurate.

cmd/cosign/cli/fulcio/fulcio.go
bobcallaway: let's do this in a separate PR, because it should also apply to calling other external services too

vaikas:
zomg, I totes missed this and I'm so sorry. I'm curious why we have two separate ports for HTTP and GRPC and not just use Duplex there, but I reckon that's an issue that should be brought up in Fulcio land.

bobcallaway:
> zomg, I totes missed this and I'm so sorry. I'm curious why we have two separate ports for HTTP and GRPC and not just use Duplex there, but I reckon that's an issue that should be brought up in Fulcio land.

mostly because gRPC support was added later and the majority of the duplex implementations I've seen simply just use HTTP/2 as a signal to route to gRPC vs HTTP interfaces, but that could/would break some clients.

bobcallaway:
https://github.com/sigstore/scaffolding/pull/463 aims to fix the test harness issue blocking this

vaikas:
> > zomg, I totes missed this and I'm so sorry. I'm curious why we have two separate ports for HTTP and GRPC and not just use Duplex there, but I reckon that's an issue that should be brought up in Fulcio land.
> 
> mostly because gRPC support was added later and the majority of the duplex implementations I've seen simply just use HTTP/2 as a signal to route to gRPC vs HTTP interfaces, but that could/would break some clients.

I think we should be fine here because at least our implementation here (and:
https://github.com/chainguard-dev/go-grpc-kit

Switches based off of the grpc headers.
https://github.com/chainguard-dev/go-grpc-kit/blob/main/pkg/duplex/server.go#L29

So, I'd be surprised if that would break HTTP clients, since I wouldn't be expecting them to be setting grpc headers.

We can add a bunch of testing for these, but I reckon it would simplify bunch of things to be able to utilize duplex. I'll try to spin up some testing for this.

bobcallaway:
if you're checking for `Content-Type` that seems reasonable - I think the code examples I had seen in the past (not the chainguard-dev one) were only switching on HTTP 1 v 2 which didn't seem client-safe. 

We did ship Fulcio v1 with the `--grpc-port` cmd line arg so if we retain that ability while also having the duplex support I think we could accommodate both.

vaikas:
> if you're checking for `Content-Type` that seems reasonable - I think the code examples I had seen in the past (not the chainguard-dev one) were only switching on HTTP 1 v 2 which didn't seem client-safe.
> 
> We did ship Fulcio v1 with the `--grpc-port` cmd line arg so if we retain that ability while also having the duplex support I think we could accommodate both.

Just to make sure I understand, you're saying that we should retain that flag on the server side and hence be able to serve on two ports?

bobcallaway:
> > if you're checking for `Content-Type` that seems reasonable - I think the code examples I had seen in the past (not the chainguard-dev one) were only switching on HTTP 1 v 2 which didn't seem client-safe.
> > We did ship Fulcio v1 with the `--grpc-port` cmd line arg so if we retain that ability while also having the duplex support I think we could accommodate both.
> 
> Just to make sure I understand, you're saying that we should retain that flag on the server side and hence be able to serve on two ports?

We could deprecate `--grpc-port` as a command line arg and find another way to express the intent when launching the server, but I dont think we should just remove the ability to run on separate ports given that Fulcio is v1.0.0 now.

vaikas:
Yeah, I just wanted to make sure I understood, ack!

haydentherapper:
@bobcallaway What is the status of the PR? Looks like most tests are passing

PR #1766, opened 2022-04-15
tuf: add debug info if tuf update fails

asraa:

#### Summary

* Adds debug info (the state of the remote repository) if a TUF update fails.

#### Ticket Link

Related https://github.com/sigstore/root-signing/issues/166

#### Release Note

```release-note

```

Example output (I manually edited one file so sigs would fail):
```
$ rm -r $HOME/.sigstore && ./cosign initialize --mirror http://localhost:8001
Error: updating local metadata and targets: error updating to TUF remote mirror: tuf: failed to decode timestamp.json: tuf: signature verification failed
remote status:{
	"mirror": "http://localhost:8001",
	"metadata": {
		"root.json": {
			"version": 2,
			"len": 5297,
			"expiration": "11 May 22 19:09 UTC"
		},
		"snapshot.json": {
			"version": 24,
			"len": 1659,
			"expiration": "29 Apr 22 00:54 UTC"
		},
		"targets.json": {
			"version": 2,
			"len": 4167,
			"expiration": "11 May 22 19:10 UTC"
		},
		"timestamp.json": {
			"version": 24,
			"len": 720,
			"expiration": "29 Apr 22 00:54 UTC"
		}
	}
}

```

### DIFF
--- a/pkg/cosign/tuf/client.go
+++ b/pkg/cosign/tuf/client.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/url"
 	"os"
 	"path"
@@ -55,10 +56,16 @@ type TUF struct {
 
 // JSON output representing the configured root status
 type RootStatus struct {
-	Local      string            `json:"local"`
-	Remote     string            `json:"remote"`
-	Expiration map[string]string `json:"expiration"`
-	Targets    []string          `json:"targets"`
+	Local    string                    `json:"local"`
+	Remote   string                    `json:"remote"`
+	Metadata map[string]MetadataStatus `json:"metadata"`
+	Targets  []string                  `json:"targets"`
+}
+
+type MetadataStatus struct {
+	Version    int    `json:"version"`
+	Size       int    `json:"len"`
+	Expiration string `json:"expiration"`
 }
 
 type TargetFile struct {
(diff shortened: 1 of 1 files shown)


pkg/cosign/tuf/client.go
```diff
@@ -118,11 +141,11 @@ func (t *TUF) getRootStatus() (*RootStatus, error) {
 		return nil, errors.Wrap(err, "getting trusted meta")
 	}
 	for role, md := range trustedMeta {
-		expires, err := getExpiration(md)
+		mdStatus, err := getMetadataStatus(md)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("getting expiration for %s", role))
+			continue
```
haydentherapper, line 146: If there was any error, shouldn't this function return an error after iterating through all roles? Otherwise it'll return no error.

pkg/cosign/tuf/client.go
```diff
@@ -360,6 +383,12 @@ func embeddedLocalStore() (client.LocalStore, error) {
 	return local, nil
 }
 
+type signedMeta struct {
```
haydentherapper, line 386: Is there a struct defined in go-tuf we can use? For timestamp, can we use https://github.com/theupdateframework/go-tuf/blob/5b81b7e5604bfae2f0dcecd041e2e44f0128c992/data/types.go#L310?

pkg/cosign/tuf/client.go
```diff
@@ -91,16 +98,32 @@ func GetRootStatus(ctx context.Context) (*RootStatus, error) {
 	return t.getRootStatus()
 }
 
+func getMetadataStatus(b []byte) (*MetadataStatus, error) {
```
haydentherapper, line 101: nit: Can we group private functions together, especially since these are all related for getting debugging info?

haydentherapper (approved):

pkg/cosign/tuf/client.go
asraa: It's a private struct, not exported in go-tuf. I don't really want to condition on metadata type though, and all metadata defines these three.

pkg/cosign/tuf/client.go
asraa: I don't want to exit early and end up not printing out the full set of metadata statuses.

I could add an error string to `MetadataStatus` though!

asraa: done, added an error in case we can't fetch status info

pkg/cosign/tuf/client.go
asraa: done

dlorenc (approved):

Issue #1769, opened 2022-04-18
Expand the cosigned webhook to support attestations, plus cue policy for CIP.


vaikas:
This is a follow on work for the Cosigned API task force. Idea being is to bring parity between 'cosign cli' and being able to specify attestation policies just like we can do with 'verify-attestation' flag.

This document was created as a starting point for the discussion by few folks, @mlieberman85 @mattmoor @hectorj2f 

I've been playing with it, and the policy bits are probably way too verbose because I don't quite know enough of cue to create sensible policies, but at least I was able to create a what I see a real world example demonstrating cases like:
 * need to support multiple attestations
 * need to support an image that must be signed by N different signatures
 * need to support that attestations of same type (say vuln scan) is attested by multiple signatures, and specifying the exact allowed signatures.

This document is a copy of the original cosigned api document, and the changes to it have been made in 'suggestion' mode to hopefully make the differences be easier to review.
https://docs.google.com/document/d/1p51OYNT1Qegc20aIwcchzXbLv7-PN8iip3OHtTOXWew/edit?usp=sharing

hectorj2f:
@vaikas I think it is a great approach ❤️  to build this solution as part of functionalities that can be used by the CLI itself. I like the idea of adding support for matching attestations as part of the same policy definition.

Also if we develop it as a SDK functionality, other integrations could benefit by using it to ensure other type of resources (custom CRs) are also signed/attested using CIPs to define the policies.

We're happy to help here.

PR #1772, opened 2022-04-18
Attestations + policy in cip.

vaikas:
#### Summary
 * Start of adding support for parity with verify-attestations in cosigned webhook
 * Add support for cue in validating a CIP _after_ signatures / attestations have been verified
 * Fix a bug where CTLog was not plumbed through for custom verification
 * Add 'name' field to authorities (gets defaulted to authority-<index>) for making it easier to write CIP level policies

Missing:
 * E2E test for doing cue validation with CIP policy. I can't figure this out, but @hectorj2f has for example rego in #1787 that we could start testing with.
 * We should add parallel validation of Policies and then within policies parallel authorities so everything doesn't have to happen sequentially.

I'd like to get this in and start experimenting so we can get real policies implemented and get
experience on what works.

Starts putting scaffolding in place for this:
#1769 

#### Ticket Link

Fixes

#### Release Note

```release-note
 * Start of adding support for parity with verify-attestations in cosigned webhook
 * Add support for cue in validating a CIP _after_ signatures / attestations have been verified
 * Fix a bug where CTLog was not plumbed through for custom verification
 * Add 'name' field to authorities (gets defaulted to authority-<index>) for making it easier to write CIP level policies
```


test/testdata/cosigned/e2e/cip-requires-two-signatures-and-two-attestations.yaml
```diff
@@ -111,20 +111,20 @@ spec:
     type: cue
     data: |
       if len(authorityMatches."keyless-att".attestations) < 2 {
-        keylessAttestationsErr: "error"
-        keylessAttestationsErr: "Did not get both keyless attestations"
+        keylessAttestationsErr: 1
+        keylessAttestationsErrMsg: "Did not get both keyless attestations"
```
vaikas, line 115: Haha, this was my super hack to raise an error because I don't understand enough of cue yet :)  So, we only execute this if if there's a failure and because there's a mismatch the policy will fail. So, all of these 'branches' really should be:
if len(...) < expected {
  raiseEvaluationError
}

But, I don't know how to do the raiseEvaluationError, so I just added a statement that will result in an error.

config/300-clusterimagepolicy.yaml
```diff
@@ -99,6 +129,9 @@ spec:
                                   type: string
                           url:
                             type: string
+                      name:
+                        description: Name is the name for this authority. Used by the CIP Policy validator to be able to reference matching signature or attestation verifications. If not specified, the name will be attestation-<index in array>
```
mattmoor, line 133: `attestation-%d`?  Why not `authority-%d`?

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_defaults.go
```diff
@@ -14,8 +14,20 @@
...
 // SetDefaults implements apis.Defaultable
-func (*ClusterImagePolicy) SetDefaults(ctx context.Context) {
+func (c *ClusterImagePolicy) SetDefaults(ctx context.Context) {
+	c.Spec.SetDefaults(ctx)
+}
+
+func (spec *ClusterImagePolicySpec) SetDefaults(ctx context.Context) {
+	for i, authority := range spec.Authorities {
+		if authority.Name == "" {
+			spec.Authorities[i].Name = fmt.Sprintf("authority-%d", i)
```
mattmoor, line 30: Same question...  `attestation` seems weird.

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation.go
```diff
@@ -144,6 +147,39 @@ func (source *Source) Validate(ctx context.Context) *apis.FieldError {
...
 }
 
+func (a *Attestation) Validate(ctx context.Context) *apis.FieldError {
+	var errs *apis.FieldError
+	if a.Name == "" {
+		errs = errs.Also(apis.ErrMissingField("name"))
+	}
+	if a.PredicateType == "" {
+		errs = errs.Also(apis.ErrMissingField("predicateType"))
+	} else if a.PredicateType != "custom" && a.PredicateType != "slsaprovenance" && a.PredicateType != "spdx" && a.PredicateType != "link" && a.PredicateType != "vuln" {
```
mattmoor, line 157: seems like a good use of `switch`?

mattmoor: Doesn't `cosign attest` support taking a full predicate type URL as well?  Should we also allow full-blown URLs?

pkg/policy/eval_test.go
```diff
@@ -0,0 +1,106 @@
...
+package policy
+
+import (
+	"context"
+	"strings"
+	"testing"
+)
+
+const (
+	customAttestation = "{\"_type\":\"https://in-toto.io/Statement/v0.1\",\"predicateType\":\"cosign.sigstore.dev/attestation/v1\",\"subject\":[{\"name\":\"registry.local:5000/cosigned/demo\",\"digest\":{\"sha256\":\"416cc82c76114b1744ea58bcbf2f411a0f2de4b0456703bf1bb83d33656951bc\"}}],\"predicate\":{\"Data\":\"foobar e2e test\",\"Timestamp\":\"2022-04-20T18:17:19Z\"}}"
```
mattmoor, line 25: These may be easier to read as multi-line strings?

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -240,85 +248,231 @@ func validatePolicies(ctx context.Context, ref name.Reference, kc authn.Keychain
...
+				return nil, errors.Wrap(err, "validate signatures with fulcio")
+			} else if len(sps) > 0 {
+				logging.FromContext(ctx).Debugf("validated signature for %s, got %d signatures", ref.Name(), len(sps))
+				return ociSignatureToPolicySignature(ctx, sps), nil
+			}
+			logging.FromContext(ctx).Errorf("no validSignatures found for %s", ref.Name())
+			return nil, fmt.Errorf("no valid signatures found with authority %s for  %s", name, ref.Name())
+		}
+	}
+	return nil, fmt.Errorf("something went really wrong here")
```
mattmoor, line 386: Might be worth explaining or commenting why?

mattmoor (commented):
Thanks for all the tests and detailed comments 🙏

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_validation.go
vaikas: I'd rather not pollute this with the switch only because I'd rather use the TODO (refactor where the options come from and use that), also attest may take the uri but verify-attest does not. It looks for the short name in the map and pulls the URL from it. Again, I think that should be done as part of a follow on PR to refactor that code a bit to pull the map out of the cli/options so it can be reused here.
https://github.com/sigstore/cosign/blob/main/pkg/policy/attestation.go#L43

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_defaults.go
vaikas: lol, I originally had it as attestation, thought it seemed wierd, so changed it to authority-%d like here, but it's correct here? Or how do you mean? Above, yes I had neglected to update the comment, but the code is right here, no?

config/300-clusterimagepolicy.yaml
vaikas: I failed to update comment, I originally had it as attestation, and realized it was not good so I changed it to authority-%d, but forgot to update the comment.

test/testdata/cosigned/e2e/cip-requires-two-signatures-and-two-attestations.yaml
```diff
@@ -0,0 +1,142 @@
...
+      data: |
+        -----BEGIN PUBLIC KEY-----
+        MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOz9FcbJM/oOkC26Wfo9paG2tYGBL
+        usDLHze93DzgLaAPDsyJrygpVnL9M6SOyfyXEsjpBTUu6uFZqHua8hwAlA==
+        -----END PUBLIC KEY-----
+    ctlog:
+      url: http://rekor.rekor-system.svc
+  policy:
+    type: cue
+    data: |
```
hectorj2f, line 97: I can get this working in a separate PR, this one is big enough for now.

pkg/policy/eval.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return nil
+}
+
+// evaluateCue evaluates a cue policy `evaluator` against `attestation`
+func evaluateCue(ctx context.Context, attestation []byte, evaluator string) error {
+	logging.FromContext(ctx).Infof("Evaluating attestation: %s", string(attestation))
+	cueCtx := cuecontext.New()
+	v := cueCtx.CompileString(evaluator)
+	return cuejson.Validate(attestation, v)
```
hectorj2f, line 59: As discussed, we have to use a different set of functions here, as cuejson.Validate expects the CUE to be self-contained and applied as a schema to the JSON.

hectorj2f: We can improve this function in a different PR.

hectorj2f (approved):
LGTM from now, let's iterate over it.

Issue #1743, opened 2022-04-12
Spec: Attestations & keyless mode for binaries 


asraa:
**Description**

This is a proposal to discuss the general Sigstore story and functionality around binaries and blobs. Many package managers are planning on adopting Sigstore tooling, especially using keyless mode as they spin up their own identity providers. 

**Problem**
How should devs signing packages and creating attestations with sigstore keyless distribute (1) the binary (2) the signed provenance or signatures and (3) cosign bundles, signing certs, etc.

**Ideas**
Right now, devs need to provide all the material needed separately to tools that can do verification. But cosign already defines a format derived from OCI spec to attach a `.att` file describing the verification material needed. However, for blobs, we don't have a digest "reference" to retrieve the blob. Otherwise, this would be a great "manifest" to pass in to verifiers that has a specification.

If we can include attestations inline optionally in the annotation of each layer (e.g. `dev.sigstore.cosign/envelope`, then we can supply only this manifest file instead of a collection of sigstore "stuff" and the attestation. For basic signatures, we already include `dev.cosignproject.cosign/signature` in the manifest.

This also has the benefit of supporting multiple attestations and sigs.

cc @laurentsimon

dlorenc:
+1 on speccing this out! I think we had another issue somewhere for it, but I'm not sure where.

lumjjb:
Would it be a stretch to include SBOM in the discussion as well? Probably will be an orthogonal problem later on, and may be helpful to have common mechanisms to influence SBOM provenance as well.

dlorenc:
SBOM is somewhat orthogonal - it's a thing that can be attested just like a binary blob. I think what Asra proposed here would just work with SBOMs as well.

laurentsimon:
> Would it be a stretch to include SBOM in the discussion as well? Probably will be an orthogonal problem later on, and may be helpful to have common mechanisms to influence SBOM provenance as well.

+1. SBOMs are just one type of attestations, and they should be discoverable/accessible from the main `.att` like other attestations (if we go that route).

Adding some links I was provided and may be relevant for the discussion:
https://github.com/in-toto/attestation/blob/main/spec/predicates/spdx.md
https://github.com/in-toto/attestation/blob/main/spec/bundle.md

laurentsimon:
ah, Dan beat me to it :-)

laurentsimon:
[...]

PR #1774, opened 2022-04-18
Add Fulcio intermediate CA certificate to intermediate pool

haydentherapper:
This certificate will be necessary for chain building from a leaf
certificate to a root once a new version of Fulcio is rolled out. For
OCI, the chain is stored in an annotation. This intermediate is
currently only needed for verify-blob when looking up the certificate
from Rekor.

For the V3 TUF Root, the intermediate will be bundled, so that it is
easily discoverable and revokable. For now, we'll simply bundle it with
Cosign. Note that intermediates are considered untrusted, so it's fine
if the intermediate is not in TUF currently, as the root that issued the
intermediate certificate is in TUF.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
+++ b/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
@@ -39,6 +39,23 @@ var fulcioTargetStr = `fulcio.crt.pem`
 // This is the v1 migrated root.
 var fulcioV1TargetStr = `fulcio_v1.crt.pem`
 
+// The untrusted intermediate CA certificate, used for chain building
+// TODO: Remove once this is bundled in TUF metadata.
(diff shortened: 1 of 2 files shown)


haydentherapper:
Would like to get this in before a 1.8 cut.

hectorj2f (approved):

dlorenc:
This will eventually make it into the on disk bundle to avoid needing to fetch from rekor right?

haydentherapper:
The comment I made about Rekor was from how verify-blob supports looking up certificates in Rekor, instead of passing them in (not ideal behavior imo, it puts additional trust in Rekor). If there's work going on to create on disk bundles for blobs, then yes, this would need to be included in the bundle.

dlorenc:
Cool, ref https://github.com/sigstore/cosign/issues/1743

PR #1790, opened 2022-04-22
Remove the dependency on v1alpha1.Identity which brings in unnecessary k8s deps.

vaikas:

#### Summary
In #1759 I reused a struct from v1alpha1.Identity, which I didn't realize brought
deps to k8s where there shouldn't have been any. This rectifies this
by re-creating that struct here.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


dlorenc (approved):

mtrmac (commented):
Oh, that was fast. Thanks!

I can confirm that this drops all `k8s.io` dependencies from an import of `github.com/sigstore/cosign/pgk/cosign`. (Some `sigs.k8s.io` dependencies remain, but that’s for a different cause.)

vaikas:
Thanks @mtrmac for checking that it indeed fixed the import issues :)

PR #1793, opened 2022-04-23
cosigned: Unify cue data and policy before evaluating it

hectorj2f:
#### Summary

This PR adds code to compile both the policy and attestations data, then unify both cue files and finally validate them. The recommendation is to avoid using cuejson.Validate which expects the cue to be self-contained and applied as a schema to the JSON.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/policy/eval_test.go
+++ b/pkg/policy/eval_test.go
@@ -75,8 +75,7 @@ const (
 		}
 	  }`
 
-	// TODO(vaikas): Enable tests once we sort this out.
-	// cipAttestation = "authorityMatches:{\"key-att\":{\"signatures\":null,\"attestations\":{\"custom-match-predicate\":[{\"subject\":\"PLACEHOLDER\",\"issuer\":\"PLACEHOLDER\"}]}},\"key-signature\":{\"signatures\":[{\"subject\":\"PLACEHOLDER\",\"issuer\":\"PLACEHOLDER\"}],\"attestations\":null},\"keyless-att\":{\"signatures\":null,\"attestations\":{\"custom-keyless\":[{\"subject\":\"PLACEHOLDER\",\"issuer\":\"PLACEHOLDER\"}]}},\"keyless-signature\":{\"signatures\":[{\"subject\":\"PLACEHOLDER\",\"issuer\":\"PLACEHOLDER\"}],\"attestations\":null}}"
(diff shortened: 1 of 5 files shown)


vaikas:
Can we try removing this and test if it now passes? 🤩 
https://github.com/sigstore/cosign/blob/main/test/e2e_test_cluster_image_policy_with_attestations.sh#L213

test/testdata/cosigned/e2e/cip-requires-two-signatures-and-two-attestations.yaml
```diff
@@ -95,48 +95,21 @@ spec:
...
+
       authorityMatches: {
         key-att: {
-          attestations: {
-            "vuln-key": [
-              {subject: "PLACEHOLDER", issuer: "PLACEHOLDER"},
-            ]
-          }
-        }
+          attestations: struct.MaxFields(1) & struct.MinFields(1)
```
vaikas, line 104: I think the key-att is causing grief. I tried with:
keyatt
and below with
keysignature

and so on, and it seemed to at least not blow up with the missing ',' error.

hectorj2f:
@vaikas @dlorenc This PR is ready for a review.

pkg/policy/eval_test.go
```diff
@@ -124,6 +123,51 @@ func TestEvalPolicy(t *testing.T) {
...
+		import "struct"
+		import "list"
+		authorityMatches: {
+		  keyatt: {
+			attestations: struct.MaxFields(1) & struct.MinFields(1)
+		  },
+		  keysignature: {
+			signatures: list.MaxItems(1) & list.MinItems(1)
+		  },
+		  if( len(authorityMatches.keylessatt.attestations) < 2) {
```
vaikas, line 163: lol we came back to this :)

vaikas (commented):
Thank you so much for grinding through this!!!

pkg/policy/eval_test.go
hectorj2f: Oh yes :), the problem comes with struct.MinFields does not throw an error as struct.MaxFields does, look at: https://pkg.go.dev/cuelang.org/go/pkg/struct. In contrast, the list.MinItems does throw an error. In my opinion it is a weird behavior in cue, i need to open an issue or ask in their repo.

vaikas: Oh snap, that's surprising to say the least.

PR #1794, opened 2022-04-23
Allow passing keys via environment variables (`env://` refs)

### LINKED ISSUE #1776: Allow keyed signing without writing to disk
Right now, `cosign sign --key` takes a file or a KMS URL. A file isn't 100% ideal for CI, since you don't want to chance other build steps (which run with the same permissions usually) to be able to read it.

Specifically, consider the scenario where the private key comes into GitHub actions as a secret. It'll be exposed to the build step as an environment variable.

Two possible solutions come to mind:

1. Add an `env://` handler for cosign keys `cosign sign --key env://MY_ENV_VAR`
2. Let `--key` take `-` to indicate STDIN. This may cause issues because `cosign sign` sometimes prompts. Maybe `cosign sign --key - --force` would work here.

znewman01:
Depends on [sigstore#407](https://github.com/sigstore/sigstore/pull/407).

#### Summary

Allow passing keys via environment variables (`env://` refs)

#### Ticket Link

Fixes #1776

#### Release Note

```release-note
Added support for providing keys via environment variables: use `--key env://MY_VAR`.
```


dlorenc:
Lgtm when the other one lands

dlorenc:
The windows failure looked like a flake cleaning up tempdirs that we've seen before. Rerunning. We might need a skip.

znewman01:
Windows test failures seem to be caused by: https://github.com/golang/go/issues/51442

I think I'll pull these into a separate test and skip it on Windows for now.

znewman01:
> The windows failure looked like a flake cleaning up tempdirs that we've seen before. Rerunning. We might need a skip.

Jinx

dlorenc (approved):

PR #1804, opened 2022-04-26
Load in intermediate cert pool from TUF

Hayden-IO:
With the v3 TUF root, the intermediate CA certificate will be included,
so that if the intermediate signing key was compromised, the
intermediate certificate could be revoked by removing it from the TUF
targets and replacing it with a trusted certificate.

This change loads the intermediate certificate from TUF. However, we
don't want to force all users to follow this structure - They may choose
to use CRLs to detect revoked intermediates. Also, I don't want to
enforce TUF usage in the Verify package. Therefore, for TUF, we lazily create
a certificate pool only if an intermediate certificate is found, and if
it's not found, then VerifyImageSignature will create a pool using the
chain provided in the annotation.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added support for loading intermediate certificates from TUF on verification
```

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
+++ b/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go
@@ -83,8 +83,8 @@ func GetIntermediates() *x509.CertPool {
 }
 
 func initRoots() (*x509.CertPool, *x509.CertPool, error) {
-	rootPool := x509.NewCertPool()
-	intermediatePool := x509.NewCertPool()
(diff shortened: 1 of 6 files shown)


haydentherapper:
cc @asraa

dlorenc:
Do you want this in 1.8?

haydentherapper:
If 1.8 hasn't been released yet, that'd be great!

dlorenc (approved):

asraa (approved):

PR #1805, opened 2022-04-26
[Cosigned] Add signature pull secrets

### LINKED ISSUE #1655: Allow specifying ImagePullSecrets in ClusterImagePolicies
**Description**

Currently `cosigned` uses ImagePullSecrets through the PodSpec-providing resources (https://github.com/sigstore/cosign/issues/801). However, there is no guarantee that signatures are stored in the same repository of the image (for instance, if `COSIGN_REPOSITORY` was set when signing).

If we can specify ImagePullSecrets in the ClusterImagePolicies, `cosigned` can use them for verification.

DennyHoang:
#### Summary
- Add `signaturePullSecrets` to `ClusterImagePolicy`'s Authorities sources.
- Allow users to define additional credentials for signature registry location. Previously, the same ImagePullSecret credentials used only therefore images and signatures needed to use the same credentials.

#### Ticket Link
Resolves https://github.com/sigstore/cosign/issues/1655

#### Release Note
```release-note
* Allow configurable signature pull secrets credentials for cosigned
```

#### Potential outstanding todo
- Add integration test scenario (Requiring more investigation from my end on how to set up credentials for locally spun up registry in our github workflow)


kiiwyz (dismissed):

cpanato:
@DennyHoang this is ready for review?

pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -115,6 +115,8 @@ type KeyRef struct {
 type Source struct {
 	// +optional
 	OCI string `json:"oci,omitempty"`
+	// +optional
```
hectorj2f, line 118: Let's add a description here `// SignaturePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the signatures used by this Source.`

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -335,18 +335,24 @@ func ValidatePolicy(ctx context.Context, ref name.Reference, cip webhookcip.Clus
 			authorityRemoteOpts := remoteOpts
 			authorityRemoteOpts = append(authorityRemoteOpts, authority.RemoteOpts...)
 
+			signaturePullSecretsOpts, err := authority.SourceSignaturePullSecretsOpts(ctx, namespace)
+			if err != nil {
+				result.errs = append(result.errs, err)
```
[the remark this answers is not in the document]
DennyHoang: added early reporting to result channel and return if it errors during creating SignaturePullSecret remote opts

pkg/cosign/kubernetes/webhook/validator.go
```diff
@@ -314,7 +314,7 @@ func validatePolicies(ctx context.Context, ref name.Reference, policies map[stri
 // signatures OR attestations if atttestations were specified.
 // Returns PolicyResult, or errors encountered if none of the authorities
 // passed.
-func ValidatePolicy(ctx context.Context, ref name.Reference, cip webhookcip.ClusterImagePolicy, remoteOpts ...ociremote.Option) (*PolicyResult, []error) {
+func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, cip webhookcip.ClusterImagePolicy, remoteOpts ...ociremote.Option) (*PolicyResult, []error) {
```
vaikas, line 317: Nit, we should document in the signature what namespace this is. It's the namespace where we allow the signature pull secrets to come from.

vaikas (approved):
Thanks, we can fix the comment in a followup. Thanks for doing this!
Thinking about how we want to add e2e tests for this. Thinking that we can create a local image in the registry we spin up, but then maybe push the signatures / attestations to another registry and use a different secret for that then?

hectorj2f (approved):
lgtm

pkg/cosign/kubernetes/webhook/validator.go
hectorj2f: +1

PR #1806, opened 2022-04-26
Support PKCS1 encoded and non-ECDSA CT log public keys

Hayden-IO:
This came up while testing out staging, which uses a PKCS1 encoded
public key. We should be flexible on the supported key format.

This also relaxes the requirement that CT log keys from TUF are ECDSA keys.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added support for non-ECDSA and PKCS1-encoded CT log public keys from TUF
```


haydentherapper:
cc @dlorenc - If 1.8 hasn't gone out yet, getting this in too would be nice!

dlorenc (dismissed):

haydentherapper:
@dlorenc - pushed another commit that removes the requirement that the key is an ECDSA key - I don't see any reason this should be enforced.

dlorenc (approved):

Issue #1785, opened 2022-04-21
Allow configurable interactive modes


wlynch:
**Description**

When toying around with calling cosign from other tools, I noticed that cosign can hang when it's invoked with keyless enabled.

e.g. if you do something like

```go
func main() {
        os.Setenv("COSIGN_EXPERIMENTAL", "1")
        out, err := exec.Command("cosign", "sign-blob", "a.txt").CombinedOutput()
        fmt.Println(string(out), err)
}
```

IIUC, this happens because of how cosign detects interactive-ness for commands - https://github.com/sigstore/cosign/blob/f2c360eb97e52fa7766ecde370f1a48b910d7404/cmd/cosign/cli/fulcio/fulcio.go#L124-L133

i.e. there is no ID token present and there's no terminal, so prompt for device flow, but stderr/stdout won't print until the command has completed so the command hangs waiting for the input that will never come. If the flow is forcibly set to `FlowNormal`, the regular browser based flow would still work and the command can complete.

For this particular case, the answer would be to use [Stdout/StderrPipe](https://github.com/sigstore/cosign/blob/f2c360eb97e52fa7766ecde370f1a48b910d7404/cmd/cosign/cli/fulcio/fulcio.go#L124-L133) to plumb through the message during execution or just call the cosign funcs directly, but I originally encountered this when trying to invoke cosign through Git plugins (https://github.com/sigstore/cosign/issues/865#issuecomment-1105641594) where this won't be possible without an upstream change to git itself.

As a workaround, I'd like to make interactive flow mode configurable to allow callers to select the flow they want. This way for tools that can't pipe stdout/stderr can still have an option to send users through a normal interactive flow if they choose (else fall back to the existing behavior).

Also open to other ideas if you have them!

PR #1818, opened 2022-04-29
Refactor fulcio signer to take in KeyOpts (take 2)

wlynch:
#### Summary

This is a roll-forward of https://github.com/sigstore/cosign/pull/1788 with the provider fix (goimports accidentally removed
`pkg/providers/all` which removed the OIDC providers in the first commit) to fix
the e2e tests.

This commit should not have change in behavior. It changes the fulcio signer
to take in the KeyOpts struct instead of passing individual parameters through,
as well as moves common code dependent on the values to the same place.

The motivation behind this change is to make it easier to add new options without
needing to plumb through another param to the already long list.

To avoid circular dependencies (sign -> fulcio -> sign), KeyOpts was moved to the
options package.

#### Ticket Link

Part of https://github.com/sigstore/cosign/issues/1785 (wanted to make this change first to make adding the option easier)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -22,7 +22,6 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/attest"
 	"github.com/sigstore/cosign/cmd/cosign/cli/generate"
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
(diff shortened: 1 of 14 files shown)


dlorenc (approved):

dlorenc:
Nice fix!

Issue #1816, opened 2022-04-29
Verify pass private Rekor check even without Rekor public key


clem844:
`cosign v1.8.0`
I use a private Rekor instance
I did not `initialize` cosign with my own TUF signed Rekor public key so it is using the public targets from `DefaultRemoteRoot = "sigstore-tuf-root"`
However, when verifying, cosign does not seem to need the Rekor public key:
```
COSIGN_EXPERIMENTAL=1 cosign verify --key cosign.pub --rekor-url http://10.20.0.26:3000 my-registry.com/artifacts-signature | jq

Verification for my-registry.com/artifacts-signature:latest --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The claims were present in the transparency log
  - The signatures were integrated into the transparency log when the certificate was valid
  - The signatures were verified against the specified public key
```
I would expect it to fail as the Rekor timestamping cannot be checked without the public key
Am I missing something?

asraa:
> I would expect it to fail as the Rekor timestamping cannot be checked without the public key

This was curious, so I reproduced it and wanted to check what was going on with cosign.

The path that this takes it the following:
1. [VerifyImageSignature](https://github.com/sigstore/cosign/blob/367c79e61db25a203f4a07dbf400b67745ce209c/pkg/cosign/verify.go#L459) with a not-nil co.SigVerifier and a not-nil co.RekorClient
2. [verifyOCISIgnature](https://github.com/sigstore/cosign/blob/367c79e61db25a203f4a07dbf400b67745ce209c/pkg/cosign/verify.go#L492) since we have a not-nil `co.SigVerifier`. This is successful.
3. Continue to [VerifyBundle](https://github.com/sigstore/cosign/blob/367c79e61db25a203f4a07dbf400b67745ce209c/pkg/cosign/verify.go#L503). This returns `bundleVerified = false` because, like you noted, you did not `cosign initialize` so the bundle was not verified with the rekor public key. 
4. Move on to validate on an online rekor client, by entry in this [conditional](https://github.com/sigstore/cosign/blob/367c79e61db25a203f4a07dbf400b67745ce209c/pkg/cosign/verify.go#L508)
5.  Validate the tlog public key [here](https://github.com/sigstore/cosign/blob/367c79e61db25a203f4a07dbf400b67745ce209c/pkg/cosign/verify.go#L514) because we have a `SigVerifier`. This returns no error.

Note that this line was not printed in your output because the bundle could not be verified
[...]

PR #1833, opened 2022-05-02
Validate tlog entry when verifying signature via public key.

wlynch:
#### Summary

Previously, when signatures were validated via public key, we only
checked for the existence of the tlog, but didn't actually validate the
entry. This refactors the code to ensure both cert + public paths go
through the same tlog entry validation.

Signed-off-by: Billy Lynch <billy@chainguard.dev>

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1816

#### Release Note

```release-note
BUG FIX: User-provided key verification with Rekor now verifies the Rekor entry as part of validation.
```


pkg/cosign/verify_test.go
```diff
@@ -345,6 +347,41 @@ func TestVerifyImageSignatureWithExistingSub(t *testing.T) {
...
+	// tlog entry for the signature during validation (even though it does not
+	// match the underlying data / key)
+	mClient := new(client.Rekor)
+	mClient.Entries = &mock.EntriesClient{}
+
+	if _, err := VerifyImageSignature(context.TODO(), ociSig, v1.Hash{}, &CheckOpts{
+		SigVerifier: sv,
+		RekorClient: mClient,
+	}); err == nil || !strings.Contains(err.Error(), "verifying inclusion proof") {
+		// TODO(wlynch): This is a weak test, since this is really failing because
```
asraa, line 374: It looks like trillian makes some fake inclusion proofs https://github.com/google/trillian/blob/3d7922afdce9fd0144e16aae08bea061c0ad1aee/merkle/logverifier/log_verifier_test.go#L45 for testing -- maybe we can do that?

asraa (approved):

Issue #1840, opened 2022-05-04
[Cosigned] glob matching is basic right now


DennyHoang:
**Description**
Currently, glob matching is just a prefix check allowing only trailing wildcarding `*`

This is an issue as glob patterns such as:
- `gcr.io` images could have suffixes for regions such as `us.gcr.io` (`*.gcr.io/image` is not supported)
- `<dockerhub account>/image` is not supported as it is resolved to `index.docker.io/...`

I would like to suggest improvement to:
- allow multiple wildcarding (i.e `*gcr.io/*`)
- maybe images without hosts should default to try to match with `index.docker.io` since that is what those images would resolve as far as I know

Related issue: https://github.com/sigstore/cosign/issues/1834

DennyHoang:
I will try to make a PR for this change

Issue #1834, opened 2022-05-04
[cosigned] image pattern selector: keep glob and remove regex


hectorj2f:
**Description**

The ImagePattern field in the ClusterImagePolicy has two optional fields Glob and Regex. There is an internal discussion to remove Regex and keep Glob.

cc @DennyHoang @vaikas @

```
type ImagePattern struct {
	// +optional
	Glob string `json:"glob,omitempty"`
	// +optional
	Regex string `json:"regex,omitempty"`
}
```

DennyHoang:
Currently, globMatch is just a suffix check right now since we support only one wildcard `*`.
That would need to be updated to be actually glob checking before removing regex.

adamd-vmw:
glob has been sufficient for our use cases to date.

hectorj2f:
@DennyHoang Please, create the issue where you explain what is missing in globMatch, so we can fix it.

vaikas:
I think the one problem that can be an issue with glob matching that I've seen personally is when pulling things off of docker.io, sometimes it's index.docker.io for example. And in gcr.io they are sometimes gcr.io, us.gcr.io, asia.gcr.io so for them I think you'd maybe want to create:
*.gcr.io/*
gcr.io/*

instead of listing them all.

hectorj2f:
I am gonna remove the image.regex field now that glob matching has been improved.

PR #1842, opened 2022-05-04
[Cosigned] Glob matching improvement

DennyHoang:
#### Summary
Instead of doing a prefix check only for glob matching, use the `filepath` implementation of `Match` to match against glob patterns.
This will also remove the restrictions for only supporting trailing wildcard `*`

Alongside this change:
- I also added defaulting to use `index.docker.io/` host if there is no host in the glob pattern and there was not a match prior. 
  This is important since when a user deploys using `myproject/nginx` for example, the image reference gets resolved to `index.docker.io/myproject/nginx`
- I also added defaulting to the `library` repository when a glob is specified without multiple path elements.
  This is important since when a user deploys `busybox` for example, the image reference gets resolved to `index.docker.io/library/busybox` 
  
Now a user should be able to able to create a cluster image policy with glob patterns:
```yaml
- busybox
- myproject/image
```

when previously they needed to prefix those as such:
```yaml
- index.docker.io/library/busybox
- index.docker.io/myproject/image
```
#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1840
Related https://github.com/sigstore/cosign/issues/1834

#### Release Note
```release-note
* Updated cosigned glob matching logic to allow for multiple wildcards
* Updated cosigned glob matching to handle dockerhub image references better

```

cc: @coyote240 @elfotografo007 @hectorj2f @vaikas

### DIFF
--- a/pkg/apis/config/glob_test.go
+++ b/pkg/apis/config/glob_test.go
@@ -15,14 +15,19 @@
 package config
 
 import (
+	"runtime"
 	"testing"
 )
 
 func TestGlobMatch(t *testing.T) {
 	tests := []struct {
-		glob  string
-		input string
-		match bool
+		glob         string
+		input        string
+		match        bool
+		errString    string
+		windowsMatch *struct {
+			match bool
+		}
 	}{
 		{glob: "foo", input: "foo", match: true},     // exact match
 		{glob: "fooo*", input: "foo", match: false},  // prefix too long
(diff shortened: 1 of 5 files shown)


pkg/apis/config/glob_test.go
```diff
@@ -20,19 +20,37 @@ import (
...
 		{glob: "fooo*", input: "foo", match: false},  // prefix too long
 		{glob: "foo*", input: "foobar", match: true}, // works
 		{glob: "foo*", input: "foo", match: true},    // works
 		{glob: "*", input: "foo", match: true},       // matches anything
 		{glob: "*", input: "bar", match: true},       // matches anything
+		{glob: "*foo*", input: "1foo2", match: true}, // matches wildcard around
+		{glob: "*repository/*", input: "repository/image", match: true},
+		{glob: "*.repository/*", input: "other.repository/image", match: true},
+		{glob: "repository/*", input: "repository/image", match: true},
+		{glob: "repository/*", input: "other.repository/image", match: false},
```
hectorj2f, line 38: Could you add some examples to test `.gcr.io` examples ?

hectorj2f (commented):
lgtm, i'd like to see some examples using gcr.io

DennyHoang:
Hmm, so Windows has `\\` as path separators instead of `/`. 

For the `glob_test.go`, I have a different expected match result for one of the test cases because Windows is different.

I'm thinking this is not a big concern because:
1. Images do not use `\\` as separators
2. cosigned isn't even installed on a windows platform afaik
3. If cosigned is installed on a windows environment, if someone specifies `gcr.io/*`, it would be a more open inclusion anyways...

If this does not get merged today, I will have to hand off this PR to @elfotografo007 as I will not be available for a bit.

hectorj2f (approved):
lgtm

Issue #1865, opened 2022-05-11
Move fulcioroots and pkg/cosign/tuf packages to sigstore/sigstore


imjasonh:
**Description**

[`github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioroots`](https://pkg.go.dev/github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioroots) contains methods to get `x509.CertPool`s representing Fulcio's root and intermediates. 

To do this, it calls [`initRoots`](https://github.com/sigstore/cosign/blob/89b9e88d3e4b6f103cd3faf2124bd3dedbc82b00/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go#L85), which based on the presence of a `SIGSTORE_ROOT_FILE` env var either loads those roots from that file or initializes a TUF client which pulls TUF config from a GCS bucket or HTTPS URL.

Based on my understanding of this code, I believe this could all move to sigstore/sigstore, and be reused by non-cosign clients (e.g., [`smimecosign`](https://github.com/wlynch/smimecosign)) from there, instead of [depending on the sigstore/cosign package](https://github.com/wlynch/smimecosign/blob/f7f4d76f3ef48ee984f316a070021fe81ed8bcdb/command_verify.go#L13).

Before I make any changes though, I wanted to check if this understanding is correct, and if folks think it's okay to move this to sigstore/sigstore (perhaps in `pkg/fulcioroots` and `pkg/tuf`)

The TUF client also carries a dependency on GCS client libraries, but I think that could also be dropped since it only makes unauthenticated requests, which can also be made to, e.g., `https://[bucket].storage.googleapis.com/[object]` unless I'm mistaken.

@haydentherapper @znewman01 I think you might have insights?

haydentherapper:
cc @asraa about TUF

There's so much we should move to sigstore/sigstore. The TUF client should be moved too at some point. I'm fine with having this moved out to sigstore/sigstore. Will `SIGSTORE_ROOT_FILE` stay in Cosign? I assume we don't want to define CLI flags in sigstore/sigstore.

Agreed about GCS, the objects can be accessed using a REST API.

imjasonh:
> Will SIGSTORE_ROOT_FILE stay in Cosign? 

I don't know enough about the use case to comment. I don't think I'd mind as a prospective consumer of the package that its behavior was dependent on an env var, especially if that was named like `SIGSTORE_` and not `COSIGN_`.

[...]

PR #1866, opened 2022-05-11
Use pkg/fulcioroots and pkg/tuf from sigstore/sigstore

imjasonh:

Depends on https://github.com/sigstore/sigstore/pull/435
Part of https://github.com/sigstore/cosign/issues/1865

#### Summary

This removes `pkg/cosign/tuf` in favor if sigstore/sigstore's new `pkg/tuf`.

This moves cosign-specific logic into an `internal` package only used by the cosign CLI. This package adds cosign-specific override behavior, and otherwise falls back to sigstore/sigstore's new `pkg/fulcioroots`.

#### Release Note

```release-note
pkg/cosign/tuf and cmd/cosign/cli/fulcio/fulcioroots packages are moved to sigstore/sigstore
```


cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -187,3 +182,81 @@ func NewClient(fulcioURL string) (api.Client, error) {
...
+		var err error
+		roots, intermediates, err = initRoots()
+		if err != nil {
+			panic(err)
+		}
+	})
+	return intermediates
+}
+
+func initRoots() (*x509.CertPool, *x509.CertPool, error) {
```
haydentherapper, line 232: Thanks for adding this here! I didn't realize this had already been added back.

imjasonh:
Do we want to keep cosign's `pkg/tuf` around with type aliases and deprecation warnings for a release, or do we think this isn't used outside of this package? Technically I don't think we make any guarantees about the Go API stability of sigstore/cosign, though we shouldn't break people unnecessarily if we think folks are depending on it.

dlorenc:
> Do we want to keep cosign's `pkg/tuf` around with type aliases and deprecation warnings for a release, or do we think this isn't used outside of cosign? Technically I don't think we make any guarantees about the Go API stability of sigstore/cosign, though we shouldn't break people unnecessarily if we think folks are depending on it.
> 
> I think gitsign is using it, but I can update that myself.

I don't mind removing it.

internal/pkg/cosign/fulcio/fulcioroots/fulcioroots.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const altRoot = "SIGSTORE_ROOT_FILE"
+
+// Get returns the Fulcio root certificate.
+//
+// If the SIGSTORE_ROOT_FILE environment variable is set, the root config found
+// there will be used instead of the normal Fulcio roots.
+//
+// If this behavior is not needed, use
```
znewman01, line 43: nit: not clear whether "this behavior" refers to "returns the Fulcio root certificate" or the bit about the env var.

znewman01 (dismissed):

internal/pkg/cosign/fulcio/fulcioroots/fulcioroots.go
haydentherapper: +1, would simply clarify that these methods support fetching from an environment variable too

internal/pkg/cosign/fulcio/fulcioroots/fulcioroots.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var (
+	rootsOnce        sync.Once
+	roots            *x509.CertPool
+	intermediates    *x509.CertPool
+	singletonRootErr error
+)
+
+const altRoot = "SIGSTORE_ROOT_FILE"
```
asraa, line 36: Move the documentation on `If the SIGSTORE_ROOT_FILE env variable is set` above to here?

Also, what is the format of this file -- notably it's a list of PEM encoded FULCIO certificates, NOT a TUF root.json.

dlorenc:
Merge conflict, otherwise lgtm and let's get this in to unblock the log jam.

dlorenc (approved):

PR #1868, opened 2022-05-11
Normalize certificate flag names

### LINKED ISSUE #1847: `--certificate` or `--cert`?
Let's sign something:

```shell
$ cosign sign-blob /dev/null --output-cert /tmp/null.crt
Error: unknown flag: --output-cert
main.go:52: error during command execution: unknown flag: --output-cert
```

Okay, must be "certificate":

```shell
$ cosign sign-blob /dev/null --output-certificate /tmp/null.crt
[...]
```

That worked! Now verify:

```shell
$ cosign verify-blob /dev/null --certificate /tmp/null.crt
Error: unknown flag: --certificate
main.go:52: error during command execution: unknown flag: --certificate
```

Wat.

```shell
$ cosign verify-blob /dev/null --cert /tmp/null.crt                                                                                
[...]
```

Oh okay ☹️ 

-----------------------------------------------------------------------

We should probably make `--*-certificate` the canonical flag name, and do what GNU CLIs do and accept unambiguous prefixes (if that's too big of a move in general, maybe just add aliases for these flags explicitly).

haydentherapper:
This changes the flag names to use certificate instead of
the abbreviated cert. To avoid breaking clients, we add
a global alias that translates between the two.

#### Ticket Link

Fixes #1847

#### Release Note

```release-note
Changed flag names to use certificate instead of abbreviated cert. Not a breaking change, as the previous flag names are aliased to the new ones.
```

### DIFF
--- a/doc/cosign_attest.md
+++ b/doc/cosign_attest.md
@@ -41,8 +41,8 @@ cosign attest [flags]
 ```
       --allow-insecure-registry                                                                  whether to allow insecure connections to registries. Don't use this for anything but testing
       --attachment-tag-prefix [AttachmentTagPrefix]sha256-[TargetImageDigest].[AttachmentName]   optional custom prefix to use for attached image tags. Attachment images are tagged as: [AttachmentTagPrefix]sha256-[TargetImageDigest].[AttachmentName]
-      --cert string                                                                              path to the X.509 certificate in PEM format to include in the OCI Signature
-      --cert-chain string                                                                        path to a list of CA X.509 certificates in PEM format which will be needed when building the certificate chain for the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and end with the root certificate. Included in the OCI Signature
+      --certificate string                                                                       path to the X.509 certificate in PEM format to include in the OCI Signature
+      --certificate-chain string                                                                 path to a list of CA X.509 certificates in PEM format which will be needed when building the certificate chain for the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and end with the root certificate. Included in the OCI Signature
   -f, --force                                                                                    skip warnings and confirmations
       --fulcio-url string                                                                        [EXPERIMENTAL] address of sigstore PKI server (default "https://fulcio.sigstore.dev")
   -h, --help                                                                                     help for attest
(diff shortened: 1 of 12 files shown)


haydentherapper:
cc @znewman01

doc/cosign_attest.md
```diff
@@ -41,8 +41,8 @@ cosign attest [flags]
 ```
       --allow-insecure-registry                                                                  whether to allow insecure connections to registries. Don't use this for anything but testing
       --attachment-tag-prefix [AttachmentTagPrefix]sha256-[TargetImageDigest].[AttachmentName]   optional custom prefix to use for attached image tags. Attachment images are tagged as: [AttachmentTagPrefix]sha256-[TargetImageDigest].[AttachmentName]
-      --cert string                                                                              path to the X.509 certificate in PEM format to include in the OCI Signature
-      --cert-chain string                                                                        path to a list of CA X.509 certificates in PEM format which will be needed when building the certificate chain for the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and end with the root certificate. Included in the OCI Signature
+      --certificate string                                                                       path to the X.509 certificate in PEM format to include in the OCI Signature
```
znewman01, line 44: should the help text reflect that these flags are aliased?

znewman01 (commented):
Is it better to do it as a normalization function or as an alias? I think viper supports `RegisterAlias` which we use elsewhere in this repo, which to me feels like a better match for what we're actually doing here.

doc/cosign_attest.md
haydentherapper: We're using `RegisterAlias` for `cosign sign-blob` to alias `output` and `output-signature`. One downside is the help text shows each of these flags on separate lines, so it looks like they're two separate flags.

Besides that, aliasing and normalizing seem pretty similar. I prefer normalize because I ideally don't want to continue to support the old flag names - Maybe we can drop the normalize function in a 2.0 release?

Another benefit of normalize over aliasing is I don't have to alias the flag in each command, there's no GlobalRegisterAlias like with normalize.

znewman01: > One downside is the help text shows each of these flags on separate lines, so it looks like they're two separate flags.

Ugh 😭 

> I prefer normalize because I ideally don't want to continue to support the old flag names

That's a sufficient argument for me.

znewman01 (approved):

dlorenc (approved):

PR #1872, opened 2022-05-13
tree: only report artifacts that are present

ribbybibby:
#### Summary

Only returns artifacts in the tree if they actually exist. I think this matches the expectations a user would have when using this command.

Also tweaks how we're using the references:
- Use the actual references we added to the map when checking them during the print phase
- Use the tag suffix from the `ociremote` package rather than redefining it in cli

#### Ticket Link

N/A

#### Release Note

```release-note
Modifies 'cosign tree' to only return artifacts that exist
```

### DIFF
--- a/cmd/cosign/cli/tree.go
+++ b/cmd/cosign/cli/tree.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):
lgtm, thanks @ribbybibby

ribbybibby:
>  It seems there are some linting issues.
@hectorj2f Looks like a timeout issue. When I run `golangci-lint run` locally it doesn't report any linting errors.

PR #1873, opened 2022-05-13
[cosigned] remove regex from the image pattern fields

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

#### Summary

As discussed in https://github.com/sigstore/cosign/issues/1834 we decided to remove regex from the image pattern fields.

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1834

#### Release Note

```release-note
cosigned: remove regex field from image pattern options.
```


vaikas:
Maybe we should create the v1beta1 API first before removing it in there instead?

hectorj2f:
@vaikas i thought the idea was to make these changes before v1beta1. That is why we can do it on alpha versions.

vaikas:
I'm not sure who's all using this, but since it's a breaking change seemed like a safer thing to roll these into v1beta1. At least that's what I've been accustomed to, but if we feel it's ok to introduce breaking changes I'm ok with it. I'll start working on the v1beta1 anyways now, so hopefully get something out by the end of the day that we can iterate on.

hectorj2f:
@vaikas i was following the indications from kubernetes api versioning where you can make these changes in alpha versions.

```
Alpha:

The version names contain alpha (for example, v1alpha1).
The software may contain bugs. Enabling a feature may expose bugs. A feature may be disabled by default.
The support for a feature may be dropped at any time without notice.
The API may change in incompatible ways in a later software release without notice.
The software is recommended for use only in short-lived testing clusters, due to increased risk of bugs and lack of long-term support.```


You can find that in https://kubernetes.io/docs/reference/using-api/.

vaikas:
@hectorj2f yeah, I know that's what k8s did, we tried to be bit more mindful of breaking changes with Knative, but I'm totes fine with this.

vaikas (approved):
Would you mind updating the sigstore/sigstore-website with the removal for the docs side of the house?

hectorj2f:
> Would you mind updating the sigstore/sigstore-website with the removal for the docs side of the house?

Sure, I'll create the PR.

PR #1909, opened 2022-05-20
Add privacy statement for PII storage

haydentherapper:
This statement has been added just before the OIDC code/normal flow,
since only in this flow we expect a user to be present. For the device
flow or when a token is provided, it's likely that Cosign is being run
in an automated environment.

This requires the user either type Y or provide a global flag for
confirmation. This should not break any existing flows in automated
environments because it's only the flow when a browser is needed, not
for the device flow or when a token is provided.

wip

#### Ticket Link

Fixes

#### Release Note

```release-note

```


haydentherapper:
@bobcallaway @dlorenc @lukehinds - Please take a look and let me know if this seems satisfactory.

dlorenc:
cc @znewman01 as well

Looks great to me, thanks!

dlorenc (approved):

lukehinds:
LGTM, let's just let it sit for a business day before merging.

lukehinds (approved):

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -92,6 +99,14 @@ func GetCert(ctx context.Context, priv *ecdsa.PrivateKey, idToken, flow, oidcIss
 		c.flow = oauthflow.NewDeviceFlowTokenGetter(
 			oidcIssuer, oauthflow.SigstoreDeviceURL, oauthflow.SigstoreTokenURL)
 	case FlowNormal:
+		fmt.Fprintln(os.Stderr, PrivacyStatement)
```
znewman01, line 102: Why not pass `PrivacyStatement` to `ConfirmPrompt`?

pkg/cosign/common.go
```diff
@@ -26,6 +26,14 @@ import (
 	"golang.org/x/term"
 )
 
+// confirmation is a global variable to store whether or not the user has provided
+// the --yes flag to skip all confirmation prompts
+var confirmation bool
```
znewman01, line 31: `confirmation` is IMO an ambiguous name.

How about `skipConfirmation` or similar?

znewman01 (commented):
I worry some folks will get in the habit of passing `-y` always to skip this prompt, but then they'll *also* skip some other prompt that *doesn't* show up every time.

Can we add a way to skip *just* this prompt? Maybe an environment variable `COSIGN_ACCEPT_PII_NOTICE=1` or something?

haydentherapper:
How do other CLIs handle this when they add a new feature that requires confirmation, but also support a flag that skips confirmation? Is an environment variable per confirmation a standard practice? To me, a “y” flag seemed more in line with the typical CLI interface. 

FWIW, the other feature using a confirmation is a destructive operation, with its own flag. I’d say that would be a good practice to follow - a unique flag for destructive, but a single flag for skipping simple acknowledgements. Thoughts?

lukehinds:
Passing in a Y is pretty common:

`dnf update -y`

znewman01:
I tend to like a lot of the guidelines at [clig.dev](https://clig.dev/#interactivity) for this.

>  I’d say that would be a good practice to follow - a unique flag for destructive, but a single flag for skipping simple acknowledgements. Thoughts?

That seems smart to me. In general, here's what I'd like to see (but I'm okay with variations here!):

- in an interactive environment
  - require confirmation for simple notifications (i.e., `[Y]/n`)
  - require explicit confirmation for destructive actions (i.e., `y/[N]`), unless a flag specific to that action is passed
- in a non-interactive environment (e.g., no TTY, or some explicit `-y` flag passed), accept defaults
  - print simple notifications, bypassing confirmation
  - error on destructive actions, unless a flag specific to that action is passed


In *this* case, I think I'd rather make the difference between displaying this message or not "am I interactive?" rather than "what flow am I using?"

haydentherapper:
I can move the check up to what calls `GetCert`, but it'll be the same effectively. For this case, even if we're in a terminal, it's not a sufficient check because a Cosign provider may have already fetched the OIDC token, so we need to be aware of state.

pkg/cosign/common.go
haydentherapper: Done

cmd/cosign/cli/fulcio/fulcio.go
haydentherapper: ConfirmPrompt printed the message after the confirmation check, and I wanted to print it before

haydentherapper:
Updated the code based on the comments - The message is now always printed, and only in an interactive session the user is prompted for input.

cmd/cosign/cli/fulcio/fulcio.go
znewman01: Unless I'm misunderstanding your claim, it doesn't seem to: 

https://github.com/sigstore/cosign/blob/03e66aad02cf7c987ea489cdeda1fd580b6b1fc6/pkg/cosign/common.go#L38-L39

I also submit that if `ConfirmPrompt` doesn't work for this use case directly, there's something wrong with it and we should fix it.

haydentherapper: %s was invisible to me 🙃

cmd/cosign/cli/options/root.go
```diff
@@ -43,4 +44,7 @@ func (o *RootOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.PersistentFlags().DurationVarP(&o.Timeout, "timeout", "t", DefaultTimeout,
 		"timeout for commands")
+
+	cmd.PersistentFlags().BoolVarP(&o.SkipConfirmation, "yes", "y", false,
+		"skip confirmation prompts")
```
znewman01, line 49: Can we provide more detail about what it does? "skip confirmations prompts for non-destructive actions" or similar?

pkg/cosign/common.go
```diff
@@ -36,6 +44,10 @@ func FileExists(filename string) bool {
 }
 
 func ConfirmPrompt(msg string) (bool, error) {
+	if skipConfirmation {
```
znewman01, line 47: It would appear that this now skips confirmation for destructive actions as well:

https://github.com/sigstore/cosign/blob/9ccf708d576f27f1a5e9248a7b95cddcd551a108/cmd/cosign/cli/clean.go#L53-L54

cmd/cosign/cli/fulcio/fulcio.go
haydentherapper: Fixed

pkg/cosign/common.go
haydentherapper: Added a simple change to ConfirmPrompt to pass in whether or not the action is destructive.

znewman01: I have a slight preference for 2 functions, `ConfirmPrompt` and `ConfirmPromptDestructive` because bool parameters are difficult to parse when reading the caller

pkg/cosign/common.go
```diff
@@ -35,7 +43,14 @@ func FileExists(filename string) bool {
...
-func ConfirmPrompt(msg string) (bool, error) {
+// ConfirmPrompt prompts the user for confirmation for an action. Supports skipping
+// the confirmation prompt when the global skipConfirmation is set, and when the
+// action is non-destructive.
+func ConfirmPrompt(msg string, destructive bool) (bool, error) {
+	if skipConfirmation && !destructive {
+		return skipConfirmation, nil
+	}
+
 	fmt.Fprintf(os.Stderr, "%s\n\nAre you sure you want to continue? [Y/n]: ", msg)
```
znewman01, line 54: We discussed `y/[N]` here (optional for this PR)

znewman01 (approved):
LGTM, my other comments in this review are just nits

haydentherapper:
Thanks for the reviews! This should address all of the comments - Added a second function, and changed to y/N (also handles both Y and y)

dlorenc (approved):

znewman01 (approved):

Issue #1903, opened 2022-05-19
Want to use GlobMatch on root level domain (e.g. "ghcr.io/*")


jdolitsky:
The GlobMatch method does not accept a pattern such as `ghcr.io/*`

Seems there is a bug on Go opened in 2015: https://github.com/golang/go/issues/11862

cc @tcnghia

jdolitsky:
Related question: is `*` meant to match anything, or just docker hub images?

imjasonh:
> Related question: is `*` meant to match anything, or just docker hub images?

Using the glob match rules, I believe it would match a dockerhub official image like `ubuntu`, but not a user owned image like `myuser/myapp`, which would need `*/*`.

I'd like to suggest that image refs like `ubuntu` get expanded into their canonical form (`index.docker.io/library/ubuntu`) and matched that way. User owned images are `index.docker.io/myuser/myapp`. go-containerregistry can parse the ref with `WeakValidation` (the default) and return the full version with `ref.Name()`.

The webhook could return a warning if the match is specified like `*`, until it becomes an error in the future. That's probably better anyway since `*` looks like "all images" and isn't using glob rules.

There's also `docker.io/library/ubuntu` and `docker.io/myuser/myapp` which are confusingly equivalent to the `index` versions. We should probably warn on those too and eventually reject that form, to avoid confusion.

imjasonh:
Re https://github.com/sigstore/cosign/issues/1903#issuecomment-1132398878 -- based on some tests with @tcnghia it sounds like we're already canonicalizing `ubuntu` into `index.docker.io/library/ubuntu` and matching the policy using that expanded form. So that's good. 🎉 

All that's left then is to warn when a policy includes `- glob: '*/*'` or `- glob: 'docker.io/*/*', since these will never match any image, regardless of how it was specified. Eventually that warning can become an error.

hectorj2f:
@imjasonh @jdolitsky I was looking to the Globing function, and generally I started to consider re-using this function from the Kubernetes core https://github.com/kubernetes/kubernetes/blob/9720d130e466f401d00e93a3c537848cbf76ca26/pkg/credentialprovider/keyring.go#L184 instead of the current function that we have. Do you have any thoughts about it ?

imjasonh:
[...]

PR #1914, opened 2022-05-23
Add support for "**" in image glob matching

imjasonh:

Fixes https://github.com/sigstore/cosign/issues/1903

This is WIP because it currently breaks a number of previous test cases, and I want to make sure this isn't considered a backward-incompatible change. Or if it is, that we're comfortable with this breakage.

```release-note
Image policy globs now support ** to match any character; matching images that are specified like "ubuntu" or "myuser/myapp" should be matched with a glob like "index.docker.io/library/*" or "index.docker.io/*/*" respectively, rather than expecting "*" or "*/*" to match as before.
```

@hectorj2f @jdolitsky

### DIFF
--- a/pkg/apis/config/glob.go
+++ b/pkg/apis/config/glob.go
@@ -16,9 +16,11 @@
 package config
 
 import (
-	"net/url"
-	"path/filepath"
+	"fmt"
+	"regexp"
 	"strings"
+
+	"github.com/google/go-containerregistry/pkg/name"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jdolitsky (approved):
awesome. if breaking change is an issue, we can deprecate and make `GlobMatchWithWarnings` 😅?

hectorj2f:
@imjasonh I believe you need to check the unit tests because they seem to be failing for the `image_policies_test`.

imjasonh:
> @imjasonh I believe you need to check the unit tests because they seem to be failing for the `image_policies_test`.

Yeah, the failures indicate the API breakage I'm talking about actually. The tests check image names like `rando`, which under this new logic would need a glob like `index.docker.io/library/rando` to match it. I'll need to relax the glob matching logic to handle this case with a warning like the others.

hectorj2f (approved):

PR #1921, opened 2022-05-25
fix: fix fetching updated targets from TUF root

### LINKED ISSUE #1899: Bug: Investigate issue in TUF root upgrades
**Description**

Investigate buggy behavior with consistent snapshots in the sigstore TUF root upgrade path:

 Verifying with an upgraded TUF root that enables consistent snapshots causes a bug in a fresh environment. Explicitly calling `cosign initialize` and then verifying avoids the bug.

Reproduced by replacing the default remote store to the `brokenv3` (consistent snapshot) branch of root-signing:
```
asraa@asraa1:~/git/cosign$ go build -o cosign ./cmd/cosign
asraa@asraa1:~/git/cosign$ COSIGN_EXPERIMENTAL=1 ./cosign verify gcr.io/distroless/base:debug 
panic: error getting targets

goroutine 1 [running]:
github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioroots.Get.func1()
	/home/asraa/git/cosign/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go:69 +0x74
sync.(*Once).doSlow(0xc00084ade0?, 0x1d?)
	/usr/local/go/src/sync/once.go:68 +0xc2
sync.(*Once).Do(...)
	/usr/local/go/src/sync/once.go:59
github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioroots.Get()
	/home/asraa/git/cosign/cmd/cosign/cli/fulcio/fulcioroots/fulcioroots.go:65 +0x31
github.com/sigstore/cosign/cmd/cosign/cli/fulcio.GetRoots(...)
	/home/asraa/git/cosign/cmd/cosign/cli/fulcio/fulcio.go:174
github.com/sigstore/cosign/cmd/cosign/cli/verify.(*VerifyCommand).Exec(0xc0005b9c30, {0x2e23cd8, 0xc000054108}, {0xc000876a90, 0x1, 0x6bc9a5?})
	/home/asraa/git/cosign/cmd/cosign/cli/verify/verify.go:113 +0x3f5
github.com/sigstore/cosign/cmd/cosign/cli.Verify.func1(0xc0007d5b80, {0xc000876a90, 0x1, 0x1})
	/home/asraa/git/cosign/cmd/cosign/cli/verify.go:115 +0x265
github.com/spf13/cobra.(*Command).execute(0xc0007d5b80, {0xc000876a70, 0x1, 0x1})
	/home/asraa/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0xc000908500)
	/home/asraa/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
	/home/asraa/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
main.main()
	/home/asraa/git/cosign/cmd/cosign/main.go:51 +0x4c
asraa@asraa1:~/git/cosign$ ls $HOME/.sigstore
root
asraa@asraa1:~/git/cosign$ rm -r $HOME/.sigstore
asraa@asraa1:~/git/cosign$ ./cosign initialize
Root status: 
 {
	"local": "/home/asraa/.sigstore/root",
	"remote": "test-sigstore-root",
	"metadata": {
		"root.json": {
			"version": 3,
			"len": 5285,
			"expiration": "10 Nov 22 02:19 UTC",
			"error": ""
		},
		"snapshot.json": {
			"version": 29,
			"len": 1975,
			"expiration": "31 May 22 17:41 UTC",
			"error": ""
		},
		"targets.json": {
			"version": 3,
			"len": 6088,
			"expiration": "10 Nov 22 02:19 UTC",
			"error": ""
		},
		"timestamp.json": {
			"version": 29,
			"len": 717,
			"expiration": "24 May 22 17:41 UTC",
			"error": ""
		}
	},
	"targets": [
		"fulcio_v1.crt.pem",
		"rekor.0.pub",
		"rekor.pub",
		"artifact.pub",
		"ctfe.pub",
		"fulcio.crt.pem",
		"fulcio_intermediate_v1.crt.pem"
	]
}
asraa@asraa1:~/git/cosign$ COSIGN_EXPERIMENTAL=1 ./cosign verify gcr.io/distroless/base:debug 

Verification for gcr.io/distroless/base:debug --
[success]
```

cc @haydentherapper 

Questions:
* Where is `GetTargets` pulling targets from in each case? 
* What target names are being pulled in each case?

asraa:

#### Summary

This change refactors the cosign TUF client and hopefully aims to simplify the logic behind embedded TUF repository and targets, and the writeable on-disk/in-memory repository and targets. Roughly, I structured this so that the cosign TUF client contains (1) A `client.LocalStore` to hold TUF repository **metadata** updates and (2) A `targetImpl` to hold downloaded and cached **target** files. 

Previously, the two were "out of sync" -- when starting with an embedded workflow, we would create an embedded root repository and an embedded targetImpl. However, the embedded targetImpl ONLY retrieved embedded targets, **not** the updated ones! 

https://github.com/sigstore/cosign/blob/5f09c424d541cf22fd3d891ccadc7f7d6c762f72/pkg/cosign/tuf/client.go#L548

Embedded workflows never used their updated targets, despite writing them to the underlying storage, and **failed** to retrieve new targets referenced by updated metadata.

Conceptually now, there are 4 main objects in the TUF client:
1. An `memoryCache` targetImpl: A map that stores target files by name.
2. A `diskCache` targetImpl: Stores targets on disk.
3. An `embeddedWrapper` targetImpl: This wraps the underlying (memory, disk) cache. By default it `Get`s embedded targets. If any targets were downloaded and `Set`, then `Get` transfers to retrieving from the underlying cache.
4. An `embeddedLocalStore` that wraps either a `MemoryLocalStore` or `FileLocalStore`. Similar to above, by default it gets embedded repo metadata, until any new metadata needed to be downloaded and set. Then any `GetMeta` operations get the cached metadata.

Other:
* To make testing easier, interfaced the default embedded repo and default remote mirror.
* Tested the addition of a new target in an update. The same test at HEAD fails to find the target.
* I tested this against a binary of cosign that pointed to the default mirror as `brokenv3`'s GCS bucket and it successfully found the new target `fulcio_interemediate_v1.crt.pem`

If this design looks good, I will continue to add testing and clean-up test code. It's a pain to manually write out TUF updates, so I'd like to unify those functions and also add testing for consistent snapshotting (note brokenv3 enabled that so I'm 90% sure that works with this fix).

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1899

#### Release Note

```release-note
* bug fix: Fixes bug in retrieving newly added verification material from TUF root.
```

cc @haydentherapper @dlorenc @znewman01


pkg/cosign/tuf/client.go
```diff
@@ -516,36 +502,162 @@ func cachedTargetsDir(cacheRoot string) string {
...
+func wrapEmbeddedLocal(s localInitFunc) client.LocalStore {
+	return &embeddedLocalStore{writeableFunc: s, writeable: nil}
+}
+
+func (e embeddedLocalStore) GetMeta() (map[string]json.RawMessage, error) {
+	if e.writeable != nil {
+		// We are using a writeable store, so use that.
+		return e.writeable.GetMeta()
+	}
+	// We haven't needed to create or write new metadata, so get the embedded metadata.
```
znewman01, line 539: Talking to myself for a bit...

I wonder whether it'd be clearer to separate out the embedded `LocalStore` implementation?

That is, you have an `embeddedLocalStore` which *just* contains the code to read the meta from the embedded.FS and `SetMeta()` returns an err always. This can be combined in a `wrapperLocalStore` or something.

This would only really move out the bottom part of GetMeta, so maybe just pull that into a private function `getMetaEmbedded`? I always prefer extracting code into methods vs. comments, because it's self-documenting and comments aren't clear whether they refer to the next line or the next 10 lines, but this is definitely personal preference.

pkg/cosign/tuf/client.go
```diff
@@ -186,7 +192,11 @@ func getRoot(meta map[string]json.RawMessage) (json.RawMessage, error) {
 		return trustedRoot, nil
 	}
 	// On first initialize, there will be no root in the TUF DB, so read from embedded.
-	trustedRoot, err := embeddedRootRepo.ReadFile(path.Join("repository", "root.json"))
+	rd, ok := GetEmbedded().(fs.ReadFileFS)
+	if !ok {
+		return nil, errors.New("fs.ReadFileFS unimplemented for embedded repo")
```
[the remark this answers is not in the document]
znewman01: This could just be that I'm bad at Go.

pkg/cosign/tuf/client.go
```diff
@@ -503,49 +489,176 @@ func rootCacheDir() string {
...
-	b, err := embeddedRootRepo.ReadFile(path.Join("repository", "targets", p))
+func wrapEmbedded(t targetImpl) targetImpl {
+	return &embeddedWrapper{writeable: t, embedded: true}
+}
+
+type embeddedWrapper struct {
+	// If we have an embedded fallback that needs updates, use
+	// the writeable targetImpl
+	writeable targetImpl
+	// Whether we are using the embedded or writeable
```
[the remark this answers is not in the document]
znewman01: I wonder if there's a better name: maybe `unmodified`? invert it and call it `updated` or `modified`?

pkg/cosign/tuf/client.go
```diff
@@ -503,49 +489,175 @@ func rootCacheDir() string {
...
+
+type embeddedLocalStore struct {
+	// This is where updated repository info will be written to, if needed.
+	writeable     client.LocalStore
+	writeableFunc localInitFunc
 }
 
-type memoryCache struct {
-	targets map[string][]byte
+var GetEmbedded = func() fs.FS {
```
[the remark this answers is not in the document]
znewman01: And then if we're going to do that, we may as well make it a parameter of the init method rather than something that needs to be mocked.

pkg/cosign/tuf/client.go
```diff
@@ -558,25 +670,75 @@ func (e *embedded) Get(p string) ([]byte, error) {
...
+		for _, entry := range entries {
+			b, err := rd.ReadFile(filepath.FromSlash(filepath.Join("repository", "targets", entry.Name())))
+			if err != nil {
+				return fmt.Errorf("reading embedded file: %w", err)
+			}
+			if err := e.writeable.Set(entry.Name(), b); err != nil {
+				return fmt.Errorf("setting embedded file: %w", err)
+			}
+		}
+		e.embedded = false
```
znewman01, line 697: IMO this makes more sense next to the `e.writeable.Set` line (i.e., you do it unconditionally). And then a comment like "now that we've updated `writeable`, we need to check it when we get targets"

znewman01 (commented):
I'm sure you saw this, but test failure/lint failure looks related.

Also, your PR description is great---can you copy a bunch of it inline?

Release note is a little inappropriate for end users:  I think "TUF client' is an implementation detail.

pkg/cosign/tuf/client.go
```diff
@@ -503,49 +489,176 @@ func rootCacheDir() string {
...
+	}
+	// We have a writeable store, so set the metadata.
+	return e.writeable.SetMeta(name, meta)
+}
+
+func (e embeddedLocalStore) DeleteMeta(name string) error {
+	if e.writeable != nil {
+		return e.writeable.DeleteMeta(name)
+	}
+	return nil
```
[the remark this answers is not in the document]
asraa: Likely you've started using your writeable cache if you're gonna hit a point in the TUF spec to DeleteMeta (https://github.com/theupdateframework/go-tuf/blob/ed6788e710fc3093a7ecc2d078bf734c0f200d8d/client/client.go#L392), but I didn't want to assume behavior of upstream TUF. That being said, you probably only ever `DeleteMeta` after `SetMeta`'ing something new, and we probably should alert if the Cosign TUF client ever tried to Delete something in the embedded repo.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: I'm pretty bad at it too! It's because `GetEmbedded` returns a `fs.FS`; there wasn't another interface that could both support `embed.FS` (for the actual prod embedded files) and a testable filesystems.

https://cs.opensource.google/go/go/+/go1.18.2:src/embed/embed.go;l=148

(embed.FS implements fs.FS)

It *shouldn't* happen in practice since prod clients are using `embed.FS` but I needed to be able to swap the embedded out for test. I could panic here, but preferred to return the error.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: Hmmmmm I don't want to lock it in because tests need to modify that on an individual basis

Otherwise I would need to pre-hook test runs, I think. But correct me if I'm wrong about that.

pkg/cosign/tuf/client.go
```diff
@@ -503,49 +489,175 @@ func rootCacheDir() string {
...
+	entries, err := ed.ReadDir("repository")
+	if err != nil {
+		return nil, err
+	}
+	rd, ok := GetEmbedded().(fs.ReadFileFS)
+	if !ok {
+		return nil, errors.New("fs.ReadFileFS unimplemented for embedded repo")
+	}
+	for _, entry := range entries {
+		if entry.Type().IsRegular() {
```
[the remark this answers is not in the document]
asraa: Because this is the metadata store, I need to avoid copying over the target directory -- that would have `ModeDir` set. Also just in case there's executables or something funky. I like your skipping paradigm!

pkg/cosign/tuf/client_test.go
```diff
@@ -308,6 +310,84 @@ func TestGetTargetsByMeta(t *testing.T) {
...
+
+	// Initialize.
+	tufObj, err := NewFromEnv(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tufObj.Close()
+
+	// Try to retrieve the newly added target.
+	targets, err := tufObj.GetTargetsByMeta(Fulcio, []string{"fooNoCustom.txt"})
```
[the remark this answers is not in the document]
asraa: I specifically want to test that it can fetch a new target WITHOUT initializing the TUF obj first and potentially having a cache. (this is what failed before). What I can do is assert that the new target was not in the embedded metadata, but I don't want to create a `tufObj` at this point.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
znewman01: Err, doesn't each test instantiate its own TUF client (via `NewFromEnv`)? You should be able to move the `GetEmbedded` call into `initializeTUF` or `wrapEmbedded`/`wrapEmbeddedLocal` just the first time. I also don't think it's that bad to have a separate constructor that *doesn't* access any environment variables that you could explicitly pass the fake FS into for tests.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
znewman01: Ahh okay, sounds like it's a limitation of the type system. Feel free to resolve

pkg/cosign/tuf/client_test.go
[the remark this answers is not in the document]
znewman01: Oops, that's a good point. Maybe a comment to that effect  (the top-level one for the test)? That says "we wanna fetch before initializing the TUF object"

pkg/cosign/tuf/client.go
asraa: I like that idea, done, but it does require making both the `wrappedEmbeddedLocalStore` and the `embeddedLocalStore` into impls of `client.LocalStore`

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: Oh!  I misunderstood what you meant by "init time" -- I thought that meant Golang "init" which happens once globally. 

I like the `initialize` idea.

pkg/cosign/tuf/client.go
```diff
@@ -44,6 +46,10 @@ const (
 	SigstoreNoCache   = "SIGSTORE_NO_CACHE"
 )
 
+var GetRemoteRoot = func() string {
```
haydentherapper, line 49: Does this need to be publicly exported, or can it be a private variable in the package?

pkg/cosign/tuf/client.go
```diff
@@ -503,49 +507,203 @@ func rootCacheDir() string {
...
-	targets map[string][]byte
+// A read-only local store using embedded metadata
+type embeddedLocalStore struct {
+	ed fs.FS
 }
 
-func (m *memoryCache) Set(p string, b []byte) error {
-	if m.targets == nil {
-		m.targets = map[string][]byte{}
+func (e embeddedLocalStore) GetMeta() (map[string]json.RawMessage, error) {
```
haydentherapper, line 547: Question: How do we use the embedded metadata (as in, what's shipped with Cosign)? Isn't it only to start the chain of TUF roots?

Would it be reasonable to only store root.json, and remove the rest? Or is go-tuf expecting a complete repository? (I thought there was support now for initializing only from a root)

haydentherapper: I'm just wondering if we can simplify the logic around embedded wrapping. We have this tri-state - Either it's in in the embedded, or it's in either memory or on disk. If we only access a file or two from embedded, just for initialization, we could potentially separate embedded from LocalStore

asraa: go-tuf is not expecting a complete repository, but we don't want to touch the remote unless we *need* an update so we want the complete repository.

it's to act as a starting local state, i guess.

haydentherapper: The timestamp will almost always be expired, so won't we always need to hit the remote?

asraa: almost always yeah, fwiw i'm willing to reconsider this in general and simplify our embedded repo state to only contain the root.json because that would align a lot more closely with normal go-tuf flows and avoid all this complex overlayed logic on whether or not we need to hit the remote...

but yeah the original idea was that if the state of the embedded repo was good we wouldn't fetch. at that point the timestamp had a fairly long expiration -- 3 weeks or so

haydentherapper: As we're likely going to be moving to more frequent updates, I'm not sure the embedded non-root metadata adds much value. Do you think it would simply the logic here? Would it remove the embeddedWrappers completely?

If you want to save a bit of remote fetching, is it possible to initialize just the LocalStore containing the targets? Then what's embedded what just be the root and targets. When we fetch the new metadata and compare target files hashes, go-tuf would find that all are locally downloaded. But this is only saving a single initial fetch, so that might not be worth it.

asraa: I tried this solution! Only the root is in the embedded, but we also embed targets and use them ONLY IF their hashes are equal to what we see in the `targets.json`.

pkg/cosign/tuf/client.go
asraa: I kind of want to make this a public global flag eventually! That would make it easy to point to preprod/staging environment. I can make it a follow up issue if you think so

As long as it's a func for now I'm fine, because I need to configure it in the tests.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
znewman01: I think I'd rather pass this in to the constructor explicitly

znewman01 (dismissed):
I think once we backport the test fixes from #1932 this is pretty close

pkg/cosign/tuf/client.go
asraa: adding this to the tracking issue explicitly https://github.com/sigstore/cosign/issues/1935

pkg/cosign/tuf/client.go
```diff
@@ -311,27 +301,41 @@ func NewFromEnv(ctx context.Context) (*TUF, error) {
...
+// Checks if the testTarget matches the valid target file metadata.
+func isValidTarget(testTarget []byte, validMeta data.TargetFileMeta) bool {
+	localMeta, err := util.GenerateTargetFileMeta(bytes.NewReader(testTarget))
 	if err != nil {
-		return err
+		return false
 	}
-	if err := os.WriteFile(cachedRemote(rootCacheDir()), b, 0600); err != nil {
-		return fmt.Errorf("storing remote: %w", err)
+	if err := util.TargetFileMetaEqual(localMeta, validMeta); err != nil {
```
haydentherapper, line 335: To also double check - This would catch if the embedded targets are modified for some reason, correct? So to verify the e2e flow:
* Cosign requests the target metadata (targets.json) from the remote
* The metadata's sig is verified using the target role from the root (which has already been updated and verified)
* For each target in targets.json, we check if the target needs to be downloaded or is present in the embedded/local cache
* For each target, we compare its hash to the hash from the metadata (this step)

pkg/cosign/tuf/client.go
```diff
@@ -475,25 +449,63 @@ func (t *TUF) updateMetadataAndDownloadTargets() error {
...
 	}
-	_, err := io.Copy(w, &dest.buf)
-	return err
+
+	// Nope -- no local matching target, go download it.
+	if w.Len() == 0 {
+		dest := targetDestination{buf: &w}
+		if err := t.client.Download(name, &dest); err != nil {
+			return fmt.Errorf("downloading target: %w", err)
+		}
```
haydentherapper, line 501: Do we need to also check `isValidTarget`, or will `Download` handle checking the target hash? (If Download doesn't, it should!)

Hayden-IO (commented):
This looks great!

pkg/cosign/tuf/client.go
```diff
@@ -143,9 +154,9 @@ func getMetadataStatus(b []byte) (*MetadataStatus, error) {
 }
 
 func (t *TUF) getRootStatus() (*RootStatus, error) {
-	local := "embedded"
-	if !t.embedded {
-		local = rootCacheDir()
+	local := rootCacheDir()
+	if noCache() {
+		local = "in-memory"
```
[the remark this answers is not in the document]
asraa: Oh, this is for logging purposes (local is a `string`): so that on initialize we print the location of the file system local store, or if in-memory, the best we can do is say the local is "in-memory"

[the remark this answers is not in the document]
haydentherapper: Oh yep, I mixed up this with the TUF client `local` var

pkg/cosign/tuf/client.go
asraa: Correct, it does check hash handling!
https://github.com/theupdateframework/go-tuf/blob/3b26aedfe985198bc88a9dda7525938c575ca046/client/client.go#L928-L934

I also put hash checking on the `Get`/retrieval side, just for robustness: https://github.com/sigstore/cosign/blob/e7bcb6961178788ceb5a2e63e33388a188b574c9/pkg/cosign/tuf/client.go#L343-L354

pkg/cosign/tuf/client.go
asraa: Yes! Mostly correct, except slight details in Target retrieval/download from https://github.com/sigstore/cosign/blob/e7bcb6961178788ceb5a2e63e33388a188b574c9/pkg/cosign/tuf/client.go#L464

* Cosign requests the target metadata (targets.json) from the remote
* The metadata's sig is verified using the target role from the root (which has already been updated and verified)
* For each target in targets.json, we check (1) if we have the target in our local cache, if it's hash is valid, do nothing. (2) if the target is present in the embedded/local cache, we compare its hash from the metadata and copy it to the cache for ease of retrieval or (3) it must need to be downloaded (and the hash is cerified in the go-tuf call). 
* On get, we retrieve the target compare its hash to the hash from the metadata (this step)

znewman01 (approved):

Hayden-IO (approved):

asraa:
cc @dlorenc I think we're all in agreement that there's follow-up here to be addressed as soon as this is merged https://github.com/sigstore/cosign/issues/1935, I'll be taking care of them one by one starting this afternoon and handing off to @vaikas for more testing of the policy controller.

dlorenc:
Yay! Thanks everyone!

asraa:
> Yay! Thanks everyone!

@dlorenc please can I have a write access review :)

dlorenc (approved):

PR #1932, opened 2022-05-30
fix: fix fetching updated targets from TUF root (based off of #1921)

vaikas:
#### Summary

This change refactors the cosign TUF client and hopefully aims to simplify the logic behind embedded TUF repository and targets, and the writeable on-disk/in-memory repository and targets. Roughly, I structured this so that the cosign TUF client contains (1) A `client.LocalStore` to hold TUF repository **metadata** updates and (2) A `targetImpl` to hold downloaded and cached **target** files. 

Previously, the two were "out of sync" -- when starting with an embedded workflow, we would create an embedded root repository and an embedded targetImpl. However, the embedded targetImpl ONLY retrieved embedded targets, **not** the updated ones! 

https://github.com/sigstore/cosign/blob/5f09c424d541cf22fd3d891ccadc7f7d6c762f72/pkg/cosign/tuf/client.go#L548

Embedded workflows never used their updated targets, despite writing them to the underlying storage, and **failed** to retrieve new targets referenced by updated metadata.

Conceptually now, there are 4 main objects in the TUF client:
1. An `memoryCache` targetImpl: A map that stores target files by name.
2. A `diskCache` targetImpl: Stores targets on disk.
3. An `embeddedWrapper` targetImpl: This wraps the underlying (memory, disk) cache. By default it `Get`s embedded targets. If any targets were downloaded and `Set`, then `Get` transfers to retrieving from the underlying cache.
4. An `embeddedLocalStore` that wraps either a `MemoryLocalStore` or `FileLocalStore`. Similar to above, by default it gets embedded repo metadata, until any new metadata needed to be downloaded and set. Then any `GetMeta` operations get the cached metadata.

Other:
* To make testing easier, interfaced the default embedded repo and default remote mirror.
* Tested the addition of a new target in an update. The same test at HEAD fails to find the target.
* I tested this against a binary of cosign that pointed to the default mirror as `brokenv3`'s GCS bucket and it successfully found the new target `fulcio_interemediate_v1.crt.pem`

If this design looks good, I will continue to add testing and clean-up test code. It's a pain to manually write out TUF updates, so I'd like to unify those functions and also add testing for consistent snapshotting (note brokenv3 enabled that so I'm 90% sure that works with this fix).

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1899

#### Release Note

```release-note
* bug fix: Fixes bug in retrieving newly added verification material from TUF root.
```


Makefile
```diff
@@ -118,7 +118,7 @@ cross:
 golangci-lint:
 	rm -f $(GOLANGCI_LINT_BIN) || :
 	set -e ;\
-	GOBIN=$(GOLANGCI_LINT_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0 ;\
+	GOBIN=$(GOLANGCI_LINT_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.0 ;\
```
cpanato, line 121: can bump to 1.46.2

vaikas:
Looks like the Windows tests are unhappy with this:
```
--- FAIL: TestGetRekorPubKeys (0.21s)
    tlog_test.go:25: Unexpected error calling GetRekorPubs, expected nil: getting trusted root: open repository\root.json: file does not exist
    tlog_test.go:28: expected 1 or more keys, got 0
Repository initialized
Staged snapshot.json metadata with expiration date: 2022-06-06 08:58:35 +0000 UTC
Staged timestamp.json metadata with expiration date: 2022-05-31 08:58:35 +0000 UTC
Committed successfully
```
Maybe a path separator issue? Seems to work on everything else just fine.

pkg/cosign/tlog.go
```diff
@@ -380,6 +385,17 @@ func VerifyTLogEntry(ctx context.Context, rekorClient *client.Rekor, e *models.L
 		rekorPubKeys[keyID] = RekorPubKey{PubKey: pubFromAPI, Status: tuf.Active}
 	}
 
+	rekorPubKeysTuf, err := GetRekorPubs(ctx)
```
vaikas, line 388: I think it might be cleaner to have all the override behaviour in one place, and move L371-386 to GetRekorPubs and have everything there. But, that would mean plumbing RekorClient through to GetRekorPubs, so wasn't sure if we wanted to do that. One thing that we _might_ want to consider is plumbing the RekorClient into the ctx passed in and pull it from there so we wouldn't have to change the signature. I'd be happy to do that if we think that makes sense.

hectorj2f (dismissed):
lgtm

dlorenc:
Nice work @vaikas and @asraa!

When @asraa looks we can just merge this one.

vaikas:
I cribbed the PR subject / description from #1921 here in case this is what we end up merging.

pkg/cosign/tuf/client.go
```diff
@@ -537,46 +563,32 @@ func (m *memoryCache) Set(p string, b []byte) error {
...
-		return nil, err
+func (m *memoryCache) Get(p string) ([]byte, error) {
+	if m.targets == nil {
+		return nil, fmt.Errorf("no cached targets available, cannot retrieve %s", p)
 	}
-	// Unfortunately go:embed appears to somehow replace our line endings on windows, we need to switch them back.
-	// It should theoretically be safe to do this everywhere - but the files only seem to get mutated on Windows so
-	// let's only change them back there.
-	if runtime.GOOS == "windows" {
-		return bytes.ReplaceAll(b, []byte("\r\n"), []byte("\n")), nil
```
znewman01, line 556: Where did this go?

znewman01 (approved):
This works for the time being, appears to fix the relevant bug.

I have some apprehensions about this whole chunk of code generally (that predate this PR), I'm going to file a cleanup bug to come back and address this because this code is pretty important and should be easy to understand/test.

znewman01:
Actually can we hold off a day on merging this? I think #1921 is pretty close

pkg/cosign/tuf/client.go
asraa: Updated https://github.com/sigstore/cosign/pull/1921/commits/410d6cf9cef895df5fc7897abb0df8e48c00c803

asraa:
> This PR does a lot, and I don't want to submit this without submitting https://github.com/sigstore/cosign/pull/1921 first. We can't rush this large PR through to unblock tests, there's a risk with breaking something else accidentally.

Hey! FWIW they're the same PR, so #1921 incorporates the CI fixes from @vaikas

dlorenc:
> > This PR does a lot, and I don't want to submit this without submitting #1921 first. We can't rush this large PR through to unblock tests, there's a risk with breaking something else accidentally.
> 
> Hey! FWIW they're the same PR, so #1921 incorporates the CI fixes from @vaikas

Yep - we only need to merge one of them. We do need to get one of them merged though as this is holding up a pretty large release.

haydentherapper:
Totally fair, but I also want to make sure we take our time reviewing the PR to not introduce other subtle bugs.

inferno-chromium:
Some folks are concerned that this might be just hiding a test failure, let's take the time to review this in depth in the original PR #1921. @lukehinds @bobcallaway for another pair of eyes. What is the release timeline, we should be able to make that one.

dlorenc:
We can close this one now, the commits got merged in with #1921.

FWIW they were always the same PR as @asraa noted, it was just done to expedite getting the tests fixed because this was blocking the release.

PR #1934, opened 2022-05-30
feat(attest): custom annotations support

### LINKED ISSUE #1773: Support for custom Annotations is missing
**Description**

It seems like `verify-attestation` allows you to check for annotation key pairs, but does not recognize the flag. It also is not "verified" content FWIW. Also, `attest` does not expose an annotation map option either, but also labels it in the help string.

@wlynch also a good starter issue maybe?

developer-guy:

#### Summary

This PR will add custom annotation support to both attest and verify-attestation commands.

#### Ticket Link

Fixes #1773 

#### Release Note

```release-note
feat(attest): custom annotations support
```

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -63,6 +63,12 @@ func Attest() *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			annotations, err := o.AnnotationsMap()
+			if err != nil {
+				return err
+			}
+
 			ko := options.KeyOpts{
 				KeyRef:                   o.Key,
 				PassFunc:                 generate.GetPass,
@@ -79,7 +85,7 @@ func Attest() *cobra.Command {
 			}
 			for _, img := range args {
 				if err := attest.AttestCmd(cmd.Context(), ko, o.Registry, img, o.Cert, o.CertChain, o.NoUpload,
-					o.Predicate.Path, o.Force, o.Predicate.Type, o.Replace, ro.Timeout); err != nil {
+					o.Predicate.Path, o.Force, o.Predicate.Type, o.Replace, ro.Timeout, annotations.Annotations); err != nil {
 					return fmt.Errorf("signing %s: %w", img, err)
 				}
 			}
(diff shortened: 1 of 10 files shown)


developer-guy:
If I understand correctly, the answer should be yes but PTAL @asraa

dlorenc:
Hmm, if they're unsigned them I'm a little worried about how this works, we can't verify them using the verify flag, can we?

Could you describe the use case a bit more? I'd be more comfortable if we could get them into the annotation.

developer-guy:
@asraa, I think we need your assistance here 🙋🏻‍♂️

developer-guy:
kindly ping @asraa

bburky:
Today the `predicateType` is inside the signed payload, inside the attestation layers themselves. If there are multiple attestation types, it is not possible to pull the manifest and select which attestations to download.

This feature is interesting to me because it would allow providing a custom hint for which layers contain a specific attestation. Alternately, a `predicateType` could be added as a standard cosign annotation to support this too if this is desirable.

If I understand correctly, today cosign has to filter attestation types after downloading every attestation layer and inspecting the `predicateType` inside the payload?  A hint in the annotations allows optimizing to only download the needed attestations.

chaospuppy:
This seems like a really useful feature in order to avoid the need to decode and parse each layer of an attestation until the desired `predicateType` is located, as @bburky described.  This limitation presents a substantial challenge to tools looking to retrieve attestation payloads that are built using languages without `cosign` libraries available in their ecosystem.  @dlorenc would it work if the `predicateType` was added as an attestation to the layer e.g.:
```
"layers": [
    {
      "mediaType": "application/vnd.dsse.envelope.v1+json",
      "size": 11452,
      "digest": "sha256:09ca22f4ce6e85c40a8a68ec0f402caeb0e815c9b21e2447c246880d2533c995",
      "annotations": {
        "dev.cosignproject.cosign/signature": ""
        "predicateType": "https://cyclonedx.org/schema"
      }
    }
  ]
```

Then, part of the verification process when a `--predicate-type` is specified as part of `cosign download attestation` command (e.g. `cosign download attestation --predicate-type https://cyclonedx.org/schema`) , the signed `predicateType` value included in the payload could be compared with the user-supplied annotation value after the layer is located using the annotation, and verification could be failed if the values don't match.  Thoughts?

developer-guy:
Can we re-open this one @dlorenc if you still consider it useful?

PR #1938, opened 2022-05-31
DNM: Investigate tuf object closure in policy controller

asraa:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -102,12 +102,12 @@ func GetRekorPubs(ctx context.Context) (map[string]RekorPubKey, error) {
 	} else {
 		tufClient, err := tuf.NewFromEnv(ctx)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error creating new TUF obj from env: %w", err)
 		}
 		defer tufClient.Close()
 		targets, err := tufClient.GetTargetsByMeta(tuf.Rekor, []string{rekorTargetStr})
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error getting rekor target by metadata: %w", err)
 		}
 		for _, t := range targets {
 			rekorPubKey, err := PemToECDSAKey(t.Target)
@@ -387,10 +387,13 @@ func VerifyTLogEntry(ctx context.Context, rekorClient *client.Rekor, e *models.L
 
 	rekorPubKeysTuf, err := GetRekorPubs(ctx)
 	if err != nil {
-		if len(rekorPubKeys) == 0 {
-			return fmt.Errorf("unable to fetch Rekor public keys from TUF repository, and not trusting the Rekor API for fetching public keys: %w", err)
-		}
-		fmt.Fprintf(os.Stderr, "**Warning** Failed to fetch Rekor public keys from TUF, using the public key from Rekor API because %s was specified", addRekorPublicKeyFromRekor)
+		return fmt.Errorf("Failed to fetch Rekor public keys from TUF: %w", err)
+		/*
+			if len(rekorPubKeys) == 0 {
+				return fmt.Errorf("unable to fetch Rekor public keys from TUF repository, and not trusting the Rekor API for fetching public keys: %w", err)
+			}
+			fmt.Fprintf(os.Stderr, "**Warning** Failed to fetch Rekor public keys from TUF, using the public key from Rekor API because %s was specified", addRekorPublicKeyFromRekor)
+		*/
 	}
 
 	for k, v := range rekorPubKeysTuf {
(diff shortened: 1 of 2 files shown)


asraa:
@vaikas Do you know if you can get the extra logging I added here into the webhook to get those logs during this call?
```
2022-06-01T00:06:51.2112046Z + echo '::group:: Verify with three CIP, one with correct Source set'
2022-06-01T00:06:51.2113741Z + kubectl create -n demo-key-remote job demo --image=registry.local:5000/policy-controller/demo@sha256:0c4f2a4309a79b60e2eff29cc870c543632c9d08cdda9345ec4d2b8a0e39dc3b
2022-06-01T00:06:51.4851921Z error: failed to create job: admission webhook "policy.sigstore.dev" denied the request: validation failed: failed policy: image-policy-key: spec.template.spec.containers[0].image
2022-06-01T00:06:51.4853529Z registry.local:5000/policy-controller/demo@sha256:0c4f2a4309a79b60e2eff29cc870c543632c9d08cdda9345ec4d2b8a0e39dc3b failed to validate public keys with authority authority-0 for registry.local:5000/policy-controller/demo@sha256:0c4f2a4309a79b60e2eff29cc870c543632c9d08cdda9345ec4d2b8a0e39dc3b: no matching signatures:
2022-06-01T00:06:51.4854302Z failed to verify signature
2022-06-01T00:06:51.4854997Z  failed to fetch Rekor public keys from TUF: error creating new TUF obj from env: creating new local store: creating cached local store: resource temporarily unavailable
2022-06-01T00:06:51.4871798Z + echo Failed to create Job with Remote Public Key with Source
2022-06-01T00:06:51.4872171Z Failed to create Job with Remote Public Key with Source
2022-06-01T00:06:51.4872691Z + exit 1
```

vaikas:
@asraa, you can see the logs here:
https://github.com/sigstore/cosign/runs/6680771665?check_suite_focus=true

If you click on the `Summary` and scroll to the bottom you see logs. Then the logs for the webhook are logs/kind-worker/pods/cosign-system_webhook-<UUID>/webhook/0.log

I can see things like this here:
```
2022-06-01T00:34:44.581452312Z stdout F Creating new TUF obj
2022-06-01T00:34:44.587554952Z stdout F Creating new TUF obj
2022-06-01T00:34:44.661711937Z stdout F Closing TUF obj
2022-06-01T00:34:44.66510416Z stdout F Creating new TUF obj
2022-06-01T00:34:44.763549904Z stdout F Closing TUF obj
2022-06-01T00:34:44.769852145Z stdout F Creating new TUF obj
```

I also created this gist that has the logs for the webhook for the entire test:
https://gist.github.com/vaikas/60cc0b57c1c63ce10023505f3f5894f7

If there are multiple policies that match a given image, we do spin up go funcs to eval them in parallel here:
https://github.com/sigstore/cosign/blob/main/pkg/cosign/kubernetes/webhook/validator.go#L256

Mayhaps the tuf code is not threadsafe that might be causing issues?

vaikas:
So, my hypothesis is that we're calling this multiple times from different threads:
https://github.com/sigstore/cosign/blob/main/pkg/cosign/tuf/client.go#L532

Which calls into the TUF code here:
https://github.com/sigstore/cosign/blob/main/pkg/cosign/tuf/client.go#L537

Into here:
https://github.com/theupdateframework/go-tuf/blob/master/client/leveldbstore/leveldbstore.go#L12
Which looks like this:
```
func FileLocalStore(path string) (tuf_client.LocalStore, error) {
	fd, err := storage.OpenFile(path, false)
	if err != nil {
		return nil, err
	}

	db, err := leveldb.Open(fd, nil)
	return &fileLocalStore{fd: fd, db: db}, err
}
```

I think the interesting part is the comments on OpenFile call:
https://github.com/syndtr/goleveldb/blob/master/leveldb/storage/file_storage.go#L89

```
// OpenFile returns a new filesystem-backed storage implementation with the given
// path. This also acquire a file lock, so any subsequent attempt to open the
// same path will fail.
//
// The storage must be closed after use, by calling Close method.
```

So, I _think_ we should not be calling the FileLocalStorage multiple times, but instead stash the returned tuf_client.LocalStore and reuse it, since according to the comments that is safe for concurrent access:
https://github.com/syndtr/goleveldb/blob/master/leveldb/db.go#L175

vaikas:
I guess we have to think about what we want to reuse. Our TUF abstraction, or the tuf.LocalStore (or something else). Since these are called from within the tlog.go, as it stands it's seemingly not threadsafe. Especially for the webhook case that's a bit problematic cause there's really no way around that given the existing abstractions.

Also have to think about what the refresh behaviour is for long running processes like the webhook, so we can't just fire and forget I reckon?

vaikas:
And of course, happy to help!

vaikas:
Perfect, was wondering about that as well, but frankly I just hadn't grokked yet what the interactions / assumptions between different stores was, but I did see the TODO but didn't quite understand what the caching behaviour should be :) I see the nocache flag and didn't understand what the significance of it was. If we could just load the entries into a map in memory, that would certainly be great.

dlorenc:
Can we close this one now?

Issue #1935, opened 2022-05-31
Cleanup for TUF code


znewman01:
To fix https://github.com/sigstore/cosign/issues/1899 @asraa submitted https://github.com/sigstore/cosign/pull/1921. @vaikas  stepped in to help with tests and adapted it to https://github.com/sigstore/cosign/pull/1932. This appears to fix the bug, and I approved it because this has been blocking the Cosign release for more than a week (the new release contains https://github.com/sigstore/cosign/pull/1894 which we should get out ASAP).

But https://github.com/sigstore/cosign/pull/1921 contains some good suggestions for cleaning up that code; we should definitely do a bunch of them. It's important that this code be clear and testable.

CC @haydentherapper

vaikas:
FYI, just to make sure you're aware, but there's this PR moving much the code sigstore/sigstore:
https://github.com/sigstore/sigstore/pull/435

asraa:
One AI for tracking: Clean up the `GetRekorPub` overrides
https://github.com/sigstore/cosign/pull/1932/files/81a86c3c4b79545d0a009ebb4d4ec77a8c1b76e8..2805823eaa317d931f5b6cb3d196aab3f52a5ed2#r884674685

asraa:
Another one: testing verification failure and TUF object closure https://github.com/sigstore/cosign/pull/1932#issuecomment-1142238581

haydentherapper:
I’m going to review 1921 today. 

i would prefer we not submit 1932 without more discussion.

asraa:
Another one: Cleanup the configurable remote https://github.com/sigstore/cosign/pull/1921/files#r882887941

asraa:
Load all targets into memory at initialize to save time reading/writing to disk: https://github.com/sigstore/cosign/pull/1921#discussion_r885966545

znewman01:
Asra, thanks for your hard work here, it's looking good 😄 

Is this actually closed?

asraa:
Re-opening, the remaining items are:
* Cleanup the `GetEmbedded` and `GetRemoteRoot`
* Consolidate the Rekor pub overrides
* Set RekorPubs in CheckOpts as we do for Fulcio roots so we aren't calling the TUF pkg from other internal packages. Makes cosign as a library more rovust.

vaikas:
I think the consolidate Rekor pub overrides has been done as well as 'Set RekorPubs in CheckOpts'.

[...]

PR #1953, opened 2022-06-02
tuf: improve TUF client concurrency and caching

asraa:
#### Summary

* Moves Rekor public key retrieval from API inside `GetRekorPubs`. 
* TUF objects always use in-memory local stores and target store. If caching is enabled, we sync with the on-disk store on (one-time) start and re-sync on any updates. We do **not** keep the local disk cache open, so we don't hog the OS file lock.
* TUF objects are now singletons (see POC from @vaikas https://github.com/sigstore/cosign/pull/1941): this means no re-initialization for each process, we can re-use the TUF object.

#### Ticket Link

Partial fix https://github.com/sigstore/cosign/issues/1935

#### Release Note

```release-note

```

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -109,7 +109,6 @@ func GetRekorPubs(ctx context.Context, rekorClient *client.Rekor) (map[string]Re
 		if err != nil {
 			return nil, err
 		}
-		defer tufClient.Close()
 		targets, err := tufClient.GetTargetsByMeta(tuf.Rekor, []string{rekorTargetStr})
 		if err != nil {
 			return nil, err
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -77,10 +77,15 @@ func getLogID(pub crypto.PublicKey) (string, error) {
 
 // GetRekorPubs retrieves trusted Rekor public keys from the embedded or cached
 // TUF root. If expired, makes a network call to retrieve the updated targets.
-// There is an Env variable that can be used to override this behaviour:
+// A Rekor client may optionally be provided in case using SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY
+// (see below).
+// There are two Env variable that can be used to override this behaviour:
 // SIGSTORE_REKOR_PUBLIC_KEY - If specified, location of the file that contains
 // the Rekor Public Key on local filesystem
-func GetRekorPubs(ctx context.Context) (map[string]RekorPubKey, error) {
+// SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY - If specified, fetches the Rekor public
+// key from the Rekor server using the provided rekorClient.
+// TODO: Rename SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY to be test-only or remove.
+func GetRekorPubs(ctx context.Context, rekorClient *client.Rekor) (map[string]RekorPubKey, error) {
 	publicKeys := make(map[string]RekorPubKey)
 	altRekorPub := os.Getenv(altRekorPublicKey)
 
@@ -121,6 +126,26 @@ func GetRekorPubs(ctx context.Context) (map[string]RekorPubKey, error) {
 			publicKeys[keyID] = RekorPubKey{PubKey: rekorPubKey, Status: t.Status}
 		}
 	}
+
+	// If we have a Rekor client and we've been told to fetch the Public Key from Rekor,
+	// additionally fetch it here.
+	addRekorPublic := os.Getenv(addRekorPublicKeyFromRekor)
+	if addRekorPublic != "" && rekorClient != nil {
+		pubOK, err := rekorClient.Pubkey.GetPublicKey(nil)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch rekor public key from rekor: %w", err)
+		}
+		pubFromAPI, err := PemToECDSAKey([]byte(pubOK.Payload))
+		if err != nil {
+			return nil, fmt.Errorf("error converting rekor PEM public key from rekor to ECDSAKey: %w", err)
+		}
+		keyID, err := getLogID(pubFromAPI)
+		if err != nil {
+			return nil, fmt.Errorf("error generating log ID: %w", err)
+		}
+		publicKeys[keyID] = RekorPubKey{PubKey: pubFromAPI, Status: tuf.Active}
+	}
+
 	if len(publicKeys) == 0 {
 		return nil, errors.New("none of the Rekor public keys have been found")
 	}
@@ -330,9 +355,7 @@ func FindTLogEntriesByPayload(ctx context.Context, rekorClient *client.Rekor, pa
 	return searchIndex.GetPayload(), nil
 }
 
-// VerityTLogEntry verifies a TLog entry. If the Env variable
-// SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY is specified, fetches the Rekor public
-// key from the Rekor server using the provided rekorClient.
+// VerityTLogEntry verifies a TLog entry.
 func VerifyTLogEntry(ctx context.Context, rekorClient *client.Rekor, e *models.LogEntryAnon) error {
 	if e.Verification == nil || e.Verification.InclusionProof == nil {
 		return errors.New("inclusion proof not provided")
@@ -365,37 +388,11 @@ func VerifyTLogEntry(ctx context.Context, rekorClient *client.Rekor, e *models.L
 		LogID:          *e.LogID,
 	}
 
-	// If we've been told to fetch the Public Key from Rekor, fetch it here
-	// first before using the TUF code below.
-	rekorPubKeys := make(map[string]RekorPubKey)
-	addRekorPublic := os.Getenv(addRekorPublicKeyFromRekor)
-	if addRekorPublic != "" {
-		pubOK, err := rekorClient.Pubkey.GetPublicKey(nil)
-		if err != nil {
-			return fmt.Errorf("unable to fetch rekor public key from rekor: %w", err)
-		}
-		pubFromAPI, err := PemToECDSAKey([]byte(pubOK.Payload))
-		if err != nil {
-			return fmt.Errorf("error converting rekor PEM public key from rekor to ECDSAKey: %w", err)
-		}
-		keyID, err := getLogID(pubFromAPI)
-		if err != nil {
-			return fmt.Errorf("error generating log ID: %w", err)
-		}
-		rekorPubKeys[keyID] = RekorPubKey{PubKey: pubFromAPI, Status: tuf.Active}
-	}
-
-	rekorPubKeysTuf, err := GetRekorPubs(ctx)
+	rekorPubKeys, err := GetRekorPubs(ctx, rekorClient)
 	if err != nil {
-		if len(rekorPubKeys) == 0 {
-			return fmt.Errorf("unable to fetch Rekor public keys from TUF repository, and not trusting the Rekor API for fetching public keys: %w", err)
-		}
-		fmt.Fprintf(os.Stderr, "**Warning** Failed to fetch Rekor public keys from TUF, using the public key from Rekor API because %s was specified", addRekorPublicKeyFromRekor)
+		return fmt.Errorf("unable to fetch Rekor public keys: %w", err)
 	}
 
-	for k, v := range rekorPubKeysTuf {
-		rekorPubKeys[k] = v
-	}
 	pubKey, ok := rekorPubKeys[payload.LogID]
 	if !ok {
 		return errors.New("rekor log public key not found for payload")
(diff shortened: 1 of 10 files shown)


dlorenc:
Should we drop WIP from the title?

dlorenc:
Cool! Nice job here :)

asraa:
cc @dlorenc @vaikas @haydentherapper @znewman01

pkg/cosign/tuf/client.go
```diff
@@ -432,9 +437,38 @@ func (t *TUF) updateMetadataAndDownloadTargets() error {
...
-		return fmt.Errorf("error updating to TUF remote mirror: %w\nremote status:%s", err, string(b))
+		return nil, fmt.Errorf("error updating to TUF remote mirror: %w\nremote status:%s", err, string(b))
+	}
+	// Success! Cache new metadata, if needed.
+	if noCache() {
+		return targets, nil
+	}
+	// Sync the in-memory local store to the on-disk cache.
+	tufDB := filepath.FromSlash(filepath.Join(rootCacheDir(), "tuf.db"))
+	diskLocal, err := tuf_leveldbstore.FileLocalStore(tufDB)
```
haydentherapper, line 450: Is this API expected to be used, or are we reaching into private APIs? I just wanna make sure this is maintainable.

pkg/cosign/tuf/client.go
```diff
@@ -528,16 +562,40 @@ func cachedTargetsDir(cacheRoot string) string {
 	return filepath.FromSlash(filepath.Join(cacheRoot, "targets"))
 }
 
+func syncLocalMeta(from client.LocalStore, to client.LocalStore) error {
+	// Otherwise populate the in-memory local store with data fetched from the cache.
```
haydentherapper, line 566: nit: This comment isn't accurate, correct? Cause this can sync to or from local/disk, depending on if it's read or write?

pkg/cosign/tuf/client.go
```diff
@@ -432,9 +437,38 @@ func (t *TUF) updateMetadataAndDownloadTargets() error {
...
+	diskLocal, err := tuf_leveldbstore.FileLocalStore(tufDB)
+	defer func() {
+		if diskLocal != nil {
+			diskLocal.Close()
+		}
+	}()
+	if err != nil {
+		return nil, fmt.Errorf("creating cached local store: %w", err)
+	}
+	if err := syncLocalMeta(t.local, diskLocal); err != nil {
```
haydentherapper, line 459: nit: Can you add a comment to be explicit that this writes data to disk from in-memory? This helps differentiate it from its other use.

pkg/cosign/tuf/client.go
```diff
@@ -306,11 +314,9 @@ func NewFromEnv(ctx context.Context) (*TUF, error) {
 
 func Initialize(ctx context.Context, mirror string, root []byte) error {
```
haydentherapper, line 315: (this comment is for the return for NewFromEnv)

Shouldn't we be pulling the root from the cache if present? Otherwise, we're always fetching all roots starting with the root of trust. I'd hope there's some mechanism in TUF to ask for the latest cached root, and it verifies from a trusted root.

pkg/cosign/tuf/client.go
```diff
@@ -213,15 +222,9 @@ func GetRootStatus(ctx context.Context) (*RootStatus, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer t.Close()
 	return t.getRootStatus()
 }
 
-// Close closes the local TUF store. Should only be called once per client.
-func (t *TUF) Close() error {
```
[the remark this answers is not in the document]
vaikas: Yup, that's right! For longer running processes (like the webhook) there should be a way to refresh the data, but @asraa has some ideas there.

pkg/cosign/tuf/client.go
```diff
@@ -46,10 +47,12 @@ const (
...
 
-// Global in-memory targets to avoid re-downloading when there is no local cache.
-// TODO: Consider using this map even when local caching to avoid reading from disk
-// multiple times (e.g. when there are multiple signatures to verify in a single call).
-var memoryTargets = map[string][]byte{}
+var (
+	// singletonTUF holds a single instance of TUF that will get reused on
+	// subsequent invocations of initializeTUF
+	singletonTUF   *TUF
+	singletonTUFMu sync.Mutex
```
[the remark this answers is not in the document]
vaikas: We probably could. I just wasn't sure if we'd be swapping the TUF object on update, or just modify parts of it on update. I think we might just be able to modify the in-mem map during updates, but as I said wasn't sure what the level of modification was right so started with this.

pkg/cosign/tuf/client.go
```diff
@@ -105,6 +108,12 @@ type remoteCache struct {
 	Mirror string `json:"mirror"`
 }
 
+func resetForTests() {
+	singletonTUFMu.Lock()
```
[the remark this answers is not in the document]
vaikas: Since some of the tests are expecting a 'fresh' TUF setup, I added this to make sure we don't get any errors from modifying it without a lock in place.

vaikas (commented):
Thanks @asraa for getting this wrangled into shape and @haydentherapper for the reviews!.

pkg/cosign/tlog.go
```diff
@@ -77,10 +77,15 @@ func getLogID(pub crypto.PublicKey) (string, error) {
...
-// There is an Env variable that can be used to override this behaviour:
+// A Rekor client may optionally be provided in case using SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY
+// (see below).
+// There are two Env variable that can be used to override this behaviour:
 // SIGSTORE_REKOR_PUBLIC_KEY - If specified, location of the file that contains
 // the Rekor Public Key on local filesystem
-func GetRekorPubs(ctx context.Context) (map[string]RekorPubKey, error) {
+// SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY - If specified, fetches the Rekor public
+// key from the Rekor server using the provided rekorClient.
+// TODO: Rename SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY to be test-only or remove.
```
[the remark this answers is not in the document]
vaikas: One thought here is that being able to fetch the public key from Rekor is part of the Rekor API, and whether this flag is here or not is only one part of getting used 'outside of testing'.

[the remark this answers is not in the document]
haydentherapper: I think it’s fine to provide an API to fetch the public key, but there needs to be an explicit strategy around root of trust management. Ideally it’d always be TUF for Sigstore, but consumers may choose to TOFU - but they must persist the key, verification logic can’t rely on fetching the key each time.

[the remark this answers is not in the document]
asraa: Agree it should be removed -- but I'd like to do this in another PR because the current webhook and tests rely on it and I need to dig in separately into that.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: For longer running processes we are also refreshing now -- this is handled in `updateClient`: we Update our local in-memory metadata, then syncMetadata to the DB.

pkg/cosign/tuf/client.go
asraa: Nice catch! Some of those comments ended up in the wrong place (as you noted in the method's internal comment)

pkg/cosign/tuf/client.go
asraa: Yes! Sorry this was meant to be in caller comment, not internal. Moved, and clarified this comment.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: Actually, I dug in while I was attempting this change: we don't end up updating the TUF client unless we re-initialize. Since this PR uses the singleton/mutex we don't actually support long-running processes. I'm going to change to `Once` since it's behavior will be the same as a mutex, and call out a follow-up to support long-running process updates in the Cleanup issue (in practice this isn't that big of an issue right now).

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: Tests within the tuf pkg will be sequential, and  we need the reset (but as you said, don't need the lock).

There is 1 test outside the TUF package that uses a custom TUF root. I'm concerned about that, that may be run in parallel. I will be refactoring that anyway to use a custom `CheckOpts` for the Rekor key rather than a custom TUF set up.

In the interim, I need to think how to lock that test.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
haydentherapper: I'm a little confused why we need the mutex.  Is it because there's a concern that `initializeTUF` will called multiple times by other processes?

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
haydentherapper: Thanks, I think Once is cleaner since it explicitly shows that the singleton should be initialized only once.

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: It shouldn't -- I switched over to `Once` initialization, and reset the `sync.Once` for tests.

The only time it will be called multiple times potentially in parallel is because of the one test outside the TUF pkg that may run in parallel with the TUF tests.

pkg/cosign/tuf/client.go
haydentherapper: Ah i had missed that line, looks good!

pkg/cosign/tuf/client.go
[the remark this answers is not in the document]
asraa: OK resolution: I added a TODO for that test that uses a custom root outside this package, expect a PR this afternoon addressing that.

pkg/cosign/tuf/client.go
```diff
@@ -50,8 +50,8 @@ const (
 var (
 	// singletonTUF holds a single instance of TUF that will get reused on
 	// subsequent invocations of initializeTUF
-	singletonTUF   *TUF
-	singletonTUFMu sync.Mutex
+	singletonTUF     *TUF
+	singletonTUFOnce = new(sync.Once)
```
haydentherapper, line 54: Wow, I actually didn't know that go had allocation via `new`. 

I don't mind this approach, but I found that someone else had the same problem and made sync but with Reset - https://github.com/matryer/resync

pkg/cosign/tuf/client.go
```diff
@@ -235,66 +233,62 @@ func GetRootStatus(ctx context.Context) (*RootStatus, error) {
...
 
-	if err := t.client.InitLocal(root); err != nil {
-		return nil, fmt.Errorf("unable to initialize client, local cache may be corrupt: %w", err)
-	}
+		// We may already have an up-to-date local store! Check to see if it needs to be updated.
+		trustedTimestamp, ok := trustedMeta["timestamp.json"]
+		if ok && !isExpiredTimestamp(trustedTimestamp) && !forceUpdate {
+			// We're golden so stash the TUF object for later use
+			singletonTUF = t
+			return
```
haydentherapper, line 280: Is this returning nil?

pkg/cosign/tuf/client.go
```diff
@@ -235,66 +233,62 @@ func GetRootStatus(ctx context.Context) (*RootStatus, error) {
...
 
-	t := &TUF{
-		mirror:   mirror,
-		embedded: embedded,
-	}
+		t.targets = newFileImpl()
+		var err error
+		t.local, err = newLocalStore()
+		if err != nil {
+			panic(err)
```
haydentherapper, line 247: Darn, is this the consequence of sync.Once? If so, this isn't viable imo, we shouldn't add a panic into the client because if a server starts depending on it, it opens up an opportunity for query of deaths. Is there a way to return an error? If not, what you had before would be fine.

haydentherapper (changes requested):

pkg/cosign/tuf/client.go
vaikas: +1 on not panicing.

haydentherapper: Could do a global variable - https://stackoverflow.com/questions/42069615/too-many-arguments-to-return-error

Just ran into this same issue in fulcioroots

pkg/cosign/tuf/client.go
asraa: It's returning out of the `sync.Once` func, but we still return `singletonTUF` from `initializeTUF`

pkg/cosign/tuf/client.go
asraa: Using the global!

pkg/cosign/tuf/client.go
asraa: yeah i saw this as well and ideally wanted to avoid a totally new dependency just for testing -- I'm not sure if there's a big downside to this approach though.

haydentherapper (approved):
Nice work!

znewman01:
LGTM, great work on this!

dlorenc (approved):

PR #1956, opened 2022-06-03
Separate RegExp matching of issuer/subject from strict

### LINKED ISSUE #1871: Cosigned policies interpreted as regex by default
> Unrelated to this PR but it's a regex??? That's a surprise
> 
> Does this open up an attack where someone passes `zjn@mail.example.com` and I go register `mailqexample.com` and can pass the check?

_Originally posted by @znewman01 in https://github.com/sigstore/cosign/pull/1869#discussion_r870839195_

I don't think this is necessarily a problem *except* users shouldn't be surprised by it or do the wrong thing by default. So maybe: make it loudly warn, or make regex behavior configurable? Or support globs instead?

vaikas:

#### Summary
Add two new fields issuerRegExp and subjectRegExp so that if the user wants RE matching of issuer/subject, it's crystal clear
that RE is being used. issuer / subject are now strict matches only.

#### Ticket Link
Fixes #1871 

#### Release Note

```release-note
CIP issuer/subject are now strict matches. Add two new fields issuerRegExp and subjectRegExp that will then use RE for matching.
```

### DIFF
--- a/.github/workflows/kind-cluster-image-policy-with-attestations.yaml
+++ b/.github/workflows/kind-cluster-image-policy-with-attestations.yaml
@@ -90,6 +90,7 @@ jobs:
         kubectl rollout status --timeout 5m --namespace cosign-system deployments/webhook
 
     - name: Run Cluster Image Policy Tests with attestations
+      timeout-minutes: 15
       run: |
         ./test/e2e_test_cluster_image_policy_with_attestations.sh
 
--- a/config/300-clusterimagepolicy.yaml
+++ b/config/300-clusterimagepolicy.yaml
@@ -276,7 +276,11 @@ spec:
                               properties:
                                 issuer:
                                   type: string
+                                issuerRegExp:
+                                  type: string
                                 subject:
+                                  type: string
+                                subjectRegExp:
                                   type: string
                           url:
                             type: string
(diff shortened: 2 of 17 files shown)


config/300-clusterimagepolicy.yaml
```diff
@@ -125,8 +125,12 @@ spec:
                               properties:
                                 issuer:
                                   type: string
+                                issuerRE:
```
hectorj2f, line 128: Suggestion: perhaps it is just me, but wouldn't it read better to use `issuerRegExp` instead of `issuerRE`?

hectorj2f (approved):
lgtm

.github/workflows/kind-cluster-image-policy-with-attestations.yaml
```diff
@@ -90,6 +90,7 @@ jobs:
         kubectl rollout status --timeout 5m --namespace cosign-system deployments/webhook
 
     - name: Run Cluster Image Policy Tests with attestations
+      timeout-minutes: 15
```
hectorj2f, line 93: timeout = 15min 🤔

hectorj2f (approved):
lgtm

PR #1971, opened 2022-06-07
FIX: update ct/otel and etcd

SantiagoTorres:
This somehow requires some fun with replace rules (perhaps a better go
mod whisperer can find a more elegant solution).

This would allow us to update opa and get rid of the two vulns on <0.40.0.


cpanato:
closing in favor of #2054

Issue #1954, opened 2022-06-02
Ability to block admission if image SBOM contains specific package (defined in Cue)


jdolitsky:
Currently, the policy admission controller is able to block admission based on the presence of certain attestations, evaluated against some Cue code. For example, from [here](https://github.com/sigstore/cosign/blob/main/test/testdata/policy-controller/e2e/cip-requires-two-signatures-and-two-attestations.yaml#L15):

```yaml
...
  policy:
    type: cue
    data: |
      package sigstore
      import "struct"
      import "list"
      authorityMatches: {
        keyatt: {
          attestations: struct.MaxFields(1) & struct.MinFields(1)
        },
...
```

I'd like the ability to, with similar syntax, prevent admission of a container with a SBOM that indicates it has a vulnerable package.

For example, the Log4Shell vulnerability ([CVE-2021-44228](https://nvd.nist.gov/vuln/detail/CVE-2021-44228)) affects several versions of the packages "log4j-api" and "log4j-core". Given that I have a attached an SBOM to my image, I'd like to be able to prevent admission based on whether a vulnerable version of one of these packages is listed in that SBOM.

Not sure what the inline Cue might look like, but something similar to `authorityMatches`, such as `packageMatches` where we can specify that specific versions of specific packages are not to be allowed.

jdolitsky:
Maybe also, as a prerequisite, the ability to block on whether the image has an SBOM at all

vaikas:
I'm sure it's a hugely unpopular opinion, but I'm curious if we might want to revisit having sbom as a separate entity from a regular attestation? We have signatures, attestations, and sboms. We even have SBOMs as attestations, in a standard in-toto format. So, I guess why not just create an SBOM as an attestation and then use the policy like you've linked to evaluate it just like any other attestation. Having a parallel way to define policies just for SBOMs just feels a bit wonky.

hectorj2f:
@jdolitsky The complexity of processing sboms (multiple formats, spdx or cyclonex ... in xml, json) and how to find those sboms is something that you don't want to do on an admission controller. I'd rather expect certain attestations to know if an image does not contain a certain package.

PR #1974, opened 2022-06-07
Add spdxjson predicate type for attestations

jdolitsky:
Related to #1954

In attempt to dive deeper into the SBOM structure in attestations, added new predicate type "spdxjson" which allows us to apply Cue policies against inner fields in JSON-formatted SPDX documents. For example:

```
predicate: {
    Data: {
        spdxVersion: "SPDX-2.2"
    }
}
```

In the case of the "spdx" type, the `Data` field is just a string which prevents further progress on this.

It's still not 100% clear how Cue can be used to detect a specific package version within an SBOM, but this gets us one step further. Happy to follow up with "cyclone" predicate type too.


dlorenc (approved):

hectorj2f (approved):

developer-guy:
Approve party! Great work @jdolitsky 👏

developer-guy (approved):

Issue #1086, opened 2021-11-20
`cosign download sbom` for multi-arch images


mattmoor:
**Description**

This came up developing: https://github.com/google/ko/pull/506

Basically what we're doing in `ko` with multi-arch builds is producing an SBOM attached to each image.  However, `cosign download sbom` only does a shallow look, and reports that there are no SBOMs.

I think that we should first check the index, but if nothing is found (or maybe regardless) walk the child images looking for additional SBOMs.  This should be trivial to do by wrapping what we have in `walk.SignedEntity` to collect things recursively.

mattmoor:
We could also take a `--platform` option to resolve a particular platform's SBOM, if we choose to default `linux/amd64`, as some clients do.

mattmoor:
cc @imjasonh FYI

ribbybibby:
I bumped into this recently. For now I'm doing the following to resolve the platform-specific image, but it might be nice if `cosign` could do this itself.
```
cosign download sbom ghcr.io/google/ko@$(crane digest ghcr.io/google/ko:latest --platform=linux/amd64)
```

It occurs to me that this isn't just an issue for `download sbom`. It could equally apply to signatures, attestations and basically any command where you're actually interested in the platform-specific image rather than the index. Or where you're interested in all the children of an index.

Perhaps it requires a more general solution for handling multi-arch images? I know `sign`, for instance, already has a `--recursive` option.

Or, alternatively, leave it up to the caller of cosign to figure out which images they want to target? Like I'm doing above with `crane`.

mattmoor:
cc @puerco given recent conversations about multi-arch.

imjasonh:
> ```
> cosign download sbom ghcr.io/google/ko@$(crane digest ghcr.io/google/ko:latest --platform=linux/amd64)
> ```

This seems like it means we should just support `cosign download sbom [IMG] --platform`, where if platform is unset, fetch the SBOM for the image or index, and if it is set, iff it's an index, fetch the SBOM attached to that platform's image.

(Same for all the attachment types, not just SBOMs)

ribbybibby:
@imjasonh That would be great. I think it could also extend more widely to commands like `triangulate`.

amouat:
This bit me recently. Whilst we're working on a solution, would it be worth rewording the error message? At the moment I get: 

```
Error: image not found in registry
main.go:52: error during command execution: image not found in registry
```

[...]

PR #1975, opened 2022-06-08
Add --platform flag to cosign sbom download 

puerco:
#### Summary

This PR adds a `--platform` flag to `cosign sbom download `. When downloading an SBOM from a multiarch index, the new flag lets the user specify the SBOM of one of the images listed in the index. 

When trying to download an SBOM from an inbox that does not have one attached, cosign will now present the user with the available architectures:

```
This multiarch image does not have an SBOM attached at the index level.
Try using --platform with one of the following architectures:
 linux/amd64, linux/arm/v5, linux/arm/v7, linux/arm64/v8, linux/386, linux/mips64le, 
 linux/ppc64le, linux/s390x, windows/amd64:10.0.20348.707, windows/amd64:10.0.17763.2928

Error: no SBOM found attached to image index
```

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1086

/cc @jdolitsky @imjasonh  @mattmoor 

#### Release Note

```release-note
- New `--platform` flag  enables downloading a specific sbom from a multi-arch image
- Improved handling of error messages when a multiarch index does not have an sbom attached
```

### DIFF
--- a/cmd/cosign/cli/download/sbom.go
+++ b/cmd/cosign/cli/download/sbom.go
@@ -28,7 +28,6 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/pkg/oci"
 	ociremote "github.com/sigstore/cosign/pkg/oci/remote"
-	"github.com/sigstore/cosign/pkg/oci/walk"
 )
 
 type platformList []struct {
@@ -91,7 +90,7 @@ func SBOMCmd(
 			)
 		}
 
-		nse, err := findSignedImage(ctx, idx, platforms[0].hash)
+		nse, err := idx.SignedImage(platforms[0].hash)
 		if err != nil {
 			return nil, fmt.Errorf("searching for %s image: %w", platforms[0].hash.String(), err)
 		}
@@ -193,26 +192,4 @@ func matchPlatform(base *v1.Platform, list platformList) platformList {
 	}
 
 	return ret
-}
-
-func findSignedImage(ctx context.Context, sii oci.SignedImageIndex, iDigest v1.Hash) (se oci.SignedEntity, err error) {
-	if err := walk.SignedEntity(ctx, sii, func(ctx context.Context, e oci.SignedEntity) error {
-		img, ok := e.(oci.SignedImage)
-		if !ok {
-			return nil
-		}
-
-		d, err := img.Digest()
-		if err != nil {
-			return fmt.Errorf("getting image digest: %w", err)
-		}
-
-		if d == iDigest {
-			se = img
-		}
-		return nil
-	}); err != nil {
-		return nil, fmt.Errorf("traversing signed entity: %w", err)
-	}
-	return se, nil
 }
--- a/cmd/cosign/cli/download/sbom.go
+++ b/cmd/cosign/cli/download/sbom.go
@@ -26,7 +26,10 @@ import (
 	ociremote "github.com/sigstore/cosign/pkg/oci/remote"
 )
 
-func SBOMCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef string, out io.Writer) ([]string, error) {
+func SBOMCmd(
+	ctx context.Context, regOpts options.RegistryOptions,
+	dnOpts options.SBOMDownloadOptions, imageRef string, out io.Writer,
+) ([]string, error) {
 	ref, err := name.ParseReference(imageRef)
 	if err != nil {
 		return nil, err
--- a/cmd/cosign/cli/download/sbom.go
+++ b/cmd/cosign/cli/download/sbom.go
@@ -17,6 +17,7 @@ package download
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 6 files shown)


imjasonh:
This conversation is making me want to finally provide a method in ggcr to filter manifests by platform, encoding all this bonkers crazy logic.

cmd/cosign/cli/download/sbom.go
```diff
@@ -66,3 +142,78 @@ func SBOMCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef stri
...
+		platforms = append(platforms, struct {
+			hash     v1.Hash
+			platform *v1.Platform
+		}{m.Digest, m.Platform})
+	}
+	return platforms, nil
+}
+
+// matchPlatform filters a list of platforms returning only those matching
+// a base. "Based" on ko's internal equivalent while it moves to GGCR.
```
imjasonh, line 166:
```suggestion
// a base. "Based" on ko's internal equivalent while it moves to GGCR.
// https://github.com/google/ko/blob/e6a7a37e26d82a8b2bb6df991c5a6cf6b2728794/pkg/build/gobuild.go#L1020
```

test/e2e_test.go
```diff
@@ -944,7 +949,7 @@ func TestAttachSBOM(t *testing.T) {
 	// Upload it!
 	must(attach.SBOMCmd(ctx, options.RegistryOptions{}, "./testdata/bom-go-mod.spdx", "spdx", imgName), t)
```
imjasonh, line 950: Should attach also take a platform? 🤔

cmd/cosign/cli/download/sbom.go
```diff
@@ -42,9 +55,72 @@ func SBOMCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef stri
...
+			if len(pl) > 0 && err == nil {
+				fmt.Fprintln(os.Stderr, "\nThis multiarch image does not have an SBOM attached at the index level.")
+				fmt.Fprintln(os.Stderr, "Try using --platform with one of the following architectures:")
+				as := []string{}
+				for _, a := range pl {
+					if a.platform.String() != "" {
+						as = append(as, a.platform.String())
+					}
+				}
+				fmt.Fprintf(os.Stderr, " %s\n\n", strings.Join(as, ", "))
```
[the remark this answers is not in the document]
puerco: Ready!

cmd/cosign/cli/download/sbom.go
```diff
@@ -42,9 +55,72 @@ func SBOMCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef stri
...
+		if nse == nil {
+			return nil, fmt.Errorf("unable to find image %s", platforms[0].hash.String())
+		}
+		se = nse
+	}
+
 	file, err := se.Attachment("sbom")
 	if err != nil {
-		return nil, err
+		if errors.Is(err, ociremote.ErrImageNotFound) {
```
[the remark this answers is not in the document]
puerco: Done, reworked the block

test/e2e_test.go
puerco: OK, I'll add `--platform` to `cosign sbom attach` in a follow-up

cmd/cosign/cli/download/sbom.go
```diff
@@ -17,16 +17,37 @@ package download
...
+
+func (pl *platformList) String() string {
+	r := []string{}
+	for _, p := range *pl {
+		r = append(r, p.platform.String())
+	}
+	return strings.Join(r, ", ")
+}
+
+func SBOMCmd(
```
imjasonh, line 47: (not blocking) It looks like the `[]string` return value is only ever read in test/e2e_test.go, and never in the actual CLI command. Is that right?

Maybe we should just have this return `error` alone, and refactor tests to not rely on the `[]string` return value.

cmd/cosign/cli/download/sbom.go
```diff
@@ -66,3 +140,79 @@ func SBOMCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef stri
...
+			}
+		}
+		ret = append(ret, p)
+	}
+
+	return ret
+}
+
+func findSignedImage(ctx context.Context, sii oci.SignedImageIndex, iDigest v1.Hash) (se oci.SignedEntity, err error) {
+	if err := walk.SignedEntity(ctx, sii, func(ctx context.Context, e oci.SignedEntity) error {
```
imjasonh, line 199: `sii.SignedImage(iDigest)` should do the same, without making you write the `walk` logic yourself. Let me know if that doesn't work for some reason.

imjasonh (approved):

cmd/cosign/cli/download/sbom.go
puerco: Yes, I agree. I think the command was meant to return more than one sbom but it is short-circuited to spit out the one sbom. We should discuss if it will be returning more than one SBOM and decide how to change this.  I think the logic to download the SBOMs (and the rest) should me moved away from the CLI to a more general package.

cmd/cosign/cli/download/sbom.go
puerco: Yay thanks for the tip @imjasonh :tada:  I pushed another commit dropping the function and switching to `SignedImage()`

dlorenc (approved):

PR #1977, opened 2022-06-08
Add cyclonedx predicate type for attestations

jdolitsky:
Follow up PR to #1974, but for CycloneDX SBOMs

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -34,6 +34,9 @@ const (
 
 	// CosignVulnProvenanceV01 specifies the type of VulnerabilityScan Predicate
 	CosignVulnProvenanceV01 = "cosign.sigstore.dev/attestation/vuln/v1"
+
+	// PredicateCycloneDX represents a SBOM using the CycloneDX standard.
+	PredicateCycloneDX = "https://cyclonedx.org/Document"
 )
 
 // CosignPredicate specifies the format of the Custom Predicate.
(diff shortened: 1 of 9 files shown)


pkg/cosign/attestation/attestation.go
```diff
@@ -34,6 +34,9 @@ const (
 
 	// CosignVulnProvenanceV01 specifies the type of VulnerabilityScan Predicate
 	CosignVulnProvenanceV01 = "cosign.sigstore.dev/attestation/vuln/v1"
+
+	// PredicateCycloneDX represents a SBOM using the CycloneDX standard.
+	PredicateCycloneDX = "https://cyclonedx.org/Document"
```
jdolitsky, line 39: Was not sure the proper value for this. This is based on `https://spdx.dev/Document` defined within in-toto: https://github.com/in-toto/in-toto-golang/blob/af1f9fb822bf841779c916a954f651e908c5a818/in_toto/model.go#L81

puerco: Filed this one on in-toto go, we could use the same value here: https://github.com/in-toto/in-toto-golang/pull/169

jdolitsky: Saw your comment there. Just changed to `/schema`

hectorj2f (approved):

dlorenc:
Needs a rebase

dlorenc (approved):

PR #2014, opened 2022-06-22
Fix #1378 create new attestation signature in replace mode if not existent

### LINKED ISSUE #1378: cosign attest --replace fails to create attestations on random.Image() generated images
While implementing the unit test for #1366, I discovered that `cosign attest --replace` silently fails to create any attestations on images created by [go-containerregistry's `random.Image()`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/random#Image). That PR added a failing `TestAttestationReplace()` test which is currently skipped because the replace option doesn't create any attestations.

This can be reproduced outside the unit tests like this (using the v1.5.0 release binary):
```bash
>random.go <<EOF
package main

import (
	"context"
	"log"

	"github.com/google/go-containerregistry/pkg/authn"
	"github.com/google/go-containerregistry/pkg/name"
	"github.com/google/go-containerregistry/pkg/v1/random"
	"github.com/google/go-containerregistry/pkg/v1/remote"
)

func main() {
	n := "localhost:5000/x/y"

	ref, err := name.ParseReference(n, name.WeakValidation)
	if err != nil {
		log.Fatal(err)
	}
	img, err := random.Image(512, 5)
	if err != nil {
		log.Fatal(err)
	}

	regClientOpts := registryClientOpts(context.Background())

	if err := remote.Write(ref, img, regClientOpts...); err != nil {
		log.Fatal(err)
	}
}

func registryClientOpts(ctx context.Context) []remote.Option {
	return []remote.Option{
		remote.WithAuthFromKeychain(authn.DefaultKeychain),
		remote.WithContext(ctx),
	}
}
EOF
# Create an image localhost:5000/x/y using random.Image()
go run random.go
# The --replace option silently fails to create any attestations on images created by random.Image()
cosign attest --replace --predicate predicate.json localhost:5000/x/y --key cosign.key
# The following command shows no attestations exist
cosign download attestation "localhost:5000/x/y"
```

The `cosign attest --replace` option works fine on other images. I tested using images copied from Docker Hub using skopeo, and was able to create and replace attestations.

Syquel:
#### Summary
Currently it is not possible to use `cosign attest` with the `--replace` option if either the attestation artifact or the attestation predicate does not exist.  
Adds the ability to `cosign attest` with the `--replace` option to create new attestation artifacts or append new predicates to an existing attestation artifact.

#### Ticket Link
Fixes #1378 

#### Release Note
```release-note
Fixed the ability of `cosign attest` with the `--replace` option to create new attestations
```

### DIFF
--- a/pkg/oci/mutate/mutate.go
+++ b/pkg/oci/mutate/mutate.go
@@ -242,7 +242,7 @@ func (si *signedImage) Attestations() (oci.Signatures, error) {
 		if err != nil {
 			return nil, err
 		}
-		return AppendSignatures(replace)
+		return ReplaceSignatures(replace)
 	}
 	return AppendSignatures(base, si.att)
 }
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -295,11 +295,7 @@ func attestVerify(t *testing.T, predicateType, attestation, goodCue, badCue stri
 	mustErr(verify(pubKeyPath, imgName, true, map[string]interface{}{"foo": "bar"}, ""), t)
 }
 
-func TestAttestationReplace(t *testing.T) {
-	// This test is currently failing, see https://github.com/sigstore/cosign/issues/1378
-	// The replace option for attest does not appear to work on random.Image() generated images
-	t.Skip()
-
+func TestAttestationReplaceCreate(t *testing.T) {
 	repo, stop := reg(t)
 	defer stop()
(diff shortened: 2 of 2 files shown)


Syquel:
Is there something left for me to do to get this PR merged?  
Would be great to have it in the next release.

dlorenc (approved):

PR #2016, opened 2022-06-22
sign-blob: bundle should work independently and respect `--output-certificate` and `--output-signature`

### LINKED ISSUE #1821: `sign-blob --bundle` clobbers `--output-{signature,certificate}`
If I pass all three of `--output-signature`, `--output-certificate`, and `--bundle`, Cosign silently only respects `--bundle`.

```shell
$ cosign sign-blob /dev/null --output-signature /tmp/keyless.sig --output-certificate /tmp/keyless.crt --bundle /tmp/keyless.bundle
[...]
$ ls /tmp/keyless.*
/tmp/keyless.bundle
```

I expect either:

1. I get all three: a signature, a bundle, a cert (preferred)
2. An error from the option parser

Guilty line looks like: https://github.com/sigstore/cosign/blob/e74f180ce697dcde62dbe12f48f1a6a26522ea77/cmd/cosign/cli/sign/sign_blob.go#L86

It appears that `cosign sign` does not support `--bundle`, though I'm not sure why not. CC @priyawadhwa who added it in https://github.com/sigstore/cosign/pull/1306

### LINKED ISSUE #2005: passing '--bundle' flag causes cosign to ignore '--output-certificate' and '--output-signature'
Good news is this behavior is the same in keyless and keyfull mode. 

to reproduce:
1) run ` COSIGN_EXPERIMENTAL=true cosign sign-blob java-service.jar --output-certificate foo.pem --output-signature foo.sig --bundle foo.bundle`
2. run `$ ls `and observe that your foo.sig and foo.pem files were never written.

Dentrax:
Fixes #1821
Fixes #2005

cc @priyawadhwa

#### Summary

```
$ COSIGN_EXPERIMENTAL=true ./cosign sign-blob /dev/null --output-certificate foo.pem --output-signature foo.sig --bundle foo.bundle

Bundle wrote in the file foo.bundle
Signature wrote in the file foo.sig
Certificate wrote in the file foo.pem
```

#### Ticket Link

Fixes

#### Release Note

```release-note
Fixed an issue where --bundle flag does not work independently with `--output-certificate` and `--output-signature` in sign-blob 
```

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -91,7 +91,10 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
 		if err != nil {
 			return nil, err
 		}
-		return []byte(signedPayload.Base64Signature), os.WriteFile(ko.BundlePath, contents, 0600)
+		if err := os.WriteFile(ko.BundlePath, contents, 0600); err != nil {
+			return nil, fmt.Errorf("create bundle file: %w", err)
+		}
+		fmt.Printf("Bundle wrote in the file %s\n", ko.BundlePath)
 	}
 
 	if outputSignature != "" {


priyawadhwa (approved):

priyawadhwa:
Thanks for fixing this @Dentrax !

PR #2018, opened 2022-06-23
Attempt to clean up pkg/cosign

imjasonh:
#### Summary

`pkg/cosign` is a bit of a kitchen sink at the moment, with [58 total imports](https://pkg.go.dev/github.com/sigstore/cosign/pkg/cosign?tab=imports), and lots of exported methods that only make sense to be used by the CLI and shouldn't be used by external packages. I started doing more and wanted to stop before making this too big. I'll send another PR to move stuff to internal packages.

- remove pkg/cosign.SetSkipConfirmation, move it from a root option to a
  per-method option when it's actually used, remove the global bool.
- add TODOs to move more methods to internal packages, or to the only
  places they're used.
- unexport some stuff that can be unexported easily.

Things like [`pkg/cosign.FindTLogEntry`](https://pkg.go.dev/github.com/sigstore/cosign/pkg/cosign#FindTlogEntry) seem generally useful, and probably belong in sigstore/sigstore, if they're not already.

#### Release Note

```release-note
Some types and methods not likely to be used outside of the cosign CLI in pkg/cosign are removed
```

### DIFF
--- a/cmd/cosign/cli/clean.go
+++ b/cmd/cosign/cli/clean.go
@@ -16,11 +16,13 @@
 package cli
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
@@ -28,7 +30,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
-	"github.com/sigstore/cosign/pkg/cosign"
 	ociremote "github.com/sigstore/cosign/pkg/oci/remote"
 )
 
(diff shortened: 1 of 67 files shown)


hectorj2f (dismissed):

dlorenc (approved):

asraa:
> Things like [pkg/cosign.FindTLogEntry](https://pkg.go.dev/github.com/sigstore/cosign/pkg/cosign#FindTlogEntry) seem generally useful, and probably belong in sigstore/sigstore, if they're not already.

I wonder if there's a good way to put this in rekor over sigstore/sigstore. It feels more appropriate there, since it's very specific to rekor's API

haydentherapper:
Sorry, just saw this PR - @imjasonh , why did we remove the global skip flag? This was intentional so we wouldn't have to worry about duplication.

haydentherapper:
Agreed, don't want to let users use this flag. If not in internal, another option would be under `cli`? I agree that it might be misleading in some cases, though I'd like it to be used to skip confirmations for all commands going forward.

imjasonh:
> Agreed, don't want to let users use this flag. If not in internal, another option would be under `cli`? I agree that it might be misleading in some cases, though I'd like it to be used to skip confirmations for all commands going forward.

That's what's potentially confusing though, many commands don't have a confirmation to skip, so it's misleading (I thought) to have the option.

I don't feel that strongly though about having a global `--skip-confirmation` flag or env option, so if you want it I can put it back, and just have that global flag do nothing most of the time, and pipe through `o.SkipConfirmation` when it's needed.

haydentherapper:
I don't feel super strongly. I'd like there to not be duplication for the prompt functions though for uniformity, so if they were centralized in one location, that'd be great.

cmd/cosign/cli/clean.go
```diff
@@ -49,9 +50,21 @@ func Clean() *cobra.Command {
 	return cmd
 }
 
+// confirmPromptDestructive prompts the user for confirmation for an action. Ignores
+// skipConfirmation.
+func confirmPromptDestructive(msg string) (bool, error) {
```
imjasonh, line 55: @haydentherapper do you mean replacing this with a `cosign.ConfirmPrompt(msg, false)`? If so that sounds like an easy fix I can send out shortly.

haydentherapper: Yea, just having one destructive prompt function, one normal confirm prompt function.

PR #2039, opened 2022-06-30
Use cosign.ConfirmPrompt more consistently

imjasonh:
Also unexport some needlessly exported consts in cli/fulcio/fulcio.go --
these should be implemented via sigstore/sigstore.

Followups from #2018 

@haydentherapper 

```release-note
NONE
```

### DIFF
--- a/cmd/cosign/cli/options/clean.go
+++ b/cmd/cosign/cli/options/clean.go
@@ -27,5 +27,6 @@ var _ Interface = (*CleanOptions)(nil)
 func (c *CleanOptions) AddFlags(cmd *cobra.Command) {
 	c.Registry.AddFlags(cmd)
 	cmd.Flags().StringVarP(&c.CleanType, "type", "", "all", "a type of clean: <signature|attestation|sbom|all> (default: all)")
+	// TODO: Rename to --skip-confirmation for consistency?
 	cmd.Flags().BoolVarP(&c.Force, "force", "f", false, "do not prompt for confirmation")
 }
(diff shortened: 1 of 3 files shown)


haydentherapper (approved):
thanks!

cmd/cosign/cli/options/clean.go
```diff
@@ -27,5 +27,6 @@ var _ Interface = (*CleanOptions)(nil)
 func (c *CleanOptions) AddFlags(cmd *cobra.Command) {
 	c.Registry.AddFlags(cmd)
 	cmd.Flags().StringVarP(&c.CleanType, "type", "", "all", "a type of clean: <signature|attestation|sbom|all> (default: all)")
+	// TODO: Rename to --skip-confirmation for consistency?
```
developer-guy, line 30: it makes sense a lot to me, may I open an issue for this? @imjasonh

imjasonh: Sure! I think we'll probably want `--force` to continue to exist for 1+ release, with a warning that it's renamed to `--skip-confirmation`, and use `--force`'s value to set `o.SkipConfirmation`. If both are set and they disagree, then fail. Does that sound right?

developer-guy: Absolutely !

developer-guy: Issue: https://github.com/sigstore/cosign/issues/2044

PR #2041, opened 2022-07-01
Refactor + Minor Fixes in key gen and key import commands

sbs2001:
Refactors:

- [X] Use the `ConfirmPrompt` function everywhere instead of rolling prompt  implementation everywhere.
- [ ] Move `pkg/cosign/common.go` to internal package.
- [ ] Maybe have a single function for writing + handling perms for  private & public key.

Fixes:

Address https://github.com/sigstore/cosign/blob/1d224d924197d3fab3f624a5cbe27f25d934e9f2/cmd/cosign/cli/generate/generate_key_pair.go#L101 and 
https://github.com/sigstore/cosign/blob/1d224d924197d3fab3f624a5cbe27f25d934e9f2/cmd/cosign/cli/importkeypair/import_key_pair.go#L52

### DIFF
--- a/cmd/cosign/cli/generate/generate_key_pair.go
+++ b/cmd/cosign/cli/generate/generate_key_pair.go
@@ -86,15 +86,12 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string, args []string) error
 	}
 
 	if cosign.FileExists("cosign.key") {
-		var overwrite string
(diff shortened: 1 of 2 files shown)


sbs2001:
@asraa  and @venafi-iw  could you please clarify what the TODO precisely means ;) Is it perm locking down  in the sense of doing something equivalent to `chattr +i <key_file>` ?

Issue #1406, opened 2022-02-06
verify-blob not finding tlog on rekor, but it is there


caarlos0:
**Description**

I might have found a bug, not sure if here, rekor, or elsewhere... or maybe my understanding of things is a bit wrong... in any case, here it goes:

First thing I noticed is that `rekor-cli` does not find any entries for some files:

```sh
$ rekor-cli search --artifact https://github.com/goreleaser/goreleaser/releases/download/v1.4.1/checksums.txt
no matching entries found

$ rekor-cli search --artifact https://github.com/goreleaser/goreleaser/releases/download/v1.4.0/checksums.txt
no matching entries found

$ rekor-cli search --artifact https://github.com/goreleaser/goreleaser/releases/download/v1.3.0/checksums.txt
no matching entries found

$ rekor-cli search --artifact https://github.com/goreleaser/nfpm/releases/download/v2.12.1/checksums.txt
Found matching entries (listed by UUID):
686fb1ee555cbc47416a99b8e8797cfe07b45946ed03df0908bcd8e4a307c78d

$ rekor-cli search --artifact https://github.com/goreleaser/nfpm/releases/download/v2.11.0/checksums.txt
no matching entries found

$ rekor-cli search --artifact https://github.com/goreleaser/nfpm/releases/download/v2.11.3/checksums.txt
no matching entries found
```

Here, it seems like after some time they are not available anymore, or at least, not searchable? 

So, if I try to verify without the PEM, it doesn't work:

```sh
$ COSIGN_EXPERIMENTAL=1 cosign verify-blob \
        --signature https://github.com/goreleaser/goreleaser/releases/download/v1.4.1/checksums.txt.sig \
       https://github.com/goreleaser/goreleaser/releases/download/v1.4.1/checksums.txt
Error: verifying blob [https://github.com/goreleaser/goreleaser/releases/download/v1.4.1/checksums.txt]: could not find a tlog entry for provided blob
main.go:46: error during command execution: verifying blob [https://github.com/goreleaser/goreleaser/releases/download/v1.4.1/checksums.txt]: could not find a tlog entry for provided blob
```

But, if I verify it **with** the PEM as well, it prints an UUID:

```sh
$ COSIGN_EXPERIMENTAL=1 cosign verify-blob \
        --cert https://github.com/goreleaser/goreleaser/releases/download/v1.4.1/checksums.txt.pem \
[...]

PR #2058, opened 2022-07-08
fix: rekor get tlog entry with uuid

asraa:

Sign blob w/ search otherwise doesn't work, because the UUID in the response of Rekor now includes a prefix shard UUID.

cc @priyawadhwa 

#### Documentation

Fixes https://github.com/sigstore/cosign/issues/1406

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/go-openapi/strfmt"
@@ -67,6 +68,10 @@ const (
 	// state ever changes, it will make lots of noise.
 	addRekorPublicKeyFromRekor = "SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY"
 )
+
+const treeIDHexStringLen = 16
+const uuidHexStringLen = 64
+const entryIDHexStringLen = treeIDHexStringLen + uuidHexStringLen
 
 // getLogID generates a SHA256 hash of a DER-encoded public key.
 func getLogID(pub crypto.PublicKey) (string, error) {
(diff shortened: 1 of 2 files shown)


priyawadhwa (approved):
nice!!

dlorenc:
Codegen or a rebase issue?

priyawadhwa:
I think an issue with compiling cosign, we might need to update to 1.18:

```
Error: ../../../go/pkg/mod/sigs.k8s.io/release-utils@v0.7.1/version/version.go:124:25: bi.Settings undefined (type *debug.BuildInfo has no field or method Settings)
note: module requires Go 1.18
```

asraa:
> I think an issue with compiling cosign, we might need to update to 1.18:

yeah, i'm seeing some problems when i run codegen locally as well. is this because rekor is at 1.18?

priyawadhwa:
I think that `sigs.k8s.io/release-util` dependency requires 1.18. looks like rekor recently pinned to a specific version of it and updated the go version at the same time -- https://github.com/sigstore/rekor/pull/904

asraa:
Confirmed this fixes goreleaser:
```
$ COSIGN_EXPERIMENTAL=1 ./cosign verify-blob   --signature https://github.com/goreleaser/goreleaser/releases/download/v1.10.3/checksums.txt.sig   https://github.com/goreleaser/goreleaser/releases/download/v1.10.3/checksums.txt
tlog entry verified with uuid: 6ad8e71d61e4757a7010ba3d31fca287db1e1e3ecdc09ebaeab317b7974f2581 index: 3045145
Verified OK
```

asraa:
@priyawadhwa @dlorenc @hectorj2f sorry this got lost after the golang bump! but this will fix the recent goreleaser issue so I am reviving it.

asraa:
Once again I have some dep problems... :/

caarlos0 (dismissed):

asraa:
@imjasonh I think I still require go 1.18 https://github.com/sigstore/cosign/pull/2093

dlorenc:
You should be able to rebase now!

asraa:
Updated, thank you!

Note the comment in the dep check: If needed, I'll see if I can split up the sharding package's sharding computation from the stuff that requires Trillian clients.

imjasonh (dismissed):

dlorenc (approved):

dlorenc (changes requested):
Actually we need to hold on the trillian dep check I think, cc @vaikas and @mattmoor

asraa:
> Actually we need to hold on the trillian dep check I think, cc @vaikas and @mattmoor

Yeah, if this can't be WIP for a little time, I can copy/paste the sharding code with a TODO to remove the copy-paste. LMK

vaikas:
I'd personally prefer to not pull that in as it has caused (well, not Trillian alone, but things with glog and other huge deps. My preference would be (I can't believe I'm saying this ;) ) cut&paste.

When you say it would be wip for little time, what's the actual fix that we would have to do?

But from this PR I don't quite get why we need to comment out the trillian bit, this doesn't seem to bring in new deps added, so I don't understand why we need to remove it.
My second question is just curiousity about leaking the the implementation details (but we don't have to discuss it here) about the trillian that sits behind rekor.

vaikas:
Also, I know folks are sensitive to deps and sizes (I know when I inadvertently pulled in something that had a k8s dep :) )

asraa:
> Or, as @vaikas says, copy and paste sharding.GetUUIDFromIDString into cosign.

Yeah exactly. I could either do this, since it's not complicated logic, but my main idea was to separate these sharding utils from the sharding code that manipulates and requires the trillian log client.

I'll resolve it internally. Before GA we will expose verification libraries with few deps in rekor anyway. And that will supersede all this code.

asraa:
All set, PTAL. Added tests for all the cases of whether Rekor will return EntryUUIDs or UUIDs to make sure matching occurs correctly.

dlorenc (approved):

Issue #1381, opened 2022-01-31
Clarifying questions re: cosign vulnerability attestation spec


luhring:
Hi! I saw @developer-guy open https://github.com/anchore/grype/issues/614. I think vulnerability scan attestations are a great idea, and I've been catching up on what exists already via these places so far:

- https://github.com/sigstore/cosign/blob/main/specs/COSIGN_VULN_ATTESTATION_SPEC.md
- https://github.com/sigstore/cosign/issues/442
- https://github.com/in-toto/attestation/issues/58

Before starting to apply this spec in a real implementation, I have a few questions that would help me better understanding how to use it.

(ALSO: I should apologize for _just now_ asking these questions and not weighing in earlier in the issue and PR discussions. Sorry about that! 🙇 )

1. The `invocation` object is shown in the spec but its fields aren't documented below like the other parts of the spec. I don't understand the role that `invocation` is intended to play in vuln scanning workflows. I believe @dlorenc [suggested](https://github.com/in-toto/attestation/issues/58#issuecomment-896218417) that we drop this object. Can it be removed from the spec? And if not, could we document the intent of this data in the spec?

2. I was following the conversation [here](https://github.com/in-toto/attestation/issues/58#issuecomment-904600218), and it looks like there's an intention to use the statement's `predicateType` field to specify not only this Cosign spec but also the type of the "wrapped" data that goes into the `result` object. Is that true? If so, could the spec example be updated to show how this relationship should work? (i.e. can we include a bit more of the in-toto statement than just the predicate itself?)

3. It looks like this specification features Trivy consistently throughout the document. (And it looks like the example attestation isn't using a wrapped _standard_, it's using Trivy's own JSON format.) Is this intentional?

4. Is `result` the right name for a field that wraps an entire scan output from another spec? I'm thinking this term might be overloaded. Those specs tend to include data that's a **superset** of "results", and then `result` tends to be an actual field _within_ the specification (i.e. to specify the vulnerability _matches_).

5. The data contained in `scanner` seems to always be included in other specifications (e.g. see [SARIF](https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json#L2897-L2923), [CycloneDX](https://github.com/CycloneDX/specification/blob/1.4/schema/bom-1.4.schema.json#L166-L202)). Is this redundant?

[...]

PR #2069, opened 2022-07-13
feat(spec): remove invocate in vuln predicate

masahiro331:
Ref: #1381 

#### Summar

Many vulnerability scanners are unaware of the type of environment in which they are used.  

The consensus in some issues for this spec seemed to be to remove invocation.

* https://github.com/in-toto/attestation/issues/58#issuecomment-896228987
* https://github.com/sigstore/cosign/issues/1381#issuecomment-1059992973

#### Release Note

* Remove invocation in vuln predicate.

#### Documentation

Need to fix the sample YAML.
https://docs.sigstore.dev/policy-controller/overview#configuring-policy-at-the-clusterimagepolicy-level

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -47,9 +47,8 @@ type CosignPredicate struct {
 
 // VulnPredicate specifies the format of the Vulnerability Scan Predicate
 type CosignVulnPredicate struct {
-	Invocation Invocation `json:"invocation"`
-	Scanner    Scanner    `json:"scanner"`
-	Metadata   Metadata   `json:"metadata"`
+	Scanner  Scanner  `json:"scanner"`
+	Metadata Metadata `json:"metadata"`
 }
 
 // I think this will be moving to upstream in-toto in the fullness of time
@@ -65,13 +64,6 @@ type CosignVulnStatement struct {
 type CycloneDXStatement struct {
 	in_toto.StatementHeader
 	Predicate interface{} `json:"predicate"`
-}
-
-type Invocation struct {
-	Parameters interface{} `json:"parameters"`
-	URI        string      `json:"uri"`
-	EventID    string      `json:"event_id"`
-	BuilderID  string      `json:"builder.id"`
 }
 
 type DB struct {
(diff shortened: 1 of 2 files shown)


masahiro331:
@jdolitsky @dlorenc 

Apologies.
Direct PR submitted in response to specification amendments.

I wish to reopen and finalize the discussion against this specification.

dlorenc:
> @jdolitsky @dlorenc
> 
> My apologies to direct PR.
> 
> I wish to reopen and finalize the discussion against this specification.

Are you hoping to merge this?

masahiro331:
Yes!!

I'm a contributor to Trivy.
It is difficult to meet these specifications when supporting the intoto_attestation format as the output of Trivy's vulnerability detection results.

If supported, it must take environment information as an argument.

As another discussion, there is a question as to which one should be set if the job that runs Trivy is different from the job that runs cosign.

masahiro331:
@dlorenc 

Yes, I hope.

I'm a contributor to Trivy.
It is difficult to meet these specifications when supporting the intoto_attestation format as the output of Trivy's vulnerability detection results.

If supported, it must take environment information as an argument.

As another discussion, there is a question as to which one should be set if the job that runs Trivy is different from the job that runs cosign.

masahiro331:
Trivy's issue.
https://github.com/aquasecurity/trivy/issues/1646

masahiro331:
@dlorenc 

Discussion on Invocation had stopped and was reopened in the PR.

If possible, we would like to check if it is optional as well as scanner.db.
Currently, there is no specification for invocation, so we don't know if it is required.

dlorenc:
Cc @puerco can you double check this one?

puerco:
I don't want to revive what appears to be done discussion just because of it, but I actually think having the parameters of the scan captured in the `invocation` is useful. They may be optional if needed but their usefulness is there.

I feel there is value in capturing the parameters and environment in a standard way. While the scanning service can expose its inner state in its own output, I think we should capture those flags from outside of the scanner because of a) trust and b) because we cannot guarantee all scanner actually embed the data in their output.

Take for example these two examples. Running Trivy with and without `--severity CRITICAL` to scan the current alpine image gives me a total of

With `--severity CRITICAL`: 0 vulnerabilities
Without that same flag: 4  vulnerabilities

Yet, the output of the scan does not capture the runtime configuration, and even if it did, I think it is easier for a system replaying the attestation to read the invocation params and environment from a standard location.

PR #2070, opened 2022-07-13
Remove replace directives in go.mod.

wlynch:
#### Summary

As @cpanato pointed out in https://github.com/sigstore/cosign/pull/2054#issuecomment-1177410669, the replace directive introduced in #2054 breaks `go install` when running outside of the
cosign directory. 

```sh
$ go install github.com/sigstore/cosign/cmd/cosign@df94451e5581d1354af086f7d1faed9f69b826cf
go: downloading github.com/sigstore/cosign v1.9.1-0.20220707161345-df94451e5581
go: github.com/sigstore/cosign/cmd/cosign@df94451e5581d1354af086f7d1faed9f69b826cf (in github.com/sigstore/cosign@v1.9.1-0.20220707161345-df94451e5581):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

This swaps the `replaces` with a `go get
go.etcd.io/etcd/v3@v3.6.0-alpha.0`, which should generally have the same
effect.

Signed-off-by: Billy Lynch <billy@chainguard.dev>

#### Release Note

NONE

#### Documentation

n/a


wlynch:
I can't assign people, but cc @cpanato @santiagotorres

wlynch:
Had some discussion on https://github.com/sigstore/policy-controller/pull/80 - cosign's usage of the alpha API might be problematic for other sigstore projects that want to rely on cosign for verification behavior but do not want to adopt an alpha API. This may make it difficult for downstream projects update past the current release. 😢

We're a bit stuck here because the otel packages made breaking changes from v0 to v1. OPA > v0.35 requires the v1 packages, but etcd <= v3.5.4 (e.g. non-alpha) needs v0.

Any ideas for how we might disentangle these deps? Perhaps we should see if we can break apart the github.com/google/certificate-transparency-go libraries so we can use the inclusion proof checking without the etcd dependency from trillian?

hectorj2f:
My 2 cents here: if the overall quality of this alpha version is sufficiently high and any breaking change is not affecting the imported functionalities, then it should be fine.

dlorenc:
I don't care about using alpha stuff, this seems good to me.

dlorenc (approved):

Issue #1660, opened 2022-03-25
Refactor cosign sign command logic into pkg


marcofranssen:
**Description**

A lot of the signing logic is currently in the `cmd` package.

https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign.go

It would be great if more of this logic can be moved into the `pkg` package so other projects can reuse it in their tools.

A good example of a command that has a more clean setup between `cmd` and `pkg` is, https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/upload/blob.go

noamichael:
This would be awesome and is the same reason I opened #1665. `go-containerregistry` has a similar structure. For example, [you can import the logic of the copy command directly](https://github.com/google/go-containerregistry/blob/efc62d802606b3e7668eafe55810b6359b352c78/pkg/crane/copy.go#L28) without any knowledge of the cli package.

znewman01:
👍 There are a lot of testing benefits to this as well

marcofranssen:
> 👍 There are a lot of testing benefits to this as well

Would indeed be great if the `os.Stdout` and `os.Stderr` in `pkg` could be replaced with an injected `io.Writer` for testing purposes.

haydentherapper:
The goal of the [sigstore-go](https://github.com/sigstore/sigstore-go) library will be a clean-room implementation of signing and verification. We welcome any small refactors in Cosign, but won't take on large scale changes at this time.

PR #2074, opened 2022-07-14
move sign pkg to pkg/sign

sallyom:
#### Summary
https://github.com/sigstore/cosign/issues/1660 

Currently the signing logic lives in `cmd/cosign/cli/sign`, this PR moves signing logic to `pkg/sign`.
Moving to pkg/sign allows other projects to include in their tools.

#### Release Note
NONE

#### Documentation
NONE


haydentherapper:
Before moving over all of these functions into `pkg`, we will need to review what's public as not all functions should be publicly available, add documentation, etc.

Issue #1462, opened 2022-02-15
Sanitize cosign's dependency tree


imjasonh:
### Problem

Cosign's module dependencies are pretty heavy.

Its [go.mod](https://github.com/sigstore/cosign/blob/f005e25466c3b6954546b718ce3a56d0efd6ec2b/go.mod) currently transitively depends on k8s.io/client-go, the AWS SDK, the Azure SDK, a GCP cloud storage API client, Docker internals, fsnotify, OpenTelemetry, a Kubernetes controller framework, a GitHub API client, cuelang, and many more.

This can make it difficult to consume for clients that don't want all that functionality, and may just want to sign some bytes using a key, push the resulting image to an OCI registry, and upload data to Rekor with certs via Fulcio.

For example, `sget`, which does relatively little compared to `cosign` -- it verifies signatures on an image and fetches a blob -- has a [go.mod file that's 280 lines](https://github.com/sigstore/sget/blob/b2bf5a609ad32c8f8aa1cfe5995681932ce17f2b/go.mod), producing a 72 MB binary 🏋️ .

Large dependency trees lead to larger built binaries (see above), larger attack surfaces, longer and more fragile builds, and (most interesting to me personally as a maintainer!) more toil for maintainers to update dependencies and sift through security alerts for those vulnerabilities.

### Possible Solutions

- Moving more dependencies into sigstore/sigstore (https://github.com/sigstore/sigstore/issues/81), and taking that opportunity to reconsider if any dependencies can be replaced by lighter-weight alternatives
- Taking better advantage of Go interfaces, especially for KMS services, and moving implementations of those interfaces into separate Go modules or separate repositories
- Splitting the cosigned webhook into a separate Go module from the `cosign` CLI, possibly even moving it to a separate repo

The `cosign` CLI would likely not get any smaller as a result of this work -- it still naturally depends on all the KMSes, K8s, cloud SDKs -- but other clients that want to do `cosign`-like things (signing, and especially verifying) wouldn't need to.

cc @luhring @jonjohnsonjr

dlorenc:
> * Splitting the cosigned webhook into a separate Go module from the `cosign` CLI, possibly even moving it to a separate repo

I don't think this will really help, but I might be misunderstanding the exact issue. The k8s libraries shouldn't show up in a downstream binary or dependency tree unless they're actually used.

imjasonh:
[...]

PR #2080, opened 2022-07-20
Remove hack/tools.go

imjasonh:
This was a relic of the time when this repo contained the cosigned
webhook (now policy-controller in its own repo).

These deps weren't included in the cosign CLI, and can just be dropped
from the repo.


imjasonh:
cc #1462

dlorenc (approved):

PR #2085, opened 2022-07-20
feat: attach: attestation: allow passing multiple payloads

### LINKED ISSUE #2052: `attach attestation` should support Attestation Bundles
**Description**

Users may have multiple attestations that they want to attach to a given image.  `cosign download attestation` supports downloading all the attestations for a given image, but `cosign attach attestation` doesn't support uploading multiple attestations.

It would be nice if `attach attestation` supported [in-toto Attestation Bundles](https://github.com/in-toto/attestation/blob/main/spec/bundle.md)

This could probably be added fairly easily with [some small modifications here](https://github.com/sigstore/cosign/blob/dbd430fce5d4b3805758bd97ec5c104b164091c8/cmd/cosign/cli/attach/attestation.go#L39).

Dentrax:

Fixes #2052 

#### Release Note

Allow passing multiple attestation payloads to "attach attestation" command


Dentrax:
```
$ ./cosign attach attestation --attestation at1.json --attestation at2.json furkanturkal/busybox:0.1.0

Using payload from: at1.json
Using payload from: at2.json
```

```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 342,
    "digest": "sha256:9641fd9d3b1b7e18ab7ed6b7fe6c232e4969043d188902aa192843cb9c3eec84"
  },
  "layers": [
    {
      "mediaType": "application/vnd.dsse.envelope.v1+json",
      "size": 4939,
      "digest": "sha256:0205d4c0d9d12706c126515b1ef452e3c7460de647aeec33d1ed01bd172e98de",
      "annotations": {
        "dev.cosignproject.cosign/signature": ""
      }
    },
    {
      "mediaType": "application/vnd.dsse.envelope.v1+json",
      "size": 5999,
      "digest": "sha256:d2a9efb2c1621823ecb3dc270a74d5ded58be7219080875be0349b625bae7206",
      "annotations": {
        "dev.cosignproject.cosign/signature": ""
      }
    }
  ]
}
```

TomHennen:
Thanks for working on this!

Looking at the example and code it looks like this doesn't follow the intent of #2052 which wants the attach command to support reading multiple attestations from a _single file_  in the [in-toto bundle](https://github.com/in-toto/attestation/blob/main/spec/bundle.md) format.

The expected behavior is that a single file would have multiple attestations contained within it (one per line) and that attach should read and attach all of them.

e.g.
```jsonl
{ "payloadType": "application/vnd.in-toto+json", "payload": "a...", "signatures": [w...] }
{ "payloadType": "application/vnd.in-toto+json", "payload": "b...", "signatures": [w...] }
```

The behavior here, letting the user specify multiple files, does seem useful though.  I wouldn't be surprised if some users would like this capability, it's just not quite what I was looking for in #2052.

WDYT?

developer-guy:
Oh, thanks, @TomHennen, we couldn't understand the issue at first, but now we are okay with the idea and will fix it soon, but @dlorenc, what do you think about this PR? It seems useful for people who might want to add multiple attestation files.

dlorenc:
Seems fine. When we support bundles, would we just detect them in the same file/flag or would that be different?

developer-guy:
> Seems fine. When we support bundles, would we just detect them in the same file/flag or would that be different?

In this PR, UX will remain the same, just add knowledge about the `jsonlines` file for attestation bundles.

cmd/cosign/cli/attach.go
```diff
@@ -84,12 +84,16 @@ func attachAttestation() *cobra.Command {
 	o := &options.AttachAttestationOptions{}
 
 	cmd := &cobra.Command{
-		Use:     "attestation",
-		Short:   "Attach attestation to the supplied container image",
-		Example: "  cosign attach attestation <image uri>",
-		Args:    cobra.ExactArgs(1),
+		Use:   "attestation",
+		Short: "Attach attestation to the supplied container image",
+		Example: `  cosign attach attestation --attestation <payload path> <image uri>
```
TomHennen, line 89: suggestion: instead of \<payload path\>, \<attestation file path\>

cmd/cosign/cli/attach.go
```diff
@@ -84,12 +84,16 @@ func attachAttestation() *cobra.Command {
...
+
+  # attach multiple attestations to a container image
```
TomHennen, line 91: suggestion: 

"attach attestations from multiple files to a container image"

cmd/cosign/cli/attach.go
```diff
@@ -84,12 +84,16 @@ func attachAttestation() *cobra.Command {
 	o := &options.AttachAttestationOptions{}
 
 	cmd := &cobra.Command{
-		Use:     "attestation",
-		Short:   "Attach attestation to the supplied container image",
-		Example: "  cosign attach attestation <image uri>",
-		Args:    cobra.ExactArgs(1),
+		Use:   "attestation",
+		Short: "Attach attestation to the supplied container image",
```
TomHennen, line 88: Is there a place to add a note that "attach attestation supports files using the in-toto bundle format" or something like that?

cmd/cosign/cli/attach/attach.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func attachAttestation(remoteOpts []ociremote.Option, signedPayload, imageRef string) error {
+	fmt.Fprintf(os.Stderr, "Using payload from: %s", signedPayload)
+	attestation, err := os.Open(signedPayload)
+	if err != nil {
+		return err
+	}
+
+	env := ssldsse.Envelope{}
```
TomHennen, line 54: I'm not super familiar with go, but should this line go within the for loop before?

cmd/cosign/cli/attach/attach.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	ssldsse "github.com/secure-systems-lab/go-securesystemslib/dsse"
+	"github.com/sigstore/cosign/cmd/cosign/cli/options"
+	"github.com/sigstore/cosign/pkg/oci/mutate"
+	ociremote "github.com/sigstore/cosign/pkg/oci/remote"
+	"github.com/sigstore/cosign/pkg/oci/static"
+	"github.com/sigstore/cosign/pkg/types"
+)
+
+func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, signedPayloads []string, imageRef string) error {
```
TomHennen, line 32: nit, instead of 'signedPayloads', should this be 'filePaths' ?

cmd/cosign/cli/attach/attach.go
```diff
@@ -0,0 +1,110 @@
...
+			return fmt.Errorf("attaching payload from %s: %w", payload, err)
+		}
+	}
+
+	return nil
+}
+
+func attachAttestation(remoteOpts []ociremote.Option, signedPayload, imageRef string) error {
+	fmt.Fprintf(os.Stderr, "Using payload from: %s", signedPayload)
+	attestation, err := os.Open(signedPayload)
```
TomHennen, line 49: nit: instead of 'attestation' should this be 'file' or 'attestationFile'

cmd/cosign/cli/attach/attach.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, payload := range signedPayloads {
+		if err := attachAttestation(ociremoteOpts, payload, imageRef); err != nil {
+			return fmt.Errorf("attaching payload from %s: %w", payload, err)
+		}
+	}
+
+	return nil
+}
+
+func attachAttestation(remoteOpts []ociremote.Option, signedPayload, imageRef string) error {
```
TomHennen, line 47: nit: instead of 'signedPayload' should this be 'filePath' ?

TomHennen (changes requested):
Thanks, this seems to work well! :)

cmd/cosign/cli/attach/attach.go
developer-guy: thank you, updated.

cmd/cosign/cli/attach/attach.go
developer-guy: thank you, updated.

cmd/cosign/cli/attach/attach.go
developer-guy: thank you, updated.

cmd/cosign/cli/attach/attach.go
developer-guy: thank you, updated.

cmd/cosign/cli/attach.go
developer-guy: thanks

cmd/cosign/cli/attach.go
developer-guy: thanks

cmd/cosign/cli/attach.go
developer-guy: Added

dlorenc (approved):

PR #2089, opened 2022-07-22
add flag to allow skipping upload to transparency log

k4leung4:
#### Summary

add flag to allow skipping upload to transparency log

we use cosign extensively with a private registry, and would like to
streamline the signing process to bypass the prompt to skip uploading to
a the transparency log with a private registry.

Signed-off-by: Kenny Leung <kleung@chainguard.dev>

#### Release Note

add flag to allow skipping upload to transparency log

usage `cosign sign --key cosign.key --tlog_upload=false <IMAGE>` to skip uploading to transparency log

#### Documentation

Yes, users can explicit decline upload to transparency log by passing the flag `--tlog_upload=false` flag. the `--force` flag takes priority if both are specified.

### DIFF
--- a/cmd/cosign/cli/options/attest.go
+++ b/cmd/cosign/cli/options/attest.go
@@ -77,6 +77,6 @@ func (o *AttestOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVarP(&o.SkipConfirmation, "yes", "y", false,
 		"skip confirmation prompts for non-destructive operations")
 
-	cmd.Flags().BoolVar(&o.NoTlogUpload, "no_tlog_upload", false,
+	cmd.Flags().BoolVar(&o.NoTlogUpload, "no-tlog-upload", false,
 		"whether to not upload the transparency log")
 }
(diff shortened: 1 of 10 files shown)


cmd/cosign/cli/options/attest.go
```diff
@@ -75,4 +76,7 @@ func (o *AttestOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVarP(&o.SkipConfirmation, "yes", "y", false,
 		"skip confirmation prompts for non-destructive operations")
+
+	cmd.Flags().BoolVar(&o.TlogUpload, "tlog_upload", true,
```
cpanato, line 80: i think it should match the other flags with composed name

for example the `--no-upload` then should be `--tlog-upload`

k4leung4: thanks for the comment.
renamed to `--no-tlog-upload`, as the `--upload` flag refers to uploading signatures and not tlog.

cpanato: thank you!

cpanato (approved):
lgtm

PR #2092, opened 2022-07-22
Remove knative/pkg deps

imjasonh:
- we were depending on knative/pkg for logging, now we just `log.Printf`
- we were depending on [depcheck](https://pkg.go.dev/knative.dev/pkg/depcheck) from knative, now depending on a [fork in a separate repo](https://github.com/depcheck-test/depcheck-test) 
- we were depending on knative/pkg for worker goroutine pooling, now just using [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) and limiting workers to to NumCPU ourselves

Dropping knative deps should drop a bunch of transitive deps as well.

#### Summary

Reducing unnecessary dependencies limits our exposure to vulnerabilities in those dependencies.

This change has no effect on binary size (`cosign` is still 83 MB 🏋️ )

#### Release Note

NONE

#### Documentation

NONE

cc @mattmoor

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -421,8 +421,6 @@ github.com/blakesmith/ar v0.0.0-20190502131153-809d4375e1fb/go.mod h1:PkYb9DJNAw
 github.com/blang/semver v3.1.0+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
 github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
 github.com/blang/semver v3.5.1+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
-github.com/blendle/zapdriver v1.3.1 h1:C3dydBOWYRiOk+B8X9IVZ5IOe+7cl+tGOexN4QqHfpE=
-github.com/blendle/zapdriver v1.3.1/go.mod h1:mdXfREi6u5MArG4j9fewC+FGnXaBR+T4Ox4J2u4eHCc=
 github.com/blizzy78/varnamelen v0.3.0/go.mod h1:hbwRdBvoBqxk34XyQ6HA0UH3G0/1TKuv5AC4eaBT0Ec=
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -672,6 +670,8 @@ github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSs
 github.com/denis-tingajkin/go-header v0.4.2/go.mod h1:eLRHAVXzE5atsKAnNRDB90WHCFFnBUn4RN0nRcs1LJA=
 github.com/denisenkom/go-mssqldb v0.11.0/go.mod h1:xbL0rPBG9cCiLr28tMa8zpbdarY27NDyej4t/EjAShU=
 github.com/denverdino/aliyungo v0.0.0-20190125010748-a747050bb1ba/go.mod h1:dV8lFg6daOBZbT6/BDGIz6Y3WFGn8juu6G+CQ6LHtl0=
+github.com/depcheck-test/depcheck-test v0.0.0-20220607135614-199033aaa936 h1:foGzavPWwtoyBvjWyKJYDYsyzy+23iBV7NKTwdk+LRY=
```
mattmoor, line 673: Should we cut a release of this, so that dependabot does its thing?

pkg/cosign/fetch.go
```diff
@@ -78,11 +78,12 @@ func FetchSignaturesForReference(ctx context.Context, ref name.Reference, opts .
 		return nil, fmt.Errorf("no signatures associated with %v: %w", ref, err)
 	}
 
-	g := pool.New(runtime.NumCPU())
 	signatures := make([]SignedPayload, len(l))
+	var g errgroup.Group
+	g.SetLimit(runtime.NumCPU())
```
mattmoor, line 83: Can you file an issue on PKG to deprecate `pool` in favor of this, and possibly move to this model for its implementation?

mattmoor (approved):

pkg/cosign/fetch.go
imjasonh: https://github.com/knative/pkg/issues/2556

go.sum
imjasonh: Sure. I don't anticipate a lot of updates to it, but I guess it'd be nice to have those be officially released (unlike knative/pkg 😆)

I won't block this PR on it though, whenever depcheck-test gets a release, dependabot will pick it up.

Issue #2091, opened 2022-07-22
x509: cannot verify signature: insecure algorithm SHA1-RSA


imjasonh:
**Description**

From head (95b74db89941e8ec85e768f639efd4d948db06cd)

```
$ go test ./pkg/cosign/
--- FAIL: TestValidateAndUnpackCertWithSCT (0.00s)
    verify_test.go:457: ValidateAndUnpackCert expected no error, got err = x509: certificate signed by unknown authority (possibly because of "x509: cannot verify signature: insecure algorithm SHA1-RSA (temporarily override with GODEBUG=x509sha1=1)" while trying to verify candidate authority certificate "Certificate Transparency CA")
    verify_test.go:464: ValidateAndUnpackCert expected no error, got err = x509: certificate signed by unknown authority (possibly because of "x509: cannot verify signature: insecure algorithm SHA1-RSA (temporarily override with GODEBUG=x509sha1=1)" while trying to verify candidate authority certificate "Certificate Transparency CA")
FAIL
FAIL	github.com/sigstore/cosign/pkg/cosign	1.422s
FAIL
```

Adding the `GODEBUG` as suggested makes the test pass:

```
GODEBUG=x509sha1=1 go test ./pkg/cosign/
ok  	github.com/sigstore/cosign/pkg/cosign	1.367s
```

```
$ go version
go version go1.18.2 darwin/arm64
```

Mostly filing this since I couldn't find any other reference to this in issues or code in any sigstore repo. Has anybody seen this before

edit: see https://github.com/golang/go/issues/41682

haydentherapper:
Looking into this. Interestingly it fails on 1.18, but not 1.19.

haydentherapper:
Need to generate new test data for certificate-transparency-go that doesn't use SHA1 for the signing algorithm digest.

dmitris:
`go test -v ./pkg/cosign` passes now without `GODEBUG=x509sha1=1` on the head version (17cc13812d8a7bfbb068ffc10fdbbdb3a9416f11) with go1.20.3 - has anything changed? @imjasonh - does it work for you now or not? `go test ./...` works for me as well.

imjasonh:
Tests also passed for me at head (17cc13812d8a7bfbb068ffc10fdbbdb3a9416f11) without the env var set. I assume @haydentherapper updated the test data to not rely on SHA1, and the env var is now unnecessary.

Thanks for trying this @dmitris ! 👍

[...]

PR #2093, opened 2022-07-22
Run tests using Go 1.18

imjasonh:
#### Summary

Go 1.19 will be out next month, which means Go 1.17 will be out of support.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,7 +23,7 @@ on:
 permissions: read-all
 
 env:
-  GO_VERSION: 1.17
+  GO_VERSION: 1.18
 
 jobs:
   unit-tests:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,12 +60,18 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           check-latest: true
       - name: Run Go tests
+        env:
+          # See #2091 for the issue describing this temp workaround.
+          GODEBUG: x509sha1=1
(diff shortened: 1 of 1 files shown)


cpanato (dismissed):

imjasonh:
This hits https://github.com/sigstore/cosign/issues/2091 😢

cpanato:
this is something we can enable only for the tests or does that affects the binary when users use this feature?

imjasonh:
> this is something we can enable only for the tests or does that affects the binary when users use this feature?

I think we could `GODEBUG=x509sha1=1` for tests, and rely on e2e tests to tell us whether this causes any issues in the actual CLI usage.

Just building cosign from HEAD with Go 1.18 and testing locally, I'm able to `COSIGN_EXPERIMENTAL=1 cosign sign && cosign verify` an image, so at least that much seems to work.

asraa:
quick ping on this if possible? it's blocking upgrading rekor so we can handle some incompatible rekor changes that prevent blob verification https://github.com/sigstore/cosign/pull/2058

imjasonh:
My understanding is we can either update our usage of the deprecated SHA-1 cert (https://github.com/sigstore/cosign/issues/2091#issuecomment-1203326239) or temporarily set the env var to ignore it during tests.

If you're blocked I'd say set the env var in tests and upgrade to Go 1.18+ to get unblocked.

dlorenc:
Added the suppression, this can unblock a few pending PRs.

dlorenc (dismissed):

dlorenc (approved):

Issue #2095, opened 2022-07-23
vulnerability attestation: change Result to interface{}


knqyf263:
Is there any specific reason `Result` is `map[string]interface{}`?
https://github.com/sigstore/cosign/blob/612826982d3695cc85f7f4e42c8c26cc1218d1a5/pkg/cosign/attestation/attestation.go#L77

We are trying to add support for [Cosign Vulnerability Scan Record Attestation](https://github.com/sigstore/cosign/blob/main/specs/COSIGN_VULN_ATTESTATION_SPEC.md) in Trivy, but our result is not `map[string]interface{}` and it makes implementation harder. Can we change it to `interface{}`?

PR #2096, opened 2022-07-23
Change Result in Vulnerability Attestation to interface{}

knqyf263:
#### Summary
Please feel free to close the PR if there is any requirement on `Result` in the Vulnerability Scan Record Attestation.

Close https://github.com/sigstore/cosign/issues/2095

#### Release Note
None

#### Documentation
None

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -71,10 +71,10 @@ type DB struct {
 }
 
 type Scanner struct {
-	URI     string                 `json:"uri"`
-	Version string                 `json:"version"`
-	DB      DB                     `json:"db"`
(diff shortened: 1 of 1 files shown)


dlorenc (approved):

dlorenc:
I think this is fine, do we need to change the docs anywhere?

knqyf263:
It is mentioned as `object`. I think it is fine. Do you want to change it to `any` or something like that?
https://github.com/sigstore/cosign/blob/95b74db89941e8ec85e768f639efd4d948db06cd/specs/COSIGN_VULN_ATTESTATION_SPEC.md?plain=1#L83

dlorenc (approved):

PR #2108, opened 2022-07-27
Resolves #522 set Created date to time of execution

### LINKED ISSUE #522: "Created" date for signature is always Dec 31, 0
When signing images and pushing them to Google Cloud Registry, the signature's "Created" date is always Dec 31, 0. 
I would have expected the "Created" date for the signature to be the date the container was signed. 

`printenv GCR_ACCOUNT | docker login -u _json_key --password-stdin gcr.io`
`cosign sign -key cosign.key $NAMESPACE:$version`

Lerentis:
#### Summary

This commit sets the Created date of the signature to the date of execution of the signing action.

#### Release Note

`-changed: Signatures do now use the created timestamp based on the local time of signing`

#### Documentation

`n/a`

### DIFF
--- a/pkg/oci/mutate/signatures.go
+++ b/pkg/oci/mutate/signatures.go
@@ -40,9 +40,12 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
 	}
 
 	img, err := mutate.Append(base, adds...)
+	if err != nil {
+		return nil, err
+	}
 
 	// Set the Created date to time of execution
-	img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
+	img, err = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/oci/mutate/signatures.go
+++ b/pkg/oci/mutate/signatures.go
@@ -16,7 +16,6 @@
 package mutate
 
 import (
-	"log"
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -42,13 +41,8 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
 
 	img, err := mutate.Append(base, adds...)
 
-	imgCfg, _ := img.ConfigFile()
-
 	// Set the Created date to time of execution
-	if imgCfg.Created.Time.IsZero() {
-		log.Println("Signed Entry Date was Zero, Setting to current time")
-		img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
-	}
+	img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
 
 	if err != nil {
 		return nil, err
--- a/pkg/oci/mutate/signatures.go
+++ b/pkg/oci/mutate/signatures.go
@@ -43,7 +43,6 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
 
 	// Set the Created date to time of execution
 	img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
-
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/oci/mutate/signatures.go
+++ b/pkg/oci/mutate/signatures.go
@@ -16,6 +16,7 @@
 package mutate
 
 import (
+	"log"
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -40,11 +41,18 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
 	}
 
 	img, err := mutate.Append(base, adds...)
+
+	imgCfg, _ := img.ConfigFile()
+
+	// Set the Created date to time of execution
+	if imgCfg.Created.Time.IsZero() {
+		log.Println("Signed Entry Date was Zero, Setting to current time")
+		img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
+	}
+
 	if err != nil {
 		return nil, err
 	}
-	// Set the Created date to time of execution
-	img, err = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
 
 	return &sigAppender{
 		Image: img,
--- a/pkg/oci/mutate/signatures.go
+++ b/pkg/oci/mutate/signatures.go
@@ -16,6 +16,8 @@
 package mutate
 
 import (
+	"time"
+
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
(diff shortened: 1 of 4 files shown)


pkg/oci/mutate/signatures.go
```diff
@@ -36,10 +38,20 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
...
+	time := v1.Time{
+		Time: time.Now(),
+	}
+
 	img, err := mutate.Append(base, adds...)
 	if err != nil {
 		return nil, err
 	}
+	// Set the Created date to time of execution
+	img, err = mutate.CreatedAt(img, time)
```
imjasonh, line 51:
```suggestion
	img, err := mutate.Append(base, adds...)
	if err != nil {
		return nil, err
	}
	// Set the Created date to time of execution
	img, err = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
```

imjasonh (commented):
This mostly lgtm, aside from the one tiny style nit.

A couple more questions though:
- Is there any kind of test we can add to check that the created date isn't zero?
- Do we also want this for attestations, SBOMs, etc?

Lerentis:
> * Is there any kind of test we can add to check that the created date isn't zero?

for a test like this one would need to mock a timestamp right? i am still learning when it comes to writing tests, so i fear this is out of my skill level as of now :/ 

> * Do we also want this for attestations, SBOMs, etc?

would make sense to keep this consistent imo, but maybe i am missing some greater context why it was implemented with this fixed date stamp before

imjasonh:
> > * Is there any kind of test we can add to check that the created date isn't zero?
> 
> for a test like this one would need to mock a timestamp right? i am still learning when it comes to writing tests, so i fear this is out of my skill level as of now :/

I don't know about everyone, but I'd personally accept a test that just checks that the created date is non-zero. Bonus points for it being any time in the last minute. Having a test that injects a fake timestamp and checks that it's exactly that date seems like overkill, but that's subjective.

> 
> > * Do we also want this for attestations, SBOMs, etc?
> 
> would make sense to keep this consistent imo, but maybe i am missing some greater context why it was implemented with this fixed date stamp before

Yeah I think we should stay consistent.

The reason it was zero before was probably some combination of (a) that was the default when using `empty.Image` and `mutate.Append`, and (b) some vague idea that the signatures should be reproducible, so that if two clients signed the same artifact with the same signature at different times they should produce the same layer bytes. But for (b) at least, that doesn't really make sense, since those two signatures would never be the same anyway.

Lerentis:
@imjasonh i hope i understood the requested changes correctly. initially i thought you were talking about a unit test and not a check about the previous date that was set 😅

pkg/oci/mutate/signatures.go
```diff
@@ -36,10 +39,21 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
...
 		})
 	}
+
 	img, err := mutate.Append(base, adds...)
+
+	imgCfg, _ := img.ConfigFile()
+
+	// Set the Created date to time of execution
+	if imgCfg.Created.Time.IsZero() {
+		log.Println("Signed Entry Date was Zero, Setting to current time")
```
imjasonh, line 49: I don't think we need to log this, this can just be the new expectation.

imjasonh:
> @imjasonh i hope i understood the requested changes correctly. initially i thought you were talking about a unit test and not a check about the previous date that was set 😅

Oh I meant we should have a unit test that when a signature/whatever gets appended, it also updates the image's created-at date to be something recent, or at least non-zero.

I think we should always update the image's created-at to be `Now()`, otherwise if I sign something today and wait 90 days and sign it again, the created-at date will be 90 days ago and Gitlab might housekeep it.

pkg/oci/mutate/signatures.go
Lerentis: removed this if statement completely with your clarification that this should be covered in a unit test and always be updated 👍

Lerentis:
ah okay then my initial understanding was correct. sorry about that.
added a simple zero unit test

pkg/oci/static/file_test.go
```diff
@@ -120,4 +120,14 @@ func TestNewFile(t *testing.T) {
...
 		}
 	})
+
+	t.Run("check date", func(t *testing.T) {
+		fileCfg, err := file.ConfigFile()
+		if err != nil {
+			t.Fatalf("FileCfg() = %v", err)
+		}
+		if fileCfg.Created.Time.IsZero() {
+			t.Errorf("Date of Signature was Zero")
```
imjasonh, line 130: Let's keep this error consistent with the other one, whatever we settle on.

imjasonh (commented):
This looks good!

I can't approve your CI run, but except for a few minor comments, this should be good to go assuming tests pass.

pkg/oci/mutate/signatures_test.go
```diff
@@ -70,4 +70,10 @@ func TestAppendSignatures(t *testing.T) {
 	} else if got, want := len(sl), 3; got != want {
 		t.Errorf("len(Get()) = %d, wanted %d", got, want)
 	}
+
+	if testCfg, err := threeSig.ConfigFile(); err != nil {
+		t.Fatalf("Get() = %v", err)
```
imjasonh, line 75:
```suggestion
		t.Fatalf("ConfigFile() = %v", err)
```

pkg/oci/static/file.go
```diff
@@ -41,6 +42,14 @@ func NewFile(payload []byte, opts ...Option) (oci.File, error) {
 	img, err := mutate.Append(base, mutate.Addendum{
 		Layer: layer,
 	})
+
```
imjasonh, line 45:
```suggestion
```

pkg/oci/static/file.go
```diff
@@ -41,6 +42,14 @@ func NewFile(payload []byte, opts ...Option) (oci.File, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Set the Created date to time of execution
+	img, err = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
+
```
imjasonh, line 52:
```suggestion
```

pkg/oci/mutate/signatures.go
```diff
@@ -36,10 +38,16 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
 			Annotations: ann,
 		})
 	}
+
 	img, err := mutate.Append(base, adds...)
+
+	// Set the Created date to time of execution
+	img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
+
```
imjasonh, line 46:
```suggestion
```

imjasonh (commented):
Only tiny tiny nits left, otherwise LGTM

pkg/oci/mutate/signatures.go
```diff
@@ -36,10 +38,15 @@ func AppendSignatures(base oci.Signatures, sigs ...oci.Signature) (oci.Signature
 			Annotations: ann,
 		})
 	}
+
 	img, err := mutate.Append(base, adds...)
+
+	// Set the Created date to time of execution
+	img, _ = mutate.CreatedAt(img, v1.Time{Time: time.Now()})
```
imjasonh, line 45: We should check both of these errs. With this, we'll return `mutate.Append`'s err, but only after calling `mutate.CreatedAt`.

imjasonh (approved):

dlorenc (approved):

PR #2114, opened 2022-07-29
Introduce a custom error type to classify errors.

mattmoor:
When checking signatures and evaluating policy, we may run into errors
for a variety of reasons unrelated to non-compliance, such as the registry
serving a 500.  This change introduces an error type that we can use to
produce and wrap errors in a way that we can distinguish between incidental
and verification errors in the calling logic.

After this change, we will be able to write:
```go
verr := &cosign.VerificationError{}
if errors.As(err, &verr) {
   // This is a verification error, handle accordingly.
} else {
  // This is something else, handle differently.
}
```

Related: https://github.com/sigstore/policy-controller/issues/109

### DIFF
--- a/pkg/cosign/errors.go
+++ b/pkg/cosign/errors.go
@@ -1,0 +1,50 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


pkg/cosign/errors.go
```diff
@@ -0,0 +1,50 @@
...
+import "fmt"
+
+var (
+	// ErrNoMatchingSignatures is the error returned when there are no matching
+	// signatures during verification.
+	ErrNoMatchingSignatures = &VerificationError{"no matching signatures"}
+
+	// ErrNoMatchingAttestations is the error returned when there are no
+	// matching attestations during verification.
+	ErrNoMatchingAttestations = &VerificationError{"no matching attestations"}
```
vaikas, line 26: This is a good pattern, then we can start adding things like if the issuer/subject or other things are not checking out.

vaikas (approved):

PR #2118, opened 2022-08-02
fix: adds envelope hash to in-toto entries in tlog entry creation

nkreiger:
Signed-off-by: Noah Kreiger <noahkreiger@gmail.com>

#### Summary

By not using the `types.NewProposedEntry` the `in-toto` rekor function was not assiging the hash key, which caused a downstream error when querying by the hash to find the log entry after it was added, making it appear as if it did not get added.

This fix will add the envelope hash to in-toto entries during tlog entry creation that will allow the entries to be queried by the hash downstream.

#### Release Note

* Bugfix adds envelope hash to in-toto entries in tlog entry creation


pkg/cosign/tlog.go
```diff
@@ -164,12 +164,15 @@ func TLogUpload(ctx context.Context, rekorClient *client.Rekor, signature, paylo
 
 // TLogUploadInTotoAttestation will upload and in-toto entry for the signature and public key to the transparency log.
 func TLogUploadInTotoAttestation(ctx context.Context, rekorClient *client.Rekor, signature, pemBytes []byte) (*models.LogEntryAnon, error) {
-	e := intotoEntry(signature, pemBytes)
-	returnVal := models.Intoto{
-		APIVersion: swag.String(e.APIVersion()),
-		Spec:       e.IntotoObj,
+	e, err := types.NewProposedEntry(context.Background(), "intoto", "0.0.1", types.ArtifactProperties{
```
hectorj2f, line 167: Wouldn’t you use the context from the input argument as it ? It sounds like a better option to me.

hectorj2f: Also I don’t feel hardcoding the intoto version as text is an improvement. Perhaps that is something to get from the api type.

asraa:
> By not using the types.NewProposedEntry the in-toto rekor function was not assiging the hash key, which caused a downstream error when querying by the hash to find the log entry

What do you mean by hash key? I'm curious if you have a reproducer or diff on the change. Before and after your change I ran 
```
COSIGN_EXPERIMENTAL=1 ./cosign attest --predicate pred.json --type slsaprovenance gcr.io/asra-ali/busybox/demo
```

and I was able to search both by payload hash. The entry contents themselves populate the same fields. (including content.hash.value and payloadHash.value)

nkreiger:
> > By not using the types.NewProposedEntry the in-toto rekor function was not assiging the hash key, which caused a downstream error when querying by the hash to find the log entry
> 
> What do you mean by hash key? I'm curious if you have a reproducer or diff on the change. Before and after your change I ran
> 
> ```
> COSIGN_EXPERIMENTAL=1 ./cosign attest --predicate pred.json --type slsaprovenance gcr.io/asra-ali/busybox/demo
> ```
> 
> and I was able to search both by payload hash. The entry contents themselves populate the same fields. (including content.hash.value and payloadHash.value)
> 
> If you meant hash key as in the merkle leaf hash then both have entry hashes that you can get log entry by UUID with.

Hi @asraa...this is how I re-create the issue with the current cosign binary

```
COSIGN_EXPERIMENTAL=1 cosign attest --predicate test.json --type custom $someimage:0.0.1

...log entry created with index:...(mine was 3089316)
```

`GET https://rekor.sigstore.dev/api/v1/log/entries?logIndex=3089316`

which returns your log, then I base64 decode the "body", which gives you the hash, and payloadHash

in this case:
```json
"content": {
            "hash": {
                "algorithm": "sha256",
                "value": "5fcf6b25abdc0a0becf5855d43361c4fad5f10290cdea213359526228bf540e8"
            },
            "payloadHash": {
                "algorithm": "sha256",
                "value": "9791b5658e1e832c889cf6518f600283365f819295916b27d0c2da0830cc975c"
            }
        },
```

then I attempt to retrieve the entry by the hash

```
curl --location --request POST 'https://rekor.sigstore.dev/api/v1/index/retrieve' \
--header 'Content-Type: application/json' \
--data-raw '{
    "hash": "sha256:5fcf6b25abdc0a0becf5855d43361c4fad5f10290cdea213359526228bf540e8"
}'
```

which returns `[]`

with my changes that fix that bug, you will get an array of UUID back running this same flow

nkreiger:
actually on further review @asraa you are right, you can use the payloadHash, with my changes it would let you search by the Hash or the PayloadHash, I was doing something hacky and was on a different version of rekor (cosign is on 0.4 I was on 0.10), so it looked like a bug to me from my local machine...if that is not desired behavior I can close this PR

asraa:
> with my changes it would let you search by the Hash or the PayloadHash, I was doing something hacky and was on a different version of rekor (cosign is on 0.4 I was on 0.10), so it looked like a bug to me from my local machine...if that is not desired behavior I can close this PR

I believe that there was an old bug that prevented old versions of Rekor from searching by content: https://github.com/sigstore/rekor/pull/870

So I think it was just Rekor versions that "fixes" the bug.

That being said, you could consider this PR unifying the entry types and upload using `types.NewProposedEntry`, but I originally thought we avoided that to avoid too many deps. I don't have a preference, just wanted to jump in to fix the PR description in case anyone searches a similar issue!

nkreiger:
@asraa thanks for posting that! That issue shows why that the changes I'm making still only allow you to search by the payloadHash on rekor 0.4, but allow you to search on either in 0.10

however, I still believe this is a bug on the cosign side (I tested the current cosign implementation against the latest rekor to confirm) due to this:

current implementation creates the intoto model, without the hash

https://github.com/sigstore/cosign/blob/main/pkg/cosign/tlog.go#L202

my proposed fix:

https://github.com/sigstore/rekor/blob/main/pkg/types/intoto/v0.0.1/entry.go#L293,L301

as you can see it is assigning the hash here, which allows you to query by it downstream

pkg/cosign/tlog.go
nkreiger: @hectorj2f correct sorry I addressed the comments by isolating it to the specific intoto function, and allowing it to propagate the default version...maybe we want to keep it a specific version though to reduce the risk of dependency changes?

asraa:
> however, I still believe this is a bug on the cosign side (I tested the current cosign implementation against the latest rekor to confirm) due to this:

Thanks Noah! This is great. Definitely +1 to add the hash in this PR. Clarify the PR description to something like "Add envelope hash to in-toto entries in tlog entry creation"?

nkreiger:
> > however, I still believe this is a bug on the cosign side (I tested the current cosign implementation against the latest rekor to confirm) due to this:
> 
> Thanks Noah! This is great. Definitely +1 to add the hash in this PR. Clarify the PR description to something like "Add envelope hash to in-toto entries in tlog entry creation"?

sure, updated

dlorenc:
Not sure why tests aren't running here, but it looks like you need to rebase now.

pkg/cosign/tlog.go
```diff
@@ -63,6 +64,8 @@ const (
 	// TODO(vaikas): Implement storing state like Rekor does so that if tree
 	// state ever changes, it will make lots of noise.
 	addRekorPublicKeyFromRekor = "SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY"
+	// define constants for the proposed entry dynamic spec mapping
+	intotoProposedEntryKey = "intoto"
```
asraa, line 68: can we use intoto.KIND? https://github.com/sigstore/rekor/blob/547eb3cef9690bc97e73e35eb3bf1f40b18a3504/pkg/types/intoto/intoto.go#L28

pkg/cosign/tlog.go
```diff
@@ -75,6 +78,18 @@ func getLogID(pub crypto.PublicKey) (string, error) {
 	return hex.EncodeToString(digest[:]), nil
 }
 
+func intotoEntry(ctx context.Context, signature, pubKey []byte) (models.ProposedEntry, error) {
+	e, err := types.NewProposedEntry(ctx, intotoProposedEntryKey, "", types.ArtifactProperties{
```
asraa, line 82: nit: add a comment that passing `""` as version uses the DefaultVersion

nkreiger: resolved to use the default version defined in the package, thought that would make more sense

pkg/cosign/tlog.go
nkreiger: done
f019a63e82219636db34d7d41e909f2ca3008197

nkreiger:
> Not sure why tests aren't running here, but it looks like you need to rebase now.

@dlorenc sorry this is my first PR for a project this big, can you elaborate on what I need to do to run the tests?

asraa (commented):
sg! intoto entry change lgtm

dlorenc:
> > Not sure why tests aren't running here, but it looks like you need to rebase now.
> 
> @dlorenc sorry this is my first PR for a project this big, can you elaborate on what I need to do to run the tests?

I don't think it's your fault - GitHub isn't running them for some reason. Can you rebase and force push? It might re-trigger them.

bobcallaway:
I'm not sure we need to change anything in cosign, as the hash is computed and added server-side before making the entry into the log: https://github.com/sigstore/rekor/blob/102dc64d4f217013dfcf789837c3581d3f03b3d7/pkg/types/intoto/v0.0.1/entry.go#L202

Are you really seeing client behavior differences between omitting it and specifying it?

nkreiger:
> I'm not sure we need to change anything in cosign, as the hash is computed and added server-side before making the entry into the log: https://github.com/sigstore/rekor/blob/102dc64d4f217013dfcf789837c3581d3f03b3d7/pkg/types/intoto/v0.0.1/entry.go#L202
> 
> Are you really seeing client behavior differences between omitting it and specifying it?

@bobcallaway yes, it does add the hash on the server side, I am not seeing that the hash doesn't exist...only that the current implementation of cosign does not add the hash before sending to the server, and it makes it so you can not query on it downstream:

https://github.com/sigstore/cosign/pull/2118#issuecomment-1202816682

see this example earlier in the PR

the only difference as far as I could tell was cosign doesn't use the types.NewProposedEntry for intoto, which adds the hash, and doing that fixes downstream query issue (if you are on latest rekor)

do you think its an issue on the rekor server side? I don't have the bandwidth to look into that today, but I can investigate if you think that is necessary

bobcallaway:
> > Are you really seeing client behavior differences between omitting it and specifying it?
> 
> I believe it is happening. IndexKeys happens BEFORE canonicalization. [sigstore/rekor#854](https://github.com/sigstore/rekor/issues/854)

So I don't think the order of the calls matters (in the insertion flow, [canonicalization](https://github.com/sigstore/rekor/blob/102dc64d4f217013dfcf789837c3581d3f03b3d7/pkg/api/entries.go#L168) happens before [`IndexKeys()`](https://github.com/sigstore/rekor/blob/102dc64d4f217013dfcf789837c3581d3f03b3d7/pkg/api/entries.go#L224) is called); but I do see the issue that is being reported here. The `V001Entry.IntotoObj` isn't ever updated to include the computed hash, but is included in the byte array that gets inserted into the log. 

IMO this should be fixed in rekor, not in cosign. 

> 
> I guess the clarification here is that this prevented searching by envelope hash:
> 
> > you can use the payloadHash, with my changes it would let you search by the Hash or the PayloadHash,

nkreiger:
> IMO this should be fixed in rekor, not in cosign.

agreed, however, I think there is still value in the PR which "unifies the entry types and upload using types.NewProposedEntry" to quote @asraa because it translates to use the

https://github.com/sigstore/rekor/blob/main/pkg/types/intoto/v0.0.1/entry.go#L264 `CreateFromArtifactProperties` and on the rekor-server side on upload of an entry of intoto type it is using Canocalize, Unmarshal, validate ... all off the same intoto entry struct, so I don't see the dependency issue that was the original concern?

lmk if you want me to close the PR though and open an issue on the rekor side

nkreiger:
thoughts? @bobcallaway

lmk if I should close the PR and move the issue, so I can stop bothering y'all :)

pkg/cosign/tlog.go
```diff
@@ -76,6 +78,18 @@ func getLogID(pub crypto.PublicKey) (string, error) {
...
+func intotoEntry(ctx context.Context, signature, pubKey []byte) (models.ProposedEntry, error) {
+	e, err := types.NewProposedEntry(ctx, intoto.KIND, intoto_v001.APIVERSION, types.ArtifactProperties{
+		ArtifactBytes:  signature,
+		PublicKeyBytes: pubKey,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return e, err
```
bobcallaway, line 90:
```suggestion
	return types.NewProposedEntry(ctx, intoto.KIND, intoto_v001.APIVERSION, types.ArtifactProperties{
		ArtifactBytes:  signature,
		PublicKeyBytes: pubKey,
	})
```

bobcallaway (commented):
one nit but otherwise LGTM

nkreiger:
> one nit but otherwise LGTM

@bobcallaway good catch, resolved, all commits signed off, lmk if there is anything else I need to do!

dlorenc:
Should be good now, thanks for your patience!

dlorenc:
Oh whoops, it looks like you might have to drop the go.mod changes, now we're getting a build failure due to downgrading rekor.

nkreiger:
@dlorenc you're right, looks like it is in the release-utils pkg bumpt to 0.7.3 that came with the rekor bump

https://github.com/sigstore/cosign/pull/2102 (there is a PR open actually to bump that pkg), maybe I could wait for that to go through? or should I revert the `go.mod` changes?

dlorenc:
No idea what's wrong with that one, I'll poke dependabot and see if that helps. If that doesn't get fixed by tomorrow or so we can merge this and fix after. Thanks!

dlorenc:
OK, #2102 is in, can you drop the replace and rebase?

nkreiger:
@dlorenc dropped

dlorenc:
Awesome, let's get this in!

dlorenc:
Thank you so much for the patience here!

dlorenc (approved):

lcarva:
FYI, I'm seeing issues with `verify-attestation` starting with this change: https://github.com/sigstore/cosign/issues/2197

csullivanupgrade:
`govulncheck` is reporting a type mismatch when using sigstore/rekor types.NewProposedEntry:

```
govulncheck: Packages contain errors:
/Users/csullivan/go/pkg/mod/github.com/sigstore/cosign@v1.11.1/pkg/cosign/tlog.go:89:19: cannot use pubKey (variable of type []byte) as [][]byte value in struct literal
```

PR #2132, opened 2022-08-04
DNM: Add repro test for https://github.com/sigstore/policy-controller/issues/130

vaikas:

#### Summary
Just trying to sort out why the policy-controller is not rejecting this policy. Put it in a PR for easier sharing, but do not merge.


dlorenc:
Can we close?

Issue #2134, opened 2022-08-05
How to get the Fulcio root certificate securely in `verify-blob`?


wata727:
First of all, thank you for the great project and everyone involved in it. I'm a maintainer of an [OSS project](https://github.com/terraform-linters/tflint) using cosign to sign release binaries. Our project uses GoReleaser for signing, and we run the following commands on the checksum:

```
COSIGN_EXPERIMENTAL=1 cosign sign-blob --output-certificate=checksum.txt.pem --output-signature=checksum.txt.keyless.sig checksum.txt
```
https://github.com/terraform-linters/tflint/blob/e7668cb182999e93a2e79534968d8ad686c14f14/.goreleaser.yml#L26-L37

To verify this, we provide a verification example as follows:

```
cosign verify-blob --cert checksums.txt.pem --signature checksums.txt.keyless.sig --certificate-github-workflow-repository=terraform-linters/tflint checksums.txt
```
https://github.com/terraform-linters/tflint/blob/e7668cb182999e93a2e79534968d8ad686c14f14/README.md#verification

However, I understand that this way does not verify the `checksum.txt.pem`'s certificate chain against the Fulcio root trust, so it is not sufficient:
https://github.com/sigstore/cosign/blob/v1.10.1/cmd/cosign/cli/verify/verify_blob.go#L122-L129

To avoid this, I understand that I need to pass the Fulcio root certificate with the `--certificate-chain` option:
https://github.com/sigstore/cosign/blob/v1.10.1/cmd/cosign/cli/verify/verify_blob.go#L131-L139

Now the question is, how can I get this root certificate securely?

I have confirmed that I can get certificates locally using TUF by `cosign initialize`. I can use this to build a certificate chain like this:

```
cosign initialize
cat ~/.sigstore/root/targets/fulcio_intermediate_v1.crt.pem > fulcio_chained.crt.pem
echo >> fulcio_chained.crt.pem
cat ~/.sigstore/root/targets/fulcio.crt.pem >> fulcio_chained.crt.pem
echo >> fulcio_chained.crt.pem
cat ~/.sigstore/root/targets/fulcio_v1.crt.pem >> fulcio_chained.crt.pem

cosign verify-blob --cert checksums.txt.pem --cert-chain fulcio_chained.crt.pem --signature checksums.txt.keyless.sig --certificate-github-workflow-repository=terraform-linters/tflint checksums.txt
Verified OK
```

[...]

PR #2139, opened 2022-08-06
Verify the certificate chain against the Fulcio root trust by default

wata727:
#### Summary
See https://github.com/sigstore/cosign/issues/2134

Currently, the verification performed by `cosign verify-blob --cert` does not verify the certificate chain, so you cannot guarantee that the certificate was correctly issued by Fulcio. So you need to explicitly pass Fulcio's root certificate via the `--cert-chain` option.

```
cosign initialize
cat ~/.sigstore/root/targets/fulcio_intermediate_v1.crt.pem > fulcio_chained.crt.pem
echo >> fulcio_chained.crt.pem
cat ~/.sigstore/root/targets/fulcio.crt.pem >> fulcio_chained.crt.pem
echo >> fulcio_chained.crt.pem
cat ~/.sigstore/root/targets/fulcio_v1.crt.pem >> fulcio_chained.crt.pem

cosign verify-blob --cert checksum.txt.pem --cert-chanin fulcio_chained.crt.pem ...
```

However, this way of getting root certificates is not secure. Ideally, it should be retrieved by a TUF client. After discussing this matter in #2134, @haydentherapper says that there is room for changing the default behavior.

This PR will change the default behavior from not verifying the certificate chain if `--cert-chain` is not passed to verifying the certificate chain against the Fulcio root trust getting by a TUF client.

#### Release Note

- Changed the default behavior from not verifying the certificate chain if `--cert-chain` is not passed to verifying the certificate chain against the Fulcio root trust getting by a TUF client.

#### Documentation

- The description of the `--certificate` option needs to be changed.

### DIFF
--- a/cmd/cosign/cli/options/certificate.go
+++ b/cmd/cosign/cli/options/certificate.go
@@ -37,7 +37,7 @@ var _ Interface = (*RekorOptions)(nil)
 // AddFlags implements Interface
 func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.Cert, "certificate", "",
-		"path to the public certificate")
+		"path to the public certificate. The certificate will be verified against the Fulcio roots if the --certificate-chain option is not passed.")
 
 	cmd.Flags().StringVar(&o.CertEmail, "certificate-email", "",
 		"the email expected in a valid Fulcio certificate")
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -119,11 +119,16 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			EnforceSCT:                   enforceSCT,
 		}
 		if certChain == "" {
-			err = cosign.CheckCertificatePolicy(cert, co)
+			// If no certChain is passed, the Fulcio root certificate will be used
+			co.RootCerts, err = fulcio.GetRoots()
 			if err != nil {
-				return err
+				return fmt.Errorf("getting Fulcio roots: %w", err)
 			}
-			verifier, err = signature.LoadVerifier(cert.PublicKey, crypto.SHA256)
+			co.IntermediateCerts, err = fulcio.GetIntermediates()
+			if err != nil {
+				return fmt.Errorf("getting Fulcio intermediates: %w", err)
+			}
+			verifier, err = cosign.ValidateAndUnpackCert(cert, co)
 			if err != nil {
 				return err
 			}
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -160,11 +160,16 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 			return err
 		}
 		if c.CertChain == "" {
-			err = cosign.CheckCertificatePolicy(cert, co)
+			// If no certChain is passed, the Fulcio root certificate will be used
+			co.RootCerts, err = fulcio.GetRoots()
 			if err != nil {
-				return err
+				return fmt.Errorf("getting Fulcio roots: %w", err)
 			}
-			pubKey, err = signature.LoadVerifier(cert.PublicKey, crypto.SHA256)
+			co.IntermediateCerts, err = fulcio.GetIntermediates()
+			if err != nil {
+				return fmt.Errorf("getting Fulcio intermediates: %w", err)
+			}
+			pubKey, err = cosign.ValidateAndUnpackCert(cert, co)
 			if err != nil {
 				return err
 			}
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -17,7 +17,6 @@ package verify
 
 import (
 	"context"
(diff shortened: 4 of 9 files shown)


wata727:
I have two things to discuss:

- Should this change be followed by similar commands such as `cosign verify`?
- Is there a good way to add tests for this implementation? `verify_blob_test.go` didn't seem to have any tests for verification.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -119,11 +119,16 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			EnforceSCT:                   enforceSCT,
 		}
 		if certChain == "" {
```
haydentherapper, line 121: We need to also update the documentation for the `--certificate` flag to note that it verifies the certificate against the Fulcio roots provided by TUF.

haydentherapper (changes requested):
The same changes need to be made in https://cs.github.com/sigstore/cosign/blob/c5fda01a8ff33ca981f45a9f13e7fb6bd2080b94/cmd/cosign/cli/verify/verify_attestation.go?q=org%3Asigstore+CheckCertificatePolicy#L141 and https://cs.github.com/sigstore/cosign/blob/95b74db89941e8ec85e768f639efd4d948db06cd/cmd/cosign/cli/verify/verify.go?q=org%3Asigstore+CheckCertificatePolicy#L162

haydentherapper:
Testing is unfortunately a bit rough. Check out `e2e_test.go`, but it might be hard to fit in a test for this, especially with TUF. We may need a TODO.

haydentherapper:
/hold, I want to give a bit more thought to this and will get back to you on Monday.

cc @znewman01

cmd/cosign/cli/verify/verify_blob.go
wata727: Fixed https://github.com/sigstore/cosign/pull/2139/commits/b4887ea391a33cedd595d3b193baae5dc15d037e

wata727:
> The same changes need to be made

Fixed https://github.com/sigstore/cosign/pull/2139/commits/eda267b9f4c323fdaa2f20330cff6ff137267b10

> Testing is unfortunately a bit rough. Check out e2e_test.go, but it might be hard to fit in a test for this, especially with TUF. We may need a TODO.

Understood. I'll take a little deeper look at the `e2e_test.go` implementation.

znewman01:
Hey, thanks for this! (and thanks for looping me in, @haydentherapper)

I haven't looked too closely at the code but this change seems like much better/safer default behavior.

I just want to think through workarounds for anyone who was relying on the old behavior. The *right* thing to do in such a case is probably to provide the public key directly, rather than via a cert; this makes it much clearer what you're asking for. Maybe we should refer users to the [`step certificate key`](https://smallstep.com/docs/step-cli/reference/certificate/key) command? IIUC it should Just Work™

Was there anything specific you wanted my feedback on?

patflynn:
I'm probably mis-interpreting this, but for cases where the user has signed a blob with a user provided custom key, why would we want to verify against the fulcio root?

haydentherapper:
@znewman01 Just looking for a second opinion! Generally this looked good to me, beyond the potentially breaking change.

@patflynn One thought is a user is very unlikely to provide a certificate in that case, they'd be providing a key. We've got a few options for providing verification material:
* No key or cert - Either:
   * Fetch the certificate+chain from OCI. Verify the certificate+chain against the TUF root
   * Fetch the certificate from Rekor (for blob signing, and I don't actually like this since Rekor is meant to be a witness to an event and not storage, but let's discuss that more in another issue). 
* A key - Use the key to verify
* A certificate+chain - Verify the chain, and extract the verification key from the cert. Don't verify against TUF.
* A certificate w/o chain - Current behavior is to just extract the verification key. My question is, is anyone providing a certificate that's not from Fulcio in this case? If the vast majority of users provide a Fulcio certificate, then we should verify against the chain from TUF.

The workaround would either be extracting the key themselves (Like Zach mentioned) or providing the full certificate chain.

patflynn:
@haydentherapper got it. that makes sense. Thanks for clarifying.

cmd/cosign/cli/options/certificate.go
```diff
@@ -37,7 +37,7 @@ var _ Interface = (*RekorOptions)(nil)
 // AddFlags implements Interface
 func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.Cert, "certificate", "",
-		"path to the public certificate")
+		"path to the public certificate. The certificate will be verified against the Fulcio roots if the --certificate-chain option is not passed.")
```
asraa, line 40: nit: Fulcio root provided in the TUF root?

asraa (approved):

haydentherapper:
LGTM, just need to fix the failing test. Thanks!

dlorenc:
The failure is unrelated and the fix is over here: https://github.com/sigstore/scaffolding/pull/277/

dlorenc:
I'll merge while we wait for the fix.

cmd/cosign/cli/options/certificate.go
wata727: I didn't mention it to TUF here because I have a concern about the description would be too long. If we should mention it, please let me know so I fix it.

Issue #2161, opened 2022-08-14
Panic in generate-key-pair when trying to write to a directory without permissions


Ikke:
**Description**

When running `cosign generate-key-pair` in a directory where it cannot write to, it results in a panic:

```
$ docker run -it --rm -e COSIGN_PASSWORD= -u1000:1000  gcr.io/projectsigstore/cosign:v1.10.1 generate-key-pair
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xd7bd7c]

goroutine 1 [running]:
github.com/sigstore/cosign/pkg/cosign.FileExists({0x27668bd?, 0x0?})
        github.com/sigstore/cosign/pkg/cosign/common.go:35 +0x5c
github.com/sigstore/cosign/cmd/cosign/cli/generate.GenerateKeyPairCmd({0x2b8bd80, 0xc0001ac000}, {0x0?, 0x0?}, {0x3f38378?, 0x0?, 0x0?})
        github.com/sigstore/cosign/cmd/cosign/cli/generate/generate_key_pair.go:88 +0x22a
github.com/sigstore/cosign/cmd/cosign/cli.GenerateKeyPair.func1(0xc0006be000?, {0x3f38378?, 0x0?, 0x0?})
        github.com/sigstore/cosign/cmd/cosign/cli/generate_key_pair.go:66 +0x4c
github.com/spf13/cobra.(*Command).execute(0xc0006be000, {0x3f38378, 0x0, 0x0})
        github.com/spf13/cobra@v1.5.0/command.go:872 +0x694
github.com/spf13/cobra.(*Command).ExecuteC(0xc00017a000)
        github.com/spf13/cobra@v1.5.0/command.go:990 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.5.0/command.go:918
main.main()
        github.com/sigstore/cosign/cmd/cosign/main.go:61 +0x4c
```

**Version**

1.10.1

znewman01:
This is closed by #2162 right?

znewman01:
Closing as fixed by #2162.

PR #2162, opened 2022-08-14
fix panic when os.Stat returns an error besides ErrNotExists

dsa0x:
Signed-off-by: Samsondeen Dare <samsondeen.dare@hashicorp.com>

#### Summary

This PR fixes a nil pointer error that occurs when `os.Stat` returns an error besides `ErrNotExist`. The issue was also reported in https://github.com/sigstore/cosign/issues/2161

#### Release Note

bug fix: fix panic when os.Stat returns an error besides ErrNotExists


dlorenc:
Looks like this needs the boilerplate fix!

pkg/cosign/common.go
```diff
@@ -27,12 +27,15 @@ import (
 )
 
 // TODO(jason): Move this to an internal package.
-func FileExists(filename string) bool {
+func FileExists(filename string) (bool, error) {
```
dlorenc, line 30: What do other semver folks think about this change? It's technically breaking, but I can't find any usage of it on GitHub.

dsa0x: Good catch. Also, to Jason's point, I think this should be in an internal package, as this behaviour doesn't need to be public to users. It's trivial to implement if they need to. That would prevent us from running into issues with breaking semver.

dlorenc: Yeah, if we're going to change the interface we might as well move it internal at the same time to prevent future issues. WDYT about making the change?

dsa0x: Sure. I can make a commit to that effect. Thanks

cmd/cosign/cli/importkeypair/import_key_pair.go
```diff
@@ -36,7 +36,12 @@ func ImportKeyPairCmd(ctx context.Context, keyVal string, args []string) error {
 		return err
 	}
 
-	if cosign.FileExists("import-cosign.key") {
+	fileExists, err := cosign.FileExists("import-cosign.key\000x")
```
cpanato, line 39: the `\000x` is expected?

cmd/cosign/cli/generate/generate_key_pair.go
```diff
@@ -85,7 +85,12 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string, args []string) error
 		return err
 	}
 
-	if cosign.FileExists("cosign.key") {
+	fileExists, err := cosign.FileExists("cosign.key")
+	if err != nil {
+		return err
```
cpanato, line 90: how about we wrap the error to give more information?

cmd/cosign/cli/importkeypair/import_key_pair.go
```diff
@@ -36,7 +36,12 @@ func ImportKeyPairCmd(ctx context.Context, keyVal string, args []string) error {
...
+	if err != nil {
+		return err
```
cpanato, line 41: here as well: how about we wrap the error to give more information?

thanks for working on this

pkg/cosign/common.go
cpanato: +1

cmd/cosign/cli/generate/generate_key_pair.go
dsa0x: done

cmd/cosign/cli/importkeypair/import_key_pair.go
dsa0x: that was a mistake. removed in [9d30c9e](https://github.com/sigstore/cosign/pull/2162/commits/9d30c9e081c91102c135b15f8d0c154873a3031b)

cmd/cosign/cli/generate/generate_key_pair.go
```diff
@@ -85,7 +86,12 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string, args []string) error
 		return err
 	}
 
-	if cosign.FileExists("cosign.key") {
+	fileExists, err := icos.FileExists("cosign.key")
+	if err != nil {
+		return errors.Wrap(err, "error checking if cosign.key exists")
```
cpanato, line 91: please use `fmt.Errorf` instead of `errors.Wrap`  the `github.com/pkg/errors` is archived and we should use the go native one.

cmd/cosign/cli/importkeypair/import_key_pair.go
```diff
@@ -36,7 +38,12 @@ func ImportKeyPairCmd(ctx context.Context, keyVal string, args []string) error {
 		return err
 	}
 
-	if cosign.FileExists("import-cosign.key") {
+	fileExists, err := icos.FileExists("import-cosign.key")
+	if err != nil {
+		return errors.Wrap(err, "error checking if import-cosign.key exists")
```
cpanato, line 43: same here: please use `fmt.Errorf`

cpanato (changes requested):
small request

cmd/cosign/cli/generate/generate_key_pair.go
```diff
@@ -85,7 +86,12 @@ func GenerateKeyPairCmd(ctx context.Context, kmsVal string, args []string) error
 		return err
 	}
 
-	if cosign.FileExists("cosign.key") {
+	fileExists, err := icos.FileExists("cosign.key")
+	if err != nil {
+		return fmt.Errorf("error checking if cosign.key exists: %w", err)
```
cpanato, line 91: small request, change the work `error` to `failed`

dsa0x: Done

internal/pkg/cosign/common.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 1: sorry not seeing before, but this should point to `2022` year

cpanato (approved):
lgtm

thanks so much
will defer to @dlorenc

Issue #2149, opened 2022-08-11
Annotation support when uploading blobs


cldmnky:
**Annotations for blobs**

We are using cosign to upload (and sign) release artifacts as blobs. We currently use something like:

`cosign upload blob -f ./release/linux/myapp:linux/amd64 -f ./release/darwin/myapp:darwin/amd64 -f ./release/windows/myapp.exe:windows/amd64 private-registry.tld/foo/myapp/myapp:{{ .releaseVersion }}`

It would be useful to be able to add (common) annotations to the descriptor(s) for the manifests to store some more information:

`cosign upload blob -a  date=xxxx -a name="a name" -f ./release/linux/myapp:linux/amd64 -f ./release/darwin/myapp:darwin/amd64 -f ./release/windows/myapp.exe:windows/amd64 private-registry.tld/foo/myapp/myapp:{{ .releaseVersion }}`

This way we can in our custom code get a bit more metadata on the releases we have stored in the registry.

I'll be happy to prepare a PR if it makes sense!

dlorenc:
Maybe a different flag to make sure it's clear these are unprotected (not signed)?

cldmnky:
Good point, I'll open a PR so we can figure out the flag name.

PR #2188, opened 2022-08-21
Add annotations for upload blob.

cldmnky:
This PR adds support for adding annotations to images and index manifests when uploading blobs (#2149)

#### Summary
We use `cosign upload blob` to upload and sign releases of a cli tool. We. would like to add a bit more metadata to our artifacts when we upload them, such as release name and other things.  

Our custom tool is then able to query or registry and display the annotations on the artifacts for our users.

Added tests for the `UploadFiles()` function since the permutations increased. The tests should cover the change.

#### Release Note

Added an `--annotations` flag to the `cosign upload blob` command that sets (unsigned) annotations on the OCI images written by the command.


pkg/cosign/remote/index.go
```diff
@@ -113,11 +113,21 @@ func UploadFiles(ref name.Reference, files []File, getMt MediaTypeGetter, remote
...
 		}
+
 		lastHash, err = img.Digest()
 		if err != nil {
 			return name.Digest{}, err
 		}
-		if err := remote.Write(ref, img, remoteOpts...); err != nil {
+
+		//  cast img to a v1.image
+		v1Img, ok := img.(v1.Image)
```
priyawadhwa, line 123: Instead of casting to a `v1.Image`, could we do something like this up above?

```go
img, err := static.NewFile(b, static.WithLayerMediaType(mt), static.WithAnnotations(annotations))
```

cldmnky: Sure!

cldmnky: Hmm `static.NewFile` did not handle the `WithAnnotations` option, added it and added a test for it.

cldmnky:
@priyawadhwa @dlorenc @mattmoor how do you want to proceed with this?

pkg/oci/static/file.go
```diff
@@ -46,6 +46,9 @@ func NewFile(payload []byte, opts ...Option) (oci.File, error) {
 		return nil, err
 	}
 
+	// Add annotations from options
+	img = mutate.Annotations(img, o.Annotations).(v1.Image)
```
priyawadhwa, line 50: Instead of calling mutate again, I think you can add the annotations in the addendum on line 42?

```go
	img, err := mutate.Append(base, mutate.Addendum{
		Layer: layer,
                Annotations: annotations,
	})
```

pkg/cosign/remote/index.go
```diff
@@ -141,6 +145,9 @@ func UploadFiles(ref name.Reference, files []File, getMt MediaTypeGetter, remote
 	}
 
 	if len(files) > 1 {
+		if annotations != nil {
+			idx = mutate.Annotations(idx, annotations).(v1.ImageIndex)
```
priyawadhwa, line 149: If we're already including annotations when the file is initially created in line 112, do we need to add them in again here?

priyawadhwa (commented):
sorry for the delayed review! left a couple comments :)

pkg/oci/static/file.go
cldmnky: Ah, yes! fixed

cldmnky: > sorry for the delayed review! left a couple comments :)

No worries!

priyawadhwa (approved):
Looks good, thank you!

Issue #2197, opened 2022-08-23
verify-attestation fails when rekor is enabled 


lcarva:
**Description**

I have an image, `quay.io/lucarval/tekton-test:914262188b`, which has been signed and attested by Tekton Chains. When using cosign 1.11.0, the `verify-attestation` command fails if `COSIGN_EXPERIMENTAL=1` is set.

```bash
$ COSIGN_EXPERIMENTAL=1 cosign-1.11.0 verify-attestation --key cosign.pub quay.io/lucarval/tekton-test:914262188b --type slsaprovenance
Error: no matching attestations:
signature not found in transparency log
 signature not found in transparency log
main.go:62: error during command execution: no matching attestations:
signature not found in transparency log
 signature not found in transparency log

$ COSIGN_EXPERIMENTAL=1 cosign-1.11.0 verify --key cosign.pub quay.io/lucarval/tekton-test:914262188b

Verification for quay.io/lucarval/tekton-test:914262188b --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The claims were present in the transparency log
  - The signatures were integrated into the transparency log when the certificate was valid
  - The signatures were verified against the specified public key

[{"critical":{"identity":{"docker-reference":"quay.io/lucarval/tekton-test"},"image":{"docker-manifest-digest":"sha256:436bd108139bb8a8656f14b935ea7b7f2c7c71e5b48462cc0d1c8b8741ebf298"},"type":"cosign container image signature"},"optional":null}]

$ cosign-1.11.0 version
...
GitVersion:    v1.11.0
GitCommit:     6bfac1a470492d8964778b1b8c41e0056bf5dbdd
GitTreeState:  clean
BuildDate:     2022-08-18T13:40:04Z
GoVersion:     go1.18.5
Compiler:      gc
Platform:      linux/amd64

$ cat cosign.pub
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEKOUTLiIJHqYyUDJLzVUwMFN7p2pu
laQ0OcWJxRTXxIZm16r8eAfkrqnDeRjCIXR9lLvYlemG15+n2+WKazUT2w==
-----END PUBLIC KEY-----
```

With the previous version of cosign, 1.10.1, all is well:

```
[...]

PR #2199, opened 2022-08-24
fix: fix typo that caused attestation verification failure

asraa:

#### Release Note

* bug: fixes a bug in container attestation verification with experimental mode in version 1.11.0

#### Documentation

Fixes https://github.com/sigstore/cosign/issues/2197
See https://github.com/sigstore/cosign/pull/2118

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -365,7 +365,7 @@ func proposedEntry(b64Sig string, payload, pubKey []byte) ([]models.ProposedEntr
 	// The fact that there's no signature (or empty rather), implies
 	// that this is an Attestation that we're verifying.
 	if len(signature) == 0 {
-		e, err := intotoEntry(context.Background(), signature, pubKey)
+		e, err := intotoEntry(context.Background(), payload, pubKey)
 		if err != nil {
 			return nil, err
 		}


asraa:
@cpanato this is a pretty bug behavior bug in cosign 1.11. May be worth a patch release.

@nkreiger @bobcallaway PTAL

bobcallaway (approved):
agreed, we should update and release this ASAP

lcarva (approved):
I have verified that this fixes the reproducer from https://github.com/sigstore/cosign/issues/2197

nkreiger:
looks good, my bad I'll never live that one down...

any reason why the e2e tests didn't catch it?

asraa:
No worries @nkreiger, it was on all of us :)

I just checked the e2e tests and it seems like there's just a SignCmd tlog test. I'll try to do a follow-up e2e test with an attestation.

cpanato (approved):

dlorenc:
Looks like there's an attestation test here: https://github.com/sigstore/cosign/blob/c61504db67076f6ba8dc13cad27eb39a3bf8d9de/test/e2e_test.go#L295

Not sure why it didn't catch this... Is it only for tlog cases?

dlorenc:
Ah - I get it now. That used to be the same code path (same HashedRekord type stored in rekor for both versions), now it's separate so we're missing the verification test for the new code path.

asraa:
> Looks like there's an attestation test here:

Hmm I think attestverify is never called with tlog explicit enabled in the tests.

dlorenc:
> Hmm I think attestverify is never called with tlog explicit enabled in the tests.

Yep - we have sign/verify with and without tlog, and attest/verify without tlog. The tlog part used to be shared code, but is separated now.

Issue #2222, opened 2022-09-05
`verify-blob` function should be implemented consistently with verify image function


hirokuni-kitahara:
**Description**

`verify-blob` is now implemented in "cmd/cosign/cli/verify" package, but this is inconsistent with `verify`.
The core functions of `verify` such as `VerifyImageSignature()` are implemented in "pkg/cosign", and it is easy for developers to invoke cosign's verification function from their own golang projects.
However, `verify-blob` does not have this type of implementation, and developers need to call `VerifyBlobCmd()` directly. This function requires some arguments that are filepath, and this makes it hard for developers to invoke verify-blob functions from some environment such as read-only container. 

The core functions of `verify-blob` should be inside "pkg/cosign" as well as verify image functions, and the functions should accept non-filepath arguments.
(`VerifyBlobCmd()` will be just a wrapper which calls the core func in "pkg/cosign" in the new implementation.)

#### Example
For example, something like below in "pkg/cosign" is what I am proposing in this issue.

```
func VerifyBlobSignature(ctx context.Context, blobBytes, sigBytes []byte, co *CheckOpts)


# reference func in pkg/cosign/verify.go
func VerifyImageSignature(ctx context.Context, sig oci.Signature, h v1.Hash, co *CheckOpts)
```

asraa:
I'll be working on one change to refactor and simplify the logic, TBA.

asraa:
Hey! We just released the security fixes and the code is refactored (and tested). If you rebase your PR (or might be easier to redo....) I think it'll be good to go!

hirokuni-kitahara:
Thank you very much! @asraa 
I will check the latest code and will update my PR.

znewman01:
@asraa : To what extend did https://github.com/sigstore/cosign/pull/2461 address this?

asraa:
> @asraa : To what extend did https://github.com/sigstore/cosign/pull/2461 address this?

I think it's totally done. The one thing required is that the interface provided here is a lot cleaner than the one currently in the pkg.

Right now it's
```
// VerifyBlobSignature verifies a blob signature.
func VerifyBlobSignature(ctx context.Context, sig oci.Signature, co *CheckOpts) (bundleVerified bool, err error) {
	// The hash of the artifact is unused.
	return verifyInternal(ctx, sig, v1.Hash{}, verifyOCISignature, co)
}
```

[...]

PR #2223, opened 2022-09-05
move `verify-blob` function to "pkg/cosign" to be consistent with `verify`

hirokuni-kitahara:
Signed-off-by: Hirokuni-Kitahara1 <hirokuni.kitahara1@ibm.com>

#### Summary

This PR solves the issue https://github.com/sigstore/cosign/issues/2222 by moving the core function of `verify-blob` from "cmd/cosign/cli/verify" package to "pkg/cosign" package.
The new function `VerifyBlobSignature()` is based on basically the same design as `VerifyImageSignature()`, which is the core function of `verify` command.
Also the arguments of `VerifyBlobSignature()` are all non-filepath information such as []byte, so it is easy for developers to invoke this function even on some read-only environment.

### DIFF
--- a/pkg/cosign/verify_blob.go
+++ b/pkg/cosign/verify_blob.go
@@ -47,6 +47,12 @@ type blobSignature struct {
 }
 
 func newBundleSignature(blobBytes []byte, b64sig string, bundle *bundle.RekorBundle) (*blobSignature, error) {
+	if blobBytes == nil {
+		return nil, errors.New("blobBytes must be non nil")
+	}
+	if b64sig == "" {
+		return nil, errors.New("b64sig must be non empty string")
+	}
 	return &blobSignature{
 		payload: blobBytes,
 		b64sig:  b64sig,
--- a/pkg/cosign/verify_blob.go
+++ b/pkg/cosign/verify_blob.go
@@ -42,18 +42,14 @@ import (
 type blobSignature struct {
 	payload []byte
 	b64sig  string
-	cert    *x509.Certificate
-	chain   []*x509.Certificate
 	bundle  *bundle.RekorBundle
 	v1.Layer
 }
 
-func newBundleSignature(blobBytes []byte, b64sig string, cert *x509.Certificate, chain []*x509.Certificate, bundle *bundle.RekorBundle) (*blobSignature, error) {
+func newBundleSignature(blobBytes []byte, b64sig string, bundle *bundle.RekorBundle) (*blobSignature, error) {
 	return &blobSignature{
 		payload: blobBytes,
 		b64sig:  b64sig,
-		cert:    cert,
-		chain:   chain,
 		bundle:  bundle,
 	}, nil
 }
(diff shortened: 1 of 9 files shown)


pkg/cosign/verify_blob.go
```diff
@@ -0,0 +1,223 @@
...
+			if isIntotoDSSE(blobBytes) {
+				verifier = dsse.WrapVerifier(verifier)
+			}
+			verifiers = append(verifiers, verifier)
+		}
+	}
+
+	var validSigExists bool
+	for _, verifier := range verifiers {
+		if err := verifyOCISignature(ctx, verifier, sig); err != nil {
```
asraa, line 129: What happens to intoto attestations?

pkg/cosign/verify_blob.go
```diff
@@ -0,0 +1,223 @@
...
+We recommend requesting the certificate/signature from the original signer of this blob and manually verifying with cosign verify-blob --cert [cert] --signature [signature].`)
+		return checkedSignatures, bundleVerified, fmt.Errorf("could not find a valid tlog entry for provided blob, found %d invalid entries", len(verifiers))
+	}
+
+	if validSigExists {
+		fmt.Fprintln(os.Stderr, "Verified OK")
+		checkedSignatures = append(checkedSignatures, sig)
+	}
+
+	if !validSigExists && co.RekorClient != nil {
```
asraa, line 148: How come the TLOG entry time will only be checked if a valid sig does not exist? `verifyOCISignature` does not check cert expiry time.

asraa (commented):
Hi! Do you think we could hold off on this for a day or two?

haydentherapper:
We should hold off on moving this code until a) the refactor we are working on is done b) there are most tests so we have confidence in moving code.

haydentherapper:
Another meta comment, seems like there are functional changes mixed into this. A refactor should be as minimal as possible, so if we did want to move this, it should be clear that we are just moving the functions, not making any changes in them.

pkg/cosign/verify_blob.go
hirokuni-kitahara: Thank you for your reviewing! The intoto DSSE verifier will only be available in keyless verification mode with my current codes, and this is simply a mistake.
I will fix this.

pkg/cosign/verify_blob.go
hirokuni-kitahara: Thank you for the information, I think we should have `VerifyBundle()` after `verifyOCISignature()` as VerifyImageSignature does. I will fix this.

hirokuni-kitahara:
@asraa Thank you for your reviewing and your comments, I will update the changes to fix some problems you mentioned in the comments.

@haydentherapper Thank you for your comment, I totally agree with that we should not mix a refactor and functional changes in 1 PR. I will remove functional changes from this PR and will create another PR for it when the timing is good.
About the refactoring you are working on, could you share some timeline like when the verify-blob codes would be ready for moving if possible?

haydentherapper:
We'll hopefully have the refactor in by either end of this week or next week.

hirokuni-kitahara:
That's great! Thank you very much @haydentherapper

Issue #2219, opened 2022-09-01
[intoto] Support intoto v0.0.2 type in rekor for upload and verification


asraa:
**Description**

This type allows full verification of the intoto entry. See https://github.com/sigstore/rekor/pull/973

We can migrate upload support to v0.0.2 and continue supporting v0.0.1 for verification. I can stage the PR for whenever Rekor server is deployed with v0.0.2 support.

haydentherapper:
We pushed the new release of Rekor to staging and ran into this. This is a blocker for releasing the new Rekor since we no longer allow the upload of 0.0.1 intoto entries, unless we want to ease that restriction.

@bobcallaway

asraa:
> This is a blocker for releasing the new Rekor since we no longer allow the upload of 0.0.1 intoto entries, unless we want to ease that restriction.

What about old clients using old cosign versions?

haydentherapper:
This is likely an issue. Clients will need to update, so we will need to give them the time to do so. I think we should relax the restriction for now so we can unblock Rekor updates.

haydentherapper:
No longer needed, as we use the DSSE type now.

PR #2226, opened 2022-09-06
use default version of intoto pluggable type

bobcallaway:
Instead of explicitly specifying the v0.0.1 type of the intoto rekor type, just use the default version (which is an empty string) and the server will use the preferred implementation.

Fixes: #2219 


haydentherapper:
cc @asraa

asraa (dismissed):

priyawadhwa (approved):
LGTM! i think the attestation test probably needs to be updated

bobcallaway:
The fix for this is blocked by https://github.com/sigstore/rekor/issues/1144

PR #2248, opened 2022-09-14
Fix e2e test failure, add test for local bundle without rekor bundle

haydentherapper:

#### Summary

This fixes the behavior of `verify-blob` when the local bundle does not contain a Rekor bundle (for example, signed without experimental mode). In this case, `verify-blob` will still run using the local bundle for the cert/sig information. 

If it does not contain one and we require a timestamp for cert validity, we fail. Adds a test for a short-lived cert with a bundle missing rekor bundle and no experimental fails. But if experimental is on, we can succeed.

#### Release Note

* fix: Fixes`verify-blob` to handle local bundles that do not contain Rekor bundles.

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -72,6 +72,7 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 	certGithubWorkflowRepository,
 	certGithubWorkflowRef string, enforceSCT bool) error {
 	var cert *x509.Certificate
+	var bundle *bundle.RekorBundle
 
 	if !options.OneOf(ko.KeyRef, ko.Sk, certRef) && !options.EnableExperimental() && ko.BundlePath == "" {
 		return &options.PubKeyParseError{}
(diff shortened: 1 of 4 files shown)


cpanato (approved):

haydentherapper:
@asraa I'm not 100% this is the right behavior. This is the case where we provide a local bundle (without experimental, so no rekor lookup) with just a signature (and it could optionally contain a cert too, but in the test, we pass a key), and the local bundle has no rekor bundle.

This is probably actually a valid use-case, that you use the local bundle just to hold the verification material.

asraa:
> @asraa I'm not 100% this is the right behavior. This is the case where we provide a local bundle (without experimental, so no rekor lookup) with just a signature (and it could optionally contain a cert too, but in the test, we pass a key), and the local bundle has no rekor bundle.

Ugh. I see this case.
Unfortunately we CAN produce a bundle out of experimental mode: https://github.com/sigstore/cosign/blob/fb51ab00a50e2fe7ec0beb558f6bbf514319457b/cmd/cosign/cli/sign/sign_blob.go#L86

So I think this should technically pass?

asraa (commented):
Fortunately this is a correctness issue, not a security concern...

haydentherapper:
Can we just add a check for experimental flag when verifying with a Rekor bundle? I think that’s the simplest fix.

haydentherapper:
That sounds good!

asraa:
Done! Updated description and release note @cpanato @haydentherapper

cpanato (approved):

cpanato:
going to merge this to check the post submit jobs and if they are green will release it

PR #2254, opened 2022-09-15
fix: add COSIGN_EXPERIMENTAL=1 for verify-bloba

developer-guy:

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -218,13 +218,10 @@ You may specify either a key, a certificate or a kms reference to verify against
 
 The signature may be specified as a path to a file or a base64 encoded string.
 The blob may be specified as a path to a file or - for stdin.`,
-		Example: `  cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--cert <cert>) --signature <sig> <blob>
+		Example: ` cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--certificate <cert>) --signature <sig> <blob>
 
   # Verify a simple blob and message
-  cosign verify-blob --key cosign.pub --signature sig msg
-
-  # Verify a simple blob with remote signature URL, both http and https schemes are supported
-  cosign verify-blob --key cosign.pub --signature http://host/my.sig
+  cosign verify-blob --key cosign.pub (--signature <sig path>|<sig url> msg)
 
   # Verify a signature from an environment variable
   cosign verify-blob --key cosign.pub --signature $sig msg
@@ -232,8 +229,8 @@ The blob may be specified as a path to a file or - for stdin.`,
   # verify a signature with public key provided by URL
   cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
 
-  # Verify a signature against a payload from another process using process redirection
-  cosign verify-blob --key cosign.pub --signature $sig <(git rev-parse HEAD)
+  # verify a signature with signature and key provided by URL
+  cosign verify-blob --key https://host.for/<FILE> --signature https://example.com/<SIG>
 
   # Verify a signature against Azure Key Vault
   cosign verify-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] --signature $sig <blob>
@@ -254,7 +251,7 @@ The blob may be specified as a path to a file or - for stdin.`,
   cosign verify-blob --key gitlab://[PROJECT_ID]  --signature $sig <blob>
 
   # Verify a signature against a certificate
-  cosign verify-blob --cert <cert> --signature $sig <blob>
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --certificate <cert> --signature $sig <blob>
 `,
 
 		Args: cobra.ExactArgs(1),
--- a/doc/cosign_verify-blob.md
+++ b/doc/cosign_verify-blob.md
@@ -18,13 +18,10 @@ cosign verify-blob [flags]
 ### Examples
 
 ```
-  cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--cert <cert>) --signature <sig> <blob>
+ cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--certificate <cert>) --signature <sig> <blob>
 
   # Verify a simple blob and message
-  cosign verify-blob --key cosign.pub --signature sig msg
-
-  # Verify a simple blob with remote signature URL, both http and https schemes are supported
-  cosign verify-blob --key cosign.pub --signature http://host/my.sig
+  cosign verify-blob --key cosign.pub (--signature <sig path>|<sig url> msg)
 
   # Verify a signature from an environment variable
   cosign verify-blob --key cosign.pub --signature $sig msg
@@ -32,8 +29,8 @@ cosign verify-blob [flags]
   # verify a signature with public key provided by URL
   cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
 
-  # Verify a signature against a payload from another process using process redirection
-  cosign verify-blob --key cosign.pub --signature $sig <(git rev-parse HEAD)
+  # verify a signature with signature and key provided by URL
+  cosign verify-blob --key https://host.for/<FILE> --signature https://example.com/<SIG>
 
   # Verify a signature against Azure Key Vault
   cosign verify-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] --signature $sig <blob>
@@ -54,7 +51,7 @@ cosign verify-blob [flags]
   cosign verify-blob --key gitlab://[PROJECT_ID]  --signature $sig <blob>
 
   # Verify a signature against a certificate
-  cosign verify-blob --cert <cert> --signature $sig <blob>
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --certificate <cert> --signature $sig <blob>
 
 ```
 


haydentherapper:
/hold

Not all of these cases require the flag.

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
...
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key gitlab://[OWNER]/[PROJECT_NAME]  --signature $sig <blob>
 
   # Verify a signature against GitLab with project id
-  cosign verify-blob --key gitlab://[PROJECT_ID]  --signature $sig <blob>
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key gitlab://[PROJECT_ID]  --signature $sig <blob>
 
   # Verify a signature against a certificate
-  cosign verify-blob --cert <cert> --signature $sig <blob>
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --cert <cert> --signature $sig <blob>
 `,
```
haydentherapper, line 258: @asraa Should we use the list you made of all possible flag combos (cert/key, experimental/no experimental, bundle, expired/unexpired cert) and have an example for each of these with more explanation?

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
...
 
   # Verify a simple blob and message
-  cosign verify-blob --key cosign.pub --signature sig msg
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature sig msg
 
   # Verify a simple blob with remote signature URL, both http and https schemes are supported
-  cosign verify-blob --key cosign.pub --signature http://host/my.sig
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature http://host/my.sig
 
   # Verify a signature from an environment variable
```
znewman01, line 229: let's axe this example? it's just explaining how the environment variables work in a shell, I think

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
...
-  cosign verify-blob --key cosign.pub --signature $sig msg
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature $sig msg
 
   # verify a signature with public key provided by URL
```
znewman01, line 232: Can we combine this with the "signature from URL" example?

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
...
-  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
 
   # Verify a signature against a payload from another process using process redirection
```
znewman01, line 235: IMO, axe this example. Maybe we can move it to the `docs/` directory of the Cosign repo. Or even a "useful tricks" section at the bottom of this doc.

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
...
-  cosign verify-blob --key cosign.pub --signature $sig <(git rev-parse HEAD)
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature $sig <(git rev-parse HEAD)
 
   # Verify a signature against Azure Key Vault
```
znewman01, line 238: I would like to combine all the KMS examples. It'd be nice to list the syntax for each, but I don't need it in the full context.

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
 
 The signature may be specified as a path to a file or a base64 encoded string.
 The blob may be specified as a path to a file or - for stdin.`,
-		Example: `  cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--cert <cert>) --signature <sig> <blob>
+		Example: ` COSIGN_EXPERIMENTAL=1 cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--cert <cert>) --signature <sig> <blob>
 
   # Verify a simple blob and message
-  cosign verify-blob --key cosign.pub --signature sig msg
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature sig msg
```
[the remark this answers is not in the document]
znewman01: +1, including the KMS ones. Anything without `--certificate` I think

cmd/cosign/cli/verify.go
```diff
@@ -218,43 +218,43 @@ You may specify either a key, a certificate or a kms reference to verify against
...
-  cosign verify-blob --key cosign.pub --signature http://host/my.sig
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature http://host/my.sig
 
   # Verify a signature from an environment variable
-  cosign verify-blob --key cosign.pub --signature $sig msg
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key cosign.pub --signature $sig msg
 
   # verify a signature with public key provided by URL
-  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
+  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
```
znewman01, line 233: `s/$sig/<sig>`

cmd/cosign/cli/verify.go
znewman01: +1 

Further, I'd like to see some organization and explanation. Right now, the text is just a laundry list of "if X, then Y"

We should explain `unset COSIGN_EXPERIMENTAL` and `COSIGN_EXPERIMENTAL=1` separately (I don't 100% understand the current behavior so please correct me, indicated uncertainty with "?")

Here's a braindump, maybe should move into Google Doc (feel free to copy it)

# no experimental

Check the signature on a blob against the given key or certificate.

## Keys

blah

## Certificates

- `--certificate` *and* `--certificate-chain` example:

   The provided certificate (and all of the certificates in the chain(?) must be unexpired right now.

   **NOTE**: this just accepts whatever the last certificate of the chain is as your root. If you want to check that it matches a fixed root certificate, use keyless mode and `SIGSTORE_ROOT_FILE` (except we don't want people to do that? what do we tell these people?)


- `--certificate` no chain

   Checks that the certificate is signed by Fulcio (must be unexpired). Will that ever work?


## Bundles

- `--bundle`

    If `--bundle` is provided, use the given Rekor bundle (see ...) to check that the time of the *signature* was while the certificate (and its ancestors??) were valid (without `--bundle`, the certificate must be *currently* valid).

    Works for keys and certs both.

   What happens if the bundle has a conflicting chain?

- `--bundle` + certs, no chain

    If there's a bundle use that to get the chain up to Fulcio? 

# experimental

## Keys

Same as above, but we check that signatures are in Rekor (I think?)

If `--bundle` provided, check that against Rekor (offline).

##  Certs

- `--certificate` and `--certificate-chain`: Like the keyfull version, except instead of automatically trusting the certificate chain, we check that the root matches Fulcio's cert.

   If you'd like to BYO-root, set `SIGSTORE_ROOT_FILE` and we'll compare the root of the chain against that.

   We hit Rekor online.

- `--certificate`, no chain: this would only work if signed by the root directly, which probably won't be the case (maybe if you BYO-root?)

## Bundles 

- `--keys` Same as above, don't hit Rekor.

- `--certificate` and `--certificate-chain` and `--bundle`. Same as above,  don't hit rekor

    TODO: what if bundle has a conflicting cert/chain?

- `--certificate` and `--bundle`: Same, but use a chain from bundle. Fails if bundle has no chain. 

    TODO: what happens if the bundle has a conflicting cert?

- `--bundle`: Same, but use cert/chain from bundle. Fails if bundle has no cert/chain.

    TODO: if the bundle just provides a key, do we trust it?

developer-guy:
I tried to address some of the reviews. Can you please take a look at @znewman01? Thx

cmd/cosign/cli/verify.go
```diff
@@ -218,28 +218,23 @@ You may specify either a key, a certificate or a kms reference to verify against
...
-  # Verify a simple blob with remote signature URL, both http and https schemes are supported
-  cosign verify-blob --key cosign.pub --signature http://host/my.sig
+  cosign verify-blob --key cosign.pub (--signature <sig path>|<sig url> msg)
 
   # Verify a signature from an environment variable
   cosign verify-blob --key cosign.pub --signature $sig msg
 
   # verify a signature with public key provided by URL
-  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
+  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
```
znewman01, line 230: Double example on this line?

I was imagining:

```
cosign verify-blog --key https://host.for/<FILE> --signature https://example.com/<SIG>
```

cmd/cosign/cli/verify.go
```diff
@@ -218,28 +218,23 @@ You may specify either a key, a certificate or a kms reference to verify against
...
-  # Verify a signature against a payload from another process using process redirection
-  cosign verify-blob --key cosign.pub --signature $sig <(git rev-parse HEAD)
 
   # Verify a signature against Azure Key Vault
-  cosign verify-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] --signature $sig <blob>
+ cosign verify-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] --signature $sig <blob>
 
   # Verify a signature against AWS KMS
-  cosign verify-blob --key awskms://[ENDPOINT]/[ID/ALIAS/ARN] --signature $sig <blob>
+  Ccosign verify-blob --key awskms://[ENDPOINT]/[ID/ALIAS/ARN] --signature $sig <blob>
```
znewman01, line 237: Typo: `Ccosign`

cmd/cosign/cli/verify.go
```diff
@@ -218,28 +218,23 @@ You may specify either a key, a certificate or a kms reference to verify against
...
   # verify a signature with public key provided by URL
-  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
+  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg  COSIGN_EXPERIMENTAL=1 cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
 
-  # Verify a signature against a payload from another process using process redirection
-  cosign verify-blob --key cosign.pub --signature $sig <(git rev-parse HEAD)
 
   # Verify a signature against Azure Key Vault
-  cosign verify-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] --signature $sig <blob>
+ cosign verify-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] --signature $sig <blob>
```
znewman01, line 234: I think we lost a space here, need to indent one more

znewman01 (commented):
I have a couple of minor issues, but otherwise I think this looks good. We may want to totally rework the docs here, but that can happen later.

dlorenc (approved):

znewman01 (approved):

PR #2256, opened 2022-09-15
fix: fix cert chain validation for verify-blob in non-experimental mode

asraa:

#### Summary

* fix: Pulls Fulcio root and intermediate when `--certificate-chain` is not passed into `verify-blob`.

In non-experimental mode, we incorrectly only loaded Fulcio roots to verify against when `certRef == ""`. 

When a `certRef != ""` in non-experimental mode, like, user provides cert with offline bundle without specifying a custom cert chain, then we would NOT load in the Fulcio roots. This would cause certificate validation failure.

Instead, whenever `certChain == ""` we will load in fulcio roots in non-experimental mode, either if you provide a cert or a bundle.

#### Release Note

* fix: Pulls Fulcio root and intermediate when `--certificate-chain` is not passed into `verify-blob`. The v1.12.0 release introduced a regression: when `COSIGN_EXPERIMENTAL` was not set, cosign `verify-blob` would check a` --certificate` (without a `--certificate-chain` provided) against the operating system root CA bundle. In this release, Cosign checks the certificate against Fulcio's CA root instead (restoring the earlier behavior).

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -149,13 +149,14 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			if err != nil {
 				return fmt.Errorf("getting Fulcio roots: %w", err)
 			}
+
 			co.IntermediateCerts, err = fulcio.GetIntermediates()
 			if err != nil {
 				return fmt.Errorf("getting Fulcio intermediates: %w", err)
 			}
 			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
 			if err != nil {
-				return err
+				return fmt.Errorf("validating certRef: %w", err)
 			}
 		} else {
 			// Verify certificate with chain
@@ -195,8 +196,15 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 				if err != nil {
 					return fmt.Errorf("getting Fulcio intermediates: %w", err)
 				}
+				co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
+			} else {
+				// Verify certificate with chain
+				chain, err := loadCertChainFromFileOrURL(certChain)
+				if err != nil {
+					return err
+				}
+				co.SigVerifier, err = cosign.ValidateAndUnpackCertWithChain(cert, chain, co)
 			}
-			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
 		}
 		if err != nil {
 			return fmt.Errorf("loading verifier from bundle: %w", err)
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -107,7 +107,7 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			co.RekorClient = rekorClient
 		}
 	}
-	if certRef == "" || options.EnableExperimental() {
+	if options.EnableExperimental() {
 		co.RootCerts, err = fulcio.GetRoots()
 		if err != nil {
 			return fmt.Errorf("getting Fulcio roots: %w", err)
@@ -145,6 +145,14 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			return err
 		}
 		if certChain == "" {
+			co.RootCerts, err = fulcio.GetRoots()
+			if err != nil {
+				return fmt.Errorf("getting Fulcio roots: %w", err)
+			}
+			co.IntermediateCerts, err = fulcio.GetIntermediates()
+			if err != nil {
+				return fmt.Errorf("getting Fulcio intermediates: %w", err)
+			}
 			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
 			if err != nil {
 				return err
@@ -178,10 +186,20 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			// check if cert is actually a public key
 			co.SigVerifier, err = sigs.LoadPublicKeyRaw(certBytes, crypto.SHA256)
 		} else {
+			if certChain == "" {
+				co.RootCerts, err = fulcio.GetRoots()
+				if err != nil {
+					return fmt.Errorf("getting Fulcio roots: %w", err)
+				}
+				co.IntermediateCerts, err = fulcio.GetIntermediates()
+				if err != nil {
+					return fmt.Errorf("getting Fulcio intermediates: %w", err)
+				}
+			}
 			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
 		}
 		if err != nil {
-			return err
+			return fmt.Errorf("loading verifier from bundle: %w", err)
 		}
 		bundle = b.Bundle
 	// No certificate is provided: search by artifact sha in the TLOG.
--- a/cmd/cosign/cli/verify/verify_blob_test.go
+++ b/cmd/cosign/cli/verify/verify_blob_test.go
@@ -434,7 +434,6 @@ func TestVerifyBlob(t *testing.T) {
 			experimental: false,
 			shouldErr:    true,
 		},
-
 		{
 			name:         "valid signature with expired certificate - experimental good rekor lookup",
 			blob:         blobBytes,
@@ -446,7 +445,6 @@ func TestVerifyBlob(t *testing.T) {
 				expiredLeafPem, true),
 			shouldErr: false,
 		},
-
 		{
 			name:         "valid signature with expired certificate - experimental bad rekor integrated time",
 			blob:         blobBytes,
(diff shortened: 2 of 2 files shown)


cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -145,6 +145,14 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
 			return err
 		}
 		if certChain == "" {
+			co.RootCerts, err = fulcio.GetRoots()
```
asraa, line 148: I could move this to a blanket `if certChain == ""` outside the switch.

We would then load it in even if we are using a pubkey and it's not needed.

znewman01: But that would add a network call where none is needed; I think we want to avoid that

asraa:
Should add test validating that we will succeed when `certRef != ""` in non-experimental mode. I think that's currently missing. One sec.

cmd/cosign/cli/verify/verify_blob.go
haydentherapper: I think this is fine as-is, it's clearer to load in the chain around the point of verification imo - This probably needs a larger rewrite anyways.

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -959,6 +957,91 @@ func TestVerifyBlobCmdWithBundle(t *testing.T) {
 	})
 }
 
+func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
+	keyless := newKeylessStack(t)
+	// Change the keyless stack.
+	_ = newKeylessStack(t)
+	t.Run("Invalid certificate root explicit certRef", func(t *testing.T) {
+		// Change the keyless stack.
+		_ = newKeylessStack(t)
```
znewman01: Don't you do that outside the test? (couple lines up)

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -178,10 +186,20 @@ func VerifyBlobCmd(ctx context.Context, ko options.KeyOpts, certRef, certEmail,
...
+				co.RootCerts, err = fulcio.GetRoots()
+				if err != nil {
+					return fmt.Errorf("getting Fulcio roots: %w", err)
+				}
+				co.IntermediateCerts, err = fulcio.GetIntermediates()
+				if err != nil {
+					return fmt.Errorf("getting Fulcio intermediates: %w", err)
+				}
+			}
 			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
```
haydentherapper, line 199: Should we handle `else` for when the cert chain is set? Otherwise, the certificate wouldn't validate if it's been signed by a non-fulcio pki

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -959,6 +957,91 @@ func TestVerifyBlobCmdWithBundle(t *testing.T) {
...
+			blobPath, /*blobRef*/
+			// GitHub identity flags start
+			"", "", "", "", "",
+			// GitHub identity flags end
+			false /*enforceSCT*/)
+		if err == nil || !strings.Contains(err.Error(), "certificate signed by unknown authority") {
+			t.Fatalf("expected error with invalid root CA, got %v", err)
+		}
+	})
+	t.Run("Invalid certificate root explicit bundle", func(t *testing.T) {
```
znewman01, line 1042: nit: maybe "Invalid certificate root, certificate fetched from bundle"? Took me a while to puzzle out what's going on

asraa:
Just added a test verifying that `certRef` provided without explicit intermediate succeeds against the fulcio roots.

cmd/cosign/cli/verify/verify_blob.go
asraa: good catch!

asraa: lemme see if I can also add a test here.

znewman01: Hm...I think it's pretty unexpected that `--certificate-chain` would mean "accept whatever cert was at the root of the chain."

This is part of why I think an explicit `--certificate-root` flag would be a good idea.

haydentherapper: We do document the the chain is the intermediates and root - https://github.com/sigstore/cosign/blob/04f509c7c95538d50a6065c2f098eada59d9f139/cmd/cosign/cli/options/certificate.go#L67-L70

asraa: done! thank you. test correctly regressed before, succeeds with change.

cmd/cosign/cli/verify/verify_blob_test.go
asraa: Fixed in latest commit

cmd/cosign/cli/verify/verify_blob.go
znewman01: TBH that help text doesn't imply to me that "we'll interpret whatever is in the last spot as the root certificate," it's ambiguous whether the tool will compare it to some other source of truth.

Plus, with `COSIGN_EXPERIMENTAL=1` set, we *do* do that check, right? Isn't that what https://github.com/sigstore/cosign/issues/1554 is about?

cmd/cosign/cli/verify/verify_blob.go
asraa: Yeah, I kept this as is to avoid the network call, sounds good. let's keep as is. Just wanted to keep code simple.

cmd/cosign/cli/verify/verify_blob.go
asraa: I thought we consider self-signed certs as roots?

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -957,6 +955,166 @@ func TestVerifyBlobCmdWithBundle(t *testing.T) {
 			t.Fatalf("expected error with mismatched issuer, got %v", err)
 		}
 	})
+	t.Run("Intermediate root not explicit in non-experimental mode", func(t *testing.T) {
```
znewman01, line 958: nit: `s/not explicit/from bundle/`?

cmd/cosign/cli/verify/verify_blob_test.go
asraa: "Invalid certificate root when specifying cert in bundle"

cmd/cosign/cli/verify/verify_blob.go
asraa: Ugh we don't, but that's what the SIGSTORE_ROOT_FILE code does.

Can we file a separate issue on that?

znewman01: If we're sure that this matches the behavior from `v1.11.x`, and we explain it more clearly in the help text, I'm okay with dealing with this problem later.

haydentherapper: Yea, let's. It's a bit unclear to me what should be the right behavior here, especially if I provide both a sigstore root and certificate chain.

haydentherapper: The behavior here should match 1.11.x (the more significant change is around checking expiration)

cmd/cosign/cli/verify/verify_blob_test.go
asraa: Hmm no: the intermediate is not in the bundle. Changed to:

`Non-explicit Fulcio roots with bundle in non-experimental mode`

asraa: hahaha implciit = non-explicit

znewman01 (approved):

cmd/cosign/cli/verify/verify_blob.go
asraa: If someone could help me update the release notes to close this out, that'd be great. I added one in. I'm not sure if anything breaks.

haydentherapper (approved):
thanks!

cmd/cosign/cli/verify/verify_blob.go
znewman01: Hm I don't see anything under "Release Notes" in the PR description.

Maybe:

> The `v1.12.0` release introduced a regression: when `COSIGN_EXPERIMENTAL` was not set, `cosign verify-blob` would check a `--certificate` (without a `--certificate-chain` provided) against the operating system root CA bundle. In this release, Cosign checks the certificate against Fulcio's CA root instead (restoring the earlier behavior).


I think we should have a summary of the current behavior (as opposed to a diff, in the release note) somewhere too; maybe that's what we can hash out in https://github.com/sigstore/cosign/pull/2254

asraa: Sorry, I put it in the summary instead. Moved, but feel free to edit.

Sg, I'll add a longer explanation

znewman01 (approved):

dlorenc (approved):

cpanato:
@asraa @dlorenc should we release a 1.12.1 patch?

haydentherapper:
@cpanato as soon as possible I think. Release notes should be specified in the PR description

PR #2260, opened 2022-09-16
fix: fixing breaking changes in  rekor v1.12.0 upgrade

developer-guy:

This PR aims to fix breaking changes in Rekor v1.12.0 pkg

#### Documentation

/cc @priyawadhwa


pkg/cosign/tlog.go
```diff
@@ -84,9 +84,17 @@ func getLogID(pub crypto.PublicKey) (string, error) {
...
+	if len(pubKey) == 0 {
+		return nil, errors.New("none of the Rekor public keys have been found")
+	}
+
+	pubKeyBytes = append(pubKeyBytes, pubKey)
+
 	return types.NewProposedEntry(ctx, intoto.KIND, intoto_v001.APIVERSION, types.ArtifactProperties{
 		ArtifactBytes:  signature,
-		PublicKeyBytes: pubKey,
+		PublicKeyBytes: pubKeyBytes,
```
asraa, line 97: had to check up on why the `rekorEntry` function here that reconstructs hashedrekord types didn't need to be changed: that directly creates a v0.0.1 hashed rekord type, instead of relying on the general entry construction.

asraa (dismissed):
LGTM minus that I think the test code also needed to be changed. Happy to push a fix for that.

asraa:
Just ran into a problem that we can't generate a rekor entry for testing for intoto v0.0.1 at 0.12.0 because v0.0.1 intoto was explicitly not supported. We need Priya's fix here: https://github.com/sigstore/rekor/commit/a6d36b8028ecdb4c047dde52b29a76b51b058b2f

asraa:
I think you may have branch protection on! (didn't expect anything less!)

But this PR is updated with two commits in my fork: https://github.com/asraa/cosign/tree/fix-rekorv112

dlorenc:
Is this one ready to go @asraa ?

haydentherapper:
This is blocking 0.12 rolling out to prod, so this should be prioritized.

dlorenc:
> This is blocking 0.12 rolling out to prod, so this should be prioritized.

what still needs to happen here? cc @asraa ?

asraa (approved):
Sorry about that! LGTM -- the tests have the change now

PR #2268, opened 2022-09-20
feat: improve the verification message

### LINKED ISSUE #2216: use proper naming for Sigstore OID information inside verification message
**Description**

we (w/@dentrax) made a development about printing cert extensions in the verification message[^1]. But, I noticed that within that optional field, the values in the form of root OID is not very user-friendly. 

```shell
COSIGN_EXPERIMENTAL=1 cosign verify ghcr.io/ossf/scorecard/v4 | jq
...
    "optional": {
      "1.3.6.1.4.1.57264.1.2": "push",
      "1.3.6.1.4.1.57264.1.3": "d13ba3f3355b958d5d62edc47282a2e7ed9fa7c1",
      "1.3.6.1.4.1.57264.1.4": "publishimage",
      "1.3.6.1.4.1.57264.1.5": "ossf/scorecard",
      "1.3.6.1.4.1.57264.1.6": "refs/heads/main",
...
```

So, we can use proper naming (sorry that I don't know the technical names of them) for them to be more useful to end users.

```shell
    "optional": {
      "githubWorkflowTrigger": "push",
      "githubWorkflowSha": "d13ba3f3355b958d5d62edc47282a2e7ed9fa7c1",
      "githubWorkflowName": "publishimage",
      "githubWorkflowRepository": "ossf/scorecard",
      "githubWorkflowRef": "refs/heads/main",

```

[^1]: https://github.com/sigstore/cosign/issues/1989

developer-guy:

Fixes #2216 

#### Release Note

feat: print the names of custom extensions (like "GitHub Workflow Trigger") rather than OIDs (like 1.3.6.1.4.1.57264.1.2) in the human-readable output of cosign verify{,-attestation}, and use field names (githubWorkflowTrigger) in keys of the JSON output of these commands (the old OID keys are kept for backwards compatibility, but they are deprecated and their use is discouraged).

> thx to @znewman01 

#### Documentation

/cc @znewman01

### DIFF
--- a/cmd/cosign/cli/verify/verify_test.go
+++ b/cmd/cosign/cli/verify/verify_test.go
@@ -1,0 +1,162 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package verify
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"sync"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/sigstore/cosign/pkg/oci"
+	"github.com/sigstore/cosign/pkg/oci/static"
+	"github.com/sigstore/cosign/test"
+	"github.com/sigstore/sigstore/pkg/signature/payload"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrintVerification(t *testing.T) {
+	// while we are adding a more human-readable output for cert extensions, on the other hand
+	// we want as backward compatible as possible, so we are keeping the old OIDs field names as well.
+	wantPayload := `
+[
+    {
+        "critical": {
+            "identity": {
+                "docker-reference": "gcr.io/baz/baz"
+            },
+            "image": {
+                "docker-manifest-digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+            },
+            "type": "cosign container image signature"
+        },
+        "optional": {
+            "1.3.6.1.4.1.57264.1.1": "oidc-issuer",
+            "1.3.6.1.4.1.57264.1.2": "myWorkflowTrigger",
+            "1.3.6.1.4.1.57264.1.3": "myWorkflowSha",
+            "1.3.6.1.4.1.57264.1.4": "myWorkflowName",
+            "1.3.6.1.4.1.57264.1.5": "myWorkflowRepository",
+            "1.3.6.1.4.1.57264.1.6": "myWorkflowRef",
+            "Issuer": "oidc-issuer",
+            "Subject": "subject",
+            "githubWorkflowName": "myWorkflowName",
+            "githubWorkflowRef": "myWorkflowRef",
+            "githubWorkflowRepository": "myWorkflowRepository",
+            "githubWorkflowSha": "myWorkflowSha",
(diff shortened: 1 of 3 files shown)


znewman01 (changes requested):
This looks great!

Three points:

1. It's technically a backwards-incompatible change. If you were relying on the output of Cosign, which *should* stay relatively stable, using `jq` or similar and looking for the OIDs, this will break that.

    I think we should do *both*: set `ss.Optional[cosign.CertExtensionGitHubWorkflowTrigger]` *and* `ss.Optional[cosign.CertExtensionMap[...]]`.

3. Is there a good way to test this? I know there's no `verify_test.go` right now. Can you make a (very simple!) one for the JSON output? 

4. `cosign verify-blob` doesn't use `PrintVerification` like `cosign verify` and `cosign verify-attestation`! It may or may not make sense to use `PrintVerification` there, but we should at least improve the output. You don't need to fix this, but it'd be great if you could file a bug.

znewman01:
Oh, and can you make the release note a little more detailed? Something like: 

> feat: print the names of custom extensions (like "GitHub Workflow Trigger") rather than OIDs (like `1.3.6.1.4.1.57264.1.2`) in the human-readable output of `cosign verify{,-attestation}`, and use field names (`githubWorkflowTrigger`) in keys of the JSON output of these commands (the old OID keys are kept for backwards compatibility, but they are deprecated and their use is discouraged).

developer-guy:
>It's technically a backwards-incompatible change. If you were relying on the output of Cosign, which should stay relatively stable, using jq or similar and looking for the OIDs, this will break that.

Done 🕺🏻

> Is there a good way to test this? I know there's no verify_test.go right now. Can you make a (very simple!) one for the JSON output?

Done 🥳

Thx for the really valuable comments and feedbacks ! 🫶

test/cert_utils.go
```diff
@@ -145,22 +145,25 @@ func GenerateLeafCertWithExpiration(subject string, oidcIssuer string, expiratio
...
-	certTemplate := &x509.Certificate{
-		SerialNumber:   big.NewInt(1),
-		EmailAddresses: []string{subject},
-		NotBefore:      time.Now().Add(-1 * time.Minute),
-		NotAfter:       time.Now().Add(time.Hour),
-		KeyUsage:       x509.KeyUsageDigitalSignature,
-		ExtKeyUsage:    []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning},
-		IsCA:           false,
-		ExtraExtensions: []pkix.Extension{{
+func GenerateLeafCert(subject string, oidcIssuer string, parentTemplate *x509.Certificate, parentPriv crypto.Signer, eexts ...pkix.Extension) (*x509.Certificate, *ecdsa.PrivateKey, error) {
```
znewman01, line 148: What do you think about renaming `eexts` to `exts` and then making the old `exts` anonymous?

```
exts = append(exts, []pkix.Extension{ ...
}
```

I could go either way, but  the name `eexts`  took me a little while to understand.

cmd/cosign/cli/verify/verify_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/stretchr/testify/assert"
+	"io"
+	"log"
+	"os"
+	"sync"
+	"testing"
+)
+
+func TestPrintVerification(t *testing.T) {
+	wantPayload := `
```
znewman01, line 43: Can you add a comment about how we want *both* the OIDs and the cleaned up names for backwards compatibility?

cmd/cosign/cli/verify/verify_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"crypto"
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
```
znewman01, line 28: Please fix lint error

cmd/cosign/cli/verify/verify_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/sigstore/sigstore/pkg/signature/payload"
+	"github.com/stretchr/testify/assert"
+	"io"
+	"log"
+	"os"
+	"sync"
+	"testing"
+)
+
+func TestPrintVerification(t *testing.T) {
```
znewman01, line 42: If you *really* want, you can add a test for the text-style output (not JSON). But that's totally optional for this PR

znewman01 (changes requested):

test/cert_utils.go
developer-guy: that makes a lot of sense ! 🫶

cmd/cosign/cli/verify/verify_test.go
developer-guy: maybe we can do this later, wdyt?

znewman01 (approved):

hectorj2f (approved):
lgtm

PR #2269, opened 2022-09-20
feat: use stdin as an input for predicate

### LINKED ISSUE #2215: read attestation from stdin in attest cmd
**Description**

We(w/@dentrax)'ve opened an issue in the Trivy project related to the signing attestation before saving it to the disk.[^1] A similar work has also been done in the Syft project.[^2] Thanks to @knqyf263, he came up with an interesting idea, if we pass attestations via a pipe in attest cmd, then all security scanners don't have to re-implement cosign functionalities. So, the UX would be like the following:

```shell
trivy image -f cosin-vuln <IMAGE> | COSIGN_EXPERIMENTAL=1 cosign attest --type vuln --predicate - <IMAGE>
```

This makes a lot of sense to me, so I've raised an issue in Cosign to discuss further; if you like the idea, I'm willing to work on it, thanks.

[^1]: https://github.com/aquasecurity/trivy/issues/2758 
[^2]: https://github.com/anchore/syft/pull/785

developer-guy:

Fixes #2215 

#### Release Note

feat: use stdin as an input for predicate

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -21,6 +21,7 @@ import (
 	_ "crypto/sha256" // for `crypto.SHA256`
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"time"
 
@@ -64,7 +65,7 @@ func uploadToTlog(ctx context.Context, sv *sign.SignerVerifier, rekorURL string,
 	return cbundle.EntryToBundle(entry), nil
 }
 
-//nolint
+// nolint
 func AttestCmd(ctx context.Context, ko options.KeyOpts, regOpts options.RegistryOptions, imageRef string, certPath string, certChainPath string,
 	noUpload bool, predicatePath string, force bool, predicateType string, replace bool, timeout time.Duration, noTlogUpload bool) error {
 	// A key file or token is required unless we're in experimental mode!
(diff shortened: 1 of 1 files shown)


cmd/cosign/cli/attest/attest.go
```diff
@@ -116,15 +117,20 @@ func AttestCmd(ctx context.Context, ko options.KeyOpts, regOpts options.Registry
 	wrapped := dsse.WrapSigner(sv, types.IntotoPayloadType)
 	dd := cremote.NewDupeDetector(sv)
 
+	var predicate []byte
```
znewman01, line 120: What do you think about making this `var predicate io.ReadCloser` so you don't have to add the `bytes.NewReader` shim below?

znewman01: You'd then just set `predicate = os.Stdin` and `predicate, err = os.Open(predicatePath)`. Remember to `defer predicate.Close()` too. Feels a little cleaner, and doesn't use up all your memory if `predicatePath` contains a huge file (at least not until deeper in the call stack).

cmd/cosign/cli/attest/attest.go
```diff
@@ -116,15 +117,20 @@ func AttestCmd(ctx context.Context, ko options.KeyOpts, regOpts options.Registry
...
 	fmt.Fprintln(os.Stderr, "Using payload from:", predicatePath)
```
znewman01, line 121: I think this `fmt.Fprintln` should go in the "else" case below: "Using payload from: -" is a little weird as a message. You *could* add a message for "Using payload from: standard input" or something, but I don't think it's necessary.

znewman01 (commented):
Nice!

cmd/cosign/cli/attest/attest.go
developer-guy: absolutely!

znewman01 (approved):
Looks great!

dlorenc (approved):

PR #2285, opened 2022-09-27
Fix: Remove an extra registry request from verification path.

mattmoor:
:bug: Remove an extra registry request when supplied with a digest.

The current logic calls `getSignedEntity` which must determine the media type of the referenced entity in order to produce an appropriately types `SignedEntity`.  However, we are only interested in the attached signatures (or attestations) and not the underlying structure, so when we are passed a digest, this roundtrip is entirely superfluous.

This changes this logic to explicitly resolve the digest (since verification wants to know the `v1.Hash`), which is free for digests.  Then we explicitly invoke `ociremote.Signatures()` on one of `ociremote.{Signature,Attestation}Tag`, which is also free (since we always pass the digest).

/kind bug

Signed-off-by: Matt Moore <mattmoor@chainguard.dev>

#### Release Note
Make signature/attestation verification slightly more efficient when passed a digest (as we always are in `policy-controller`).

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -118,19 +118,6 @@ type CheckOpts struct {
 	Identities []Identity
 }
 
-func getSignedEntity(signedImgRef name.Reference, regClientOpts []ociremote.Option) (oci.SignedEntity, v1.Hash, error) {
-	se, err := ociremote.SignedEntity(signedImgRef, regClientOpts...)
-	if err != nil {
-		return nil, v1.Hash{}, err
-	}
-	// Both of the SignedEntity types implement Digest()
-	h, err := se.(interface{ Digest() (v1.Hash, error) }).Digest()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


imjasonh (dismissed):

pkg/cosign/verify.go
```diff
@@ -625,22 +620,30 @@ func loadSignatureFromFile(sigRef string, signedImgRef name.Reference, co *Check
...
+	digest, err := ociremote.ResolveDigest(signedImgRef, co.RegistryClientOpts...)
 	if err != nil {
 		return nil, false, err
 	}
-	atts, err := se.Attestations()
+	h, err := v1.NewHash(digest.Identifier())
+	if err != nil {
+		return nil, false, err
+	}
+	st, err := ociremote.SignatureTag(digest, co.RegistryClientOpts...)
```
mattmoor, line 642: oops

dlorenc (approved):

PR #2286, opened 2022-09-27
Add attest-blob command

priyawadhwa:
Built on top of https://github.com/sigstore/cosign/pull/1265, thank you mlieberman85!

This is just the signing piece, we will need `cosign verify-attest-blob` as well which I'll do in a follow-up PR.
Once that's in I'll add in the unit and integration tests!


vaikas:
Thanks so much for doing this! Would it be possible to add an e2e test for this functionality like here (hopefully mainly copy&paste if we want a new test leg, or just add a new test case in the existing one (sign/verify) (not worrying about the now overbaked testfile name :) ):
https://github.com/sigstore/cosign/blob/main/.github/workflows/kind-verify-attestation.yaml#L91

I'd be happy to add this myself but might not get to it for day or twoish

vaikas:
Oh, absolutely, that's perfetto!

vaikas (approved):

Issue #1905, opened 2022-05-20
Signing images by digest that don't exist yet


imjasonh:
**Description**

It should be possible to sign an image specified by digest that doesn't (yet) exist in the registry, so long as the user has permission to push to the repository. As an example, I should be able to `cosign sign registry.biz/myuser/image@sha256:abcdef` even if no image with that digest exists yet.

Signing `registry.biz/myuser/image:tag` should still require that that tag exists, because `cosign` needs to be able to lookup the digest the tag points to. The `registry.biz/myuser` repository also must exist, and be accessible to the user, because that's where we'll push the signature.

Pre-signing images because an image build tool can write a signature to the repository _before_ the image being signed is pushed, for example if I enforce a policy that only signed images can exist in my registry. Signing after pushing introduces a race condition, especially if signing fails for some reason (e.g., I close the OIDC browser tab and walk away), leading to unsigned images.

This used to work in a previous version of `cosign`, but seems to have regressed (possibly https://github.com/sigstore/cosign/pull/1616 @puerco?). If we decide to fix this, we should introduce a test that guards against future regressions. Re #1616 we should also make sure that you can sign the same non-existent digest multiple times successfully.

Currently:

```
$ COSIGN_EXPERIMENTAL=1 cosign sign gcr.io/imjasonh/blah@sha256:$(head -n 10 /dev/urandom | sha256sum | cut -d' ' -f1)
Generating ephemeral keys...
Retrieving signed certificate...
Your browser will now be opened to:
https://oauth2.sigstore.dev/auth/auth?access_type=online&client_id=sigstore&code_challenge=REDACTED&code_challenge_method=S256&nonce=REDACTED&redirect_uri=http%3A%2F%2Flocalhost%3A58857%2Fauth%2Fcallback&response_type=code&scope=openid+email&state=REDACTED
Successfully verified SCT...
Error: signing [gcr.io/imjasonh/blah@sha256:7145ec233c4ebd21780be57aef96e3e18894d92a00755c3fd8974f5e82c244c8]: accessing image: entity not found in registry
main.go:52: error during command execution: signing [gcr.io/imjasonh/blah@sha256:7145ec233c4ebd21780be57aef96e3e18894d92a00755c3fd8974f5e82c244c8]: accessing image: entity not found in registry
```

thesayyn:
[...]

PR #2288, opened 2022-09-27
fix: allow signing of images that don't exist

thesayyn:
#### Summary

#1905 

#### Release Note

fix: allow signing of images that don't exist

#### Documentation

cc @imjasonh

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -22,6 +22,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
@@ -29,6 +30,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 
 	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioverifier"
@@ -44,6 +46,7 @@ import (
 	"github.com/sigstore/cosign/pkg/cosign/pkcs11key"
 	cremote "github.com/sigstore/cosign/pkg/cosign/remote"
 	"github.com/sigstore/cosign/pkg/oci"
+	ociempty "github.com/sigstore/cosign/pkg/oci/empty"
 	"github.com/sigstore/cosign/pkg/oci/mutate"
 	ociremote "github.com/sigstore/cosign/pkg/oci/remote"
 	"github.com/sigstore/cosign/pkg/oci/walk"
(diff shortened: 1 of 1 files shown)


cmd/cosign/cli/sign/sign.go
```diff
@@ -152,10 +155,23 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Regist
...
+			var se oci.SignedEntity
+
+			_, err := remote.Head(digest)
+			var te *transport.Error
+			if errors.As(err, &te) && te.StatusCode == http.StatusNotFound {
+				emptyimg, err := ociempty.SignedImage(ref)
+				if err != nil {
+					return fmt.Errorf("accessing image: %w", err)
+				}
+				se = emptyimg
```
imjasonh, line 167: Just checking: Is this going to mean that any attachments get hung off the digest of an empty image? Or does `SignedImage(digest)` handle the mapping of digest->attachments, and ignore the digest of its actual image contents?

An e2e test might help here, something that attaches some content to `@sha256:$(head -n /dev/urandom | sha256sum -)` and gets it back by that same random digest.

thesayyn: Yes, it will override any attachment images since it assumes that the image and all of the side images (sbom, sig, etc) are absent simply by doing a HEAD call for the image by digest. should it check for the presence of attachments as well to prevent the override?

thesayyn: Also, i am not sure where to put the e2e test. i have checked `test/e2e` folder but couldn't find a place where we assert the behavior with an ephemeral registry present. Could you or someone else put me in right direction?

imjasonh: > Yes, it will override any attachment images since it assumes that the image and all of the side images (sbom, sig, etc) are absent simply by doing a HEAD call for the image by digest. should it check for the presence of attachments as well to prevent the override?

Yeah I think that's right. I should be able to attach 2+ things to a non-existent image, e.g., a signature and an SBOM, or 2x signatures.

thesayyn: in that case `ociempty.SignedImage` is not going to cover. we need a better version of `ociremote.SignedEntity`

thesayyn: or adjust `ociremote.SignedEntity` in a way that doesn't fail when the image isn't present on the remote as you originally suggested.

thesayyn:
cc @imjasonh @dlorenc this is still relevant.

thesayyn:
@znewman01 can we add relevant labels to keep the bot away?

imjasonh:
Sorry for letting this slip.

The "good" news (or, news, anyway) is that cosign's OCI interactions are changing to support the OCI v1.1 referrers API, and I think we can simplify a lot of this code and registry interaction along the way, which will also make this change easier.

I've never liked the SignedEntity abstraction, and I think we can drop it. In v1.1 API mode, signing <some digest> (whether it exists or not) is as simple as pushing a manifest with `"subject": "<some digest>"`), and code already in cosign will be responsible for either also updating `:sha256-<some digest>`, or detecting that the registry will do it for us.

This might mean that signing non-existent things is only supported when in v1.1 mode, unless we want to untangle SignedEntity to support it. I think waiting for v1.1 mode to become the default is fine, personally.

PR #2295, opened 2022-09-28
bump mongo dep to avoid retracted version

imjasonh:
#### Summary

cosign depends on mongo via `github.com/go-openapi/strfmt`

mongo has retracted v0.10.0 due to a possible data corruption bug (that I don't believe affects cosign in any way at all), which means folks that `go get ./...` from the cosign repo, or any dependent of cosign, see this scary sounding warning:

```
go: warning: go.mongodb.org/mongo-driver@v1.10.0: retracted by module author: Contains a possible data corruption bug in RewrapManyDataKey when using libmongocrypt versions less than 1.5.2.
go: to switch to the latest unretracted version, run:
	go get go.mongodb.org/mongo-driver@latest
```

This bumps our dep on mongo-driver using:

```
go get -u github.com/go-openapi/strfmt@latest
```

...so the warning goes away.

see also https://github.com/ko-build/ko/issues/832

#### Release Note

NONE

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -267,7 +267,7 @@ require (
 	go.etcd.io/etcd/server/v3 v3.6.0-alpha.0 // indirect
 	go.etcd.io/etcd/tests/v3 v3.6.0-alpha.0 // indirect
 	go.etcd.io/etcd/v3 v3.6.0-alpha.0 // indirect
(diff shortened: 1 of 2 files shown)


cpanato:
@imjasonh ping :)

Issue #2294, opened 2022-09-28
[bug] verify-blob & rekor shards: search all shards for proposed entries


asraa:
**Description**

https://github.com/fluxcd/source-controller/issues/899#issuecomment-1261295343

After rekor sharded, the current Rekor tree is moved to a new one. 

When verifying blobs, we search the Rekor log proposed entries:
https://github.com/sigstore/cosign/blob/727e3e1d4af2856126f67c014d5bd2cf4b29a94a/pkg/cosign/tlog.go#L392

and expect that exactly **one** entry was retrieved.

Rekor CURRENTLY only searches the **CURRENT** tree for an entry, and will fail if the leaf is not present on the current log. That is causing the flux verify-blob failure after the shard rotation:
```
could not find entry in tlog: signature not found in transparency logError: verifying blob [./checksums.txt]: signature not found in transparency log
```

The entry did not exist on the current log.

As a fix (TBD), rekor will search ALL previous shards and may return more than one search entry. Because of this, the following changes need to happen in cosign:
1. Allow multiple entries returned from [SearchLogQuery](https://github.com/sigstore/cosign/blob/727e3e1d4af2856126f67c014d5bd2cf4b29a94a/pkg/cosign/tlog.go#L409)
2. In the cosign case, we need to start using a function like `tlogFindEarliestEntry` to replace https://github.com/sigstore/cosign/blob/727e3e1d4af2856126f67c014d5bd2cf4b29a94a/cmd/cosign/cli/verify/verify_blob.go#L410: the earliest entry with the earliest integrated time will always suffice for expiration checks.
3. In the container case, [these](https://github.com/sigstore/cosign/blob/727e3e1d4af2856126f67c014d5bd2cf4b29a94a/pkg/cosign/verify.go#L358-L384) funcs also need to be fixed, again, to handle multiple return entries, and selecting the earliest one for the certificate validation.

Note: old clients MAY be OK after the rekor fix: this will only affect users who are verifying blobs (or online lookups for containers) with cosign AND there were duplicate uploads on multiple shards. However, if there were re-uploads or duplicate entries across shards, old clients will break if rekor responds with multiple entries.

cc @haydentherapper @bobcallaway @priyawadhwa @lukehinds @stefanprodan 

**Version**

All

asraa:
Does anyone have some time to fix? If no one can volunteer I can "promise" my next free time goes to this, but... you know...

PR #2297, opened 2022-09-30
fix: make tlog entry lookups for online verification shard-aware

asraa:

Fixes https://github.com/sigstore/cosign/issues/2294 as described in the issue.


asraa:
The build here successfully verifies https://github.com/fluxcd/source-controller/issues/899

asraa:
adding a multiple rekor entry blob test real quick

asraa:
all set!

BTW are there any regression tests against OCI container verification with online lookups?

priyawadhwa (approved):
nice!

PR #2316, opened 2022-10-06
feat: add insecure option to name options

### LINKED ISSUE #2290: Support for insecure mode in verifier
cosign has a flag `--allow-insecure-registry` that set `InsecureSkipVerify=true` on in the underlying `tls` config. We would like to have another option that set the scheme to `http` the same ways `crane` does it with its `insecure` flag.

The primary use case is testing.

developer-guy:

Fixes #2290 

#### Summary

This PR will add an insecure option to the name options and use it while parsing the reference of the image name to set the _HTTP_ scheme.

#### Release Note

* adds an insecure option to sign & verify commands' name options to set the _HTTP_ scheme for insecure registries properly

### DIFF
--- a/cmd/cosign/cli/options/registry.go
+++ b/cmd/cosign/cli/options/registry.go
@@ -38,6 +38,7 @@ type Keychain = authn.Keychain
 // RegistryOptions is the wrapper for the registry options.
 type RegistryOptions struct {
 	AllowInsecure      bool
+	AllowHTTPRegistry  bool
 	KubernetesKeychain bool
 	RefOpts            ReferenceOptions
 	Keychain           Keychain
@@ -48,7 +49,10 @@ var _ Interface = (*RegistryOptions)(nil)
 // AddFlags implements Interface
 func (o *RegistryOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.AllowInsecure, "allow-insecure-registry", false,
-		"whether to allow insecure connections to registries. Don't use this for anything but testing")
+		"whether to allow insecure connections to registries (e.g., with expired or self-signed TLS certificates). Don't use this for anything but testing")
+
+	cmd.Flags().BoolVar(&o.AllowHTTPRegistry, "allow-http-registry", false,
+		"whether to allow using HTTP protocol while connecting to registries. Don't use this for anything but testing")
 
 	cmd.Flags().BoolVar(&o.KubernetesKeychain, "k8s-keychain", false,
 		"whether to use the kubernetes keychain instead of the default keychain (supports workload identity).")
@@ -69,6 +73,14 @@ func (o *RegistryOptions) ClientOpts(ctx context.Context) ([]ociremote.Option, e
 		opts = append(opts, ociremote.WithTargetRepository(targetRepoOverride))
 	}
 	return opts, nil
+}
+
+func (o *RegistryOptions) NameOptions() []name.Option {
+	var nameOpts []name.Option
+	if o.AllowHTTPRegistry {
+		nameOpts = append(nameOpts, name.Insecure)
+	}
+	return nameOpts
 }
 
 func (o *RegistryOptions) GetRegistryClientOpts(ctx context.Context) []remote.Option {
--- a/cmd/cosign/cli/attach/attach.go
+++ b/cmd/cosign/cli/attach/attach.go
@@ -36,7 +36,7 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, signed
 	}
 
 	for _, payload := range signedPayloads {
-		if err := attachAttestation(ociremoteOpts, payload, imageRef); err != nil {
+		if err := attachAttestation(ociremoteOpts, payload, imageRef, regOpts.NameOptions()); err != nil {
 			return fmt.Errorf("attaching payload from %s: %w", payload, err)
 		}
 	}
@@ -44,7 +44,7 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, signed
 	return nil
 }
 
-func attachAttestation(remoteOpts []ociremote.Option, signedPayload, imageRef string) error {
+func attachAttestation(remoteOpts []ociremote.Option, signedPayload, imageRef string, nameOpts []name.Option) error {
 	fmt.Fprintf(os.Stderr, "Using payload from: %s", signedPayload)
 	attestationFile, err := os.Open(signedPayload)
 	if err != nil {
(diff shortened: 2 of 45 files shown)


dlorenc (dismissed):

hectorj2f (approved):
Lgtm

developer-guy:
There are some other places that we need to do the same thing, I guess 🤔

developer-guy:
kindly ping @hectorj2f @dlorenc

znewman01 (changes requested):
I'm actually a little worried about this change.

Right now, `--allow-insecure-registry` ignores registries with bad/insecure TLS certs, but still connects over HTTPS. IIUC, the [`name.Insecure` option for go-containerregistry](https://pkg.go.dev/github.com/google/go-containerregistry@v0.11.0/pkg/name#Insecure) instead *forces* HTTP.

So (again, IIUC) after this change, the original use for `--allow-insecure-registry` will no longer work.

There's a couple ways to solve it that I can think of:

1. I'm totally wrong, in which case this is fine.
2. Add a new flag. This is pretty annoying in that it bloats the flags, but it's really explicit.
3. Allow users to specify that they want to use `http` by modifying the image ref. For instance, instead of `gcr.io/foo:latest` you could use `http://gcr.io/foo:latest`. This means we're implicitly making a new, spec-noncompliant form of image references which I'm a little skeptical about.
4. Magic: if `AllowInsecureRegistry` is passed, *cosign* could try to connect to the registry both ways and see which one works.

In any case, I think I'd like to see some manual testing done.

imjasonh:
I'd vote for (2) over (3). We have a similar problem in crane and ko, and just saying there should be a separate flag if you want http seems the best option. This also gives us a go option for http-but-not-also-bad-https.

Do we think we need a warning for a release or two before we change the semantics of "insecure registry"

imjasonh:
@jonjohnsonjr

jonjohnsonjr:
> spec-noncompliant

I'm 90% sure there's no spec 🙃 

Is there prior art for other clients that need to do this?

znewman01:
> Do we think we need a warning for a release or two before we change the semantics of "insecure registry"

Are we proposing that (for Cosign)? I think we can introduce a new flag without changing existing behavior, we just probably want better docs. Or are you talking GGCR?

imjasonh:
Sorry I misunderstood, you're right.

developer-guy:
kindly ping. What has to be done before merging this? 🙋🏻‍♂️

znewman01:
I think we want to separate this behavior into a distinct flag. So `--allow-insecure-registry` should continue to work exactly as it does *before* this PR, and the new flag (`--allow-http-registry` maybe?) would enable this behavior.

developer-guy:
that should be okay now @znewman01

cmd/cosign/cli/options/registry.go
```diff
@@ -50,6 +51,9 @@ func (o *RegistryOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.AllowInsecure, "allow-insecure-registry", false,
 		"whether to allow insecure connections to registries. Don't use this for anything but testing")
```
znewman01, line 52: Maybe worth distinguishing a little bit more now that we have a new flag:

> whether to allow insecure connections to registries (e.g., with expired or self-signed TLS certificates)...

znewman01 (changes requested):
I'm tempted to say that we should move this into `options/registry.go` like https://github.com/sigstore/cosign/blob/32f6f1ae4940806440a6b86d05928485f2e18ed2/cmd/cosign/cli/options/registry.go#L74

That'd save you a few lines of typing everywhere else. WDYT?

Also, I assume you see the lint errors and docgen error

znewman01 (approved):
Awesome, looks great!

bkumar08:
> I think we want to separate this behavior into a distinct flag. So `--allow-insecure-registry` should continue to work exactly as it does _before_ this PR, and the new flag (`--allow-http-registry` maybe?) would enable this behavior.

Hi All,
I have tried to test new flag  '--allow-http-registry' with insecure harbor registry with cosign prerelease **cosign_2.0.0.rc.0** . images are getting signed and pushed to insecure registry as expected with this flag. 
but i see a P3 bug on **cosign_2.0.0.rc.0**  version.
Issue:  --allow-http-registry flag is added under consig help 
Expected Behaviour:  --allow-http-registry flag should be added under cosign help
example:

# cosign version

GitVersion:    **v2.0.0-rc.0**
GitCommit:     a827922053c48283fb458b7b6bc51be3f477ec4d
GitTreeState:  clean
BuildDate:     2022-12-15T00:07:35Z
GoVersion:     go1.19.4
Compiler:      gc
Platform:      linux/amd64

**# cosign help**
A tool for Container Signing, Verification and Storage in an OCI registry.

Usage:
  cosign [command]

Available Commands:
  attach                  Provides utilities for attaching artifacts to other artifacts in a registry
  attest                  Attest the supplied container image.
  attest-blob             Attest the supplied blob.
  clean                   Remove all signatures from an image.
  completion              Generate completion script
  copy                    Copy the supplied container image and signatures.
  dockerfile              Provides utilities for discovering images in and performing operations on Dockerfiles
  download                Provides utilities for downloading artifacts and attached artifacts in a registry
  env                     Prints Cosign environment variables
  generate                Generates (unsigned) signature payloads from the supplied container image.
  generate-key-pair       Generates a key-pair.
  help                    Help about any command
  import-key-pair         Imports a PEM-encoded RSA or EC private key.
  initialize              Initializes SigStore root to retrieve trusted certificate and key targets for verification.
  load                    Load a signed image on disk to a remote registry
  login                   Log in to a registry
  manifest                Provides utilities for discovering images in and performing operations on Kubernetes manifests
  policy                  subcommand to manage a keyless policy.
  public-key              Gets a public key from the key-pair.
  save                    Save the container image and associated signatures to disk at the specified directory.
  sign                    Sign the supplied container image.
  sign-blob               Sign the supplied blob, outputting the base64-encoded signature to stdout.
  tree                    Display supply chain security related artifacts for an image such as signatures, SBOMs and attestations
  triangulate             Outputs the located cosign image reference. This is the location cosign stores the specified artifact type.
  upload                  Provides utilities for uploading artifacts to a registry
  verify                  Verify a signature on the supplied container image
  verify-attestation      Verify an attestation on the supplied container image
  verify-blob             Verify a signature on the supplied blob
  verify-blob-attestation Verify an attestation on the supplied blob
  version                 Prints the version

Flags:
  -h, --help                 help for cosign
      --output-file string   log output to a file
  -t, --timeout duration     timeout for commands (default 3m0s)
  -d, --verbose              log debug output

Additional help topics:
  cosign piv-tool                This cosign was not built with piv-tool support!
  cosign pkcs11-tool             This cosign was not built with pkcs11-tool support!

Use "cosign [command] --help" for more information about a command.

"--allow-http-registry" flag is not listed in above output.
Do we need create any bug to get address this issue. if yes then let me know will P3 bug for same.

Thanks !

znewman01:
It doesn't appear on the top-level `cosign help` output because not every command supports this flag.

If you check `cosign sign --help` for instance it should appear

Issue #2236, opened 2022-09-08
feat: Cosign env should show all possible environment variables.


todaywasawesome:
**Description**

I'm trying to track down all the possible environment variables as I learn how to use cosign but they're not listed anywhere. It's not uncommon in other tools that use environment variables to have that command show all values set and unset. 

So running `cosign env` would show something like:
```
COSIGN_PASSWORD=
...
```

This would make it a lot easier to learn how to use cosign.

asraa:
>  I believe there are some called libraries whose behavior is affected by environment variables: SOFTHSM_*, GOOGLE_SERVICE_ACCOUNT_NAME, etc. We can try to do those best effort perhaps. For https://github.com/sigstore/sigstore we can do a similar trick as in cosign.

Yes! The TUF client pkg exists in sigstore/sigstore, and the behavior there is also affected by env vars.

In the future, that will not be the case and all env vars will be at the top level in cosign. I have a TUF client redesign in the works, but it will likely not happen for the next 3 weeks.

xmudrii:
Is this still available? I’d like to try my luck contributing to cosign and this seems like a nice starter. 🙂

znewman01:
@xmudrii: 👍 

If you need guidance, let me know. My last comment shows a couple ways to get started.

xmudrii:
I created #2322 which combines easy and hard modes with the difference that I didn't create a new CLI command, but used `cosign env` (as described in the ticket description).

> ban Getenv in the codebase (with a linter?), make callers go through a common pkg/env.go interface to access any environment variables, and have a hardcoded allowlist there.

The PR should have accomplished everything but banning Getenv-related functions. I did some research on banning functions and it doesn't seem possible to ban functions natively with Go. Creating a linter is definitely an option and the workflow would be:
- create a linter based on golangci-lint requirements and guidelines
- open source the said linter
- integrate the linter with golangci-lint
  - A potential blocker here can be are golangci-lint folks willing to accept our linter. Otherwise, it doesn't seem too complicated to add a new linter, there's some work required, but it's not impossible
  - This might take some time to implement, integrate, and release, so I wouldn't block on this. If we agree to proceed with this, we can integrate linter later

> (Even better: parse all env vars for a given CLI command at the entry-point and pass them down through the call stack—IMHO Getenv in library code is a smell).

[...]

PR #2322, opened 2022-10-09
Implement 'env' package to handle environment variables in cosign 

xmudrii:
#### Summary

This PR implements a new `env` package to handle environment variables in cosign. This PR/implementation targets only `COSIGN_` environment variables, but it can be extended to handle other environment variables as well. This implementation is a combination of easy and hard mode described here: https://github.com/sigstore/cosign/issues/2236#issuecomment-1241050406

The `env` package implements:

- `Variable` type which is used to define an environment variable
- `VariableOpts` type which is used to describe an environment variable
- A slice with known environment variables and their descriptions (used for printing)
- `Getenv(Variable) string` function which is a wrapper around `os.Getenv` for given `Variable`
- `LookupEnv(Variable) (string, bool)` function which is a wrapper around `os.LookupEnv` for given `Variable`
- `PrintEnv()` function which is used to print all known variables, their values, and descriptions

All `os.Getenv`/`os.LookupEnv` calls on `COSIGN_` variables are replaced with `env.Getenv` and `env.LookupEnv`. There are some other `os.Getenv` calls on non-cosign variables, but those will be handled in a follow up PR (if this turns out to be an approach we want to follow).

This PR also modifies `cosign env` command with some additional features. First of all, the command is now using `env.PrintEnv()` to print the known cosign variables instead of searching Environ. However, we still search Environ for `SIGSTORE_` and `TUF_` variables. Additionally, **there are two new flags**:

- `--show-descriptions` (default `true`) -- used to enable descriptions for known cosign variables
- `--show-sensitive-values` (default `false`) -- by default, `cosign env` is going to show asterisk signs if variable is marked as sensitive. Turning this flag on will show values for sensitive variables

Example 1 (with descriptions and hidden sensitive values, default):

```shell
COSIGN_PASSWORD=test ./cosign env
```

```
# COSIGN_DOCKER_MEDIA_TYPES to be used with registries that do not support OCI media types
COSIGN_DOCKER_MEDIA_TYPES=""
# COSIGN_EXPERIMENTAL enables experimental cosign features
COSIGN_EXPERIMENTAL=""
# COSIGN_PASSWORD overrides password inputs with this value
COSIGN_PASSWORD="******"
# COSIGN_PKCS11_MODULE_PATH is PKCS11 module-path
COSIGN_PKCS11_MODULE_PATH=""
# COSIGN_PKCS11_PIN to be used if PKCS11 PIN is not provided
COSIGN_PKCS11_PIN=""
# COSIGN_REPOSITORY can be used to store signatures in an alternate location
COSIGN_REPOSITORY=""
```

Example 2 (with sensitive values):

```shell
COSIGN_PASSWORD=test ./cosign env --show-sensitive-values
```

```
# COSIGN_DOCKER_MEDIA_TYPES to be used with registries that do not support OCI media types
COSIGN_DOCKER_MEDIA_TYPES=""
# COSIGN_EXPERIMENTAL enables experimental cosign features
COSIGN_EXPERIMENTAL=""
# COSIGN_PASSWORD overrides password inputs with this value
COSIGN_PASSWORD="test"
# COSIGN_PKCS11_MODULE_PATH is PKCS11 module-path
COSIGN_PKCS11_MODULE_PATH=""
# COSIGN_PKCS11_PIN to be used if PKCS11 PIN is not provided
COSIGN_PKCS11_PIN=""
# COSIGN_REPOSITORY can be used to store signatures in an alternate location
COSIGN_REPOSITORY=""
```

One missing piece that is mentioned in the issue, but is not implemented, is a linter to forbid using `os.Getenv` and related functions. This is something that we need to discuss and investigate a bit more. Also, there are other environment variables such as `SIGSTORE_` and `TUF_` that are not handled, but this is something that can be discussed and handled later. I believe this PR should be a solid foundation for any further work on this topic.

Fixes part of https://github.com/sigstore/cosign/issues/2236

#### Release Note

- Implement `env` package used to register and handle environment variables used by cosign
- `cosign env` now prints all registered `COSIGN_` environment variables
- Add `--show-descriptions` (default `true`) flag to `cosign env` used to show description for each registered `COSIGN_` environment variable
- Add `--show-sensitive-values` (default `false`) flag to `cosign env` used to hidh values for `COSIGN_` environment variables marked as sensitive

#### Documentation

I'm not sure, but I guess mostly like yes.


xmudrii:
cc @znewman01 PTAL ^^ :slightly_smiling_face:

xmudrii:
I have an issue with running `make docgen`:

```
go run -tags pivkey,pkcs11key,cgo ./cmd/help/
# pkg-config --cflags  -- libpcsclite
Package libpcsclite was not found in the pkg-config search path.
Perhaps you should add the directory containing `libpcsclite.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libpcsclite' found
pkg-config: exit status 1
make: *** [Makefile:176: docgen] Error 2
```

Any idea how can I fix this? I'm running Ubuntu 22.04.

cpanato:
> I have an issue with running `make docgen`:
> 
> ```
> go run -tags pivkey,pkcs11key,cgo ./cmd/help/
> # pkg-config --cflags  -- libpcsclite
> Package libpcsclite was not found in the pkg-config search path.
> Perhaps you should add the directory containing `libpcsclite.pc'
> to the PKG_CONFIG_PATH environment variable
> No package 'libpcsclite' found
> pkg-config: exit status 1
> make: *** [Makefile:176: docgen] Error 2
> ```
> 
> Any idea how can I fix this? I'm running Ubuntu 22.04.

you need to install the `libpcsclite-dev` - `apt install libpcsclite-dev`

pkg/cosign/env/env.go
```diff
@@ -0,0 +1,120 @@
...
+			Sensitive:   false,
+		},
+		VariableRepository: {
+			Description: "can be used to store signatures in an alternate location",
+			Sensitive:   false,
+		},
+	}
+)
+
+func Getenv(name Variable) string {
```
znewman01, line 74: Should this enforce that `name` is in `environmentVariables`? IMO okay to `panic` if not. (Same for `LookupEnv`)

Also should it enforce that `name` starts with `COSIGN_`?

pkg/cosign/env/env.go
```diff
@@ -0,0 +1,120 @@
...
+	return os.Getenv(name.String())
+}
+
+func LookupEnv(name Variable) (string, bool) {
+	return os.LookupEnv(name.String())
+}
+
+func PrintEnv(showDescription, showSensitive bool) {
```
znewman01, line 82: Perhaps this should be moved to `cli/`? WDYT?

pkg/cosign/env/env.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+)
+
+type Variable string
+
+type VariableOpts struct {
```
znewman01, line 27: nit: types might be nice here so users know to pass `=1` or `=true` vs. `=my password`. You could do this directly in the descriptions too.

znewman01 (commented):
Great idea to redact sensitive information by default 👍

pkg/cosign/env/env.go
xmudrii: > Should this enforce that name is in environmentVariables? IMO okay to panic if not. (Same for LookupEnv)

I wasn't sure if it's okay to panic or not, but I think in this case it's acceptable. I'll add this change.

> Also should it enforce that name starts with COSIGN_?

I thought that we might want to use this mechanism for some non-cosign environment variables. Considering that we're not doing this right now, I'm going to enforce this and we can always remove it if needed.

pkg/cosign/env/env.go
xmudrii: The reason why I didn't want this in `cli/` is to avoid publicly exposing `environmentVariables` slice. This is because I want to avoid someone changing the slice outside the env package. If this is important, I can workaround it with some getter or I can make the slice public.

pkg/cosign/env/env.go
xmudrii: Great idea, will add it!

xmudrii:
@znewman01 I should have addressed all your comments. I'm trying to configure `forbidigo` and after that the PR should be ready.

xmudrii:
@znewman01 I added `forbidigo` to golangci-lint. There are some places where we still use `os.Getenv` with non-cosign environment variables. I handled that in two ways:

- if the file/package has many os.Getenv/LookupEnv calls, ignore that file or package to `.golangci.yml` file
- if the file/package has only one or few os.Getenv/LookupEnv calls, use in-line `//nolint:forbidigo` comment

Please take a look at this and let me know if this approach is okay or if I should change anything.

znewman01 (commented):
I'm confused about how verification will work. Verifiers need to reconstruct the entry. Do they have enough information to do so, and can we do so unambiguously?

Specifically:

- Every `timestamp_verification_data` in a given bundle should contain signatures for the *same* Rekor entry, right?
- If so, why is there a `KindData` in *each* `timestamp_verification_data`?
- There are cases where the artifact + bundle doesn't have the information we need to reconstruct the Rekor entry (e.g., if we only have RFC 3161 timestamps, there's no KindVersion), right?

And two broader questions. First, how entry-type-aware should applications/client libraries be? The Alpine Rekor entries include metadata from the PKGINFO. AFAICT, these are just unvalidated metadata and duplicate information from the artifact. To verify the Rekor signature, you'd need to parse the .apk file and format the entry yourself, which feels like a lot of work. Then, the signature lives inside the artifact.

I feel like I'm missing some context on the design philosophy of Rekor entries. Here's how I think about it:

- A Rekor entry contains:
  - some signed data (or a reference to such data)
  - the signature over that data (or the signature could be in theh artifact itself)
  - optionally with unverified metadata that can be used for searching.
- The entry can contain the data itself or just a hash; it can contain the signature itself, or the signature can live inside the data.
- Users can use the unverified metadata for searching, but it should be considered untrusted.
- After verification, users can trust the artifact along with *some* of the other metadata (exactly which metadata depends on the verification procedure used from the type).
- They still cannot trust the unsigned metadata, unless they reconstruct it from the artifact (at which point, they should just use the metadata from the artifact).

Why aren't we putting the whole entry into `bundle.content`? IIRC, two reasons. First, to avoid sticking unvalidated public keys in the bundle. Second, it's a lot of duplication with the VerificationMaterial. Are we better off fixing the Rekor entry types themselves to remove the VerificationMaterial? Rekor wouldn't be able to validate the entries, but it already can't, in a lot of cases.

--------------------------------------------------------------------------------

Sorry that I keep coming back with more issues, but fresh eyes always help me see a little better. Despite my comments I don't *actually* want to block this forever :smile:. But I also think it's critical to get this right. Could we push *something* marked experimental pretty soon (with a self-destruct clock, to prevent it from being entrenched), then try to build client libraries with it? I think that will help us shake out the rought edges.

Also, you may have noticed that I'm always going to err on the side of wanting more types. Feel free to push back on that!

.golangci.yml
```diff
@@ -41,6 +51,16 @@ issues:
     linters:
     - errcheck
     - gosec
+    - forbidigo
+  - path: pkg/cosign/env
+    linters:
+    - forbidigo
+  - path: pkg/providers/
```
znewman01, line 58: I actually think most of these are worth documenting. IMO the goal of `cosign env` is to ask "what from my environment is possibly affecting Cosign" or to ask "what knobs do I have to affect Cosign's behavior?"

(Same for `pkg/cosign/git/gitlab`.)

WDYT?

(I think it's reasonable to address that in a later PR, before closing the accompanying issue.)

cmd/cosign/cli/fulcio/fulcioverifier/ctl/verify.go
```diff
@@ -74,7 +74,7 @@ func ContainsSCT(cert []byte) (bool, error) {
 func VerifySCT(ctx context.Context, certPEM, chainPEM, rawSCT []byte) error {
 	// fetch SCT verification key
 	pubKeys := make(map[[sha256.Size]byte]logIDMetadata)
-	rootEnv := os.Getenv(altCTLogPublicKeyLocation)
+	rootEnv := os.Getenv(altCTLogPublicKeyLocation) //nolint:forbidigo
```
znewman01, line 77: IMO this is worth documenting.

pkg/cosign/env/env.go
znewman01: Good thought. I think it's prudent to avoid letting callers mutate that slice.

I think it's worth it to add a getter, though.

pkg/cosign/env/env_test.go
```diff
@@ -0,0 +1,214 @@
...
+package env
+
+import (
+	"io"
+	"os"
+	"testing"
+)
+
+func TestGetenv(t *testing.T) {
+	os.Setenv(VariableExperimental.String(), "1")
```
znewman01, line 25: Nit: TBH for a test, I think it might be more readable to hardcode `COSIGN_EXPERIMENTAL`. It makes it clearer what you're testing.

I'll go with your judgment here, though. WYDT?

pkg/cosign/tlog.go
```diff
@@ -110,7 +110,7 @@ func intotoEntry(ctx context.Context, signature, pubKey []byte) (models.Proposed
 // TODO: Rename SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY to be test-only or remove.
 func GetRekorPubs(ctx context.Context, rekorClient *client.Rekor) (map[string]RekorPubKey, error) {
 	publicKeys := make(map[string]RekorPubKey)
-	altRekorPub := os.Getenv(altRekorPublicKey)
+	altRekorPub := os.Getenv(altRekorPublicKey) //nolint:forbidigo
```
znewman01, line 113: IMO this is worth including.

cmd/cosign/cli/env.go
```diff
@@ -21,29 +21,41 @@ import (
...
 			return nil
 		},
 	}
+
+	o.AddFlags(cmd)
+	return cmd
 }
 
-func getEnv() []string {
+func getExternalEnv() []string {
```
znewman01, line 53: I think it's worth folding this into `PrintEnv`, there are some subtle scenarios to test.

cmd/cosign/cli/env.go
```diff
@@ -21,29 +21,41 @@ import (
...
+	cmd := &cobra.Command{
 		Use:   "env",
 		Short: "Prints Cosign environment variables",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			for _, e := range getEnv() {
+			// Print COSIGN_ environment variables
+			env.PrintEnv(o.ShowDescriptions, o.ShowSensitiveValues)
+
+			// Print external environment variables (SIGSTORE_ and TUF_)
```
znewman01, line 40: Maybe worth printing a comment about these (`# These are environment variables that Cosign doesn't specifically know about, but with prefixes that look like they might be used by some of Cosign's dependencies` or something)

cmd/cosign/cli/env.go
```diff
@@ -21,29 +21,41 @@ import (
...
+			for _, e := range getExternalEnv() {
 				fmt.Println(e)
```
znewman01, line 42: This is just printing the name, right? Want to avoid sensitive values I think.

cmd/cosign/cli/env.go
```diff
@@ -21,29 +21,41 @@ import (
...
 }
 
-func getEnv() []string {
+func getExternalEnv() []string {
 	out := []string{}
 	for _, e := range os.Environ() {
 		// Prefixes to look for. err on the side of showing too much rather
 		// than too little. We'll only output things that have values set.
 		for _, prefix := range []string{
-			"COSIGN_",
```
znewman01, line 46: I wonder if you want to keep `COSIGN_`-prefixed variables, and do something like `if strings.HasPrefix(e, prefix) && e not in env.allEnvVars`? (<--pseudocode)

internal/pkg/cosign/fulcio/fulcioroots/fulcioroots.go
```diff
@@ -62,7 +62,7 @@ func initRoots() (*x509.CertPool, *x509.CertPool, error) {
 	// intermediatePool should be nil if no intermediates are found
 	var intermediatePool *x509.CertPool
 
-	rootEnv := os.Getenv(altRoot)
+	rootEnv := os.Getenv(altRoot) //nolint:forbidigo
```
znewman01, line 65: IMO worth documenting

pkg/cosign/env/env.go
```diff
@@ -0,0 +1,141 @@
...
+	VariableExperimental     Variable = "COSIGN_EXPERIMENTAL"
+	VariableDockerMediaTypes Variable = "COSIGN_DOCKER_MEDIA_TYPES"
+	VariablePassword         Variable = "COSIGN_PASSWORD"
+	VariablePKCS11Pin        Variable = "COSIGN_PKCS11_PIN"
+	VariablePKCS11ModulePath Variable = "COSIGN_PKCS11_MODULE_PATH"
+	VariableRepository       Variable = "COSIGN_REPOSITORY"
+)
+
+var (
+	environmentVariables = map[Variable]VariableOpts{
```
znewman01, line 47: One thing I'm noticing: this design centralizes the environment variables.

You could also consider a registry, where various parts of Cosign would register their own environment variables. (Closer to where they're used). I think *many* of these are used *a lot of places*, so that approach may not be great :shrug: 

I think this is fine as-is, I just want to make a note so that this decision is explicit.

pkg/cosign/env/env.go
```diff
@@ -0,0 +1,141 @@
...
+			Sensitive:   false,
+		},
+	}
+)
+
+func mustRegisterEnv(name Variable) {
+	if _, ok := environmentVariables[name]; !ok {
+		panic(fmt.Sprintf("environment variable %q is not registered in pkg/cosign/env", name.String()))
+	}
+	if !strings.HasPrefix(name.String(), "COSIGN_") {
```
znewman01, line 85: Welp, now that I've asked you to Enforce the `COSIGN_` prefix we might want to start doing `SIGSTORE_` and `TUF_` variables. Can you allowlist all three prefixes? You could also get rid of this altogether in favor of the "is in `environmentVariables`" check

I think in the medium-to-long-term we probably want to do something similar to this pattern in the upstream sigstore/TUF repositories, and then we could really go back to only having `COSIGN_` variables (we'd call `sigstore/pkg/env.LookupEnv` where needed).

pkg/cosign/env/env.go
```diff
@@ -0,0 +1,141 @@
...
+		panic(fmt.Sprintf("environment varialbe %q must start with COSIGN_ prefix", name.String()))
```
znewman01, line 86: Typo: variable

pkg/cosign/env/env_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+package env
+
+import (
+	"io"
+	"os"
+	"testing"
+)
+
+func TestGetenv(t *testing.T) {
```
znewman01, line 24: Can we add some tests that `recover()` from the panics of invalid variables?

cmd/cosign/cli/env.go
znewman01: (if `o.ShowDescriptions` is set)

pkg/cosign/tlog.go
```diff
@@ -150,7 +150,7 @@ func GetRekorPubs(ctx context.Context, rekorClient *client.Rekor) (map[string]Re
 
 	// If we have a Rekor client and we've been told to fetch the Public Key from Rekor,
 	// additionally fetch it here.
-	addRekorPublic := os.Getenv(addRekorPublicKeyFromRekor)
+	addRekorPublic := os.Getenv(addRekorPublicKeyFromRekor) //nolint:forbidigo
```
znewman01, line 153: Same here.

pkg/cosign/git/github/github.go
```diff
@@ -44,7 +44,7 @@ func New() *Gh {
 
 func (g *Gh) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) error {
 	var httpClient *http.Client
-	if token, ok := os.LookupEnv("GITHUB_TOKEN"); ok {
+	if token, ok := os.LookupEnv("GITHUB_TOKEN"); ok { //nolint:forbidigo
```
znewman01, line 47: IMO worth documenting

pkg/blob/load.go
```diff
@@ -53,7 +53,7 @@ func LoadFileOrURL(fileRef string) ([]byte, error) {
 			}
 		case "env://":
 			envVar := parts[1]
-			value, found := os.LookupEnv(envVar)
+			value, found := os.LookupEnv(envVar) //nolint:forbidigo
```
znewman01, line 56: This is a legitimate exception, because it's user-provided, nice find.

Can we add a justification for the lint skip?

```
// Most of Cosign should use `env.LookupEnv` (see #2236) to restrict us to known environment variables (usually `$COSIGN_*`.
// However, in this case, `envVar` is user-provided and not one of the allow-listed env vars.
```

pkg/cosign/env/env.go
znewman01: > I'm going to enforce this and we can always remove it if needed.

+1 to that thought

znewman01 (changes requested):
This is a really great first contribution to Cosign, thank you so much!

Some changes requested, mostly minor.

xmudrii:
@znewman01 Thank you so much for the review! I agree with all those comments. It’s already late for me (I’m in the CEST time zone), so I’ll address those comments beginning of the next week if that’s okay. 🙂

znewman01:
> @znewman01 Thank you so much for the review! I agree with all those comments. It’s already late for me (I’m in the CEST time zone), so I’ll address those comments beginning of the next week if that’s okay. 🙂

No rush! Have a great weekend

.golangci.yml
xmudrii: > I actually think most of these are worth documenting. IMO the goal of cosign env is to ask "what from my environment is possibly affecting Cosign" or to ask "what knobs do I have to affect Cosign's behavior?"

@znewman01 Just to make sure I have a clear understanding -- do you mean that I should also add those variables to the slice of known/registered variables (`environmentVariables` in the `env` package)? As a result of that, `cosign env` is also going to print variables like `GITHUB_TOKEN`, `GITLAB_TOKEN`, etc...

In that case, we have to remove checking if the variable starts with `COSIGN_` (https://github.com/sigstore/cosign/pull/2322#discussion_r991726750), but I think it makes sense to do that.

znewman01: Precisely, I think you should add them all to `environmentVariables`.

(And yes, I think the `COSIGN_` prefix check doesn't make sense anymore. Sorry I asked you to add it and now am asking you to take it away!)

cmd/cosign/cli/env.go
xmudrii: No, this is printing `key=value`. We can't really tell if it's sensitive or not in this case.

pkg/cosign/env/env.go
xmudrii: This is now moved to `cli/`.

.golangci.yml
xmudrii: Fixed.

cmd/cosign/cli/env.go
xmudrii: This is done, I'm just not sure if I tested everything expected.

cmd/cosign/cli/env.go
xmudrii: That's a good point, I added something similar.

cmd/cosign/cli/env.go
xmudrii: This is a great idea, I added something like that.

pkg/cosign/env/env.go
xmudrii: I went with some in-between approach here. I added a new field to `VariableOpts` -- `External`. If variable is not external (`External: false`), then it must start with `COSIGN_`, otherwise we don't care.

pkg/cosign/env/env_test.go
xmudrii: Added.

pkg/blob/load.go
xmudrii: Added.

xmudrii:
@znewman01 I think that I should have addressed all comments. Please let me know if I forgot something or if you have any additional comments. I tried to document all variables as best as I could, but I'm not sure have I got all descriptions and expectations correctly, so please take a look at that.

cmd/cosign/cli/env_test.go
```diff
@@ -16,22 +16,204 @@
...
+			showSensitiveValues: false,
+			expectedOutput:      expectedWithNonRegisteredEnvNoDesc,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Set needed environment variables
+			for k, v := range tt.environmentVariables {
+				os.Setenv(k, v)
```
znewman01, line 196: Will this test run into trouble if I have `COSIGN_FOO=1` set when I run the tests?

cmd/cosign/cli/env_test.go
```diff
@@ -16,22 +16,204 @@
...
+
+	expectedWithNonRegisteredEnv = `# COSIGN_TEST1 is the first test variable
+# Expects: test1 value
+COSIGN_TEST1="abcd"
+# COSIGN_TEST2 is the second test variable
+# Expects: test2 value
+COSIGN_TEST2=""
+# Environment variables below are not registered with cosign,
+# but might still influence cosign's behavior.
+COSIGN_TEST3=abcd
```
znewman01, line 69: Hm...I think the safe thing to do is to treat everything as sensitive unless told otherwise (`--show-sensitive`).

I could just `export COSIGN_MY_ROOT_PASSWORD=hunter2` and it'd get printed out. That's *kind of* my fault, but 🤷. You could imagine a situation where `sigstore/sigstore` adds a new variable for a password and we don't put it into our map by the next Cosign release, which would be actually bad.

It's a little bit user-unfriendly but definitely safe.

znewman01, line 69: Do we want quotes here? `COSIGN_TEST3="abcd"`

cmd/cosign/cli/env.go
znewman01: I should have written down what I meant, ha. You covered everything I can think of, *except* if we want to make external environment variables "sensitive" by default. Then we need a test case with `--show-sensitive` and one without

cmd/cosign/cli/env_test.go
znewman01: Or get rid of quotes for the other variables. `env` doesn't show quotes, even if the variable has a space in it. 

```shell
$ env
env | head -n 3
TERM_SESSION_ID=w1t0p0:674E0F26-D173-4E6D-A95C-F26D8DEE667D
SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.g6K1fscsWe/Listeners
LC_TERMINAL_VERSION=3.4.15
ITERM_PROFILE=Hotkey Window
```

Don't feel strongly either way but consistency is nice 🙂

pkg/cosign/env/env.go
znewman01: nice!

znewman01 (changes requested):
So close! This is looking great. Just a couple small things

cmd/cosign/cli/env_test.go
xmudrii: Yes, it will. This is a great point and something that I missed.

I tried to solve this, I'm not super proud of the fix, but it's in this commit: https://github.com/sigstore/cosign/pull/2322/commits/5c8b0b0af40927cf4b95d0ad767a78158311892a

It's not the best way to solve it, and I'm open to reverting that commit if you don't like the solution.

cmd/cosign/cli/env_test.go
xmudrii: I agree it's definitely safer. It should be fixed now. I also added a few test cases for this.

xmudrii: I also think consistency is nice. I removed quotes, it should be the same as for `env` and `os.Environ()` in Go.

xmudrii:
@znewman01 I addressed your comments, PTAL when you have some time. :slightly_smiling_face:

znewman01 (dismissed):

cmd/cosign/cli/env_test.go
znewman01: :shipit:

znewman01:
Small lint issue, once you fix that I think we can merge!

xmudrii:
@znewman01 Fixed. :slightly_smiling_face:

znewman01 (approved):

znewman01:
Nice, if tests pass this should merge automatically!

Thank you so much! 🙏  Great first PR, I wish everybody sent one like this 😄

xmudrii:
@znewman01 I'm so happy to see this PR merged! Thanks a lot for going through this with me, I really appreciate it!

PR #2332, opened 2022-10-12
Add --output-attestation flag to attest-blob and remove experimental signing

priyawadhwa:
We can add in experimental signing later, but for now I want to stick with just the `--key` flag so it's easier to add comprehensive tests

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -40,7 +40,7 @@ import (
 )
 
 // nolint
-func AttestBlobCmd(ctx context.Context, ko options.KeyOpts, artifactPath string, artifactHash string, certPath string, certChainPath string, predicatePath string, predicateType string, timeout time.Duration, outputSignature string) error {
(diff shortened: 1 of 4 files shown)


dlorenc (approved):

haydentherapper:
Hey @priyawadhwa, this seems like a breaking change to drop keyless functionality. If we end up cutting releases before all experimental flags are removed (which I think we did already), then this is going to cause issues for users who expected keyless signing to work. I'm also a little concerned that logic might change in between dropping keyless functionality and adding it back - I'm particularly concerned about introducing vulnerabilities accidentally in verification.

Can we create an issue tracking the removal of the experimental flag for each subcommand?

haydentherapper:
I've realized this is a new command, not an existing one, so ignore the first part of the comment. I think we should still discuss more about removing the experimental flag. Also, somewhat related, I think we should figure out how to make sure all sign/verify commands offer the same set of features.

priyawadhwa:
> Also, somewhat related, I think we should figure out how to make sure all sign/verify commands offer the same set of features.

yep, agreed. i think the code needs a lot of refactoring, and i didn't want to make mistakes with this new command trying to add in every feature at once. so i started with just signing with keys, and planned on adding in other things one at a time (also to make it easier to review!). +1 to creating an issue for removing the experimental flag!

Issue #2330, opened 2022-10-12
allow setting flag values via environment variables


priyawadhwa:
When I'm using staging I end up having to pass in flags for rekor, fulcio, oauth, every time I use cosign. It would be nice to be able to set flag values with env vars instead so I don't have to keep repeating flags. I think viper supports something like this but have to look into it!

PR #2346, opened 2022-10-17
Bind flags as prefixed env variables

hectorj2f:

#### Summary

There is an increase in the amount of flags used in our commands. Therefore we should think about binding these flags as environment variables that would help us when using the commands. 

closes https://github.com/sigstore/cosign/issues/2330

#### Release Note

`Allow setting flag values via environment variables`

### DIFF
--- a/cmd/cosign/cli/options/root.go
+++ b/cmd/cosign/cli/options/root.go
@@ -16,10 +16,16 @@
 package options
 
 import (
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
 )
+
+const EnvPrefix = "COSIGN"
 
 // RootOptions define flags and options for the root cosign cli.
 type RootOptions struct {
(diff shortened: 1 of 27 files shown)


cmd/cosign/cli/options/root.go
```diff
@@ -45,3 +51,27 @@ func (o *RootOptions) AddFlags(cmd *cobra.Command) {
...
+			_ = v.BindEnv(f.Name, flagToEnvVar(f.Name))
+		}
+		if !f.Changed && v.IsSet((f.Name)) {
+			val := v.Get(f.Name)
+			_ = cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val))
+		}
+	})
+}
+
+func flagToEnvVar(f string) string {
```
priyawadhwa, line 74: Could you add a small unit test for these functions?

priyawadhwa (commented):
This looks great @hectorj2f !!

cmd/cosign/cli/options/root.go
hectorj2f: Done :)

cpanato (approved):
looks cool

PR #2357, opened 2022-10-19
switch to reusable workflow for license/vuln scan

bobcallaway:

### DIFF
--- a/.github/workflows/depsreview.yml
+++ b/.github/workflows/depsreview.yml
@@ -20,9 +20,5 @@ permissions:
 
 jobs:
   dependency-review:
-    runs-on: ubuntu-latest
-    steps:
-      - name: 'Checkout Repository'
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8  # v3
-      - name: 'Dependency Review'
-        uses: actions/dependency-review-action@fd675ced9c17f1393071e1a2e685ab527e585a0c # v2
+    name: License and Vulnerability Scan
+    uses: sigstore/community/.github/workflows/reusable-dependency-review.yml@9b1b5aca605f92ec5b1bf3681b1e61b3dbc420cc


dlorenc (approved):

.github/workflows/depsreview.yml
```diff
@@ -20,9 +20,5 @@ permissions:
 
 jobs:
   dependency-review:
-    runs-on: ubuntu-latest
-    steps:
-      - name: 'Checkout Repository'
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8  # v3
-      - name: 'Dependency Review'
-        uses: actions/dependency-review-action@fd675ced9c17f1393071e1a2e685ab527e585a0c # v2
+    name: License and Vulnerability Scan
```
hectorj2f, line 23: q: shouldn't this be under steps ?

hectorj2f: it seems there is a problem with the github status. It ran how you can see here: https://github.com/sigstore/cosign/actions/runs/3281401375/jobs/5403359684

bobcallaway:
What’s the problem? I’m not seeing what you’re referring to on quick
glance.

hectorj2f:
It seems the dependency-review status is not reported back, although it seems the action ran successfully.

Screenshot 2022-10-19 at 15 02 17

.github/workflows/depsreview.yml
asraa: Reusable workflows can't be invoked inside steps; I've had somewhat similar problems on a Required status check that expects a particular job name and status to be posted.

You wil likely need to change the required status check in the branch protection settings to actually be from this one:
Dependency Review / License and Vulnerability Scan / Scan dependencies for license compliance and vulnerabilities (pull_request) Successful in 5s

Consider using an id name for the jobs inside the reusable workflow, I think that also complicates it

cpanato:
ok looks like the issue is the github check have a wrong name

.github/workflows/depsreview.yml
hectorj2f: @asraa Thanks for the clarification.

Issue #2367, opened 2022-10-20
Cosign doesn't use the correct CycloneDX predicate


lehors:
**Description**

The CycloneDX predicate defined in in-toto-golang has changed and cosign needs to move to the latest version to be compatible with other tools like syft. See https://github.com/in-toto/in-toto-golang/pull/188

**Version**

GitVersion:    1.12.0
GitCommit:     8483d6c71f153f38f237ba79c88d0fda6306e6e3
GitTreeState:  "clean"
BuildDate:     2022-09-14T15:41:15Z
GoVersion:     go1.19.1
Compiler:      gc
Platform:      darwin/amd64

developer-guy:
Yeah, AFAIK, @puerco already knows that problem, and will fix it soon 🙋🏻‍♂️

lehors:
Fixed by PR #2368

PR #2368, opened 2022-10-20
Bump in-toto-golang dependency to fix incompatibility issue

lehors:
#### Summary

cosign is incompatible with other tools like syft because it doesn't use the correct CycloneDX predicate.
This changes the in-toto-golang dependency to the latest version so that cosign uses the correct value of the CycloneDX predicate.

This addresses Issue #2367

#### Release Note

Update in-toto-golang dependency to get the correct CycloneDX predicate

#### Documentation

N.A.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/google/go-containerregistry v0.11.0
 	github.com/google/go-github/v45 v45.2.0
-	github.com/in-toto/in-toto-golang v0.3.4-0.20220709202702-fa494aaa0add
+	github.com/in-toto/in-toto-golang v0.4.1-0.20221018183522-731d0640b65f
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/manifoldco/promptui v0.9.0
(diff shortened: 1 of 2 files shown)


developer-guy (approved):
lgtm

hectorj2f (approved):
thanks 💪🏻

adityasaky:
Looks like `go get` arbitrarily assigns a version number when pulling from a commit since the last tag. We just tagged v0.5.0 which includes this change, so it should probably be bumped again?

lehors:
> Looks like `go get` arbitrarily assigns a version number when pulling from a commit since the last tag. We just tagged v0.5.0 which includes this change, so it should probably be bumped again?
> 
> https://github.com/in-toto/in-toto-golang/releases/tag/v0.5.0

The version numbers assigned by go get aren't arbitrary. Go get assigns pseudo version numbers based on the commit number and related timestamp in a systematic manner. But I'm happy to bump the dependency again to use the official version.

lehors:
Ok, I see @puerco did it so we're all set.

PR #2370, opened 2022-10-20
Replace accidentally-removed public constants

znewman01:
These were removed in #2322, which made it an accidentally-semver-breaking change. This unbreaks semver.

These should be removed at the next major version release or sooner if we decide keeping semver is not important.

Signed-off-by: Zachary Newman <zjn@chainguard.dev>

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/pkg/oci/mediatypes.go
+++ b/pkg/oci/mediatypes.go
@@ -21,6 +21,11 @@ import (
 	"github.com/sigstore/cosign/pkg/cosign/env"
 )
 
+const (
+	// Deprecated: use `pkg/cosign/env/VariableDockerMediaTypes` instead.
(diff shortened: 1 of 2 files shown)


hectorj2f (approved):

dlorenc:
If no one complained it's probably fine to leave them out IMO.

znewman01:
> If no one complained it's probably fine to leave them out IMO.

It hasn't shipped yet so nobody had an opportunity to complain  😛 

My preference is to leave these in for now, but [adopt a policy around versioning](https://github.com/sigstore/cosign/discussions/2365) that let's us make breaking changes. Then we can make it clear in our [versioning docs](https://docs.sigstore.dev/api-stability/) and then all bets are off.

Issue #2138, opened 2022-08-05
verify-blob can't verify DSSE blobs with COSIGN_EXPERIMENTAL


wlynch:
**Description**

@eddiezane and I stumbled on this today - verify-blob has logic for handling DSSE messages 

https://github.com/sigstore/cosign/blob/128f8fbd5f1bed0f7f1069d01c7e33cd52f6c381/cmd/cosign/cli/verify/verify_blob.go#L234-L237

but when this is combined with COSIGN_EXPERIMENTAL, it fails:

```sh
$ cat test.json.sig 
{"payloadType":"application/vnd.in-toto+json","payload":"eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjAuMSIsInByZWRpY2F0ZVR5cGUiOiIiLCJzdWJqZWN0IjpbeyJuYW1lIjoiIiwiZGlnZXN0Ijp7InNoYTI1NiI6IjVmMzY0OTgwZjc0NDAwZjg3NzNmODJjZjU4ZWM3ODQ0YTExZDI3Y2YifX1dLCJwcmVkaWNhdGUiOnsibmFtZSI6InVuaXQtdGVzdC0xIiwic3RhdHVzIjoic3VjY2VzcyJ9fQ==","signatures":[{"keyid":"","sig":"MEUCIEha6COOH9fAeqZE36P2nlpmT5dmN/DQdi6ffhdEI5s1AiEAlVTKIYxPAQLxbTUBjOIdFJLzkLf9wgSGH2VuTdx4k9U="}]}

$ COSIGN_EXPERIMENTAL=1 cosign verify-blob --signature test.json.sig test.json.sig --cert=<(rekor-cli get --log-index 3121421 --format json | jq -r .Body.IntotoObj.publicKey | base64 -d)
Error: verifying blob [test.json.sig]: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest  &{Code:400 Message:verifying signature: failed to verify signature: %!w(<nil>)}
main.go:62: error during command execution: verifying blob [test.json.sig]: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest  &{Code:400 Message:verifying signature: failed to verify signature: %!w(<nil>)}

[...]

PR #2389, opened 2022-10-26
[verify-blob] deprecate DSSE support in verify-blob

asraa:

This only worked when the bundle was provided or if this was not a keyless signature. https://github.com/sigstore/cosign/issues/2138

This *should* be supported in verify-blob-attestation. @priyawadhwa want me to take a quick stab at adding keyless?


priyawadhwa:
> This should be supported in verify-blob-attestation. @priyawadhwa want me to take a quick stab at adding keyless?

@asraa that would be awesome, thank you!

haydentherapper:
Removing label since this is scheduled for 2.0

asraa:
Going to reopen a second clearer PR

PR #2393, opened 2022-10-31
verify: fetch rekor public keys before verification

asraa:

Removes calls to `GetRekorPubs` in `pkg/`. This fetches the rekor public keys before verification, making verification commands easier to test in other environments.

TODO: verify blob is currently all in cli/, so there are still some internal TUF calls there. TBD.

**THIS SHOULD BE MERGED INTO A v2 BRANCH**

@vaikas 

Release note:
```
* feat!: Updates function signature for `cosign.GetRekorPubs` to remove unncessary `RekorClient` parameter.
* feat: Adds a field `RekorPubKeys` in `CheckOpts` that configured the trusted rekor public keys needed for verifying tlog entries. This may be populated with the public TUF instance using `cosign.GetRekorPubs`.
* breaking change: Updates function signatures for `VerifyTLogEntry` and `VerifyBundle` to take a map of trusted Rekor public keys to verify against. This may be populated with the public TUF instance using `cosign.GetRekorPubs`.
```


pkg/cosign/tlog.go
```diff
@@ -402,9 +427,9 @@ func FindTLogEntriesByPayload(ctx context.Context, rekorClient *client.Rekor, pa
 	return searchIndex.GetPayload(), nil
 }
 
-// VerityTLogEntry verifies a TLog entry.
+// VerityTLogEntry verifies a TLog entry against the trusted map of rekorPubKeys.
```
vaikas, line 430: nit: What's the key in the Map?

pkg/cosign/tlog.go
```diff
@@ -138,6 +138,27 @@ func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey,
 	return publicKeys, nil
 }
 
+// rekorPubsFromClient returns a RekorPubKey keyed by the log ID from the Rekor client.
+// NOTE: This should **not** be used in the verification path, but may be used in the
```
vaikas, line 142: nit: 'should not' or 'must not'? I'm just curious with the emphasis on the not, if we should just make it a strong statement to avoid misuse?

pkg/cosign/tlog.go
```diff
@@ -138,6 +138,27 @@ func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey,
...
+// sign path to validate return responses are consistent from Rekor.
+func rekorPubsFromClient(rekorClient *client.Rekor) (map[string]RekorPubKey, error) {
+	publicKeys := make(map[string]RekorPubKey)
+	pubOK, err := rekorClient.Pubkey.GetPublicKey(nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch rekor public key from rekor: %w", err)
+	}
+	pubFromAPI, err := PemToECDSAKey([]byte(pubOK.Payload))
```
vaikas, line 150: this is just like L103-111, wondering if that could be hoisted out to a function that we can tell call from here (and hopefully remove the ENV stuff from GetRekorPubs but do the equivalent call to it from 'cli' portion instead). I assume at least that would be desirable?
We can do it in a follow up, but just curious if that even makes sense?

pkg/cosign/verify.go
```diff
@@ -873,7 +884,9 @@ func CheckExpiry(cert *x509.Certificate, it time.Time) error {
 	return nil
 }
 
-func VerifyBundle(ctx context.Context, sig oci.Signature, rekorClient *client.Rekor) (bool, error) {
+// This verifies an offline bundle contained in the sig against the trusted
```
vaikas, line 887: nit: what's the key in the map.

vaikas: Note, I only jotted it down (or tried to) in the public methods.

vaikas (dismissed):
Thank you so much!! Looking so much lovelier with this refactor :)

pkg/cosign/tlog.go
```diff
@@ -91,7 +91,7 @@ func intotoEntry(ctx context.Context, signature, pubKey []byte) (models.Proposed
 // There are two Env variable that can be used to override this behaviour:
 // SIGSTORE_REKOR_PUBLIC_KEY - If specified, location of the file that contains
 // the Rekor Public Key on local filesystem
-func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey, error) {
+func GetRekorPubs(ctx context.Context) (map[string]RekorPubKey, error) {
```
znewman01, line 94: This is the only reason it's a breaking change, right? I'll bring this up in the community meeting tomorrow but I'd love an official decision that Cosign's library API isn't subject to semver guarantees, then we wouldn't have to wait for v2 I think

pkg/cosign/verify.go
znewman01: This is a good point. That suggests to me it's worth wrapping the `map[string]RekorPubKey` type into a newtype or struct

znewman01 (commented):
+1 to Ville's comments, LGTM otherwise

pkg/cosign/verify.go
asraa: > map[string]RekorPubKey type into a newtype or struct

Right: I'm not even sure if it's the right type to expose to the `CheckOpts` struct, since it's fairly opaque to users. Thinking about this again, maybe it's more worth it to make a more user-friendly constructor for this type? Would address Villie's comment above on repeated code. Lemme try that out

pkg/cosign/tlog.go
asraa: There's also `VerifyBundle` and `VerifyTLogEntry`: I really did want to make this change at some point (I can leave as TODOs though): it's really odd that those verification func signatures currently require an online rekor client: they only require a trusted rekor key.

priyawadhwa:
@asraa we decided to branch off 1.0 into the `1.0-fork` branch, so you can merge this into `main` whenever you're ready!

pkg/cosign/tlog.go
haydentherapper: Given the state of the library part of cosign, I'm not overly concerned with getting this in now vs 2.0.

pkg/cosign/tlog.go
asraa: must not, IMO. Yeah: I made this a private func on purpose so that users wouldn't be able to do this (but on the other hand, it would be useful for testing with a custom Rekor). updated the comment.

pkg/cosign/tlog.go
asraa: addressed in commit, but only to factor out the logic and make it easier to construct a map of trusted rekor pub keys outside TUF

pkg/cosign/verify.go
asraa: done

asraa:
Addressed comments, let me know if there's more

pkg/cosign/tlog.go
```diff
@@ -57,6 +57,13 @@ type RekorPubKey struct {
 	Status tuf.StatusKind
 }
 
+// This is a map of RekorPubKeys indexed by log ID that's used in verification
```
vaikas, line 60: Love it. Another approach we could've taken is just to define a new type without an embedded Keys in it, something like:
type TrustedRekorPubKeys map[string]RekorPubKey

And that would've provided us with the documentation, but this is totes fine with me. By doing it this way in case there's something else that we may need in the future, we can add it easier here :)

pkg/cosign/verify_test.go
```diff
@@ -248,9 +250,17 @@ func TestVerifyImageSignatureWithNoChain(t *testing.T) {
 	opts := []static.Option{static.WithCertChain(pemLeaf, []byte{}), static.WithBundle(rekorBundle)}
 	ociSig, _ := static.NewSignature(payload, base64.StdEncoding.EncodeToString(signature), opts...)
 
+	logId, _ := getLogID(sv.Public)
+	ecdsaKey, _ := sv.PublicKey()
+
 	// TODO(asraa): Re-enable passing test when Rekor public keys can be set in CheckOpts,
 	// instead of relying on the singleton TUF instance.
-	verified, err := VerifyImageSignature(context.TODO(), ociSig, v1.Hash{}, &CheckOpts{RootCerts: rootPool})
+	verified, err := VerifyImageSignature(context.TODO(), ociSig, v1.Hash{}, &CheckOpts{
```
vaikas, line 258: You have no idea how excited I'm about this :)

pkg/cosign/tlog.go
```diff
@@ -91,24 +98,19 @@ func intotoEntry(ctx context.Context, signature, pubKey []byte) (models.Proposed
...
 		raw, err := os.ReadFile(altRekorPub)
 		if err != nil {
 			return nil, fmt.Errorf("error reading alternate Rekor public key file: %w", err)
 		}
-		extra, err := PemToECDSAKey(raw)
-		if err != nil {
-			return nil, fmt.Errorf("error converting PEM to ECDSAKey: %w", err)
-		}
-		keyID, err := getLogID(extra)
+		publicKeys.AddRekorPubKey(raw, tuf.Active)
```
vaikas, line 110: err = publicKeys.AddRekorPubKey(raw, tuf.Active)

pkg/cosign/tlog.go
```diff
@@ -119,23 +121,34 @@ func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey,
 			return nil, err
 		}
 		for _, t := range targets {
-			rekorPubKey, err := PemToECDSAKey(t.Target)
+			publicKeys.AddRekorPubKey(t.Target, t.Status)
```
vaikas, line 124: err = publicKeys.AddRekorPubKey(t.Target, t.Status)

pkg/cosign/tlog.go
```diff
@@ -119,23 +121,34 @@ func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey,
...
+// rekorPubsFromClient returns a RekorPubKey keyed by the log ID from the Rekor client.
+// NOTE: This **must not** be used in the verification path, but may be used in the
+// sign path to validate return responses are consistent from Rekor.
+func rekorPubsFromClient(rekorClient *client.Rekor) (*TrustedRekorPubKeys, error) {
+	publicKeys := NewTrustedRekorPubKeys()
+	pubOK, err := rekorClient.Pubkey.GetPublicKey(nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch rekor public key from rekor: %w", err)
+	}
+	publicKeys.AddRekorPubKey([]byte(pubOK.Payload), tuf.Active)
```
vaikas, line 147: err = publicKeys.AddRekorPubKey([]byte(pubOK.Payload), tuf.Active)

vaikas (commented):
Awesomesauce! Just couple of lint warnings about names and from my pov it's good to go.

pkg/cosign/tlog.go
asraa: Yep, originally I used that approach of `type TrustedRekorPubKeys map[string]RekorPubKey`, but it ended up being ugly to do things like `AddKeyToTrustedRekorPubKeys` without supplying them as args rather than a method on a struct

pkg/cosign/verify_test.go
asraa: Hahahah see my TODO comment above too, I'll fix that in a follow-up!

pkg/cosign/tlog.go
asraa: done in latest commit w/ the others!

pkg/cosign/verify_test.go
vaikas: Yes, I saw that, and was referring to the TODO :)

pkg/cosign/tlog.go
vaikas: ah, perfect, yeah, I like this way personally better anyways :) Just jotted it down for prosperity (alternatives, considered lol)

pkg/cosign/tlog.go
```diff
@@ -436,12 +450,7 @@ func VerifyTLogEntry(ctx context.Context, _ *client.Rekor, e *models.LogEntryAno
 		LogID:          *e.LogID,
 	}
 
-	rekorPubKeys, err := GetRekorPubs(ctx, nil)
-	if err != nil {
-		return fmt.Errorf("unable to fetch Rekor public keys: %w", err)
-	}
-
-	pubKey, ok := rekorPubKeys[payload.LogID]
+	pubKey, ok := rekorPubKeys.Keys[payload.LogID]
```
vaikas, line 453: I'm so sorry, I missed this last time around :(

We should check for rekorPubKeys for being nil now that it's inside the wrapper struct. We do it in VerifyBundle, but not here.

vaikas: Maybe right after checking that e.Verification is not nil? L425?

asraa: Nice catch, thanks!

vaikas (dismissed):
Thank you!!!

znewman01 (approved):
This LGTM. Is there a code path we can hit for *online* verification?

asraa:
>  Is there a code path we can hit for online verification?

What do you mean? I'm not sure it's relevant for this PR. We're discussing the design for the necessary flags/behavior here: https://docs.google.com/document/d/1E-cB0OdABTi4BXnRV9oBzdbUgRk3cJ-jw0E5KGE3Eiw/edit?resourcekey=0-TCQAQ1H_pAiRzC7avDt1LA#

znewman01:
> > Is there a code path we can hit for online verification?
> 
> What do you mean? I'm not sure it's relevant for this PR. We're discussing the design for the necessary flags/behavior here: https://docs.google.com/document/d/1E-cB0OdABTi4BXnRV9oBzdbUgRk3cJ-jw0E5KGE3Eiw/edit?resourcekey=0-TCQAQ1H_pAiRzC7avDt1LA#

I think what I mean is `VerifyTlogEntry` now doesn't take a RekorClient.  What's my entry point in `pkg/cosign/verify.go` if I want to do online verification?  I guess...`FindTlogEntry`? That feels very manual, I would like a `Verify*()` call instead.

But agreed it's not super relevant for this PR, I was just wondering because now there's no RekorClient getting passed in

asraa:
> I was just wondering because now there's no RekorClient getting passed in

You already have all the online data inside the tlog entry to do the validation. This is just the verification portion, which does internally validate the provided inclusion proof. 

You're right though that the cosign API is unclear. :)

asraa:
I can't repro the failure locally :| Still trying

asraa:
@priyawadhwa @vaikas Do either of you know if there's a difference in the environments running `./test/e2e_test.sh` locally or on CI? https://github.com/sigstore/cosign/actions/runs/3373442220/workflow#L123

In CI we set up scaffolding first, but either way the tests run in `test/e2e_test.go` use the public Rekor: https://github.com/sigstore/cosign/blob/main/test/e2e_test.go#L73

Unless somehow they're not. If they're using the public Rekor (and it is on my laptop, since I printed the log ID of the failing bundle), then I wouldn't need to customize the trust root. However, on CI it fails saying the trust root has no matching Rekor keys.

asraa:
>  However, on CI it fails saying the trust root has no matching Rekor keys.

I think @loosebazooka helped deduce that CI must be running with the local scaffolding `.sigstore` and root of trust. I have no idea how this test was working before then.

asraa:
Resolved: the problem was that this test didn't add TUF root public keys because it's not a keyless verification. I added a TODO. We'll have this resolved when we have the online proof option for checking existing PKI log entries.

vaikas:
Looks like there was some flakiness checking out the repo, retrying.

pkg/cosign/tlog.go
```diff
@@ -390,13 +404,17 @@ func FindTlogEntry(ctx context.Context, rekorClient *client.Rekor,
 	return results, nil
 }
 
-// VerityTLogEntry verifies a TLog entry.
-// The argument *client.Rekor is unused and may be nil.
-func VerifyTLogEntry(ctx context.Context, _ *client.Rekor, e *models.LogEntryAnon) error {
+// VerityTLogEntry verifies a TLog entry against a map of trusted rekorPubKeys indexed
+// by log id.
+func VerifyTLogEntry(e *models.LogEntryAnon, rekorPubKeys *TrustedRekorPubKeys) error {
```
znewman01, line 409: Nit: maybe `func VerifyTLogEntryOffline`?

cmd/cosign/cli/verify/verify.go
```diff
@@ -159,6 +159,11 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 			return fmt.Errorf("getting Fulcio intermediates: %w", err)
 		}
 	}
+	// TODO: We only need this when we are doing online verification.
```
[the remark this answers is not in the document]
asraa: I can gate it with --skip-tlog-verify though.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -137,6 +137,10 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		if err != nil {
 			return fmt.Errorf("getting Fulcio intermediates: %w", err)
 		}
+		co.RekorPubKeys, err = cosign.GetRekorPubs(ctx)
```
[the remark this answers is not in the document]
asraa: Yeah, I'll add comments around all of these, and around the Fulcio root CAs as well (being needed for keyless cert verification)

pkg/cosign/tlog.go
asraa: Might as well since we're changing API. Online only means FindTlogEntry(Online) and run it through the same procedure as offline verification right now. We don't hold any client state to really do an "online verification"

asraa:
Looking at some e2e test failures:
```
2022-12-06T15:16:31.6270119Z     e2e_test.go:161: no matching signatures:
2022-12-06T15:16:31.6271235Z         searching log query: [POST /api/v1/log/entries/retrieve][404] searchLogQuery default  &{Code:404 Message:no responses found}
```

znewman01 (commented):
lgtm other than test failures

asraa:
> lgtm other than test failures

Yep -- Hayden alerted me that it was the SkipTlog issues, they're in https://github.com/sigstore/cosign/pull/2505 too, that I can rebase on

cmd/cosign/cli/verify/verify.go
```diff
@@ -161,6 +163,21 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 			return fmt.Errorf("getting Fulcio intermediates: %w", err)
 		}
 	}
+	if !co.SkipTlogVerify {
```
haydentherapper, line 166: Rekor client init is done above. We just need to remove the duplication.

asraa: Ah thanks!

asraa: done!

znewman01 (approved):

pkg/cosign/tlog.go
```diff
@@ -119,23 +120,32 @@ func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey,
...
 	}
 
-	return publicKeys, nil
+	return &publicKeys, nil
+}
+
+// rekorPubsFromClient returns a RekorPubKey keyed by the log ID from the Rekor client.
+// NOTE: This **must not** be used in the verification path, but may be used in the
+// sign path to validate return responses are consistent from Rekor.
+func rekorPubsFromClient(rekorClient *client.Rekor) (*TrustedRekorPubKeys, error) {
```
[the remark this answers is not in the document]
haydentherapper: It's also worth considering that this now doubles the amount of traffic to Rekor from Cosign. Is this needed, or can we add a follow up to plumb through TUF to verify the response?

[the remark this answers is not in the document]
asraa: There's a follow-up issue posted. I can get to it after the attestation support :/ 
it doesn't really double, but I get that it's an extra call on the sign path.

haydentherapper (approved):

PR #2397, opened 2022-11-01
Breaking change: replace --no-tlog-upload flag with --tlog-upload flag

priyawadhwa:
In 1.0, users could upload to the tlog and use a signing key by setting the COSIGN_EXPERIMENTAL env var. 

Now that the env var is being removed, we still need to give users a way to sign with a key and upload to the tlog.

```
cosign sign --key cosign.key --tlog-upload myimage

```

Replacing --no-tlog-upload with --tlog-upload gives us this opportunity, and is more intuitive.

#### Release Note
Breaking change: replace --no-tlog-upload flag with --tlog-upload flag (which by default is set to `false`)

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -67,7 +67,7 @@ func uploadToTlog(ctx context.Context, sv *sign.SignerVerifier, rekorURL string,
 
 // nolint
 func AttestCmd(ctx context.Context, ko options.KeyOpts, regOpts options.RegistryOptions, imageRef string, certPath string, certChainPath string,
-	noUpload bool, predicatePath string, force bool, predicateType string, replace bool, timeout time.Duration, noTlogUpload bool) error {
+	noUpload bool, predicatePath string, force bool, predicateType string, replace bool, timeout time.Duration, tlogUpload bool) error {
 	// A key file or token is required unless we're in experimental mode!
 	if options.EnableExperimental() {
 		if options.NOf(ko.KeyRef, ko.Sk) > 1 {
@@ -159,7 +159,7 @@ func AttestCmd(ctx context.Context, ko options.KeyOpts, regOpts options.Registry
 	}
 
 	// Check whether we should be uploading to the transparency log
-	if sign.ShouldUploadToTlog(ctx, ko, digest, force, noTlogUpload) {
+	if sign.ShouldUploadToTlog(ctx, ko, digest, force, tlogUpload) {
 		bundle, err := uploadToTlog(ctx, sv, ko.RekorURL, func(r *client.Rekor, b []byte) (*models.LogEntryAnon, error) {
 			return cosign.TLogUploadInTotoAttestation(ctx, r, signedPayload, b)
 		})
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -72,7 +72,6 @@ func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Refere
 		return true
 	}
 	// If we aren't using keyless signing and --tlog-upload=false, return
-	// keyless signing currently requires uploading to the tlog (this will change once user-supplied timestamps for cert verification are supported)
 	if !keylessSigning(ko) && !tlogUpload {
 		return false
 	}
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -66,18 +66,13 @@ const TagReferenceMessage string = `WARNING: Image reference %s uses a tag, not 
     images by tag will be removed in a future release.
 `
 
-func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, force, noTlogUpload bool) bool {
+func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, force, tlogUpload bool) bool {
 	// We are forcing publishing to the Tlog.
 	if force {
 		return true
 	}
-	// Check whether to not upload Tlog.
-	if noTlogUpload {
-		return false
-	}
-
-	// If we aren't using keyless signing, we can skip the tlog
-	if !keylessSigning(ko) {
+	// If we aren't using keyless signing and --tlog-upload=false, return
+	if !keylessSigning(ko) && !tlogUpload {
 		return false
 	}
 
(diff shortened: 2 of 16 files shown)


asraa:
A comment for the future, although this hasn't changed. Can you help update the doc below? What does `--verify-tlog` mean? If the user specifies a `--bundle`, is that OK? Or does this mean `--reuiqre-tlog-inclusion`, like an online inclusion proof?
  
https://docs.google.com/document/d/1E-cB0OdABTi4BXnRV9oBzdbUgRk3cJ-jw0E5KGE3Eiw/edit?resourcekey=0-TCQAQ1H_pAiRzC7avDt1LA#

cmd/cosign/cli/attest/attest.go
```diff
@@ -159,7 +159,7 @@ func AttestCmd(ctx context.Context, ko options.KeyOpts, regOpts options.Registry
 	}
 
 	// Check whether we should be uploading to the transparency log
-	if sign.ShouldUploadToTlog(ctx, ko, digest, force, noTlogUpload) {
+	if sign.ShouldUploadToTlog(ctx, ko, digest, force, tlogUpload) {
```
asraa, line 162: what's the difference between this and `fource`?

asraa (commented):
Do we need this with the force flag?

cmd/cosign/cli/verify/verify.go
```diff
@@ -114,7 +115,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		co.ClaimVerifier = cosign.SimpleClaimVerifier
 	}
 
-	if c.keylessVerification() {
+	if c.keylessVerification() || c.VerifyTlog {
```
asraa, line 118: Oops, just realized you did add this in: This is currently implementing forced online I think, at least on the CLI layer. I would rename it to that, because if there's a bundle in the OCI registry, does that "count"?

cmd/cosign/cli/sign/sign.go
```diff
@@ -66,18 +66,13 @@ const TagReferenceMessage string = `WARNING: Image reference %s uses a tag, not
...
 	}
-	// Check whether to not upload Tlog.
-	if noTlogUpload {
-		return false
-	}
-
-	// If we aren't using keyless signing, we can skip the tlog
-	if !keylessSigning(ko) {
+	// If we aren't using keyless signing and --tlog-upload=false, return
+	if !keylessSigning(ko) && !tlogUpload {
```
haydentherapper, line 75: keyless = true, tlog = true - upload
keyless = true, tlog = false - don't upload
keyless = false, tlog = true - upload
keyless = false, tlog = false - don't upload

So keyless doesn't matter then, correct? We can just set tlog-upload's default to true.

haydentherapper: Ah, I guess a problem here is we don’t want to default to be true, for the managed key cases. So we should check keyless. 

In the code, I think we still need to handle the case of keyless = true, and tlog = false

cmd/cosign/cli/attest/attest.go
priyawadhwa: i believe the goal of force is to get around the user prompt that happens in `ShouldUploadToTlog`

https://github.com/sigstore/cosign/blob/1c15b03a6e9e47f5c85a8fd7ff98bf75e910df6a/cmd/cosign/cli/sign/sign.go#L88-L92

haydentherapper: We should rename force to be more descriptive

cmd/cosign/cli/sign/sign.go
priyawadhwa: > Ah, I guess a problem here is we don’t want to default to be true, for the managed key cases. 

Yah, exactly! 

> I think we still need to handle the case of keyless = true, and tlog = false

in that case, verification would always fail, so i feel like we shouldn't allow users to sign that way to begin with.

cmd/cosign/cli/attest/attest.go
priyawadhwa: i think we can just delete it in a separate PR, we have a `SkipConfirmation` field we can rely on instead

https://github.com/sigstore/cosign/blob/04f509c7c95538d50a6065c2f098eada59d9f139/cmd/cosign/cli/options/sign.go#L35

seems like the intention was to remove `--force` and replace it with that based on this comment back in July!

https://github.com/sigstore/cosign/pull/2039#discussion_r911985088

cmd/cosign/cli/verify/verify.go
priyawadhwa: I'm not sure I understand the issue completely 🤔 Do you mean the keylessVerification check should actually be a forcedOnlineVerification check?

cmd/cosign/cli/attest/attest.go
asraa: Gotcha, cool: so `force` will be `skipConfirmation` then in another PR!

cmd/cosign/cli/sign/sign.go
asraa: That'll change if they're allowed to get a user-supplied timestamp, but they'd be using cosign libraries at that point and not CLIs.
I think this checks fine, but you might want to add a comment saying that we must upload for keyless.

cmd/cosign/cli/verify/verify.go
asraa: I think it's confusing too... `c.VerifyTlog` is setting a RekorClient for ONLINE lookups. Not just any tlog verification (in case it comes from a bundle).

cmd/cosign/cli/sign/sign.go
priyawadhwa: sg, added the comment!

cmd/cosign/cli/verify/verify.go
asraa: Does it mean "any tlog verification, including from a bundle"?

priyawadhwa: that's what i intended!

asraa: My personal preference is allowing the client to specify additional online verification, rather than just any

When Rekor was sharded, we never detected that w/ probers because we never had a way to explicitly say go check Rekor online for a proof and all OCI images have a bundle. If we use this flag, we never get the path of supporting bundle verification and online inclusion proofs (which are stronger)

I don't think it'll force online -- if a bundle is supplied and works i think it exits early. but please check!

asraa: TBH want to do the verify path in a follow-up? the sign path in this PR looks fine to me

haydentherapper: +1 to doing verify in a follow up

priyawadhwa: sure, sign was my initial plan, but the e2e tests won't pass without verify. i can just keep those portions commented out for now.

haydentherapper: What if we construct a valid bundle for the test?

priyawadhwa: i think it'll probably depend on what flags/behavior we implement in the verify PR, i think it makes more sense to update the e2e test in that one (right now the test is intended to confirm that signing/verifying with a managed key + online tlog lookup works)

asraa (approved):
LGTM! 

Could you update the release note with what the default is for the flag?

hectorj2f (approved):

PR #2399, opened 2022-11-01
Breaking change: remove --force flag from sign and attest and rely on --yes flag to skip confirmation

priyawadhwa:
Note, this doesn't remove the --force flag from `cosign clean` and `cosign copy`, which use it to override destruction of images/signatures.

#### Release Note
Breaking change: remove --force flag from sign and attest and rely on --yes flag to skip confirmation


cpanato (approved):
lgtm

znewman01 (approved):

PR #2400, opened 2022-11-02
Breaking change: Change SCT verification behavior to default to enforcement

### LINKED ISSUE #2382: Require embedded SCT on verification by default
**Description**

Currently, SCT verification happens on:
* Signing, either by checking an embedded SCT (public good Fulcio) or a detached SCT returned in a header. It's skipped if an insecure flag is provided
* Verification, if either the `enforce-sct` flag is set or if an embedded SCT is present.

Signing is good - Always check, unless you explicitly provide an insecure flag. However, for verification, you must present a flag to require the check, which is not ideal - This was added because previously issued certificates did not include an embedded SCT and the Cosign metadata did not include a way to persist the detached SCT.

I propose:
1. We require that an SCT be embedded or explicitly provided for the detached case. For the detached case, we could either add new annotation metadata, or simply require it be passed by flag. Given the detached case is only for private deployments, I would probably go with flag for now.
1. We remove the `enforce-sct` flag and leverage an `insecure` flag to skip the check on verification.

This would be a part of Cosign 2.0, since it's a breaking change.

Code:
* https://github.com/sigstore/cosign/blob/d795dcb4ca44472896fec39b6ece4bafaee28f58/pkg/cosign/verify.go#L204-L215
* https://github.com/sigstore/cosign/blob/d795dcb4ca44472896fec39b6ece4bafaee28f58/cmd/cosign/cli/options/certificate.go#L77

Hayden-IO:
The public good CA always returns an embedded SCT. If you want to skip the check,
you now must provide a flag to do so. Before, you had to provide a flag
to enforce the check.

Fixes https://github.com/sigstore/cosign/issues/2382

#### Release Note

Changed SCT verification to enforce that the Fulcio certificate has been logged


pkg/cosign/verify_sct_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"os"
+	"testing"
+
+	ct "github.com/google/certificate-transparency-go"
+	"github.com/google/certificate-transparency-go/testdata"
+	"github.com/google/certificate-transparency-go/tls"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+// TODO: Move back into verify_test.go once the test cert has been regenerated
```
haydentherapper, line 34: FYI - This was because the test cert is signed with a SHA1 digest algorithm, and golang throws an error in testing. To override, you need to set `GODEBUG`, but this can't be set in the test itself. I've tried various things, but the easiest was to simply move this test to its own file with a build tag - Note that this is still tested in GHA tests.

haydentherapper:
Resolved merge conflict

haydentherapper:
Bump!

priyawadhwa (approved):
looks great!

asraa:
@haydentherapper I think this might be breaking tests at head: I think the leaf cert generation doesn't attach an SCT. I have no idea how tests passed though!

```
--- FAIL: TestValidateAndUnpackCertSuccessWithOtherNameSan (0.00s)
    verify_test.go:658: ValidateAndUnpackCert expected no error, got err = certificate does not include required embedded SCT and no detached SCT was set
```

PR #2401, opened 2022-11-02
Refactor attest command to make it easier to set args

priyawadhwa:


cpanato (approved):
looks great! thanks for that

PR #2402, opened 2022-11-02
bump scaffolding to 0.4.12

bobcallaway:

### DIFF
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -36,7 +36,7 @@ jobs:
 
     env:
       KO_DOCKER_REPO: "registry.local:5000/policy-controller"
-      SCAFFOLDING_RELEASE_VERSION: "v0.4.8"
+      SCAFFOLDING_RELEASE_VERSION: "v0.4.12"
       GO111MODULE: on
       GOFLAGS: -ldflags=-s -ldflags=-w
       KOCACHE: ~/ko


bobcallaway:
@vaikas @priyawadhwa this model of pinning to back versions of rekor/fulcio as well as older versions of cosign via scaffolding feels very broken to me. Are we actually getting meaningful coverage from these tests in PRs?

vaikas:
That's a great point @bobcallaway and something that we did think about on how to best handle but never got around to it. Do you think we should be grabbing the latest or head for things like Fulcio/Rekor?

priyawadhwa (approved):

bobcallaway:
In rekor, @priyawadhwa implemented a check that ran rekor-cli against 3 versions (HEAD, and 2 latest published tags). For cosign, I would want to run it against prod, and perhaps a matrix of 3 versions of both rekor & fulcio for adding coverage to those running private instances of rekor and/or fulcio.

vaikas:
Ok, that's doable. But only against released versions, correct?

bobcallaway:
> Ok, that's doable. But only against released versions, correct?
> […](#)
> On Wed, Nov 2, 2022 at 12:49 PM Bob Callaway ***@***.***> wrote: In rekor, @priyawadhwa implemented a check that ran rekor-cli against 3 versions (HEAD, and 2 latest published tags). For cosign, I would want to run it against prod, and perhaps a matrix of 3 versions of both rekor & fulcio for adding coverage to those running private instances of rekor and/or fulcio. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you modified the open/close state.Message ID: ***@***.***>

yup

PR #2411, opened 2022-11-03
Deprecate --certificate-email flag. Make --certificate-identity and -…

### LINKED ISSUE #2056: Discourage (disallow?) naked `cosign verify` invocations
Inspired by https://github.com/sigstore/sigstore-python/issues/155

`cosign verify` (in keyless mode) without any `--cert-email` provided will succeed when literally anybody has signed an image/blob. This is almost never the right behavior: evil@hacker.com can sign any blob you can.

Instead, we should require *some* verification policy:

- `--cert-email` (*and* `--cert-oidc-provider`; see #1947)
- a CUE policy
- `--insecure`, which will do the current behavior today (possibly a hidden flag)

(Keyfull mode is okay, because you're passing the key material or CA in explicitly)

kpk47:
…-certificate-oidc-issuer required for verification.

--certificate-email is now an alias for --certificate-identity

Signed-off-by: kpk47 <kkris@google.com>

#### Summary

Closes #2056


cmd/cosign/cli/options/certificate.go
```diff
@@ -41,14 +40,14 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
...
 	_ = cmd.Flags().SetAnnotation("certificate", cobra.BashCompFilenameExt, []string{"cert"})
 
-	cmd.Flags().StringVar(&o.CertEmail, "certificate-email", "",
-		"the email expected in a valid Fulcio certificate")
+	cmd.Flags().StringVar(&o.CertIdentity, "certificate-email", "",
+		"DEPRECATED: Use --certificate-identity instead. The email expected in a valid Fulcio certificate")
 
 	cmd.Flags().StringVar(&o.CertIdentity, "certificate-identity", "",
-		"the identity expected in a valid Fulcio certificate. Valid values include email address, DNS names, IP addresses, and URIs.")
+		"Required. The identity expected in a valid Fulcio certificate. Valid values include email address, DNS names, IP addresses, and URIs.")
```
woodruffw, line 47: This is a tiny thing, but is it possible to make these options mutually exclusive? Passing both of them at once implies user error/confusion, since only one takes effect at a time.

cmd/cosign/cli/options/certificate.go
```diff
@@ -42,13 +42,13 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
 	_ = cmd.Flags().SetAnnotation("certificate", cobra.BashCompFilenameExt, []string{"cert"})
 
 	cmd.Flags().StringVar(&o.CertEmail, "certificate-email", "",
```
haydentherapper, line 44: I'd just remove this flag actually. This is going to be going into 2.0, so we can break things.

woodruffw: +1 -- I was thinking in the context of sigstore-python where we'll remove a whole bunch of things right before 1.0, but removal right now makes much more sense in the context of cosign 2.0.

kpk47: Done.

cmd/cosign/cli/options/certificate.go
```diff
@@ -41,14 +40,11 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
 		"path to the public certificate. The certificate will be verified against the Fulcio roots if the --certificate-chain option is not passed.")
 	_ = cmd.Flags().SetAnnotation("certificate", cobra.BashCompFilenameExt, []string{"cert"})
 
-	cmd.Flags().StringVar(&o.CertEmail, "certificate-email", "",
-		"the email expected in a valid Fulcio certificate")
-
 	cmd.Flags().StringVar(&o.CertIdentity, "certificate-identity", "",
-		"the identity expected in a valid Fulcio certificate. Valid values include email address, DNS names, IP addresses, and URIs.")
+		"Required. The identity expected in a valid Fulcio certificate. Valid values include email address, DNS names, IP addresses, and URIs.")
```
haydentherapper, line 44: Is there enforcement that both of these flags are provided?

cmd/cosign/cli/verify.go
```diff
@@ -92,12 +93,15 @@ against the transparency log.`,
 				return err
 			}
 
+			if o.CertVerify.CertIdentity == "" || o.CertVerify.CertOidcIssuer == "" {
```
haydentherapper, line 96: I think we instead want this in the Exec functions, like https://github.com/sigstore/cosign/blob/1c15b03a6e9e47f5c85a8fd7ff98bf75e910df6a/cmd/cosign/cli/verify/verify.go#L78

You also need this in the other verify commands, for manifest, dockerfile, and verify-blob

cmd/cosign/cli/options/certificate.go
znewman01: Outdated comment, we're just axeing `--certificate-email` altogether

znewman01 (commented):
Overall LGTM. Needs a rebase before we can merge.

I'd be interested in putting this validation all in the same place. Maybe factor out some of these arguments into a new `options.VerifyOpts` and have a function to parse them into `cosign/verify.CheckOpts`. But I'm okay doing that later.

znewman01 (dismissed):
adding approving review so tests run

znewman01 (dismissed):
approving for tests

znewman01:
test failure looks related!

znewman01 (dismissed):
approve for tests, ping me once you're ready for full review

kpk47:
@znewman01 This is ready for review now. The biggest changes are in test/sign_blob_test.sh, where I had to switch the keyless entry to using a local key and drop the verification step for the self-signed certificate. I believe the test logic is still sound (sign and verify with cosign, upload signature with self-signed cert, check that the first entry is still verifiable).

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -104,16 +103,18 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 		return err
 	}
 
+	if c.KeyRef == "" && (c.CertIdentity == "" || c.CertOIDCIssuer == "") {
```
haydentherapper, line 106: Can you add a test for failure if these aren't set?

pkg/cosign/verify.go
```diff
@@ -251,14 +244,11 @@ func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Ver
...
 
-	if err := validateCertIdentity(cert, co); err != nil {
-		return err
-	}
-
 	if err := validateCertExtensions(ce, co); err != nil {
 		return err
 	}
 	issuer := ce.GetIssuer()
+	fmt.Fprintf(os.Stderr, "cert issuer: %s\n", issuer)
```
haydentherapper, line 251: remove?

test/sign_blob_test.sh
```diff
@@ -29,14 +29,19 @@ BLOB=verify-experimental-blob
...
 cat $BLOB
 
+pass="$RANDOM"
+export COSIGN_PASSWORD=$pass
+./cosign generate-key-pair
+signing_key=cosign.key
+verification_key=cosign.pub
+
 echo "Sign the blob with cosign first and upload to rekor"
-$COSIGN_CLI sign-blob --output-certificate blob.cert --output-signature blob.sig $BLOB
```
haydentherapper, line 33: What is the reason for needing to change this? Do we lose coverage of the keyless flow with this change? I think both are useful as e2e tests.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -94,6 +93,10 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 		return &options.PubKeyParseError{}
 	}
 
+	if c.CertRef != "" && (c.CertIdentity == "" || c.CertOIDCIssuer == "") {
```
znewman01, line 96: I don't *love* the duplication here. Could we pull this out into a common "VerificationIdentity" struct that goes into the `VerifyBlobCmd` (and other `*BlobCmd` structs) and stick the check there?

cmd/cosign/cli/verify/verify.go
```diff
@@ -94,23 +93,26 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		c.HashAlgorithm = crypto.SHA256
 	}
 
+	if c.CertRef != "" && (c.CertIdentity == "" || c.CertOidcIssuer == "") {
```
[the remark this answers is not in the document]
kpk47: That makes sense to me, but how should it behave with a local bundle? IIUC, the bundle wraps the cert and additional metadata. Do we also need `--cert-identity` and `--cert-oidc-issuer` in that case?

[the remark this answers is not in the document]
znewman01: >  Do we also need --cert-identity and --cert-oidc-issuer in that case?

IMO: Yes, we do.

We don't want to trust the identity from the bundle; we want to cross-reference it with the `--cert-*` flags.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -76,21 +75,24 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		return &options.PubKeyParseError{}
 	}
 
+	if c.CertRef == "" && (c.CertIdentity == "" || c.CertOidcIssuer == "") {
```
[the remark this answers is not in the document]
kpk47: Yes, this was a typo.

cmd/cosign/cli/verify/verify_blob.go
kpk47: I moved the check into CheckVerifyOptions. WDYT? I'm not convinced it's the best solution, but it _is_ already referenced by all the Check*Cmd structs.

cmd/cosign/cli/verify/verify_blob.go
kpk47: Done.

pkg/cosign/verify.go
kpk47: Done

test/sign_blob_test.sh
kpk47: I may be able to keep this keyless depending on how GitHub OIDC mints tokens, though we'd need some extra work to make it work when run locally.

What's the intended usage for this test? I don't see it referenced in any workflows, so I added it to tests.yaml.

cmd/cosign/cli/verify/verify_blob.go
znewman01: I like this overall, I think. I'm finding it more readable.

I was imagining sticking the `KeyRef` inside `CertVerifyOptions` as well so we wouldn't have to duplicate the `if KeyRef != ""` check. Then, the `Identities()` would return basically a sum-type of `cosign.Identity | PublicKey` you could `switch` over that. WDYT?

Part of what you're running up against is that this whole thing needs a big refactor (not your responsibility), gross. Thanks for your patience :)

cmd/cosign/cli/verify/verify.go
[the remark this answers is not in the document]
znewman01: I think this has made some e2e tests start failing.

cmd/cosign/cli/verify/verify_blob.go
kpk47: I'm not sure I like that solution since it's pulling (a tiny) part of key parsing into a struct that's about certificate verification.

I filed https://github.com/sigstore/cosign/issues/2462 with my suggestions for how to go about the larger refactor. How about we deal with the duplication in this PR and then I address the larger refactor in a follow-on change?

cmd/cosign/cli/verify/verify.go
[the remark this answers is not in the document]
znewman01: Pinging this—I don't want the `--bundle` flag to turn off the requirement for a key or identity.

znewman01 (dismissed):
approving for tests only

haydentherapper:
Bumping PR, @kpk47, what is left? You'll need to rebase since we had to change import paths to include `v2`

kpk47:
Sorry, this fell off my plate. There isn't any major work left here, just fixing up a broken integration test. I've got time blocked out to finish this in the next few days.

test/sign_blob_test.sh
kpk47: I got this to work by treating the values of certificate-identity and certificate-oidc-issuer as regexps and then having the test pass in a match-all pattern.

cmd/cosign/cli/options/certificate.go
```diff
@@ -82,3 +81,13 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
...
 }
+
+func (o *CertVerifyOptions) Identities() ([]cosign.Identity, error) {
+	if o.CertIdentity == "" {
+		return nil, errors.New("--certificate-identity is required for verification in keyless mode")
+	}
+	if o.CertOidcIssuer == "" {
+		return nil, errors.New("--certificate-oidc-issuer is required for verification in keyless mode")
+	}
+	return []cosign.Identity{{IssuerRegExp: o.CertOidcIssuer, Issuer: o.CertOidcIssuer, SubjectRegExp: o.CertIdentity, Subject: o.CertIdentity}}, nil
```
haydentherapper, line 92: We don't need to set `IssuerRegExp` and `SubjectRegExp` here, since the case statement just ignores them if they're not set. 

I think we'd need another set of flags to support verifying by regex (for example, an email could contain a period), but I think we could add this later on. Given this is a CLI interface, I don't expect needing to verify a regex.

EDIT: I see you are leveraging these flags as a regex later on. Do we want to just go ahead and add flags for certificate-identity-regex and certificate-oidc-issuer-regex now?

pkg/cosign/verify.go
```diff
@@ -315,13 +304,19 @@ func CheckCertificatePolicy(cert *x509.Certificate, co *CheckOpts) error {
 						break
 					}
 				}
+				if regex.MatchString(cert.Subject.String()) {
```
haydentherapper, line 307: Same here, a certificate subject is not the same as the subject alternative name, and we don't expect fulcio certificates to ever set the certificate subject.

test/sign_blob_test.sh
```diff
@@ -33,10 +33,9 @@ echo "Sign the blob with cosign first and upload to rekor"
 $COSIGN_CLI sign-blob --output-certificate blob.cert --output-signature blob.sig $BLOB
 
 echo "Verifying ..."
-$COSIGN_CLI verify-blob --signature blob.sig --cert blob.cert $BLOB
+$COSIGN_CLI verify-blob --signature blob.sig --cert blob.cert --certificate-identity '.*' --certificate-oidc-issuer '.*' $BLOB
```
haydentherapper, line 36: As I noted above, I think we shouldn't overload this flag to support regex, and should add a separate flag. Otherwise, you run into potential risks with a character representing both regex and a certain character (like a period, valid for emails)

pkg/cosign/verify.go
kpk47: Do we care about non-Fulcio certificates that don't include SANs? We're using such a certificate in test/sign_blob_test.sh, so I figured it must be a supported use case.

pkg/cosign/verify.go
```diff
@@ -289,11 +278,11 @@ func CheckCertificatePolicy(cert *x509.Certificate, co *CheckOpts) error {
 			case identity.IssuerRegExp != "":
 				if regex, err := regexp.Compile(identity.IssuerRegExp); err != nil {
 					return fmt.Errorf("malformed issuer in identity: %s : %w", identity.IssuerRegExp, err)
-				} else if regex.MatchString(issuer) {
+				} else if regex.MatchString(issuer) || regex.MatchString(cert.Issuer.String()) {
```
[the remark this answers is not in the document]
kpk47: Done

cmd/cosign/cli/options/certificate.go
kpk47: Done. Added the new flags.

test/sign_blob_test.sh
kpk47: Done

pkg/cosign/verify.go
haydentherapper: I don't think we should. The bring-your-own-pki story needs to be fleshed out and we'd need to think through how to represent a subject (do we do something like `CN=foo;O=bar`, or do we support flags for each subject component?)

Do you want me to help regenerate a test certificate?

Hayden-IO (commented):
Just need to rerun `make docgen`, this looks great!

pkg/cosign/verify.go
kpk47: Yes, please.

haydentherapper:
@kpk47 Can you update `test/testdata/test_blob_cert.pem` with the following:

```
-----BEGIN CERTIFICATE-----
MIIBdDCCARqgAwIBAgIUZw7gQ6T/IgmiMD1AWB2OTIIVH1owCgYIKoZIzj0EAwIw
ADAeFw0yMjEyMjEwMDIwNThaFw0zMjEyMTgwMDIwNThaMAAwWTATBgcqhkjOPQIB
BggqhkjOPQMBBwNCAAR1Q4hB1jtagrdsVxygtDa/rli00U7n/1I/NSw8yoMRQ+MO
AjRhg3gtcV0tha34L6150qJirQHbfocsao8X6wFmo3IwcDAdBgNVHQ4EFgQUx3Wb
0LwCWoGsl0FUpeQb3M4MukkwHwYDVR0jBBgwFoAUx3Wb0LwCWoGsl0FUpeQb3M4M
ukkwEgYDVR0TAQH/BAgwBgEB/wIBATAaBgNVHREEEzARgQ9mb29AZXhhbXBsZS5j
b20wCgYIKoZIzj0EAwIDSAAwRQIhALXG7XS5TIFLp+jLSxjuRk1Tj5MfE+y9x92Z
YPMbi9GZAiAmfEe0+q5l3PnI6zliOG5kG6EcS80QQgQmPcFvRZWOvw==
-----END CERTIFICATE-----

```

And update https://github.com/sigstore/cosign/blob/main/test/testdata/README.md, the openssl command, to be :

```
openssl req -key test/testdata/test_blob_private_key -x509 -days 3650 -out cert.pem -new -nodes -subj "/" -addext "subjectAltName = email:foo@example.com"
```

You can remove all the output below the command too, `-subj "/"` handles that so you aren't prompted.

You made need to update tests to check for this example SAN. Feel free to rerun the command too if you have a preferred email value.

pkg/cosign/verify.go
haydentherapper: added comment

Hayden-IO (dismissed):
LGTM, thanks so much for working on this!

Hayden-IO (dismissed):

haydentherapper:
@znewman01  once tests finish, do you want to merge?

znewman01:
E2e tests are failing, once we get those fixed I can merge (will take a quick skim today too)

Hayden-IO (approved):
test failure is unrelated

znewman01 (approved):

dmitris:
@znewman01 do you know if the docs on https://docs.sigstore.dev/cosign/keyless/ were updated for this change or not?

haydentherapper:
Docs are being actively updated for cosign 2.0. Keyless will likely be removed as it’s not duplicated with the verify doc, which has been updated.

PR #2418, opened 2022-11-07
verify: combine verification flows with substituteable verification fns

asraa:

This combines the verification flows for OCI attestations and blobs.

This will make it easier to add flags for validation, and cleanup the logic when the experimental flag is removed.

There is further type-specific logic contained in side bundle verification, it would be nice to separate those out as well.

You don't need to merge this, but I really hated the double logic. Please suggest something else if it would be nicer!

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -121,24 +121,15 @@ type CheckOpts struct {
 	Identities []Identity
 }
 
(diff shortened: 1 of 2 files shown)


priyawadhwa (approved):
nice!

hectorj2f (approved):
Lgtm, thanks

Issue #2331, opened 2022-10-12
Allow using a TSA instead of Rekor to validate certs


priyawadhwa:
Right now, cosign assumes that every Fulcio cert has an associated entry in Rekor. The time that the entry was added to Rekor is used to verify that the signature happened during the cert's validity period.

We can get the same guarantee with a TSA. We'll need to refactor cosign to allow using a TSA instead of Rekor. We'll also want to store this TSA in the bundle so offline verification still works.

It would be nice to give users the option to choose between Rekor or the TSA!

znewman01:
CC @haydentherapper I know you've talked about this before, I thought there was somewhere this was being tracked but I can't find it.

haydentherapper:
This came up in https://github.com/sigstore/cosign/issues/2194, when discussing how to handle expired certificates. Either a Rekor entry or a signed timestamp should suffice as a countersignature over time. A TSA doesn't offer public auditability however, so it's not exactly equivalent to a Rekor entry. A verifier may want both a timestamp and Rekor entry - The timestamp to provide a third-party assertion over time, and the Rekor entry to provide an auditable record.

I've been working on spinning up a timestamp authority - https://github.com/sigstore/timestamp-authority. Sigstore will deploy one, and any other interested community members can spin up their own TSA too.

Support for timestamps within Rekor entries, to provide immutable timestamps and replace the integrated time, is something that @cdris is working on! Ceri will be sharing a doc with the community shortly about how we can currently verify entries with a TSA, propose how to include timestamps in Rekor entries, and where we want to go from there.

The bundle that's been proposed does have support for RFC3161 timestamps, as we plan to add them into Rekor entries in the near future.

We don't have an open issue yet - Related is a discussion on Roughtime (https://github.com/sigstore/rekor/issues/831), but @cdris will open an issue in Rekor soon.

asraa:
> It would be nice to give users the option to choose between Rekor or the TSA!

We can start with an env var for trusting a TSA root CA, but we also need a plan for inserting a TSA root CA into TUF. Just FYI: if Sigstore deploys one, no problem, but as we do this work I'll track changes we'll need fro mthe TUF side.

haydentherapper:
Tracking issue - https://github.com/sigstore/root-signing/issues/466

hectorj2f:
FYI: I'm working on this feature in case anyone else is working on a similar direction.

asraa:
@hectorj2f how do you plan to approach it? 

[...]

PR #2422, opened 2022-11-08
Define a new sigstore Bundle type to include additional verification data and materials

hectorj2f:
#### Summary

As part of this effort https://github.com/sigstore/cosign/issues/2331, I'm proposing renaming the original RekorBundle type to a more generic sigstore Bundle type that includes additional data. The new fields added to the sigstore Bundle are inspired from the protobuf specification here: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto. 

The next actions include adding support to use a Timestamp Authority server (sigstore/timestamp-authority) to guarantee a signature happened during the cert's validity period.

### DIFF
--- a/pkg/cosign/bundle/bundle.go
+++ b/pkg/cosign/bundle/bundle.go
@@ -1,0 +1,86 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundle
+
+import (
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+type Bundle struct {
+	VerificationData
+	VerificationMaterial
+}
+
+// VerificationData contains extra data that can be used to verify things
+// such as transparency logs and timestamped verifications.
+type VerificationData struct {
+	// RekorPayload holds metadata about recording a Signature's ephemeral key to
+	// a Rekor transparency log.
+	// Use Payload instead of TransparencyLogEntry to keep backwards compatibility.
+	Payload RekorPayload
+
+	// TimestampVerificationData holds metadata about a timestamped verification.
+	TimestampVerificationData
+}
+
+// VerificationMaterial captures details on the materials used to verify
+// signatures or any additional timestamped verifications.
+type VerificationMaterial struct {
+	// A chain of X.509 certificates.
+	CertBytes []byte
+
+	// PublicKeyIdentifier optional unauthenticated hint on which key to use.
+	PublicKeyIdentifier string
+}
+
+// TimestampVerificationData contains various timestamped data following RFC3161.
+type TimestampVerificationData struct {
+	// SignedEntryTimestamp holds metadata about a timestamped counter signature over the artifacts signature.
+	SignedEntryTimestamp []byte
+
+	// EntryTimestampAuthority contains the recorded entry from timestamp authority server.
+	EntryTimestampAuthority []byte
+}
+
+func EntryToBundle(tLogEntry *models.LogEntryAnon, signedEntryTimestamp, entryTimestampAuthority, certBytes []byte, pubKeyID string) *Bundle {
+	b := &Bundle{}
+	// If none of the verification data is configured then return nil
+	if (tLogEntry == nil || tLogEntry.Verification == nil) && len(entryTimestampAuthority) == 0 {
+		return nil
+	}
+	// Add Transparency log entry and a signed timestamp value
+	if tLogEntry != nil && tLogEntry.Verification != nil {
+		b.Payload = RekorPayload{
+			Body:           tLogEntry.Body,
+			IntegratedTime: *tLogEntry.IntegratedTime,
+			LogIndex:       *tLogEntry.LogIndex,
+			LogID:          *tLogEntry.LogID,
+		}
+		b.SignedEntryTimestamp = tLogEntry.Verification.SignedEntryTimestamp
+
+		if len(signedEntryTimestamp) > 0 {
+			b.SignedEntryTimestamp = signedEntryTimestamp
+		}
+	}
+	// Set the EntryTimestampAuthority from the timestamp authority server
+	if len(entryTimestampAuthority) > 0 {
+		b.EntryTimestampAuthority = entryTimestampAuthority
+	}
+	if len(certBytes) > 0 || pubKeyID != "" {
+		b.CertBytes = certBytes
+		b.PublicKeyIdentifier = pubKeyID
+	}
+	return b
+}
--- a/pkg/oci/static/options_test.go
+++ b/pkg/oci/static/options_test.go
@@ -26,7 +26,7 @@ import (
 )
 
 func TestOptions(t *testing.T) {
-	bundle := &bundle.RekorBundle{}
+	bundle := &bundle.Bundle{}
 
 	tests := []struct {
 		name string
@@ -87,7 +87,7 @@ func TestOptions(t *testing.T) {
 			LayerMediaType:  ctypes.SimpleSigningMediaType,
 			ConfigMediaType: types.OCIConfigJSON,
 			Annotations: map[string]string{
-				BundleAnnotationKey: "{\"SignedEntryTimestamp\":null,\"Payload\":{\"body\":null,\"integratedTime\":0,\"logIndex\":0,\"logID\":\"\"}}",
+				BundleAnnotationKey: "{\"Payload\":{\"body\":null,\"integratedTime\":0,\"logIndex\":0,\"logID\":\"\"},\"SignedEntryTimestamp\":null,\"EntryTimestampAuthority\":null,\"CertBytes\":null,\"PublicKeyIdentifier\":\"\"}",
 			},
 			Bundle: bundle,
 		},
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -47,7 +47,7 @@ import (
 
 type tlogUploadFn func(*client.Rekor, []byte) (*models.LogEntryAnon, error)
 
-func uploadToTlog(ctx context.Context, sv *sign.SignerVerifier, rekorURL string, upload tlogUploadFn) (*cbundle.RekorBundle, error) {
+func uploadToTlog(ctx context.Context, sv *sign.SignerVerifier, rekorURL string, upload tlogUploadFn) (*cbundle.Bundle, error) {
 	rekorBytes, err := sv.Bytes(ctx)
 	if err != nil {
 		return nil, err
@@ -62,7 +62,7 @@ func uploadToTlog(ctx context.Context, sv *sign.SignerVerifier, rekorURL string,
 		return nil, err
 	}
 	fmt.Fprintln(os.Stderr, "tlog entry created with index:", *entry.LogIndex)
-	return cbundle.EntryToBundle(entry), nil
+	return cbundle.EntryToBundle(entry, []byte{}, []byte{}, []byte{}, ""), nil
 }
 
 // nolint
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -79,7 +79,7 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
 			return nil, err
 		}
 		fmt.Fprintln(os.Stderr, "tlog entry created with index:", *entry.LogIndex)
-		signedPayload.Bundle = cbundle.EntryToBundle(entry)
+		signedPayload.Bundle = cbundle.EntryToBundle(entry, []byte{}, []byte{}, []byte{}, "")
 	}
 
 	// if bundle is specified, just do that and ignore the rest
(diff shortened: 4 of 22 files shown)


pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+	VerificationMaterial
+}
+
+// VerificationData contains extra data that can be used to verify things
+// such as transparency logs and timestamped verifications.
+type VerificationData struct {
+	// RekorPayload holds metadata about recording a Signature's ephemeral key to
+	// a Rekor transparency log.
+	// Use Payload instead of TransparencyLogEntry to keep backwards compatibility.
+	Payload RekorPayload
```
hectorj2f, line 32: I kept the `Payload` field instead of renaming it to `TransparencyLogEntry`, as specified in the protobuf spec, to keep backwards compatibility.

pkg/oci/static/options_test.go
```diff
@@ -87,7 +87,7 @@ func TestOptions(t *testing.T) {
 			LayerMediaType:  ctypes.SimpleSigningMediaType,
 			ConfigMediaType: types.OCIConfigJSON,
 			Annotations: map[string]string{
-				BundleAnnotationKey: "{\"SignedEntryTimestamp\":null,\"Payload\":{\"body\":null,\"integratedTime\":0,\"logIndex\":0,\"logID\":\"\"}}",
+				BundleAnnotationKey: "{\"Payload\":{\"body\":null,\"integratedTime\":0,\"logIndex\":0,\"logID\":\"\"},\"SignedEntryTimestamp\":null,\"EntryTimestampAuthority\":null,\"CertBytes\":null,\"PublicKeyIdentifier\":\"\"}",
```
hectorj2f, line 90: This is the main aspect of the new bundle. Whenever we start using a TSA the bundle will look more like:

```
"Bundle":{"Payload":{"body":null,"integratedTime":0,"logIndex":0,"logID":""},"SignedEntryTimestamp":null,"EntryTimestampAuthority":"MIIEbjADAgEAMIIEZQYJKoZIhvcNAQcCoIIEVjCCBFICAQExDTALBglghkgBZQMEAgEwgdQGCyqGSIb3DQEJEAEEoIHEBIHBMIG+AgEBBgkrBgEEAYO/MAIwMTANBglghkgBZQMEAgEFAAQgLFsLP5xZKtHz5LM1jdA+K9GQFwI20N5BnnKOOgsPpvcCFQDKNZaXbkFPJh0AD1YsJ27y+SllNxgTMjAyMjExMDcxNzEzNDQrMDEwMDADAgEBAhReYQUVtj8oPw/AXX8xVrnIZfq6Y6A0pDIwMDEOMAwGA1UEChMFbG9jYWwxHjAcBgNVBAMTFVRlc3QgVFNBIFRpbWVzdGFtcGluZ6CCAckwggHFMIIBaqADAgECAhRHCu9dHKS97mFo1cH5neJubRibujAKBggqhkjOPQQDAjAoMQ4wDAYDVQQKEwVsb2NhbDEWMBQGA1UEAxMNVGVzdCBUU0EgUm9vdDAeFw0yMjExMDMxMTUzMThaFw0zMTExMDMxMTU2MThaMDAxDjAMBgNVBAoTBWxvY2FsMR4wHAYDVQQDExVUZXN0IFRTQSBUaW1lc3RhbXBpbmcwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAATo3W6NQrpx5D8z5IvgD2DlAgoJMF4KPY9Pj4UfFhfOq029ryszXp3460Z7N+x86bDvyjVrHaeiPnl1HO9Q52zso2owaDAOBgNVHQ8BAf8EBAMCB4AwHQYDVR0OBBYEFHSIhDdTGIsodML/iUOhx7hgo/K7MB8GA1UdIwQYMBaAFBoZYijuouZCvKDtBd0eCyaU2HWoMBYGA1UdJQEB/wQMMAoGCCsGAQUFBwMIMAoGCCqGSM49BAMCA0kAMEYCIQCmPVr5kwYe4Jg9PGO6apgfzSrKAtESgNHpAbE3iIvJhQIhAJIGNxshJcC8LXHRrVWM77no3d3GguSvR01OAPZwE2pqMYIBmDCCAZQCAQEwQDAoMQ4wDAYDVQQKEwVsb2NhbDEWMBQGA1UEAxMNVGVzdCBUU0EgUm9vdAIURwrvXRykve5haNXB+Z3ibm0Ym7owCwYJYIZIAWUDBAIBoIHqMBoGCSqGSIb3DQEJAzENBgsqhkiG9w0BCRABBDAcBgkqhkiG9w0BCQUxDxcNMjIxMTA3MTYxMzQ0WjAvBgkqhkiG9w0BCQQxIgQgGQ4kDSvQ8UJgZaIIKWZg9R0L06St8OBAYrj2VGSq7rcwfQYLKoZIhvcNAQkQAi8xbjBsMGowaAQgXqxJD0nAgg6en9P1bRrU7+6tzxOMn3YThreg7uR6T7EwRDAspCowKDEOMAwGA1UEChMFbG9jYWwxFjAUBgNVBAMTDVRlc3QgVFNBIFJvb3QCFEcK710cpL3uYWjVwfmd4m5tGJu6MAoGCCqGSM49BAMCBEcwRQIhALiLTG7glyyiiHJ33DsWIi24rHpcSnW/EeLfNqaXty4VAiBG7moX2xoOwOfJm8DZveTox9tpcV6NnIN/3LzGs8eSgA==","CertBytes":null, "PublicKeyIdentifier": ""}
```

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+
```
hectorj2f, line 33: I am awaiting feedback before adding other fields such as the `messageSignature` included in the spec.

hectorj2f:
@asraa @priyawadhwa @haydentherapper I'd like to hear your thoughts about this proposal 👀 . I decided to adapt the current implementation to the fields in the protobuf spec. Also, I included a new field EntryTimestampAuthority that will contains the response as received from the TSA server.

haydentherapper:
If we are beginning to adopt the Sigstore bundle format, I would prefer we use exactly what has been discussed and approved. Otherwise we now have a third format effectively. If we want to deviate, we should discuss the change in the protobuf specs repo first imo

cc @kommendorkapten

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+// limitations under the License.
+
+package bundle
+
+import (
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+type Bundle struct {
+	VerificationData
```
priyawadhwa, line 22: What's the difference between `VerificationData` and `VerificationMaterial`? I wonder if we could come up with more descriptive names so it's more clear :)

priyawadhwa (commented):
this looks really good! just one question.

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+// signatures or any additional timestamped verifications.
+type VerificationMaterial struct {
+	// A chain of X.509 certificates.
+	CertBytes []byte
+
+	// PublicKeyIdentifier optional unauthenticated hint on which key to use.
+	PublicKeyIdentifier string
+}
+
+// TimestampVerificationData contains various timestamped data following RFC3161.
```
haydentherapper, line 48: As we discussed in the protobuf-specs repo, SETs are not just timestamp verification data, hence why we went with this nested format - https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L53-L59

pkg/cosign/bundle/bundle.go
asraa: It might be better to port that comment to the upstream bundle, but these are terms being used by those protos.
Data is extra auxiliary data for verification (timestamps, rekor entries), but verification material is the direct material (key hint, certificates)

pkg/cosign/bundle/bundle.go
asraa: Those fields overlap with `LocalSignedPayload` too. The top level fields in that would ideally be removed, or else it would open mismatch issues.

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// EntryTimestampAuthority contains the recorded entry from timestamp authority server.
+	EntryTimestampAuthority []byte
+}
+
+func EntryToBundle(tLogEntry *models.LogEntryAnon, signedEntryTimestamp, entryTimestampAuthority, certBytes []byte, pubKeyID string) *Bundle {
+	b := &Bundle{}
+	// If none of the verification data is configured then return nil
+	if (tLogEntry == nil || tLogEntry.Verification == nil) && len(entryTimestampAuthority) == 0 {
+		return nil
```
asraa, line 61: why return nil over returning an empty bundle? id be worried about accidental dereferences later

hectorj2f: I agree with you but `tLogEntry == nil || tLogEntry.Verification == nil` is reusing the old behavior that also returned a nil bundle.

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+// TimestampVerificationData contains various timestamped data following RFC3161.
+type TimestampVerificationData struct {
+	// SignedEntryTimestamp holds metadata about a timestamped counter signature over the artifacts signature.
+	SignedEntryTimestamp []byte
+
+	// EntryTimestampAuthority contains the recorded entry from timestamp authority server.
+	EntryTimestampAuthority []byte
+}
+
+func EntryToBundle(tLogEntry *models.LogEntryAnon, signedEntryTimestamp, entryTimestampAuthority, certBytes []byte, pubKeyID string) *Bundle {
```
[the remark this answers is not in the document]
hectorj2f: Yes, we need tests but I wanted to firstly get some feedback, as I expected things might change.

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+	PublicKeyIdentifier string
+}
+
+// TimestampVerificationData contains various timestamped data following RFC3161.
+type TimestampVerificationData struct {
+	// SignedEntryTimestamp holds metadata about a timestamped counter signature over the artifacts signature.
+	SignedEntryTimestamp []byte
+
+	// EntryTimestampAuthority contains the recorded entry from timestamp authority server.
+	EntryTimestampAuthority []byte
```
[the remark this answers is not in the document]
hectorj2f: Yes, it makes sense. I like it.

pkg/cosign/bundle/bundle.go
hectorj2f: Good point! That could easily happen.

pkg/cosign/bundle/bundle.go
hectorj2f: I am gonna port these and other questions to the upstream bundle. Thanks.

pkg/cosign/bundle/bundle.go
hectorj2f: I created one issue here https://github.com/sigstore/protobuf-specs/issues/13. Hopefully we can get more ppl involved on this discussion.

pkg/cosign/bundle/bundle.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// TimestampVerificationData holds metadata about a timestamped verification.
+	TimestampVerificationData
+}
+
+// VerificationMaterial captures details on the materials used to verify
+// signatures or any additional timestamped verifications.
+type VerificationMaterial struct {
+	// A chain of X.509 certificates.
+	CertBytes []byte
```
znewman01, line 42: I'd expect this to be a `crypto/x509.Certificate`

I guess really...is this struct a *data format* or an object? My preference is always to move parsing as early as possible so we don't have to handle parse errors in the middle of the program logic.

znewman01 (commented):
This is a very-much breaking change to the Golang API. Can we make it official that this API isn't supported by any versioning guarantees?

See also https://github.com/sigstore/cosign/discussions/2365#discussioncomment-3925982

kommendorkapten:
The Sigstore bundle from https://github.com/sigstore/protobuf-specs is in a decent shape I would say. Is it worthy to cut a release in the current stage? Probably no, but I don't think we are miles away. For the [npm build provenance](https://github.com/npm/rfcs/pull/626) we are working on two different implementations of the bundle, generation and verification. One in Go, and one in JavaScript. So we are getting real world testing of the usability _now_. My feeling is that we will see some small changes in the imminent future, like https://github.com/sigstore/protobuf-specs/pull/10, and cut a release within a few weeks. All the work on the bundle we are doing is expected to be part of [sigstore-js](https://github.com/sigstore/sigstore-js) and [sigstore-go](https://github.com/sigstore/sigstore-go).

Beyond the first release, there are a few expected breaking changes, but the horizon for them is a bit unclear, as we are waiting for e.g. Rekor SET V2. 
The bundle is versioned, so it shouldn't be an issue for clients to support multiple versions. If `cosign` waits until the next bundle release for adoption, I think that can be reasonable too, as long as the this PR doesn't stray away too much and makes the adoption harder.

hectorj2f:
Thanks everyone for the review and suggestions. I'll keep this PR open so I can take over on these changes whenever there aren't major breaking changes expected from the protobuf spec.

In the meantime, I'll try to move on with https://github.com/sigstore/cosign/issues/2331, and open a different PR with a smaller change that just renames the struct name from `RekorBundle to Bundle`, and adds a new field `EntryRFC3161` to record the timestamped entry from `sigstore/timestamp-authority`.

kommendorkapten:
If it's of any help, we are ready to merge this PR https://github.com/sigstore/protobuf-specs/pull/12 which adds the generated Go code to the repository, so you don't need to rewrite any structs -- you can import them directly from that repository.

hectorj2f:
Closing this for now! I'll re-open it when starting this work again.

Issue #2419, opened 2022-11-07
Remove non-GA APIs from experimental sign-blob verification flow


patflynn:
**Description**

This issue is to ensure we remove the use of `/api/v1/index/retrieve` from cosign [verify-blob flow](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L235) before we launch this version of the command to stable. 

Another option is to get Rekor to commit to supporting the API.

asraa:
I think there were some users who were still relying on this flow. I agree, we discussed this at some point during the cosign blob verification outputs. I can put out a quick PR to do that, we should label in release notes HEAVILY. 

For e.g. here were users failing when that endpoint broke: https://github.com/sigstore/cosign/issues/1406

haydentherapper:
We can remove it for 2.0. But it would be worth understanding what the customer workflows are and if we can support those in other ways

asraa:
GoReleaser updated docs to pass in the signing cert, at least that is done.

PR #2425, opened 2022-11-08
BREAKING: Remove non-GA APIs from sign-blob verification

asraa:

Fixes https://github.com/sigstore/cosign/issues/2419

Requires users to provide a bundle or a certificate and signature so that we can recreate the Rekor entry and not search by payload hash using the non-GA Rekor endpoint.

#### Release Note

* verify-blob: Removes the use of a non-GA API from sign-blob. Users are required to provide a `--bundle` OR both the `--signature` and `--certificate`/`--key` to the `verify-blob` command.


haydentherapper (commented):
YAY

znewman01 (commented):
Love to see it:
Screen Shot 2022-11-08 at 2 03 03 PM
 



lint failure is shockingly actually related to the change for once

hectorj2f (approved):

PR #2427, opened 2022-11-09
Add --offline flag to force offline verification

priyawadhwa:


pkg/cosign/verify.go
```diff
@@ -677,6 +680,11 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 		return false, fmt.Errorf("unable to verify bundle: %w", err)
 	}
 
+	// If the --offline flag was specified, fail here
+	if !bundleVerified && co.Offline {
```
asraa, line 684: i think the line above needs to be changed, or else the bundle verification error output would not make sense
```
	if err != nil && co.RekorClient == nil {
```

like if the bundle is not verified and we have a RekorClient set (which can still happen with offline mode so far), then the bundle verification error doesn't get propogated.

asraa, line 684: Right now `VerifyBundle` checks if there's a bundle AND verifies the bundle, and right now it seems like `err` is returned if there is an error checking, but `nil` error can also be returned when there is NO bundle. 

Maybe it can be changed to:
* if hasBundle(sig) -> verifyBundle -> if err, return err (always fail if the provided bundle is bad)
* if offline -> return now, offline verification failed
* check online

@znewman01 make sense?

asraa: Arguably you shouldn't have  RekorClient set in offline, so the two checks might merge if you enforce that.

haydentherapper: It seems like currently, setting `rekor-url` to `""` is equivalent to an offline flag. Do we want to remove that assumption? I think it should be an error to set `rekor-url=""`, and that the source of truth for whether we do offline or not will be this new `offline` flag

haydentherapper: +1 to this flow!

test/e2e_test.go
```diff
@@ -1566,3 +1567,72 @@ func TestAttestBlobSignVerify(t *testing.T) {
...
+
+	imgRef, _, cleanup := mkimage(t, img1)
+	defer cleanup()
+
+	_, privKeyPath, pubKeyPath := keypair(t, td)
+
+	ctx := context.Background()
+
+	// Sign image1 and store the entry in rekor
+	// remoteOpts := ociremote.WithRemoteOptions(registryClientOpts(ctx)...)
```
vaikas, line 1586: nit: remove this?

test/e2e_test.go
```diff
@@ -1566,3 +1567,72 @@ func TestAttestBlobSignVerify(t *testing.T) {
...
+		SkipConfirmation: true,
+	}
+	must(sign.SignCmd(ro, ko, so, []string{img1}), t)
+	// verify image1 online and offline
+	must(verify(pubKeyPath, img1, true, nil, ""), t)
+	verifyCmd := &cliverify.VerifyCommand{
+		KeyRef:      pubKeyPath,
+		Offline:     true,
+		CheckClaims: true,
+		RekorURL:    "https://rekor.sigstore.dev",
```
vaikas, line 1600: nit: is this necessary for offline?

pkg/cosign/verify.go
priyawadhwa: yah since rekor-url has a default value RekorClient is pretty much always set (maybe people set rekor-url="" but i haven't really seen evidence of that). 

> the source of truth for whether we do offline or not will be this new offline flag
+1

i can update the error so that it propagates.

test/e2e_test.go
priyawadhwa: nope, thanks!

pkg/cosign/verify.go
priyawadhwa: sgtm, updated

priyawadhwa:
@asraa ptal whenever you can! :)

pkg/cosign/verify.go
znewman01: wait, is it updated? I don't see it

priyawadhwa: @znewman01 i believe it's updated here

znewman01: maybe a GH UI hiccup, weird. thx!

znewman01 (commented):
This LGTM but will let someone else pull the trigger on approval

pkg/cosign/verify.go
asraa: I think it is now. I hate that `VerifyBundle` has the double failure meaning (no bundle vs bad bundle), but it checks out now.

If there's no bundle we'll return false, nil.

asraa (approved):

PR #2435, opened 2022-11-10
Rename RekorBundle type to Bundle

hectorj2f:

#### Summary

As a result of conversations in https://github.com/sigstore/cosign/pull/2422, we could rename the RekorBundle struct to Bundle. This bundle struct will start including fields to handle different verification policies such Rekor entries and/or RFC3161 timestamped entries.

### DIFF
--- a/pkg/cosign/bundle/bundle.go
+++ b/pkg/cosign/bundle/bundle.go
@@ -16,9 +16,9 @@ package bundle
 
 import "github.com/sigstore/rekor/pkg/generated/models"
 
-// RekorBundle holds metadata about recording a Signature's ephemeral key to
+// Bundle holds metadata about recording a Signature's ephemeral key to
 // a Rekor transparency log.
-type RekorBundle struct {
+type Bundle struct {
 	SignedEntryTimestamp []byte
 	Payload              RekorPayload
 }
@@ -30,11 +30,11 @@ type RekorPayload struct {
 	LogID          string      `json:"logID"`
 }
 
-func EntryToBundle(entry *models.LogEntryAnon) *RekorBundle {
+func EntryToBundle(entry *models.LogEntryAnon) *Bundle {
 	if entry.Verification == nil {
 		return nil
 	}
-	return &RekorBundle{
+	return &Bundle{
 		SignedEntryTimestamp: entry.Verification.SignedEntryTimestamp,
 		Payload: RekorPayload{
 			Body:           entry.Body,
--- a/pkg/cosign/fetch.go
+++ b/pkg/cosign/fetch.go
@@ -34,13 +34,13 @@ type SignedPayload struct {
 	Payload         []byte
 	Cert            *x509.Certificate
 	Chain           []*x509.Certificate
-	Bundle          *bundle.RekorBundle
+	Bundle          *bundle.Bundle
 }
 
 type LocalSignedPayload struct {
-	Base64Signature string              `json:"base64Signature"`
-	Cert            string              `json:"cert,omitempty"`
-	Bundle          *bundle.RekorBundle `json:"rekorBundle,omitempty"`
+	Base64Signature string         `json:"base64Signature"`
+	Cert            string         `json:"cert,omitempty"`
+	Bundle          *bundle.Bundle `json:"Bundle,omitempty"`
 }
 
 type Signatures struct {
--- a/pkg/cosign/verify_test.go
+++ b/pkg/cosign/verify_test.go
@@ -204,7 +204,7 @@ func signEntry(ctx context.Context, t *testing.T, signer signature.Signer, entry
 	return signature
 }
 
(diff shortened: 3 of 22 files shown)


cpanato (dismissed):

pkg/cosign/bundle/bundle.go
```diff
@@ -16,9 +16,9 @@ package bundle
 
 import "github.com/sigstore/rekor/pkg/generated/models"
 
-// RekorBundle holds metadata about recording a Signature's ephemeral key to
+// Bundle holds metadata about recording a Signature's ephemeral key to
```
haydentherapper, line 19: Comment needs to be updated

haydentherapper:
Can I throw out another question? What about making another annotation on the OCI image rather than try to fit this into the Rekor bundle? I feel like that might be cleaner, and wouldn’t affect the existing bundle format. 

cc @asraa @znewman01

priyawadhwa:
I think either option would work fine, but not strongly opinionated on this one since we plan on ripping it out soon and replacing it either way.

pkg/cosign/fetch.go
```diff
@@ -34,13 +34,13 @@ type SignedPayload struct {
...
+	Bundle          *bundle.Bundle
 }
 
 type LocalSignedPayload struct {
-	Base64Signature string              `json:"base64Signature"`
-	Cert            string              `json:"cert,omitempty"`
-	Bundle          *bundle.RekorBundle `json:"rekorBundle,omitempty"`
+	Base64Signature string         `json:"base64Signature"`
+	Cert            string         `json:"cert,omitempty"`
+	Bundle          *bundle.Bundle `json:"Bundle,omitempty"`
```
priyawadhwa, line 43: nit: lowercase bundle

```
json:"bundle,omitempty"`
```

pkg/cosign/verify_test.go
```diff
@@ -245,9 +245,9 @@ func TestVerifyImageSignatureWithNoChain(t *testing.T) {
 	pe, _ := proposedEntry(base64.StdEncoding.EncodeToString(signature), payload, pemLeaf)
 	entry, _ := rtypes.UnmarshalEntry(pe[0])
 	leaf, _ := entry.Canonicalize(ctx)
-	rekorBundle := CreateTestBundle(ctx, t, sv, leaf)
+	Bundle := CreateTestBundle(ctx, t, sv, leaf)
```
priyawadhwa, line 248: nit: maybe call this testBundle to conform with go style guidelines?

priyawadhwa (commented):
Looks great, just a couple nits

asraa:
I would also vote for an "experimental" new annotation: we dont know if the new bundle format is totally stable yet, so we'll have issues with  backwards compatibility if we start using it pre-emptyively without any guarantees..

priyawadhwa:
> we dont know if the new bundle format is totally stable yet, so we'll have issues with backwards compatibility if we start using it pre-emptyively without any guarantees..

I don't think this PR is using the new bundle format, it's just refactoring the existing one.

hectorj2f:
Just to clarify, i don't intend any big refactoring. I wanted to rename this object to Bundle and add a new field for the timestamp entry. I believe we are open to add new fields without considering breaking changes, right?

asraa:
Oh, got it. Sorry I guess that's fine.

I think renaming is breaking... I'm sorry. I'm not sure, I think I'm confused and not sure the best way to make this change. It's either add in the old logic, or begin implementing the new bundle

priyawadhwa:
@asraa it's pretty confusing i think we've been using "bundle" to refer to a few different things 😅 

Basically I think the purpose of these PRs is just to add in support for the TSA, and it's modifying the existing bundle format to accomplish that since the new protobuf-based bundle format is [a few weeks away](https://github.com/sigstore/cosign/pull/2422#issuecomment-1308349779). The plan is once the new one is ready we'll replace this. I'd prefer not to block the TSA feature on the protobuf bundle work though, just in case there are any unexpected delays.

I think this PR in particular is a breaking change in the Golang API, but I don't think it's too big of a deal since it'll be replaced soon anyway (and we'll have to replace the RekorBundle type at that point anyway).

haydentherapper:
I would prefer we don’t have an intermediate step where we create this bundle format and then turn it into another bundle struct. I think this adds in more confusion cause now we have to consider which version of the bundle we are referring to.

What’s the cost of implementing this as its own annotation?

haydentherapper:
Just thinking through how we’ve explained the bundle annotation so far, we’ve always referred to it as the Rekor response. I would prefer we don’t overload it with things that aren’t related to Rekor. 

When we implement the new bundle format, we will probably want to use a new annotation anyways.

hectorj2f:
Closing this PR! I'll change the approach and add all the code changes to handle a new annotation. I'll also add cli flags when using a timestamp authority server and TSA bundle.

Thanks for the feedback and discussion.

PR #2439, opened 2022-11-10
Add a new time-stamping bundle to record rfc3161 verification data

hectorj2f:
#### Summary

This change is related to https://github.com/sigstore/cosign/issues/2331 and a continuation of the work started in https://github.com/sigstore/cosign/pull/2435. We aim to add a new bundle to keep the existing RekorBundle due to backwards compatibility issues. This new TSABundle will be used in a different annotation. This TSABundle stores the response of an entry from a time-stamping RFC3161 server encoded in base64. The goal is to reuse this field together with other data to verify an artifact signature against a time-stamping server.

### DIFF
--- a/pkg/cosign/bundle/tsa.go
+++ b/pkg/cosign/bundle/tsa.go
@@ -1,0 +1,33 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundle
+
+// TSABundle holds metadata about timestamp RFC3161 verification data.
+type TSABundle struct {
+	// SignedRFC3161Timestamp contains a RFC3161 signed timestamp provided by a time-stamping server.
+	// Clients MUST verify the hashed message in the message imprint
+	// against the signature in the bundle. This is encoded as base64.
+	SignedRFC3161Timestamp []byte
+}
+
+// TimestampToTSABundle receives a base64 encoded RFC3161 timestamp.
+func TimestampToTSABundle(timestampRFC3161 []byte) *TSABundle {
+	if timestampRFC3161 != nil {
+		return &TSABundle{
+			SignedRFC3161Timestamp: timestampRFC3161,
+		}
+	}
+	return nil
+}
--- a/pkg/cosign/bundle/tsa_test.go
+++ b/pkg/cosign/bundle/tsa_test.go
@@ -1,0 +1,48 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundle
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/go-openapi/strfmt"
+)
+
+func TestTSABundle(t *testing.T) {
+	testCases := []struct {
+		name                  string
+		timestampRFC3161Entry []byte
+		expectedTSABundle     *TSABundle
+	}{{
+		name:                  "nil timestamp entry",
+		timestampRFC3161Entry: nil,
+		expectedTSABundle:     nil,
+	}, {
+		name:                  "timestamp entry",
+		timestampRFC3161Entry: strfmt.Base64([]byte("signature")),
+		expectedTSABundle: &TSABundle{
+			SignedRFC3161Timestamp: strfmt.Base64([]byte("signature")),
+		},
+	}}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/cosign/bundle/bundle.go
```diff
@@ -21,6 +21,9 @@ import "github.com/sigstore/rekor/pkg/generated/models"
 type Bundle struct {
 	SignedEntryTimestamp []byte
 	Payload              RekorPayload
+
+	// RFC3161Entry contains the recorded entry from timestamp authority servers.
+	RFC3161Entry []byte
```
haydentherapper, line 26: Will this be the raw bytes, or base64 encoded like the SET?

haydentherapper, line 26: the comment should clarify the format, I’d recommend the same comment we have discussed in the protobuf specs

hectorj2f: I was thinking of raw bytes.

I haven't find any base64 reference for the fields in the specs 🤔. But I am open to encode/decode these bytes to base64.
 https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto
 https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto

znewman01: I think @haydentherapper's point was that the `SignedEntryTimestamp []byte` above is actually base64-encoded (I think?). I don't think it *should* be, but putting a new, non-base64-encoded value side-by-side with it in the `Bundle` struct makes it confusing.

Also, when we're dumping this to the OCI registry, do we have to base64-encode at that point? 

Can we just make sure that comments make this all crystal clear?

znewman01:
The thing that concerns me out about adding something to the bundle here is that we're committing to supporting it indefinitely, right? If somebody signs with Cosign version (N+1), which has support for this, and we soon want to rip it out, we can't—we're going to need to verify Cosign version (N+1) bundles in Cosign version (N+2) and even version (N+100). So doing this creates a permanent maintenance burden for functionality that we want to replace almost immediately.

That doesn't mean we shouldn't do it—but I want to be sure that the benefit of supporting this feature a month early outweighs the permanent maintenance burden. (One argument *for* merging is that we *already have a maintenance burden  associated with the legacy bundle format anyway, so this is a pretty minor thing to keep around. Further, the code we'll need to keep around is pretty simple—it'll just parse the old bundle format into the new one.)

haydentherapper:
BTW, check out https://github.com/sigstore/cosign/pull/2435#issuecomment-1310824569, we agreed that this will be implemented in a new annotation/separate flag rather than in the existing bundle

hectorj2f:
@haydentherapper I added the new bundle so we can add a specific annotation for this bundle.

haydentherapper:
Will take a look Monday, thanks!

priyawadhwa (dismissed):
lgtm, thanks @hectorj2f !

pkg/cosign/bundle/tsa.go
```diff
@@ -0,0 +1,34 @@
...
+package bundle
+
+// TSABundle holds metadata about timestamp RFC3161 verification data.
+type TSABundle struct {
+	// SignedRFC3161Entry contains a RFC3161 signed timestamps provided by a time-stamping server.
+	// Clients MUST verify the hashed message in the message imprint
+	// against the signature in the bundle. This is encoded as base64.
+	SignedRFC3161Entry []byte
+
+	// TODO: Additional fields fir verification data such as CertBytes
```
haydentherapper, line 24: Do we need this TODO? Verification data should be passed out of band.

pkg/cosign/bundle/tsa.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func EntryToTSABundle(timestampRFC3161Entry []byte) *TSABundle {
```
haydentherapper, line 27: Clarify that this takes in a base64 encoded 3161 timestamp?

haydentherapper, line 27:
```suggestion
func TimestampToTSABundle(timestampRFC3161 []byte) *TSABundle {
```

haydentherapper: I don't think we should use the term 'entry' given the timestamp isn't admitted into a log

pkg/cosign/bundle/tsa.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+package bundle
+
+// TSABundle holds metadata about timestamp RFC3161 verification data.
+type TSABundle struct {
+	// SignedRFC3161Entry contains a RFC3161 signed timestamps provided by a time-stamping server.
+	// Clients MUST verify the hashed message in the message imprint
+	// against the signature in the bundle. This is encoded as base64.
+	SignedRFC3161Entry []byte
```
haydentherapper, line 22:
```suggestion
	SignedRFC3161Timestamp []byte
```

haydentherapper: same here, as these aren't entries

pkg/cosign/bundle/tsa.go
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundle
+
+// TSABundle holds metadata about timestamp RFC3161 verification data.
+type TSABundle struct {
+	// SignedRFC3161Entry contains a RFC3161 signed timestamps provided by a time-stamping server.
```
haydentherapper, line 19:
```suggestion
	// SignedRFC3161Timestamp contains a RFC3161 signed timestamp provided by a time-stamping server.
```

pkg/cosign/bundle/tsa.go
hectorj2f: Whenever it gets stored in a tlog, it'll make more sense. Also we could suffix it with response as it contains information about the raw taken, hashed message, time...

hectorj2f: done

pkg/cosign/bundle/tsa.go
hectorj2f: I am considering adding one of the certs used in the chain. But I wanted to discussed it before. I'm removing this comment for now.

hectorj2f:
@haydentherapper I've addressed all your comments. I'm going to merge this PR when ready, so I can move on with other PRs.

cpanato (approved):

PR #2449, opened 2022-11-11
Better feedback if specifying incompatible argument on `cosign sign --attachment`

ChaosInTheCRD:
Signed-off-by: chaosinthecrd <thomas.meadows@jetstack.io>

This is a tiny change, but when using the `cosign sign --attachment` flag, it didn't seem clear why I was getting an error back. Of course the reason why is that the only attachment string it will accept is `sbom`, which is making me wonder whether at least temporarily this would be better as a boolean flag `--sbom-attachment` until there are other attachment types.

I have changed the error message, just as a suggestion as I feel it makes it more clear as to what is going on as opposed to 

```
Error: flag: help requested
main.go:62: error during command execution: flag: help requested
```

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -16,7 +16,6 @@
 package cli
 
 import (
-	"flag"
 	"fmt"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


znewman01 (approved):
Lint is related, annoyingly. Sorry!

Otherwise LGTM

ChaosInTheCRD:
hopefully that's good to go 😄

znewman01 (approved):

PR #2454, opened 2022-11-14
Add TSA bundle annotation

hectorj2f:

#### Summary

This change is related to https://github.com/sigstore/cosign/issues/2331. This PR adds a new annotation for the TSA Bundle.

### DIFF
--- a/pkg/oci/internal/signature/layer.go
+++ b/pkg/oci/internal/signature/layer.go
@@ -33,7 +33,7 @@ const (
 	certkey      = "dev.sigstore.cosign/certificate"
 	chainkey     = "dev.sigstore.cosign/chain"
 	BundleKey    = "dev.sigstore.cosign/bundle"
-	TSABundleKey = "dev.sigstore.cosign/bundle"
+	TSABundleKey = "dev.sigstore.cosign/3161timestamp"
 )
 
 type sigLayer struct {
--- a/pkg/oci/internal/signature/layer.go
+++ b/pkg/oci/internal/signature/layer.go
@@ -29,10 +29,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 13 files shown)


pkg/oci/internal/signature/layer.go
```diff
@@ -29,10 +29,11 @@ import (
...
 const (
-	sigkey    = "dev.cosignproject.cosign/signature"
-	certkey   = "dev.sigstore.cosign/certificate"
-	chainkey  = "dev.sigstore.cosign/chain"
-	BundleKey = "dev.sigstore.cosign/bundle"
+	sigkey       = "dev.cosignproject.cosign/signature"
+	certkey      = "dev.sigstore.cosign/certificate"
+	chainkey     = "dev.sigstore.cosign/chain"
+	BundleKey    = "dev.sigstore.cosign/bundle"
+	TSABundleKey = "dev.sigstore.cosign/bundle"
```
priyawadhwa, line 36: I think you might have copy/pasted BundleKey here!

priyawadhwa (commented):
nice! just left one comment.

haydentherapper (commented):
Nice work!

pkg/oci/internal/signature/layer.go
hectorj2f: Good catch! I named here as `tsabundle` https://github.com/sigstore/cosign/pull/2454/files#diff-a4859b65c42ca40ac004766c29a8b1279097c2d7909552f77885af38876729c8R36. But I am not good at naming, so we can use 3161timestamp instead.

priyawadhwa (approved):
lgtm, tests looks great!

Issue #2255, opened 2022-09-15
COSIGN_EXPERIMENTAL and offline mode


znewman01:
`COSIGN_EXPERIMENTAL` was introduced AFAICT for two reasons:

1. The Sigstore *idea*, *interface*, and *implementation* were still experimental
2. The Sigstore *infrastructure* wasn't reliable/didn't have guarantees.

So to start, anything that touched Sigstore infrastructure (TUF roots, Fulcio, Rekor) was gated: users needed to set `COSIGN_EXPERIMENTAL=1` to use them.

Over time, we've blurred the line a little; without turning on experimental mode:

- You can pass a Rekor bundle and cosign will check it.
- To do that, you need the Rekor public key, which you get via TUF (this goes over the network!)
- Similarly, you can provide a cert signed by Fulcio, and cosign will check it against Fulcio's key
-  It gets that key via TUF

There's a question about "what is the point of `cosign verify-blob` without `COSIGN_EXPERIMENTAL` but with a provided certificate?"

I don't want to answer that, at least not here. It's hopefully going to be an obsolete question soon: we're relatively comfortable with the idea of Sigstore, and we're targeting general availability (GA) for the infrastrcucture soon: https://github.com/sigstore/fulcio/issues/766 and https://github.com/sigstore/rekor/issues/1005 . **But**, at least some folks have been deliberately not providing this flag to indicate "don't try to hit the network." That's *sort-of* what it does right now: "I don't try to query Rekor/Fulcio, but I *know* about them, and also I'll get their keys via TUF."

That suggests to me that, post-dropping-`COSIGN_EXPERIMENTAL`, there should be some mechanism to verify offline. Maybe a `--offline` flag, or even a `verify-offline` command (which is nice because now we can *only* show users the relevant flags). If we do that, we probably want to have some requirement on how out-of-date your TUF roots can be (maybe user-configurable?) and instructions on how to grab them from `cosign initialize`.

haydentherapper:
+1 to an offline flag! Here's a copy paste of a comment I left on a doc when we were looking into the Cosign CVE:

With no flag, I'd propose the behavior be:
* Default to fetching from Rekor without any flags
* To avoid a call to Rekor, either
--- provide a bundle
--- add a flag explicitly opting out of the rekor call

znewman01:
+1 to that as the default behavior.

Then, the additional changes for an offline mode:

- use cached TUF state (after checking their age), which gives you Rekor/Fulcio root certs
[...]

PR #2457, opened 2022-11-14
Remove experimental mode from sign-blob and verify-blob

priyawadhwa:

Ref #2255 

#### Release Note
Remove experimental mode from sign-blob and verify-blob


asraa (approved):

haydentherapper (approved):
Nice!!

PR #2460, opened 2022-11-15
add support for tsa signing and verification of images

hectorj2f:

#### Summary

In relation to https://github.com/sigstore/cosign/issues/2331, this PR adds support for image signing and verification using a RFC3161 time-stamping server **instead** of the transparency log.

NOTE: Rekor and TSA cannot be used together for the moment.

For testing purposes, we have integrated using https://github.com/sigstore/timestamp-authority. So we expect you to run a TSA server for testing:

```
$ curl http://localhost:3000/api/v1/timestamp/certchain > ts_chain.pem

$ cosign sign <YOUR_IMAGE> --tlog-upload=false --timestamp-server-url=http://localhost:3000

$ cosign verify <YOUR_IMAGE> --timestamp-server-url=http://localhost:3000 --timestamp-cert-chain=ts_chain.pem
```

#### Release Note

Add support for signing images using a RFC3161 time-stamping server


hectorj2f:
@priyawadhwa @haydentherapper I am considering to rename any reference of TSABundle to 3161Timestamp. Any thoughts ?.

pkg/cosign/verify.go
```diff
@@ -928,6 +946,43 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, rekorClient *client.Re
...
+	} else if bundle == nil {
+		return false, nil
+	}
+
+	b64Sig, err := sig.Base64Signature()
+	if err != nil {
+		return false, fmt.Errorf("reading base64signature: %w", err)
+	}
+
+	fmt.Println("Verifying TSA Bundle")
```
hectorj2f, line 962: Perhaps we want to avoid printing this message (it will be printed per validated signature).

znewman01: If you have it, it should go to STDERR

pkg/cosign/verify.go
```diff
@@ -928,6 +946,43 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, rekorClient *client.Re
...
+	}
+	certPool := x509.NewCertPool()
+	ok := certPool.AppendCertsFromPEM(pemBytes)
+	if !ok {
+		return false, fmt.Errorf("error parsing response into Timestamp while appending certs from PEM")
+	}
+
+	err = tsaverification.VerifyTimestampResponse(bundle.SignedRFC3161Timestamp, bytes.NewReader(sigBytes), certPool)
+	if err != nil {
+		return false, fmt.Errorf("unable verifyTSRWithPEM: %w", err)
```
hectorj2f, line 980: I need to fix this error msg.

cmd/cosign/cli/sign/sign.go
```diff
@@ -231,7 +234,16 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 	if sv.Cert != nil {
 		s = ifulcio.NewSigner(s, sv.Cert, sv.Chain)
 	}
-	if ShouldUploadToTlog(ctx, ko, digest, ko.SkipConfirmation, tlogUpload) {
+
+	// TODO: For the moment you can only use the timestamped service OR the transparency log.
+	if ko.TSAServerURL != "" {
```
haydentherapper, line 239: I would prefer to support both and not have this as a TODO. Timestamps are simply a way to have a third-party provide time, but don't provide transparency. Having this as a TODO risks changing the security model of Sigstore.

Following the existing code, can we have the `tsa.NewSigner` be wrapped by `irekor.NewSigner`?

cc @asraa

cmd/cosign/cli/options/verify.go
```diff
@@ -76,6 +79,12 @@ func (o *VerifyOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&o.LocalImage, "local-image", false,
 		"whether the specified image is a path to an image saved locally via 'cosign save'")
+
+	cmd.Flags().StringVar(&o.TSAServerURL, "timestamp-server-url", "",
+		"url to a timestamp RFC3161 server, default none")
+
+	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-cert-chain", "",
```
haydentherapper, line 86: Can you clarify that this should contain the entire chain from the TSA certificate leaf at index 0 to the root at index n

internal/pkg/cosign/tsa/signer.go
```diff
@@ -0,0 +1,137 @@
...
+	tsaclient "github.com/sigstore/timestamp-authority/pkg/generated/client"
+	ts "github.com/sigstore/timestamp-authority/pkg/generated/client/timestamp"
+)
+
+func GetTimestampedSignature(sigBytes []byte, tsaClient *tsaclient.TimestampAuthority) ([]byte, error) {
+	requestBytes, err := createTimestampAuthorityRequest(sigBytes, crypto.SHA256, "")
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println("Calling TSA authority ...")
```
haydentherapper, line 45: Remove?

internal/pkg/cosign/tsa/signer.go
```diff
@@ -0,0 +1,137 @@
...
+	params := ts.NewGetTimestampResponseParams()
+	params.SetTimeout(time.Second * 30)
```
haydentherapper, line 47: Can we reduce this, this seems like a long timeout?

internal/pkg/cosign/tsa/signer.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, fmt.Errorf("unable to get the response: %w", err)
+	}
+
+	// validate that timestamp is parseable
+	ts, err := timestamp.ParseResponse(respBytes.Bytes())
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Fprintln(os.Stderr, "Timestamp authority entry created with time:", ts.Time)
```
haydentherapper, line 62:
```suggestion
	fmt.Fprintln(os.Stderr, "Timestamp fetched with time: ", ts.Time)
```

internal/pkg/cosign/tsa/signer.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, nil, err
+	}
+
+	return newSig, pub, nil
+}
+
+func createTimestampAuthorityRequest(artifactBytes []byte, hash crypto.Hash, policyStr string) ([]byte, error) {
+	reqOpts := &timestamp.RequestOptions{
+		Hash:         hash,
+		Certificates: true, // if the timestamp response should contain a certificate chain
```
haydentherapper, line 110:
```suggestion
		Certificates: true, // if the timestamp response should contain the timestamp certificate
```

haydentherapper: small nit, the chain is not included, only the leaf

internal/pkg/cosign/tsa/signer_test.go
```diff
@@ -0,0 +1,95 @@
...
+		t.Fatalf("cosign.GeneratePrivateKey() failed: %v", err)
+	}
+	s, err := signature.LoadECDSASignerVerifier(priv, crypto.SHA256)
+	if err != nil {
+		t.Fatalf("signature.LoadECDSASignerVerifier(key, crypto.SHA256) failed: %v", err)
+	}
+	return s
+}
+
+func TestSigner(t *testing.T) {
```
haydentherapper, line 48: Nice tests!

pkg/cosign/verify.go
```diff
@@ -928,6 +946,43 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, rekorClient *client.Re
...
+	}
+
+	return true, nil
```
haydentherapper, line 983: We also need to verify the certificate expiration as part of verification, like what we do in the rekor bundle - https://cs.github.com/sigstore/cosign/blob/44a7117e0cee495a163c92e9f347efb2d9b747c7/pkg/cosign/verify.go#L907

pkg/cosign/verify.go
```diff
@@ -674,6 +683,15 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 			return bundleVerified, err
 		}
 	}
+	// TODO: Verify TSA only and return the result with the tsa bundle
```
haydentherapper, line 686: Like what I mentioned above, the timestamp augments the Rekor bundle rather than supplants it. It's a way to fetch a timestamp from another source. I'd like the verification flow to look more like what's been done in verifyblob, but we should do that refactor outside of this PR imo. For now, I'd propose something like:
* Add in TS verification and check certificate expiration in VerifyTSABundle
* If verification fails, exit
* If succeeds, pass a flag to `VerifyBundle` that says "don't check the timestamp" ([here](https://cs.github.com/sigstore/cosign/blob/44a7117e0cee495a163c92e9f347efb2d9b747c7/pkg/cosign/verify.go?q=verifybundle+org%3Asigstore#L907)) since it was already checked during TS verification. And also [here](https://cs.github.com/sigstore/cosign/blob/44a7117e0cee495a163c92e9f347efb2d9b747c7/pkg/cosign/verify.go?q=tlogValidateCertificate+org%3Asigstore#L462) when calling Rekor online, we don't need to check expiration if the TS is valid and expiration passes

I think we should still check Rekor/bundle if a timestamp is provided, unless we add flags to control this behavior to skip a Rekor check.

cc @asraa

haydentherapper (changes requested):

cmd/cosign/cli/options/verify.go
```diff
@@ -132,6 +141,9 @@ type VerifyBlobOptions struct {
 	Rekor               RekorOptions
 	Registry            RegistryOptions
 	CommonVerifyOptions CommonVerifyOptions
+
+	TSAServerURL     string
```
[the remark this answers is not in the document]
hectorj2f: I was doubting to add these flags without supporting its functionality for the other verify commands.

internal/pkg/cosign/tsa/signer.go
hectorj2f: I believe this is the default timeout value used by the timestamp-cli.

cmd/cosign/cli/sign/sign.go
hectorj2f: @haydentherapper @asraa Could we add support for both as part of a follow-up PR ? Likewise I will advocate for supporting the TSA as an alternative to the tlog. Some users may not want transparency.

internal/pkg/cosign/tsa/signer.go
haydentherapper: We should update that too. For the sigstore TSA we'll aim for <1s p99s, for third parties, <10 or <5 should be reasonable

cmd/cosign/cli/options/verify.go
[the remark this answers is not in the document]
priyawadhwa: i'm pretty sure all verify commands end up calling `verifyInternal` where the actual TSA verification happens, so adding in the flags should just work!

cmd/cosign/cli/sign/sign.go
haydentherapper: As long as the follow up is before we cut a new release, sure. Is it a large change to support both? I would have thought we could support both by just wrapping the tsa signer in the rekor signer struct.

If a user doesn't want to use the transparency log, that should be controlled via an explicit flag (`tlog-upload`) rather than the usage of timestamping. I just want to make sure we're clear on the messaging and security model for Sigstore, that timestamping provides an alternative to how to verify expiration of a certificate. Rekor currently plays two roles, transparency and timestamping, so using a TSA only changes the second. 

There should be a limited number of reasons to not use Rekor - private artifacts are the main one. Otherwise, you really should be uploading entries to Rekor.

priyawadhwa: > There should be a limited number of reasons to not use Rekor - private artifacts are the main one. Otherwise, you really should be uploading entries to Rekor.

+1. We should still support this use case though. We can default to using the TSA+Rekor together, and provide an opt-in flag for folks who want to sign private artifacts and verify against the TSA but not Rekor.  I think because of this use case it's more logical to have a separate TSA signer/verifier than to wrap it into Rekor. 

> Could we add support for both as part of a follow-up PR ?

sounds good! as a reviewer I definitely prefer lots of smaller PRs :) there's no plan to cut any releases at the moment so should be fine.

asraa: This all sounds good to me: I think everyone laid it out right. We can support TSA+Rekor, with a control knob on removing hte Rekor upload.

For verification, we can start verifying TSA's on conditional requirements, soon always validating them in the long tail. The verification flow will look the same, we still will use a Rekor promise of inclusion offline, but instead use the verified TSA as the time check.

pkg/cosign/verify.go
asraa: +1 I think if we should still verify the Rekor bundle. Feel free to followup with a flag to control this behavior (--skip-tlog-verify)

hectorj2f: Yes, blob support will be done as part of a different PR.

I'll add that `skip-tlog-verify` flag as part of a follow-up PR.

cmd/cosign/cli/sign/sign.go
hectorj2f: I extracted the following list of action items as result of this thread:

* Add support for **TSA and Rekor together**.
* Users could rely on **TSA-only** whenever they set `--tlog-upload=false`. We'll add a warning message and documentation about the limitations of using the timestamp verification **instead** of the tlog verification.
* Add a **`--skip-tlog-verify`** flag to our **`cosign verify*`** commands to skip the tlog verification done by default.

internal/pkg/cosign/tsa/signer_test.go
hectorj2f: Thanks 🦾

pkg/cosign/verify.go
hectorj2f: I added support for checking the certificate expiration although I thought we could also have some helper functions in the timestamp-authority to run these validations (meaning the cert and timestamp.Time). wdyt ?

internal/pkg/cosign/tsa/signer.go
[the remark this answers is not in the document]
hectorj2f: I removed it. It looks bad when verifying multiple signatures for one image.

hectorj2f:
@haydentherapper @priyawadhwa I've addressed all your comments. Please, have a look.

On the other hand I have a list of actions items to tackle in a different PR https://github.com/sigstore/cosign/pull/2460#discussion_r1023190745.

pkg/cosign/verify.go
```diff
@@ -928,6 +947,58 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, rekorClient *client.Re
...
+
+	sigBytes, err := base64.StdEncoding.DecodeString(b64Sig)
+	if err != nil {
+		return false, fmt.Errorf("reading DecodeString: %w", err)
+	}
+	pemBytes, err := os.ReadFile(filepath.Clean(tsaCertChainPath))
+	if err != nil {
+		return false, fmt.Errorf("error reading certification chain path file: %w", err)
+	}
+	certPool := x509.NewCertPool()
```
haydentherapper, line 976: Can you add a TODO to make sure we update this logic once we finish https://github.com/sigstore/timestamp-authority/issues/121, because this isn't a sufficient check (since this sets the leaf as a the "trusted" cert)

pkg/cosign/verify.go
```diff
@@ -928,6 +947,58 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, rekorClient *client.Re
...
+	cert, err := sig.Cert()
+	if err != nil {
+		return false, err
+	}
+
+	sigBytes, err := base64.StdEncoding.DecodeString(b64Sig)
+	if err != nil {
+		return false, fmt.Errorf("reading DecodeString: %w", err)
+	}
+	pemBytes, err := os.ReadFile(filepath.Clean(tsaCertChainPath))
```
haydentherapper, line 972: Can you have a TODO to support certs from TUF? Just to make sure we come back to this once we spin up a timestamp server

haydentherapper (commented):
LGTM

pkg/cosign/verify.go
hectorj2f: Done

pkg/cosign/verify.go
hectorj2f: Done

priyawadhwa (approved):
nice!

PR #2461, opened 2022-11-15
refactor: share logic between verify-blob, verify, and verify-attestation

asraa:

* Cleanup: Unifies the logic around blob verification and OCI 
* Cleanup: Updates the OCI logic to check for the timestamp using the bundle (@hectorj2f you can also add a condition to replace `validityTime` with the TSA)
* Security: Fixes an issue at HEAD: with COSIGN_EXPERIMENTAL removed, OCI validation succeeds when there is no bundle provided in offline mode, EVEN IF the certificate was expired (basically we skip the cert expiration check if  there is no call to Rekor or a bundle).

Follow-ups:
* Fix the security issue when using TSA

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -16,49 +16,29 @@
 package verify
 
 import (
-	"bytes"
 	"context"
 	"crypto"
-	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
-	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
-	"time"
 
-	"github.com/go-openapi/runtime"
 	ssldsse "github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
 	"github.com/sigstore/cosign/pkg/blob"
 	"github.com/sigstore/cosign/pkg/cosign"
-	"github.com/sigstore/cosign/pkg/cosign/bundle"
 	"github.com/sigstore/cosign/pkg/cosign/pivkey"
 	"github.com/sigstore/cosign/pkg/cosign/pkcs11key"
+	"github.com/sigstore/cosign/pkg/oci/static"
 	sigs "github.com/sigstore/cosign/pkg/signature"
-	"github.com/sigstore/sigstore/pkg/tuf"
 
 	ctypes "github.com/sigstore/cosign/pkg/types"
-	"github.com/sigstore/rekor/pkg/generated/client"
-	"github.com/sigstore/rekor/pkg/generated/models"
-	"github.com/sigstore/rekor/pkg/pki"
-	"github.com/sigstore/rekor/pkg/types"
-	"github.com/sigstore/rekor/pkg/types/hashedrekord"
-	hashedrekord_v001 "github.com/sigstore/rekor/pkg/types/hashedrekord/v0.0.1"
-	"github.com/sigstore/rekor/pkg/types/intoto"
-	intoto_v001 "github.com/sigstore/rekor/pkg/types/intoto/v0.0.1"
-	"github.com/sigstore/rekor/pkg/types/rekord"
-	rekord_v001 "github.com/sigstore/rekor/pkg/types/rekord/v0.0.1"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
-	"github.com/sigstore/sigstore/pkg/signature/dsse"
-	signatureoptions "github.com/sigstore/sigstore/pkg/signature/options"
 )
 
 func isb64(data []byte) bool {
@@ -88,14 +68,20 @@ type VerifyBlobCmd struct {
 // nolint
 func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 	var cert *x509.Certificate
-	var bundle *bundle.RekorBundle
+	var chain []*x509.Certificate
+	opts := make([]static.Option, 0)
 
 	// Require a certificate/key OR a local bundle file that has the cert.
+	if options.NOf(c.KeyRef, c.CertRef, c.Sk, c.BundlePath) == 0 {
+		return fmt.Errorf("please provide a cert to verify against via --certificate or a bundle via --bundle")
+	}
+
+	// We can't have both a key and a security key
 	if options.NOf(c.KeyRef, c.Sk) > 1 {
 		return &options.KeyParseError{}
 	}
 
-	sig, err := signatures(c.SigRef, c.BundlePath)
+	sig, err := base64signature(c.SigRef, c.BundlePath)
 	if err != nil {
 		return err
 	}
@@ -125,13 +111,16 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 			}
 			co.RekorClient = rekorClient
 		}
-		co.RootCerts, err = fulcio.GetRoots()
-		if err != nil {
-			return fmt.Errorf("getting Fulcio roots: %w", err)
-		}
-		co.IntermediateCerts, err = fulcio.GetIntermediates()
-		if err != nil {
-			return fmt.Errorf("getting Fulcio intermediates: %w", err)
+		// Use default TUF roots if a cert chain is not provided.
+		if c.CertChain == "" {
+			co.RootCerts, err = fulcio.GetRoots()
+			if err != nil {
+				return fmt.Errorf("getting Fulcio roots: %w", err)
+			}
+			co.IntermediateCerts, err = fulcio.GetIntermediates()
+			if err != nil {
+				return fmt.Errorf("getting Fulcio intermediates: %w", err)
+			}
 		}
 	}
 
@@ -161,29 +150,11 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 		if err != nil {
 			return err
 		}
-		if c.CertChain == "" {
-			co.RootCerts, err = fulcio.GetRoots()
-			if err != nil {
-				return fmt.Errorf("getting Fulcio roots: %w", err)
-			}
-
-			co.IntermediateCerts, err = fulcio.GetIntermediates()
-			if err != nil {
-				return fmt.Errorf("getting Fulcio intermediates: %w", err)
-			}
-			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
-			if err != nil {
-				return fmt.Errorf("validating certRef: %w", err)
-			}
-		} else {
+		if c.CertChain != "" {
 			// Verify certificate with chain
-			chain, err := loadCertChainFromFileOrURL(c.CertChain)
+			chain, err = loadCertChainFromFileOrURL(c.CertChain)
 			if err != nil {
 				return err
-			}
-			co.SigVerifier, err = cosign.ValidateAndUnpackCertWithChain(cert, chain, co)
-			if err != nil {
-				return fmt.Errorf("verifying certRef with certChain: %w", err)
 			}
 		}
 		if c.SCTRef != "" {
@@ -193,7 +164,8 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 			}
 			co.SCT = sct
 		}
-	case c.BundlePath != "":
+	}
+	if c.BundlePath != "" {
 		b, err := cosign.FetchLocalSignedPayloadFromPath(c.BundlePath)
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


pkg/cosign/verify.go
```diff
@@ -820,6 +837,11 @@ func VerifyLocalImageAttestations(ctx context.Context, path string, co *CheckOpt
 	return verifyImageAttestations(ctx, atts, h, co)
 }
 
+func VerifyImageAttestation(ctx context.Context, att oci.Signature, h v1.Hash, co *CheckOpts) (
```
asraa, line 840: We can remove this light wrapper if the blob verification moves inside pkg/

pkg/cosign/verify.go
```diff
@@ -675,27 +654,65 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
...
 
-	// If the --offline flag was specified, fail here
+	if bundleVerified {
+		// Update with the verified bundle's integrated time.
+		validityTime, err = getBundleIntegratedTime(sig)
+	}
+
+	// If the --offline flag was specified, fail here. bundleVerified returns false with
+	// no error when there was no bundle provided.
+	// TODO: You can be offline when you are verifying with a public key. This shouldn't
```
asraa, line 674: @priyawadhwa WDYT?

asraa: I think it's solved by the flag that says "don't check the tlog", and it should remain here for now
This is an error saying "Signature has not been checked against a tlog"

pkg/cosign/verify.go
```diff
@@ -647,23 +622,27 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
...
 		}
-		// If the chain annotation is not present or there is only a root
-		if chain == nil || len(chain) <= 1 {
-			co.IntermediateCerts = nil
-		} else if co.IntermediateCerts == nil {
-			// If the intermediate certs have not been loaded in by TUF
-			pool := x509.NewCertPool()
-			for _, cert := range chain[:len(chain)-1] {
-				pool.AddCert(cert)
+		// If the there is only a root. If there is no chain annotation present, preserve the
```
asraa, line 625: I had to change this logic to make this work for blobs, where we are using the TUF intermediates/roots or supplying the whole chain via the CLI. I'll revisit this @haydentherapper

asraa (commented):
Going to just move the blob verifies to the package. That will further simplify.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -171,20 +171,12 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 			if err != nil {
 				return fmt.Errorf("getting Fulcio intermediates: %w", err)
 			}
-			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
```
haydentherapper, line 174: Is this not still needed?

asraa: Yeah, it happens inside the verification command here: https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L636

asraa: I think I'll make it clearer by removing more code and putting it into pkg/

pkg/cosign/verify.go
```diff
@@ -867,6 +889,16 @@ func CheckExpiry(cert *x509.Certificate, it time.Time) error {
...
 }
 
+func getBundleIntegratedTime(sig oci.Signature) (time.Time, error) {
+	bundle, err := sig.Bundle()
+	if err != nil {
+		return time.Now(), err
+	} else if bundle == nil {
+		return time.Now(), nil
+	}
+	return time.Unix(bundle.Payload.IntegratedTime, 0), nil
```
hectorj2f, line 899: This time will be extracted from the RFC3161 timestamp for the TSA option.

hectorj2f:
@asraa I opened a PR adding support for the blob verification for TSA and/or Rekor here: https://github.com/sigstore/cosign/pull/2464

pkg/cosign/verify.go
asraa: That's for rekor bundle integrated time extraction, but yes, we would need another option for extracting the validityTime from the TSA in the outer call.

hectorj2f: yep. It makes sense.

hectorj2f:
These changes look good to me. 

> but can we use this refactored shared logic for verification?

This depends on our choice for blob signing. If we choose to generate two bundles. How do we expect the verification to work ?

asraa:
>  If we choose to generate two bundles. How do we expect the verification to work ?

I think the same as OCI. Here's a flow:
1. Verify the cryptographic signature on the artifact using the public key verifier
2. If a Rekor bundle is provided or required, verify it and extract the integrated time.
3. If a TSA bundle is provided or requried, verify it and extract the TSA time. This will take precedence over the Rekor time (since it's verifiable)
4. If a certificate is used, check the certificate against the time (either Rekor or TSA, but TSA if both are provided)

asraa:
Just updated to fix the tests. I had to do two things:
* OCI does not compare certificates/pubkeys in the bundle, so our blob tests doing that with a tampered bundle don't fail as expected
* Another tests fails because of verification doesn't REQUIRE a TLOG check if it's not needed (i.e. for public keys). I can change that. Please let me know.

hectorj2f:
@asraa Please, rebase your branch. I have added some changes to use a flag skip-tlog-verify as proposed. It also supports both validations tsa and rekor. Let me know if you find any issue so I can help.

asraa (commented):
@hectorj2f @priyawadhwa @znewman01 @kpk47 i would appreciate a review! 

Some of the verification logic is now clobbered with a few PRs merged into main, and I will put out a follow-up & more tests (particularly: if you skip TLOG verify and use a TSA you also end up skipping the short-lived expiry validation now)

pkg/cosign/verify.go
```diff
@@ -697,33 +680,83 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 		return bundleVerified, nil
```
haydentherapper, line 680: So the plan for the next PR is to move the above TSABundle verification below, and use the time from it?

Does providing a signed timestamp supersede the timestamp in the Rekor bundle? Or do we check both and warn the certificate can be verified with one and not the other?

pkg/cosign/verify.go
```diff
@@ -1030,6 +1071,34 @@ func compareSigs(bundleBody string, sig oci.Signature) error {
...
+		return fmt.Errorf("failed to extract key from bundle: %w", err)
+	}
+
+	decodeSecond, err := base64.StdEncoding.DecodeString(bundleKey)
+	if err != nil {
+		return fmt.Errorf("decoding base64 string %s", bundleKey)
+	}
+
+	// Compare the PEM bytes, to spurious ignore newlines.
+	pemFirst, _ := pem.Decode(pemBytes)
```
haydentherapper, line 1091: Should we err if `rest` is not just newlines?

pkg/cosign/verify.go
asraa: Actually, I hadn't considered that, but I like that: let's verify with both, and expect both to pass. 

Eventually we can phase out Rekor time or something, but for now we expect both to work, so let's validate against both.

asraa: I'll follow-up in the next PR

pkg/cosign/verify.go
asraa: Good point: you could stick in some spurious data. It's unexpected and prob doesn't change properties, but let's do it anyway

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -239,12 +234,12 @@ func TestVerifyBlob(t *testing.T) {
...
-		cert        *x509.Certificate
-		bundlePath  string
+		name       string
+		blob       []byte
+		signature  string
+		key        []byte
+		cert       *x509.Certificate
+		bundlePath string
 		// If online lookups to Rekor are enabled
 		experimental bool
```
haydentherapper, line 244: Most of these tests set the experimental flag, but we've begun to remove it - should we remove it from the tests too? (Separate refactor prolly)

asraa: Yeah, I'll add in a separate PR: it actually does nothing now :)

pkg/cosign/verify.go
asraa: Oops this broke tests... fixing now

hectorj2f (approved):
LGTM, we have simplified the verification logic a lot.

pkg/cosign/verify.go
asraa: Fixed

hectorj2f (dismissed):

pkg/cosign/verify.go
```diff
@@ -697,33 +680,83 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
...
+		// Update with the verified bundle's integrated time.
+		validityTime, err = getBundleIntegratedTime(sig)
+		if err != nil {
+			return false, fmt.Errorf("error getting bundle integrated time: %w", err)
+		}
+	}
+
+	// If the --offline flag was specified, fail here. bundleVerified returns false with
+	// no error when there was no bundle provided.
+	// TODO: You can be offline when you are verifying with a public key. This shouldn't
```
haydentherapper, line 703: Could you clarify what's meant by this?

haydentherapper (approved):

pkg/cosign/verify.go
asraa: Will also follow-up in the re-review of the verification main function.

But this means that: if you are verifying with a normal PKI system, and using --offline, then you MIGHT not require timestamping/tlog verification. In that case, we might not want to return an error. I'm not sure.

haydentherapper: If you have a key, then timestamping isn't required, but tlog verification is still useful to know that the artifact is auditable. If you have a cert, then both timestamping and auditable is useful. Not sure if that helps haha.

asraa: I think I agree -- I think the TODO can be removed then, will remove in the fix-up. I'll try to update the top-level verification logic comments as well.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -88,14 +68,20 @@ type VerifyBlobCmd struct {
 // nolint
 func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 	var cert *x509.Certificate
-	var bundle *bundle.RekorBundle
+	var chain []*x509.Certificate
+	opts := make([]static.Option, 0)
 
 	// Require a certificate/key OR a local bundle file that has the cert.
+	if options.NOf(c.KeyRef, c.CertRef, c.Sk, c.BundlePath) == 0 {
```
znewman01, line 75: Should this be `!= 1`? 

Oh, I guess not because `BundlePath` isn't really a full "bundle" yet. That's annoying.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -88,14 +68,20 @@ type VerifyBlobCmd struct {
...
+		return fmt.Errorf("please provide a cert to verify against via --certificate or a bundle via --bundle")
+	}
+
+	// We can't have both a key and a security key
```
znewman01, line 79: What about a key/sk and a cert?

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -193,220 +164,100 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
+			return err
+		}
+	}
+	var chainPEM []byte
+	if chain != nil {
+		// Set the last one in the co.RootCerts
+		if co.RootCerts == nil {
+			co.RootCerts = x509.NewCertPool()
+		}
+		co.RootCerts.AddCert(chain[len(chain)-1])
```
znewman01, line 220: Does this mean that if we pass a chain, we automatically trust it?

That feels wrong to me.

znewman01 (commented):
(I owe you a more detailed pass, just a couple things that jumped out at first)

cmd/cosign/cli/verify/verify_blob.go
haydentherapper: Before the refactoring, this was a way to pass in a chain as an alternative to fetching from the TUF root. Without a chain, we'd default to the TUF metadata, otherwise we'd use the chain passed in.

One bit of feedback is there's a lot of places we're checking `chain != nil` or `c.CertChain != ""`, so it's a little hard to quickly see exactly where we're getting the trust root from.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -193,220 +164,100 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
+		// Use the rest as the cert chain.
+		chainPEM, err = cryptoutils.MarshalCertificatesToPEM(chain)
+		if err != nil {
+			return err
 		}
 	}
+	opts = append(opts, static.WithCertChain(certPEM, chainPEM))
```
haydentherapper, line 227: What happens if `chain` is not set, in the case of using the default TUF chain [here](https://github.com/sigstore/cosign/blob/08277261bfdbfe8085c1a580cb3d6f59a6fea97a/cmd/cosign/cli/verify/verify_blob.go#L114)? `chainPEM` will not be set, is that an issue?

haydentherapper: I think the purpose was that the roots were loading in earlier, and if you have an intermediate in your chain, we'd automatically populate an intermediate pool. Had to review https://github.com/sigstore/cosign/commit/7402eb46db00a78e1aa7aec53f711a8a504d6407 where I added this

cmd/cosign/cli/verify/verify_blob.go
haydentherapper: To Zach's point, there are three ways you could pass a root in (this is also all applicable to intermediates):
* SIGSTORE_ROOT_FILE, which will get read in in fulcio.GetRoots
* tuf (either the sigstore tuf root or a self-hosted one), read in in fulcio.GetRoots
* Passing in the cert chain here

I'm not opposed if we want to force the user to set SIGSTORE_ROOT_FILE, as this supports both roots and intermediates (added in https://github.com/sigstore/cosign/pull/1749). I don't recall why this was added this way - likely just a product of iterative development and some overlapping features haha. 

That said, I'm also not opposed to supporting the chain flag too. It really is no different than SIGSTORE_ROOT_FILE, but if it makes things easier, we could drop it. Just wanna make sure we don't lose any supported features.

cmd/cosign/cli/verify/verify_blob.go
haydentherapper: I bet this was added originally because we didn't have the intermediate in the TUF root, so the idea was that we'd fetch the intermediate from the chain automatically (and we treated the intermediate as "untrusted", just a part of chain building)

(Item to myself, write better PR messages)

pkg/cosign/verify.go
```diff
@@ -660,23 +640,26 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
...
 		}
-		// If the chain annotation is not present or there is only a root
-		if chain == nil || len(chain) <= 1 {
-			co.IntermediateCerts = nil
-		} else if co.IntermediateCerts == nil {
-			// If the intermediate certs have not been loaded in by TUF
-			pool := x509.NewCertPool()
-			for _, cert := range chain[:len(chain)-1] {
-				pool.AddCert(cert)
+		// If there is no chain annotation present, we preserve the pools set in the CheckOpts.
```
haydentherapper, line 643: I'd want to think this through a bit more, but I wonder if this can be removed. Basically, force the caller to set CheckOpts and don't just set intermediates based on the chain. Then you could also say that the chain on the OCI struct isn't needed.

Worried this is going to cause some issue if it's removed though...

cmd/cosign/cli/verify/verify_blob.go
asraa: Yeah: this is because in the BLOB case you pass in the cert chain, so you DO trust it "out of band".
In the OCI case, the cert chain is provided on the registry. This was the only way I could get parity on the behavior. the cert chain passed here does represent the full trust chain.

cmd/cosign/cli/verify/verify_blob.go
asraa: > I think the purpose was that the roots were loading in earlier, and if you have an intermediate in your chain, we'd automatically populate an intermediate pool. Had to review https://github.com/sigstore/cosign/commit/7402eb46db00a78e1aa7aec53f711a8a504d6407 where I added this

Exactly. Before we actually reset intermediates, but in using blobs, we actually wanted to preserve the TUF roots that were populated already.

pkg/cosign/verify.go
asraa: > Then you could also say that the chain on the OCI struct isn't needed.

How would you know the intermediates though? I guess right now we can fetch via TUF (I remember before we couldn't, which is why we ended up in this situation).

cmd/cosign/cli/verify/verify_blob.go
asraa: Agree

asraa: key, sk, and cert are mutually exclusive. Changed to that condition.

pkg/cosign/verify.go
haydentherapper: Yea, I think what’s here now sounds good, I wouldn’t want to change anything in a refactor.

cmd/cosign/cli/verify/verify_blob.go
asraa: > One bit of feedback is there's a lot of places we're checking chain != nil or c.CertChain != "", so it's a little hard to quickly see exactly where we're getting the trust root from.

Just realized you mentioned this so I went and checked. There's only one place that `c.CertChain` is used, and that's to populated `chain` -- after that everything is populated from `chain`. Cleaned that up.

asraa: LMK if there's anything else here, or we can resolve.

asraa:
@haydentherapper @znewman01 lemme know if you have time to do another pass! since it's working around the verification logic, would love to have it in. I have a follow-up issue on cleaning up the verification main method.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,252 +149,109 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
 	}
-
-	// Performs all blob verification.
-	if err := verifyBlob(ctx, co, blobBytes, sig, cert, bundle); err != nil {
-		return err
+	// Set a cert chain if provided.
+	var chainPEM []byte
+	if c.CertChain != "" {
+		chain, err := loadCertChainFromFileOrURL(c.CertChain)
+		if err != nil {
```
haydentherapper, line 194: nit: if chain == nil, that seems like an error to me, because we expect a chain

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -161,252 +149,109 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
+			return err
+		}
+		if chain != nil {
+			// Set the last one in the co.RootCerts. This is trusted, as its passed in
+			// via the CLI.
+			if co.RootCerts == nil {
+				co.RootCerts = x509.NewCertPool()
+			}
+			co.RootCerts.AddCert(chain[len(chain)-1])
+			// Use the rest as the cert chain in the signature object.
```
haydentherapper, line 204: nit: It's not the rest, it's the entire chain (intermediates and root) - this is intentional, correct?

asraa: Yep it's intentional, I'll fix the comment. The cert chain in the signature object is taken "with a grain of salt" (omits the last cert)

cmd/cosign/cli/verify/verify_blob.go
asraa: we can do that

pkg/cosign/verify.go
```diff
@@ -697,33 +680,83 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 		return bundleVerified, nil
 	}
 
-	bundleVerified, err = VerifyBundle(ctx, sig, co.RekorClient)
+	// 2. Check the validity time of the signature.
+	// This is the signature creation time. As a default upper bound, use the current
+	// time.
+	validityTime := time.Now()
+
+	bundleVerified, err = VerifyBundle(ctx, sig, co, co.RekorClient)
```
haydentherapper, line 688: This is related to the commented out test - What happens if I am verifying a signature with just a key without a bundle? Does this return false?

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -255,24 +250,28 @@ func TestVerifyBlob(t *testing.T) {
...
 		},
+		/* TODO:
+		This currently passes without error because we don't require an experimental
+		lookup for public keys. Add this test back in when we have a --verify-tlog-online
 		{
 			name:         "valid signature with public key - experimental no rekor fail",
 			blob:         blobBytes,
 			signature:    blobSignature,
-			sigVerifier:  signer,
+			key:          pubKeyBytes,
```
haydentherapper, line 264: looking through the code, i am unsure why this isn't erroring out. in verifyInternal, it seems like we should be falling back to an online lookup if bundle verification fails

haydentherapper: Never mind, dug into the code more, I see that it doesn't err out, just returns that the bundle wasn't verified.

haydentherapper (approved):

hectorj2f (approved):

cmd/cosign/cli/verify/verify_blob.go
znewman01: Filed https://github.com/sigstore/cosign/issues/2472 as a follow up

PR #2463, opened 2022-11-16
Support TSA and Rekor verifications

hectorj2f:
#### Summary

This PR adds support for signing and verification of OCI images using TSA and Rekor verifications.

related to: https://github.com/sigstore/cosign/issues/2331,

### DIFF
--- a/cmd/cosign/cli/options/verify.go
+++ b/cmd/cosign/cli/options/verify.go
@@ -85,12 +85,6 @@ func (o *VerifyOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&o.LocalImage, "local-image", false,
 		"whether the specified image is a path to an image saved locally via 'cosign save'")
-
-	cmd.Flags().StringVar(&o.TSAServerURL, "timestamp-server-url", "",
-		"url to a timestamp RFC3161 server, default none")
-
-	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-cert-chain", "",
-		"path to certificate chain PEM file for the Timestamp Authority")
 }
 
 // VerifyAttestationOptions is the top level wrapper for the `verify attestation` command.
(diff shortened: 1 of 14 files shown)


cmd/cosign/cli/options/verify.go
```diff
@@ -34,6 +35,9 @@ func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-cert-chain", "",
 		"path to certificate chain PEM file for the Timestamp Authority")
+
+	cmd.Flags().BoolVar(&o.SkipTlogVerify, "skip-tlog-verify", false,
```
haydentherapper, line 39: I’d prefer a scarier name, I really want to avoid people setting this accidentally. Can we preface it with “insecure”?

hectorj2f: I added in both the flag and the description.

pkg/cosign/verify.go
```diff
@@ -684,27 +687,26 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
...
 			return false, fmt.Errorf("unable to verify TSA bundle: %w", err)
 		}
-
-		return bundleVerified, nil
 	}
 
-	bundleVerified, err = VerifyBundle(ctx, sig, co.RekorClient)
-	if err != nil {
-		return false, fmt.Errorf("error verifying bundle: %w", err)
+	if !co.SkipTlogVerify {
```
priyawadhwa, line 697: nit: beyond this point everything is about tlog verification, so i think we can just skip it all by doing something like

```
if co.SkipTlogVerify {
  return true, nil
}
```

priyawadhwa (approved):
nice!

haydentherapper:
Make sure to sync with @asraa about the verification refactor since there’s overlap

asraa (commented):
@hectorj2f you now removed the short-lived certificate verification when using a TSA... when I merged I noticed this. I won't fix in that, but I'll make a separate fix since it's logic fix and not a refactor.

PR #2464, opened 2022-11-16
Blob signing and verification using a RFC3161 time-stamping server

hectorj2f:
#### Summary

This PR adds support for signing and verification of blobs using a RFC3161 time-stamping server. This PR generates two different bundles when signing by passing the flags `--tsa-bundle` and `--bundle`. However, the blob verification only uses one of the bundle files (to be discussed).

Note that this PR needs https://github.com/sigstore/cosign/pull/2463 to be merged first.

**List of open questions:**

* **Verification:** Should the verification use two bundle flags `--tsa-bundle` and `--bundle`? Or should we only accept a single bundle file during the validation ?.

* **Signing:** Should the existing bundle contain the verification data from the TSA (SignedRFC3161Tiemstamp and the Tlog entry) all together ?

```
{
	"base64Signature": "xxxx/8tmKUTTOHLBTHn",
	"cert": "xx==",
	"rekorBundle": {
		"SignedEntryTimestamp": "xxx+T3qDZtg0=",
		"Payload": {
			"body": "xxxxxxx=",
			"integratedTime": 1,
			"logIndex": 1,
			"logID": "xxxxx"
		}
	},
	"tsaBundle": {
		"SignedRFC3161Timestamp": "xxxx"
	}
}
```


cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -315,7 +385,18 @@ func verifyBlob(ctx context.Context, co *cosign.CheckOpts,
 		}
 		validityTime = time.Unix(bundle.IntegratedTime, 0)
 		fmt.Fprintf(os.Stderr, "tlog entry verified offline\n")
-	// b. We can make an online lookup to the transparency log since we don't have an entry.
+		// b. We can make an online lookup to the transparency log since we don't have an entry.
+	case tsaBundle != nil:
+		// TODO: Verify TSA only and return the result with the tsa bundle
```
asraa, line 390: you would need to set the validity time here

hectorj2f:
I am pausing this work until https://github.com/sigstore/cosign/pull/2461 gets merged!

hectorj2f:
@priyawadhwa @haydentherapper This PR is ready for a review. I was planning to increase the test coverage and perhaps add this experimental flag/env var that we discussed, as part of future PRs.

cmd/cosign/cli/options/verify.go
```diff
@@ -141,6 +141,8 @@ type VerifyBlobOptions struct {
 	Rekor               RekorOptions
 	Registry            RegistryOptions
 	CommonVerifyOptions CommonVerifyOptions
+
+	RFC3161TimestampPath string
```
priyawadhwa, line 145: we could add this to `CommonVerifyOptions` alongside `TSACertChainPath`

cmd/cosign/cli/sign/sign.go
```diff
@@ -73,6 +72,7 @@ func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Refere
 	// Check if TSA signing is enabled and tlog upload is disabled
 	if !tlogUpload && tsaServerURL != "" {
 		fmt.Fprintln(os.Stderr, "\nWARNING: skipping transparency log upload")
+		return false
```
priyawadhwa, line 75: Looks like tlog-upload defaults to false, for they keyless signing+rekor case I think we will return false here before we get a chance to check for keyless signing.

https://github.com/sigstore/cosign/blob/44a7117e0cee495a163c92e9f347efb2d9b747c7/cmd/cosign/cli/options/signblob.go#L75

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -65,7 +67,21 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
...
+		b64Sig := []byte(base64.StdEncoding.EncodeToString(sig))
+
+		respBytes, err := tsa.GetTimestampedSignature(b64Sig, clientTSA)
+		if err != nil {
+			return nil, err
+		}
+
+		signedPayload.RFC3161Timestamp = cbundle.TimestampToRFC3161Timestamp(respBytes)
+	}
+	if ShouldUploadToTlog(ctx, ko, nil, ko.SkipConfirmation, tlogUpload, ko.TSAServerURL) {
```
priyawadhwa, line 84: I just realized we actually don't need to pass in `ko.SkipConfirmation` or `ko.TSAServerURL` since we pass in `ko` 😓 No need to change, we can do it some other time.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -94,10 +97,41 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
+		// TODO: Update this logic once https://github.com/sigstore/timestamp-authority/issues/121 gets merged.
+		// This relies on untrusted leaf certificate.
+		tsaCertPool := x509.NewCertPool()
+		ok := tsaCertPool.AppendCertsFromPEM(pemBytes)
+		if !ok {
+			return fmt.Errorf("error parsing response into Timestamp while appending certs from PEM")
+		}
+		co.TSACerts = tsaCertPool
+		// Use default TUF roots if a cert chain is not provided.
+		if c.CertChain == "" {
```
priyawadhwa, line 124: seems like this code is duplicated for the TSA and in the next `if` statement for rekor verification. We could pull it out to happen afterwards?

priyawadhwa (commented):
tests look great! just a few comments.

haydentherapper:
Cc @asraa

cmd/cosign/cli/options/verify.go
hectorj2f: @priyawadhwa RFC3161TimestampPath is only used by blob verification, I am not sure it follows the mission of common verify options.

cmd/cosign/cli/sign/sign.go
hectorj2f: For keyless signing + rekor, `tsaServerURL=""` will be empty so it won't enter here.

cmd/cosign/cli/sign/sign_blob.go
hectorj2f: yes, it can be simplified. 👍🏻

cmd/cosign/cli/sign/sign.go
priyawadhwa: nevermind! i assumed we had a default set for `timestamp-server-url` since we do for fulcio and rekor, but i see now that we don't. please disregard!

cmd/cosign/cli/verify/verify_attestation.go
hectorj2f: Actually I think we don't need it here.

cmd/cosign/cli/options/verify.go
priyawadhwa: ah gotcha yah let's leave it then, thanks!

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -65,7 +67,21 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
 
 	signedPayload := cosign.LocalSignedPayload{}
 
-	if ShouldUploadToTlog(ctx, ko, nil, ko.SkipConfirmation, tlogUpload, "") {
+	if ko.TSAServerURL != "" {
+		clientTSA, err := tsaclient.GetTimestampClient(ko.TSAServerURL)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create TSA client: %w", err)
+		}
+		b64Sig := []byte(base64.StdEncoding.EncodeToString(sig))
```
haydentherapper, line 75: Should we document somewhere that we're generating a timestamp over the base64 signature, not the raw bytes? This was somewhat discussed in https://github.com/sigstore/timestamp-authority/issues/116 - Cosign is being opinionated in using the base64-encoded sig, which I agree with, just think we should be explicit about this.

Update the SPEC docs for the OCI annotations?

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -82,6 +98,20 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
 		signedPayload.Bundle = cbundle.EntryToBundle(entry)
 	}
 
+	// if bundle is specified, just do that and ignore the rest
+	if ko.RFC3161TimestampPath != "" {
```
haydentherapper, line 102: Should we require that if `ko.TSAServerURL` is set, `ko.RFC3161TimestampPath` must also be? Otherwise it's unnecessarily fetching a timestamp without persisting it.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -82,6 +98,20 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
...
+		signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(sig)
+
+		contents, err := json.Marshal(signedPayload)
```
haydentherapper, line 105: Shouldn't this be writing `signedPayload.RFC3161Timestamp`, not `signedPayload.Base64Signature`?

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -72,16 +73,16 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 	opts := make([]static.Option, 0)
 
 	// Require a certificate/key OR a local bundle file that has the cert.
-	if options.NOf(c.KeyRef, c.CertRef, c.Sk, c.BundlePath) == 0 {
-		return fmt.Errorf("please provide a cert to verify against via --certificate or a bundle via --bundle")
+	if options.NOf(c.KeyRef, c.CertRef, c.Sk, c.BundlePath, c.RFC3161TimestampPath) == 0 {
```
haydentherapper, line 76: This doesn't seem correct, `RFC3161TimestampPath` should contain a timestamp, not a code-signing certificate.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -180,6 +205,31 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 		}
 		opts = append(opts, static.WithBundle(b.Bundle))
 	}
+	if c.RFC3161TimestampPath != "" {
+		b, err := cosign.FetchLocalSignedPayloadFromPath(c.RFC3161TimestampPath)
+		if err != nil {
+			return err
+		}
+		// Note: RFC3161 timestamp does not set the certificate.
```
haydentherapper, line 213: I'm confused by this, `b.Cert` is always `""`, so isn't this never run?

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -273,6 +323,12 @@ func base64signature(sigRef string, bundlePath string) (string, error) {
 			return "", err
 		}
 		targetSig = []byte(b.Base64Signature)
+	case rfc3161TimestampPath != "":
```
haydentherapper, line 326: I left a comment above about this, but `rfc3161TimestampPath` should be storing a timestamp, not a signature, correct?

haydentherapper: Further, a 3161 timestamp should be in addition to the bundle, this makes it either/or.

haydentherapper (commented):
@hectorj2f Left some comments, happy to implement the changes if you are busy.

haydentherapper:
Follow up PR https://github.com/sigstore/cosign/pull/2499

PR #2465, opened 2022-11-17
Use the TSA x509 certificate pool as parameter to CheckOpts

hectorj2f:

#### Summary

CheckOpts should use a x509 certificate pool instead of a file path. Likewise, this allows to move the validation of the certificate pool back to the cmd package.

#### Release Note

NONE


haydentherapper (approved):
Agreed, this looks good! This'll probably need to get updated to separate roots/intermediates later on, but we can punt for now until the TODO is done.

PR #2469, opened 2022-11-18
Rename tsa bundle to RFC3161timestamp

hectorj2f:

#### Summary

As discussed, we decided to rename the TSABundle to RFC3161Timestamp as it represents better its data used for signing and verification.

### DIFF
--- a/pkg/oci/internal/signature/layer.go
+++ b/pkg/oci/internal/signature/layer.go
@@ -29,11 +29,11 @@ import (
 )
 
 const (
-	sigkey       = "dev.cosignproject.cosign/signature"
-	certkey      = "dev.sigstore.cosign/certificate"
-	chainkey     = "dev.sigstore.cosign/chain"
-	BundleKey    = "dev.sigstore.cosign/bundle"
-	TSABundleKey = "dev.sigstore.cosign/3161timestamp"
+	sigkey              = "dev.cosignproject.cosign/signature"
+	certkey             = "dev.sigstore.cosign/certificate"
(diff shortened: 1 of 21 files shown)


pkg/oci/internal/signature/layer.go
```diff
@@ -29,11 +29,11 @@ import (
...
-	sigkey       = "dev.cosignproject.cosign/signature"
-	certkey      = "dev.sigstore.cosign/certificate"
-	chainkey     = "dev.sigstore.cosign/chain"
-	BundleKey    = "dev.sigstore.cosign/bundle"
-	TSABundleKey = "dev.sigstore.cosign/3161timestamp"
+	sigkey              = "dev.cosignproject.cosign/signature"
+	certkey             = "dev.sigstore.cosign/certificate"
+	chainkey            = "dev.sigstore.cosign/chain"
+	BundleKey           = "dev.sigstore.cosign/bundle"
+	RFC3161TimestampKey = "dev.sigstore.cosign/3161timestamp"
```
asraa, line 36: should we make this rfc3161?

asraa (commented):
small comment about the annotation name, otherwise lgtm

pkg/oci/internal/signature/layer.go
hectorj2f: yes, I was thinking the same. 👍🏻

hectorj2f: Fixed!

priyawadhwa (approved):

PR #2470, opened 2022-11-18
Improved log output on cosign clean

ChaosInTheCRD:
Signed-off-by: chaosinthecrd <thomas.meadows@jetstack.io>

Hi everyone! 👋 

I noticed that when using `cosign clean`, it was indicating that the logs that it was deleting objects that weren't there. I saw that there is some logic to try and determine the response of the `remote.Delete` request, but this doesn't seem to be working as expected.

For this reason, I decided that it would be more reliable to perform a `remote.Get` as a more reliable means of confirming the existence of the object.

### DIFF
--- a/cmd/cosign/cli/clean.go
+++ b/cmd/cosign/cli/clean.go
@@ -93,15 +93,16 @@ func CleanCmd(ctx context.Context, regOpts options.RegistryOptions, cleanType, i
 	}
 
 	for _, t := range cleanTags {
+    if _, err := remote.Get(t, remoteOpts...); err != nil {
+      var te *transport.Error
+      if errors.As(err, &te) && te.StatusCode == http.StatusNotFound {
+        continue
+      }
+    }
 		if err := remote.Delete(t, remoteOpts...); err != nil {
 			var te *transport.Error
-			if errors.As(err, &te) && te.StatusCode == http.StatusNotFound {
-				// If the tag doesn't exist, some registries may
-				// respond with a 404, which shouldn't be considered an
-				// error.
-			} else {
-				fmt.Fprintf(os.Stderr, "could not delete %s from %s\n: %v\n", t, imageRef, err)
-			}
+			errors.As(err, &te)
+			fmt.Fprintf(os.Stderr, "could not delete %s from %s\n: %v\n", t, imageRef, err)
 		} else {
 			fmt.Fprintf(os.Stderr, "Removed %s from %s\n", t, imageRef)
 		}


znewman01:
Hey, thanks for this!

> I saw that there is some logic to try and determine the response of the remote.Delete request, but this doesn't seem to be working as expected.

I would like to determine what's going on there first before we merge this—which opens us up to race conditions and doubles the number of roundtrips. I don't think they're race conditions that would ever matter, but still 🙂 This seems like a bug in the registries, or maybe we're misunderstanding how the registry is supposed to respond in these cases—I'd like to fix those as a root cause.

Can you get a dump of what happens in this case (that is, what's in `err`?)? If it seems like

ChaosInTheCRD:
@znewman01 glad to hear this - I agree with you. I was conscious of doubling the requests... but in the case where I was using this feature to clean an image in Google Artifact Registry, the `.Delete()` wasn't giving back an error of any kind. It succeeded with the `.Delete()`, hence why it gave the positive response.

I feel that while **yes**, it adds a request, but it seems to be more fool-proof? I get your concern about the race-condition. In that case though, the failure would be printing out that it managed to successfully delete the object... which it, current is doing?

I will have a dig.

znewman01:
> @znewman01 glad to hear this - I agree with you. I was conscious of doubling the requests... but in the case where I was using this feature to clean an image in Google Artifact Registry, the `.Delete()` wasn't giving back an error of any kind. It succeeded with the `.Delete()`, hence why it gave the positive response.
> 

Hmmmm, interesting. @imjasonh @jdolitsky is this correct behavior for an OCI registry? Does GAR claim to be OCI compliant?

> I feel that while **yes**, it adds a request, but it seems to be more fool-proof? 

Sure, contingent on the registry being wrong, it's more fool-proof 😛

> I get your concern about the race-condition. In that case though, the failure would be printing out that it managed to successfully delete the object... which it, current is doing?

Yeah, I don't think the race here is cataclysmic for that reason. There's another race where the "image" gets added after the Get but that's basically what happens if you add the image after the Delete. (There's a bigger race condition here where someone adds multiple entries, and you wind half of them deleted but that was the case before this change.) I just would prefer to do it in one shot if possible (avoid the extra network round-trip).

----------------------------------------------------------

So I think I'm okay to merge this, but we should put a comment about why ("workaround for GAR bug") and ideally a place to track GAR getting fixed.

(You also need to fix lint 🙂 )

imjasonh:
cc @jonjohnsonjr

ChaosInTheCRD:
sorry - just coming back to this.

I haven't had any further time to test. @jonjohnsonjr or @imjasonh , maybe you have a bit more context here? Is what I have done a terrible idea?

I am happy to add a comment and fix the lint as @znewman01 suggested, but I understand if we don't want to merge this for now.

Thanks!

jonjohnsonjr:
> is this correct behavior for an OCI registry?

AFAICT it's unspecified.

I think returning success here is based on an attempt to make DELETE idempotent. Looking around the internet, most folks seem to think that the _behavior_ being idempotent is separate from the response code being the same, which makes sense to me...

znewman01:
ok, seems like we should just put this hack in. thanks all!

znewman01:
This is still a good idea. @ChaosInTheCRD probably needs a rebase and a lint fix (sorry, my fault for sleeping on it) but I can approve after that

PR #2474, opened 2022-11-21
respect tlog-upload flag with TSA

hectorj2f:

#### Summary

`--tlog-upload` flag was ignored when relying on time-stamping verification. We should allow disabling the tlog entry creation when relying on time-stamping.

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -69,7 +69,11 @@ const TagReferenceMessage string = `WARNING: Image reference %s uses a tag, not 
     images by tag will be removed in a future release.
 `
 
-func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, skipConfirmation, tlogUpload bool) bool {
+func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, skipConfirmation, tlogUpload bool, tsaServerURL string) bool {
+	// Check if TSA signing is enabled and tlog upload is disabled
+	if !tlogUpload && tsaServerURL != "" {
+		fmt.Fprintln(os.Stderr, "\nWARNING: skipping transparency log upload")
+	}
 	// If we aren't using keyless signing and --tlog-upload=false, return
 	if !keylessSigning(ko) && !tlogUpload {
 		return false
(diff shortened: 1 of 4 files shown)


cmd/cosign/cli/sign/sign.go
```diff
@@ -243,7 +243,11 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 
 		s = tsa.NewSigner(s, clientTSA)
 	}
-	if ShouldUploadToTlog(ctx, ko, digest, ko.SkipConfirmation, tlogUpload) {
+
+	// Check if TSA signing is enabled and tlog upload is disabled
+	if !tlogUpload && ko.TSAServerURL != "" {
```
priyawadhwa, line 248: could we put this check in `ShouldUploadToTlog` itself so that it's respected in all places that call that function?

hectorj2f: I wanted to wait until all the different parts using this function will have support for TSA. Otherwise it looks like adding a tsa parameter that is only used for images at the moment.

priyawadhwa: It should still work though right, since tsaurl will never be set for the other commands? We'd be avoiding creating tech debt and it'll still work correctly.

priyawadhwa (approved):
thanks!

cmd/cosign/cli/sign/sign.go
hectorj2f: I don't think there is any tech debt yet, but this is something easy to solve for everyone.

haydentherapper:
Hey @hectorj2f @priyawadhwa, I think we should revert this (or I'll remove it in my new PR). 

Getting a timestamp should not affect whether or not you upload a rekor entry. It's not an alternative to signature transparency, it's only an alternative to where the timestamp comes from.

priyawadhwa:
I disagree, for private artifacts users may not want to store anything in a transparency log and because the artifacts are private they aren't looking for signature transparency. Using a TSA instead of Rekor is a good option in that case. I think it's important to support both cases, and we've made it clear via flags that using a TSA is an insecure option.

haydentherapper:
They should be setting `tlog-upload=false` in that case then. Only on verification is there an `insecure-*` flag, so it's not clear on signing that not uploading to the tlog is fundamentally changing the security model of Sigstore.

haydentherapper:
Or maybe we have a global `private-artifact` flag that changes behavior?

I don't think the default should be not uploading to the transparency log however. The defaults must be secure by default, and removing the tlog upload isn't.

priyawadhwa:
> They should be setting tlog-upload=false in that case then. 

That sounds good to me, that way it's explicit on sign and then we have `insecure` on verify. I'm ok with any reasonable combo of flags, my main point was just that the feature should be supported.

haydentherapper:
Agreed! I see the issue in the previous code, that if keyless=true and tlog-upload=false, we don't respect tlog-upload. I'll submit a fix to always respect tlog-upload, regardless of if a TSA is used. Does that sound good?

priyawadhwa:
SGTM

PR #2479, opened 2022-11-23
Fix SPDX json media type

puerco:
#### Summary

This commit modifies the mediatype used by cosign when attaching SPDX SBOMs encoded in json. It will now use the [registered mediatype](https://www.iana.org/assignments/media-types/text/spdx) text/spdx plus the json suffix:  text/spdx+json

This fixes attaching those SBOMs in Quay plus possibly other registries which validate media types:

```

Failed validating 'enum' in schema['properties']['layers']['items']['properties']['mediaType']:
    {'description': 'The MIME type of the referenced manifest',
     'enum': ['application/vnd.oci.image.layer.v1.tar',
              'application/vnd.oci.image.layer.v1.tar+gzip',
              'application/vnd.oci.image.layer.v1.tar+zstd',
              'application/vnd.oci.image.layer.nondistributable.v1.tar',
              'application/vnd.oci.image.layer.nondistributable.v1.tar+gzip',
              'application/vnd.dev.cosign.simplesigning.v1+json',
              'application/vnd.dsse.envelope.v1+json',
              'text/spdx',
              'text/spdx+xml',
              'text/spdx+json',
              'application/vnd.syft+json',
              'application/vnd.cyclonedx',
              'application/vnd.cyclonedx+xml',
              'application/vnd.cyclonedx+json',
              'application/vnd.in-toto+json',
              'application/tar+gzip',
              'application/vnd.cncf.helm.chart.content.v1.tar+gzip',
              'application/vnd.oci.image.layer.v1.tar+gzip',
              'application/vnd.cncf.openpolicyagent.policy.layer.v1+rego',
              'application/vnd.cncf.openpolicyagent.data.layer.v1+json'],
     'type': 'string'}
```

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>

Thanks to @sandipanpanda for reporting the issue.

#### Release Note

cosign will now use `text/spdx+json` as media type when attaching SPDX SBOMs encoded in json.

### DIFF
--- a/pkg/types/media.go
+++ b/pkg/types/media.go
@@ -27,7 +27,7 @@ const (
 	SyftMediaType          = "application/vnd.syft+json"
 	SimpleSigningMediaType = "application/vnd.dev.cosign.simplesigning.v1+json"
 	SPDXMediaType          = "text/spdx"
-	SPDXJSONMediaType      = "spdx+json"
+	SPDXJSONMediaType      = "text/spdx+json"
 	WasmLayerMediaType     = "application/vnd.wasm.content.layer.v1+wasm"
 	WasmConfigMediaType    = "application/vnd.wasm.config.v1+json"
 )


hectorj2f (approved):
lgtm

hectorj2f:
My only concern is related to any dependency that used the old media type. Perhaps we could keep both media types during some time.

puerco:
@hectorj2f I get it, but how? Short of adding a `--type=wrong-spdx` I don't see a way to preserve it. Specifying `spdx` in  type flag when calling `cosign attach sbom` doesn't even let you control the media type directly as it does some magic here to determine it based on the file extension:
https://github.com/sigstore/cosign/blob/fdeb16e647a6cecad86837df0b4ac1a8187a495b/cmd/cosign/cli/options/attach.go#L88-L95

Also, let's keep in mind that the incomplete media type is, well, wrong :)

dlorenc:
Yeah, I get the concern about compatibility but correctness is probably more important here.

dlorenc (approved):

cpanato (approved):

hectorj2f:
It sounds good. Perhaps we want to **highlight** this change in the coming release notes.

PR #2482, opened 2022-11-25
RFC: When verifying, only use data after it is considered acceptable

mtrmac:
#### Summary
`pkg/cosign.verifyInternal` keeps growing conditions (A) at the end of the function, long after already relying (B) on data that is constrained by these conditions. It effectively makes the outcome of B indeterminate until A happens; that’s hard to track, and A might not happen at all if the function returns early. This makes me _nervous_: personally for me, it makes following the operation and rules of the code much harder than I think is necessary.

So, this is a move somewhat towards, but not completely, the verification organization suggested in https://github.com/sigstore/cosign/pull/1648#discussion_r856479474.

Most data in `verifyInternal`, and closest utility functions, is now stored with variables named `untrusted…`. Only after _all_ checks on that data are done, it may be stored in a variable named `acceptable…`.

Testing relies on existing unit tests; no new tests were added.

See individual commit messages for details. Most of the commits are trivial variable renames.

Behavior changes:
- Certificate expiry is now not tested if we don’t rely on the certificate at all (if the user provides a trusted public key)
- If we do rely on the certificate, we now always test expiry against _some_ time. Previously that didn’t happen if there was no RFC 3161 timestamp and Rekor log presence was opted out. (I.e. this should allow using ordinary X.509 CA certificates without relying on a timestamp / Rekor at all, the way the TLS ecosystem does.)
- If there is a RFC 3161 timestamp, and Rekor log presence is processed but not required, we only require the certificate to be valid as of the timestamp, not as of the current time.

Basic summary of the ordering changes:
- The verification flow now:
  - _First_ ensures timestamp / Rekor presence; that obtains a trusted signature creation timestamp, if any
  - Then, if necessary, validates the certificate (requires the timestamp from previous step)
  - Cryptographically verifies the signature/payload
  - Only at the very end, processes the payload
- `VerifyBundle` now _first_ ensures that the entry was actually stored on Rekor, and only afterwards uses it to see whether it matches the supposedly-logged signature.

#### Release Note
Certificate expiry checking was made a bit more accurate.

#### Documentation
N/A.

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -164,7 +164,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		if err != nil {
 			return fmt.Errorf("getting Fulcio roots: %w", err)
 		}
-		co.IntermediateCerts, err = fulcio.GetIntermediates()
+		co.UntrustedIntermediateCerts, err = fulcio.GetIntermediates()
 		if err != nil {
 			return fmt.Errorf("getting Fulcio intermediates: %w", err)
 		}
@@ -205,7 +205,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 			if err != nil {
 				return fmt.Errorf("getting Fulcio roots: %w", err)
 			}
-			co.IntermediateCerts, err = fulcio.GetIntermediates()
+			co.UntrustedIntermediateCerts, err = fulcio.GetIntermediates()
 			if err != nil {
 				return fmt.Errorf("getting Fulcio intermediates: %w", err)
 			}
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -264,18 +264,10 @@ func ValidateAndUnpackCert(untrustedCert *x509.Certificate, co *CheckOpts) (sign
 	return verifier, nil
 }
 
-// CheckCertificatePolicy checks that the certificate subject and issuer match
-// the expected values.
-func CheckCertificatePolicy(correctlySignedCert *x509.Certificate, co *CheckOpts) error {
+// validateIssuerPolicy checks that the certificate is valid per co.Identities.
+// It should only be called after the certificate has been determined to be correctly signed.
+func validateIssuerPolicy(correctlySignedCert *x509.Certificate, co *CheckOpts) error {
 	ce := CertExtensions{Cert: correctlySignedCert}
-
-	if err := validateCertSubject(correctlySignedCert, co); err != nil {
-		return err
-	}
-
-	if err := validateCertExtensions(ce, co); err != nil {
-		return err
-	}
 	issuer := ce.GetIssuer()
 	// If there are identities given, go through them and if one of them
 	// matches, call that good, otherwise, return an error.
@@ -330,6 +322,24 @@ func CheckCertificatePolicy(correctlySignedCert *x509.Certificate, co *CheckOpts
 			}
 		}
 		return &VerificationError{"none of the expected identities matched what was in the certificate"}
+	}
+	return nil
+}
+
+// CheckCertificatePolicy checks that the certificate subject and issuer match
+// the expected values.
+func CheckCertificatePolicy(correctlySignedCert *x509.Certificate, co *CheckOpts) error {
+	ce := CertExtensions{Cert: correctlySignedCert}
+
+	if err := validateCertSubject(correctlySignedCert, co); err != nil {
+		return err
+	}
+
+	if err := validateCertExtensions(ce, co); err != nil {
+		return err
+	}
+	if err := validateIssuerPolicy(correctlySignedCert, co); err != nil {
+		return err
 	}
 	return nil
 }
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -481,12 +481,12 @@ func ValidateAndUnpackCertWithChain(cert *x509.Certificate, chain []*x509.Certif
 }
 
 func tlogValidateEntry(ctx context.Context, client *client.Rekor, rekorPubKeys *TrustedRekorPubKeys,
-	sig oci.Signature, pem []byte) (*models.LogEntryAnon, error) {
-	b64sig, err := sig.Base64Signature()
+	untrustedSig oci.Signature, pem []byte) (*models.LogEntryAnon, error) {
+	b64sig, err := untrustedSig.Base64Signature()
 	if err != nil {
 		return nil, err
 	}
-	payload, err := sig.Payload()
+	payload, err := untrustedSig.Payload()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1062,24 +1062,24 @@ func VerifyRFC3161Timestamp(untrustedSig oci.Signature, tsaCerts *x509.CertPool)
 		return nil, fmt.Errorf("reading base64signature: %w", err)
 	}
 
-	var tsBytes []byte
+	var untrustedTSAArtifact []byte
 	if len(untrustedB64Sig) == 0 {
 		// For attestations, the Base64Signature is not set, therefore we rely on the signed payload
 		untrustedPayload, err := untrustedSig.Payload()
 		if err != nil {
 			return nil, fmt.Errorf("reading the payload: %w", err)
 		}
-		tsBytes = untrustedPayload
+		untrustedTSAArtifact = untrustedPayload
 	} else {
 		// create timestamp over raw bytes of signature
 		untrustedRawSig, err := base64.StdEncoding.DecodeString(untrustedB64Sig)
 		if err != nil {
 			return nil, err
 		}
-		tsBytes = untrustedRawSig
+		untrustedTSAArtifact = untrustedRawSig
 	}
 
-	err = tsaverification.VerifyTimestampResponse(untrustedTimestamp.SignedRFC3161Timestamp, bytes.NewReader(tsBytes), tsaCerts)
+	err = tsaverification.VerifyTimestampResponse(untrustedTimestamp.SignedRFC3161Timestamp, bytes.NewReader(untrustedTSAArtifact), tsaCerts)
 	if err != nil {
 		return nil, fmt.Errorf("unable to verify TimestampResponse: %w", err)
 	}
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1050,10 +1050,10 @@ func VerifyBundle(sig oci.Signature, co *CheckOpts) (bool, error) {
 // returns the timestamp value.
 // It returns (nil, nil) if there is no timestamp, or (nil, err) if there is an invalid timestamp.
 func VerifyRFC3161Timestamp(untrustedSig oci.Signature, tsaCerts *x509.CertPool) (*timestamp.Timestamp, error) {
-	ts, err := untrustedSig.RFC3161Timestamp()
+	untrustedTimestamp, err := untrustedSig.RFC3161Timestamp()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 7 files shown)


znewman01 (commented):
Hey, thanks for this! Very good idea overall and I'm strongly in favor.

I'd like even better to enforce this at the type system layer, rather than simply via naming. You might be able to do some cool encapsulation tricks. (Don't make it possible to instantiate a `Foo` struct from outside the module; instead, you must create an `UnverifiedFoo` and then turn it into a `Foo`.)

But in the shorter term—is there any chance you could break this up into 2 or 3 smaller PRs? I find once I'm reviewing more than 100–200 lines of code my eyes gloss over and I find it difficult to actually make sure the changes are correct.

haydentherapper:
+1, for PRs that both refactor and modify behavior, we should separate those into separate PRs, especially for critical parts of the code base like verify. Very minimal test changes for refactors, with more unit tests for behavior changes. 

> If we do rely on the certificate, we now always test expiry against some time. Previously that didn’t happen if there was no RFC 3161 timestamp and Rekor log presence was opted out. (I.e. this should allow using ordinary X.509 CA certificates without relying on a timestamp / Rekor at all, the way the TLS ecosystem does.)

Is this correct? We default to the current time when checking expiration time. If no bundle or timestamp is provided, we use the current time. There should be no case when cert expiration is not checked, but let me know if you found one.

> If there is a RFC 3161 timestamp, and Rekor log presence is processed but not required, we only require the certificate to be valid as of the timestamp, not as of the current time.

This sounds like a good change! Note that if both a bundle and 3161 time stamp are present, we want to check the times for both. Maybe we make this configurable in the future, but for now, I’d like to continue checking both, because we do expect both to be valid.

mtrmac:
> Hey, thanks for this! Very good idea overall and I'm strongly in favor.
> 
> I'd like even better to enforce this at the type system layer, rather than simply via naming. You might be able to do some cool encapsulation tricks. (Don't make it possible to instantiate a `Foo` struct from outside the module; instead, you must create an `UnverifiedFoo` and then turn it into a `Foo`.)

Go doesn’t have a strong enough visibility system for that, other than using separate subpackages — and I think that would spread the core verification logic over too many separate files, making the end resulting logic hard to follow. (That’s especially the case for certificates, which really have several mostly-independent correctness criteria, like the various subject restrictions that can be configured.)

There’s certainly much more that could be done (e.g. maybe the `oci.Signature` type should name all fields `Untrusted…`). I wanted to get the core verification logic ordered, and I can’t immediately spend significantly more time on a style proposal beyond that.

> But in the shorter term—is there any chance you could break this up into 2 or 3 smaller PRs? I find once I'm reviewing more than 100–200 lines of code my eyes gloss over and I find it difficult to actually make sure the changes are correct.

Sure, the set of commits should be correct at every point, so It can split in arbitrary places.

asraa:
> Sure, the set of commits should be correct at every point, so It can split in arbitrary places.

@mtrmac could we start with the timestamp check changes in the verification flow and do some renaming as follow-ups once there's no functional changes? 

I'm looking at
https://github.com/sigstore/cosign/pull/2482/commits/3b21d42db128878485000b9c49c37f021267ee11
https://github.com/sigstore/cosign/pull/2482/commits/eb1f06001b95a0c99a69fe06415315614d3f0d6a
https://github.com/sigstore/cosign/pull/2482/commits/0a343bcbe759a8fbe341bb805db3cf50fec6d0d6

pkg/cosign/verify.go
```diff
@@ -611,117 +641,145 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
...
+		} else {
+			// If the --offline flag was specified, fail here. bundleVerified returns false with
+			// no error when there was no bundle provided.
+			// TODO: You can be offline when you are verifying with a public key. This shouldn't
+			// error out, but maybe should be a log message.
+			if co.Offline {
+				return false, fmt.Errorf("offline verification failed")
+			}
+
+			if co.RekorClient != nil {
```
asraa, line 692: I think this check was always very confusing to me. It seems like if you aren't skipping TLOG verification, then if there's no rekor client, somehow you can still pass without checking the tlog.

do you think maybe it should be an error that co.RekorClient is nil when we have to do tlog verification and never got a bundle?

asraa (commented):
this is looking great! i've just reviewing the verify main command.

cmd/cosign/cli/verify/verify.go
```diff
@@ -154,7 +154,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		if err != nil {
 			return fmt.Errorf("getting Fulcio roots: %w", err)
 		}
-		co.IntermediateCerts, err = fulcio.GetIntermediates()
+		co.UntrustedIntermediateCerts, err = fulcio.GetIntermediates()
```
haydentherapper, line 157: Could we keep the same name, `IntermediateCerts`?

I would prefer we don't used "Untrusted" in the name. This term comes from openssl from what I've seen, and I personally don't think it's correct to only use intermediate certificates for chain building ("untrusted"). Intermediates should be trusted, or at least come from a trusted source.

haydentherapper (changes requested):

cmd/cosign/cli/verify/verify.go
asraa: This one is confusing. These particular ones are coming from a trusted source, so I can agree to see it named the same. The ones found on the registry through CertChain are untrusted intermediates

haydentherapper:
My 2cents: I see multiple types of refactoring in this PR - Name changes, reordering, etc. Given the criticality of this package, I think we should scope down the PR even more. I see some of the commits are behavior changes too - These definitely should be in their own PR, especially since I don't see tests that accompany them (which suggests we don't have sufficient testing in Cosign, which we don't :) ).

pkg/cosign/verify.go
mtrmac: Right now, the API is such that external callers can call `VerifyImageSignature` and get `(bundleVerified = false, err = nil)`. So this PR doesn’t change that, it’s a large enough set of changes already.

I agree that it seems dangerous and undesirable; IMHO external users of `pkg/cosign` should just provide a policy, and not be relied upon to correctly enforce other conditions on top. Conceptually, I think of “change the API philosophy” and “change the implementation philosophy” are fairly separate conversations — OTOH, admittedly, the question of the right timestamp to use for certificate expiry means the two can’t separated that cleanly.

cmd/cosign/cli/verify/verify.go
mtrmac: Yes, see `co.UntrustedIntermediateCerts = untrustedPool`. So, in that sense, I think changing the name of `CheckOpts` to `UntrustedIntermediateCerts` has been a success exactly because the trusted nature of those coming from `fulcio.GetIntermediates` is insufficient to make the field trusted.

(I’m also not sure what a “trusted intermediate” means. Is that a root of trust, i.e. a non-intermediate root? Or an intermediate that we still require to be correctly signed by a root of trust, i.e. something where we don’t actually require any trust?)

---

Actually, what I _really_ think is that the caller-provided policy should be immutable and should not be modified to hold single-instance data within `verify.go` at all. Then we might even more precisely differentiate between
- completely-untrusted signature-carried data
- externally Fulcio-provided data (acceptable if the connection to Fulcio is trusted)
- completely locally-provided 100%-trusted policy

and potentially make different decisions based on those details. That would be a different PR.

mtrmac:
Right, I have rebased this first to have a baseline from which I can extract a smaller subset.

Then I intend to keep this PR to be a catch-all that merges branches of the smaller subsets, keeping a record of what is merged/being reviewed / outstanding (and dropping the rejected parts). If/as time allows.

cmd/cosign/cli/verify/verify.go
haydentherapper: For a core library function, I think we should not be opinionated if the certificates are trusted or untrusted. For example, we don't know where the root certificate came from either when the checkopts is populated. I don't see a reason to distinguish between untrusted and trusted in the CheckOpts. Internally, we can name variables differently based on their source.

mtrmac: > I think we should not be opinionated if the certificates are trusted or untrusted.

I… don’t know what that would mean.

The verification code _must_ be certain about which data is a part of the root of trust that is, by necessity, assumed valid, and which data is part of the external attack surface that is, by default, suspect. At the very least these two categories are 100% distinct. There might be even more categories in between (e.g. the intermediate certificates might be more trusted than the underlying signatures, but they are still potentially a part of the attack surface).

(And then the API should help the caller of the API express its assumptions clearly, and to help expose any incorrect assumptions the caller might be making.

From that point of view, calling the certificates “untrusted” is _good_ for the caller, because it tells the caller the caller doesn’t need to make much effort at all to ensure validity of the data. “A trusted system is one whose failure would break a security policy”.)

---

I don’t care at all about the individual naming of the options. If you don’t want the field to be named `Untrusted`, and to instead have a `CheckOpts.{TrustedSigVerifier,TrustedRootCerts}`, sure, that might be even more explicit. If you think the distinction between trusted and untrusted data is sufficient to express in comments and doesn’t need to be in field names, that’s not my preference but I can live with that just fine.

But I read the above as suggesting that it doesn’t make a difference whether the data is trusted or not, and I just don’t know what to do with that. Probably I’m completely misunderstanding your point.

haydentherapper:
@mtrmac Do you think it'd be possible to have a PR just with the behavior changes first? We wanted to cut a release candidate soon for cosign 2.0, and one of the behavior changes in particular is a blocker for release imo:

```
If we do rely on the certificate, we now always test expiry against some time. Previously that didn’t happen if there was no RFC 3161 timestamp and Rekor log presence was opted out. (I.e. this should allow using ordinary X.509 CA certificates without relying on a timestamp / Rekor at all, the way the TLS ecosystem does.)
```

This should result in a verification failure. Reproduced with `cosign verify-blob --certificate blob.cert --signature blob.sig --insecure-skip-tlog-verify blob`, using an expired cert, it says it succeeds.

fyi @priyawadhwa

mtrmac:
> @mtrmac could we start with the timestamp check changes in the verification flow and do some renaming as follow-ups once there's no functional changes?
> 
> I'm looking at [3b21d42](https://github.com/sigstore/cosign/commit/3b21d42db128878485000b9c49c37f021267ee11) [eb1f060](https://github.com/sigstore/cosign/commit/eb1f06001b95a0c99a69fe06415315614d3f0d6a) [0a343bc](https://github.com/sigstore/cosign/commit/0a343bcbe759a8fbe341bb805db3cf50fec6d0d6)

---

> @mtrmac Do you think it'd be possible to have a PR just with the behavior changes first? We wanted to cut a release candidate soon for cosign 2.0, and one of the behavior changes in particular is a blocker for release imo:
> 
> ```
> If we do rely on the certificate, we now always test expiry against some time. Previously that didn’t happen if there was no RFC 3161 timestamp and Rekor log presence was opted out. (I.e. this should allow using ordinary X.509 CA certificates without relying on a timestamp / Rekor at all, the way the TLS ecosystem does.)
> ```
> 
> This should result in a verification failure. Reproduced with `cosign verify-blob --certificate blob.cert --signature blob.sig --insecure-skip-tlog-verify blob`, using an expired cert, it says it succeeds.

https://github.com/sigstore/cosign/pull/2504: as it happens, I was planning to add that one on top of the previous request.

If you want to urgently backport the other change, feel free to; I’m afraid I’m done for the week.

mtrmac:
This PR is now based on top of #2504 (and includes an explicit merge commit). So I’m marking it as a draft to direct reviews and conversations to the #2504 subset for now.

mtrmac:
Sure, rebased.

Do you have a preference for a subsection to propose next? Some options:
- Pure variable renames
- The certificate validation functions, leading up to the split of `validateCertIssuanceAndSubject` from `ValidateAndUnpackCert`
- The reordering of `VerifyBundle` (probably smallest)

Issue #2476, opened 2022-11-21
Custom Annotation Support for Attestations


chaospuppy:
**Description**

Currently, when using `cosign download attestation`, every attestation attached to the artifact is returned, which forces additional parsing to take place client-side using tools like `jq` to decode and filter out only the desired attestation (based on `predicateType`, for example).  Adding support for annotations on artifact layers would achieve two things:
1. It would make retrieval of the desired attestation using the `cosign` CLI trivial (`cosign download attestation --annotation foo=bar <image>`)
2. Verifying a particular attestation would also be trivial with `cosign verify-attestation --annotation <image>`
3. It would have the side effect of making the annotations visible as part of what gets returned by the `v2` registry API, which means any language with basic support for making requests would be able to locate the desired attestation (albeit without attestation verification via `cosign verify-attestatation`)

This idea had been brought up before in #1773 and a PR was opened for it in #1934.  That PR was ultimately closed [due to concerns](https://github.com/sigstore/cosign/pull/1934#issuecomment-1152762144) over how the annotation itself could be verified as part of the payload.  I had [suggested](https://github.com/sigstore/cosign/pull/1934#issuecomment-1294279234) that perhaps the annotation is specified in both the StatementHeader for the layer and in the payload itself.  In this way, the annotation could be used to filter the desired layer without the need to download all attestations and decode them using an external tool like `jq`, and the `verify-attestation` would be performed by first validating the payload signature and then making sure the user-provided `--annotation` value matched the annotation in the signed payload.

I'd be willing to contribute a PR to try to provide this, if it sounds like a good idea.

znewman01:
@asraa WDYT?

chaospuppy:
I’ve created #2484, which partially solves this by reusing the predicateType as an annotation on the attestation manifest.  This exposes the predicateType to simple GET requests, which helps external tools not written in Go have an easier time of finding a particular attestation,  and can also be used to be more selective about which attestation should be output with `cosign download attestation —predicate-type <string>`.  Arbitrary annotations are probably still desirable for some use cases though.

PR #2484, opened 2022-11-27
2476 predicate type download

chaospuppy:
Relates to #2476 
#### Summary

This PR will add a `predicateType` annotation to the manifest of each attestation.  This annotation can then be used with the `cosign download attestation` command to output only the desired annotation, using a new `—predicate-type` flag.  If the flag is used, but no `predicateType` annotation is matched, an error is raised.  If the flag is not provided, all attestations are returned, as they are currently.
#### Release Note

New Feature: filter attestations output from `cosign download attestation` command by using the `—predicate-type` flag.
#### Documentation

This will require an update to the README for the `cosign download` command, which will be updated provided this seems like a good idea.

### DIFF
--- a/cmd/cosign/cli/download.go
+++ b/cmd/cosign/cli/download.go
@@ -84,19 +84,21 @@ func downloadSBOM() *cobra.Command {
 
 func downloadAttestation() *cobra.Command {
 	o := &options.RegistryOptions{}
+	ao := &options.AttestationDownloadOptions{}
 
 	cmd := &cobra.Command{
 		Use:              "attestation",
 		Short:            "Download in-toto attestations from the supplied container image",
-		Example:          "  cosign download attestation <image uri>",
+		Example:          "  cosign download attestation <image uri> [--predicate-type]",
 		Args:             cobra.ExactArgs(1),
 		PersistentPreRun: options.BindViper,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return download.AttestationCmd(cmd.Context(), *o, args[0])
+			return download.AttestationCmd(cmd.Context(), *o, *ao, args[0])
 		},
 	}
 
 	o.AddFlags(cmd)
+	ao.AddFlags(cmd)
 
 	return cmd
 }
--- a/cmd/cosign/cli/download/attestation.go
+++ b/cmd/cosign/cli/download/attestation.go
@@ -26,7 +26,7 @@ import (
 	"github.com/sigstore/cosign/v2/pkg/cosign"
 )
 
-func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef string) error {
+func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOptions options.AttestationDownloadOptions, imageRef string) error {
 	ref, err := name.ParseReference(imageRef, regOpts.NameOptions()...)
 	if err != nil {
 		return err
@@ -39,10 +39,20 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, imageR
 	if err != nil {
 		return err
 	}
-	attestations, err := cosign.FetchAttestationsForReference(ctx, ref, ociremoteOpts...)
+
+	var predicateURI string
+	if attOptions.PredicateType != "" {
+		predicateURI, err = options.ParsePredicateType(attOptions.PredicateType)
+		if err != nil {
+			return err
+		}
+	}
+
+	attestations, err := cosign.FetchAttestationsForReference(ctx, ref, predicateURI, ociremoteOpts...)
 	if err != nil {
 		return err
 	}
+
 	for _, att := range attestations {
 		b, err := json.Marshal(att)
 		if err != nil {
--- a/cmd/cosign/cli/download/attestation.go
+++ b/cmd/cosign/cli/download/attestation.go
@@ -40,15 +40,15 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOpt
 		return err
 	}
 
-	var predicateURI string
+	var predicateType string
 	if attOptions.PredicateType != "" {
-		predicateURI, err = options.ParsePredicateType(attOptions.PredicateType)
+		predicateType, err = options.ParsePredicateType(attOptions.PredicateType)
 		if err != nil {
 			return err
 		}
 	}
 
-	attestations, err := cosign.FetchAttestationsForReference(ctx, ref, predicateURI, ociremoteOpts...)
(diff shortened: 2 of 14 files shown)


pkg/cosign/fetch.go
```diff
@@ -138,20 +138,42 @@ func FetchAttestationsForReference(ctx context.Context, ref name.Reference, opts
...
+			attestations = append(attestations, a)
+			return nil
 		})
 	}
 	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 
+	if len(attestations) == 0 && predicateURI != "" {
+		return nil, fmt.Errorf("no layers of predicate type %s found", predicateURI)
```
hectorj2f, line 174: you might want to use the following error message: `no attestations with predicate type '%s' found`

cmd/cosign/cli/download/attestation.go
```diff
@@ -34,10 +34,20 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, imageR
 	if err != nil {
 		return err
 	}
-	attestations, err := cosign.FetchAttestationsForReference(ctx, ref, ociremoteOpts...)
+
+	var predicateURI string
+	if attOptions.PredicateType != "" {
+		predicateURI, err = options.ParsePredicateType(attOptions.PredicateType)
```
hectorj2f, line 40: Suggestion: I'd keep name it as `predicateType` instead of `predicateURI`.

cmd/cosign/cli/download.go
```diff
@@ -84,19 +84,21 @@ func downloadSBOM() *cobra.Command {
...
 	cmd := &cobra.Command{
 		Use:              "attestation",
 		Short:            "Download in-toto attestations from the supplied container image",
-		Example:          "  cosign download attestation <image uri>",
+		Example:          "  cosign download attestation <image uri> [--predicate-type]",
 		Args:             cobra.ExactArgs(1),
 		PersistentPreRun: options.BindViper,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return download.AttestationCmd(cmd.Context(), *o, args[0])
+			return download.AttestationCmd(cmd.Context(), *o, *ao, args[0])
```
hectorj2f, line 96: Please, run `make docgen`

cmd/cosign/cli/options/download.go
```diff
@@ -22,10 +22,22 @@ type SBOMDownloadOptions struct {
...
 // AddFlags implements Interface
 func (o *SBOMDownloadOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.Platform, "platform", "",
 		"download SBOM for a specific platform image")
 }
+
+// AddFlags implements Interface
+func (o *AttestationDownloadOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.PredicateType, "predicate-type", "",
+		"download attestation matching supplied predicateType")
```
hectorj2f, line 42: nit: `matching supplied predicate type`

chaospuppy:
Resolved all threads @hectorj2f

pkg/cosign/fetch.go
```diff
@@ -138,20 +138,42 @@ func FetchAttestationsForReference(ctx context.Context, ref name.Reference, opts
...
+			if !ok || pt != predicateType {
+				continue
+			}
+		}
+		att := att
+		var a AttestationPayload
 		g.Go(func() error {
 			attestPayload, _ := att.Payload()
-			return json.Unmarshal(attestPayload, &attestations[i])
+			err := json.Unmarshal(attestPayload, &a)
```
hectorj2f, line 161: Question: Wouldn't it be an alternative to extract the predicateType from the attestation payload instead of relying on annotations ?

hectorj2f (commented):
Regardless of the proposed solution, we are missing tests here.

pkg/cosign/fetch.go
chaospuppy: It would be, but doing so requires decoding each payload first to inspect the predicateType, which is both more memory-intensive and mean that third-party tools (written in languages other than Go) that are trying to find the desired attestation must download each attestation blob and decode the payload as well.  

Adding the annotation means the desired attestation layer can be found using basic `GET` requests against the docker v2 API, which means other tools can integrate more easily with the artifacts produced by Cosign.

hectorj2f: Yes, that makes sense although this command will be misleading as the flag describes an action that could not happen. If the command flag aims to download attestations with a predicate type, if there aren't annotations for these attestations it will return an empty list even if there are attestations.

chaospuppy: Ah good point, I will update the help text and add unit tests this week.  I think the behavior makes sense though - if the flag is provided and no *annotations* are matched, throw an error to avoid confusion around why no attestations were returned, else if no flag was provided, return all attestations.

chaospuppy: Well, on second thought - this MR will ensure that the `predicateType` annotation *always* exists.  So the behavior would be the same.  The only exception to this is legacy attestations put in place before this PR.  Either way, the text should be clarified.

chaospuppy:
Looks like I have a few more places to set arguments to FetchAttestationsForReference as well.

chaospuppy:
E2E tests have been updated, as has the help text for `--predicate-type`.

chaospuppy:
@hectorj2f any outstanding issues on this PR that you see?

chaospuppy:
Hi again @hectorj2f, just pinging on this again to see if it could be merged in.

cmd/cosign/cli/options/download.go
```diff
@@ -22,10 +22,22 @@ type SBOMDownloadOptions struct {
...
 // AddFlags implements Interface
 func (o *SBOMDownloadOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.Platform, "platform", "",
 		"download SBOM for a specific platform image")
 }
+
+// AddFlags implements Interface
+func (o *AttestationDownloadOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.PredicateType, "predicate-type", "",
+		"download attestation with matching predicateType annotation")
```
hectorj2f, line 42: Future suggestion: this could be an array of matching annotations.

hectorj2f:
I added more reviewers to avoid missing anything here. It looks good to me.

znewman01 (approved):
Sorry for the delay! Seems ok to me

znewman01:
BTW you may have to rebase to get tests to run, sorry! I'll re-approve quick this time.

chaospuppy:
@znewman01 no worries, rebased.

znewman01 (approved):

Issue #2497, opened 2022-11-30
cosign attest-blob prints both the base64 encoded DSSE envelope and the attestation to the output


asraa:
**Description**

```
$  cosign attest-blob --predicate predicate.json --key cosign.key test.txt --output-file attestation.jsonl
$ cat attestation.jsonl 
eyJwYXlsb2FkVHlwZSI6ImFwcGxpY2F0aW9uL3ZuZC5pbi10b3RvK2pzb24iLCJwYXlsb2FkIjoiZXlKZmRIbHdaU0k2SW1oMGRIQnpPaTh2YVc0dGRHOTBieTVwYnk5VGRHRjBaVzFsYm5RdmRqQXVNU0lzSW5CeVpXUnBZMkYwWlZSNWNHVWlPaUpqYjNOcFoyNHVjMmxuYzNSdmNtVXVaR1YyTDJGMGRHVnpkR0YwYVc5dUwzWXhJaXdpYzNWaWFtVmpkQ0k2VzNzaWJtRnRaU0k2SW5SbGMzUXVkSGgwSWl3aVpHbG5aWE4wSWpwN0luTm9ZVEkxTmlJNkltSmlZakE0TURObVpXSTBaRFkyWVdFME9URTNPVGhpTkRGbE5XTTNZV05tT1dJNE9EVTVPRFl6TURrek5qUmlNR1UyWlRWaFpESTJPVGd3WkRCaE5USWlmWDFkTENKd2NtVmthV05oZEdVaU9uc2lSR0YwWVNJNkluc2dYQ0ppZFdsc1pGUjVjR1ZjSWpvZ1hDSjRYQ0lzSUZ3aVluVnBiR1JsY2x3aU9pQjdJRndpYVdSY0lqb2dYQ0l5WENJZ2ZTd2dYQ0p5WldOcGNHVmNJam9nZTMwZ2ZWeHVJaXdpVkdsdFpYTjBZVzF3SWpvaU1qQXlNaTB4TVMwek1GUXlNRG94Tnpvd00xb2lmWDA9Iiwic2lnbmF0dXJlcyI6W3sia2V5aWQiOiIiLCJzaWciOiJNRU1DSDFHZk1rTWFoMy9FazlReVM2NDFGL1dOY2FWZ2s1MlBYZWZPSEErR3JQUUNJQ3RXY2lSdE5Od21nak04RHhWUkY4KzUvYy9qQlUxYUNLanlFVnRmOHJScCJ9XX0=
[...]

PR #2498, opened 2022-11-30
fix: removes attestation payload from attest-blob's output & no base64 encoding

asraa:
Fixes https://github.com/sigstore/cosign/issues/2497

* fix!: This writes and prints the signature (DSSE envelope) output as the JSON, NOT base64-encoded. Previously, this would base64-encode the envelope
* fix!: This removes printing the attestation JSON to the std output. This confuses piping the output, since now both the DSSE envelope and the attestation are printed.
* fix!: `verify-blob-attestation` now expects a JSON DSSE envelope (NOT base64-encoded)

BREAKINGCHANGE
* `verify-blob-attestation` expects the JSON DSSE envelope. Previously, this was expected to be base64-encoded.
* `attest-blob` outputs only the JSON DSSE envelope. Previously, it also printed the intoto attestation payload and the DSSE envelope was additionally base64-encoded.

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"context"
 	"crypto"
-	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -127,7 +126,6 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		return errors.Wrap(err, "signing")
 	}
 
-	sig = []byte(base64.StdEncoding.EncodeToString(sig))
 	if c.OutputSignature != "" {
 		if err := os.WriteFile(c.OutputSignature, sig, 0600); err != nil {
 			return fmt.Errorf("create signature file: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -127,7 +126,6 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		return errors.Wrap(err, "signing")
 	}
 
-	sig = []byte(base64.StdEncoding.EncodeToString(sig))
 	if c.OutputSignature != "" {
```
priyawadhwa, line 129: IIRC we don't set a default value for `OutputSignature`. WDYT of defaulting to a file called `cosign_signature` or something so that it's always stored somewhere?

priyawadhwa (approved):
LGTM!

cmd/cosign/cli/attest/attest_blob.go
asraa: It's always printing to STDOUT which is also the convention for `sign-blob` so people can just cat it into the file. I think it's OK like that?

hectorj2f (approved):

cmd/cosign/cli/attest/attest_blob.go
priyawadhwa: sgtm!

PR #2499, opened 2022-11-30
Updates to Timestamp signing and verification

Hayden-IO:
* Switch to using the raw signature rather than base64 signature for OCI and blob signing
* For blob signing, write only the timestamp to disk, not the LocalSignedPayload (since that's already written with the bundle. I want to avoid conflating the Rekor bundle with the TS)
* For blob verification, expect only a timestamp in the file. If you don't pass a bundle, you'll need to also pass the signature by flag
* Some nits from the previous PR

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -72,19 +72,20 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 
 	signedPayload := cosign.LocalSignedPayload{}
 
+	var rfc3161Timestamp *cbundle.RFC3161Timestamp
 	if ko.TSAServerURL != "" {
 		clientTSA, err := tsaclient.GetTimestampClient(ko.TSAServerURL)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create TSA client: %w", err)
 		}
-		b64Sig := []byte(base64.StdEncoding.EncodeToString(sig))
 
-		respBytes, err := tsa.GetTimestampedSignature(b64Sig, clientTSA)
+		respBytes, err := tsa.GetTimestampedSignature(sig, clientTSA)
 		if err != nil {
 			return nil, err
 		}
 
-		signedPayload.RFC3161Timestamp = cbundle.TimestampToRFC3161Timestamp(respBytes)
+		rfc3161Timestamp = cbundle.TimestampToRFC3161Timestamp(respBytes)
+		// TODO: Consider uploading RFC3161 TS to Rekor
 	}
 	if ShouldUploadToTlog(ctx, ko, nil, tlogUpload) {
 		rekorBytes, err = sv.Bytes(ctx)
@@ -103,18 +104,15 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 		signedPayload.Bundle = cbundle.EntryToBundle(entry)
 	}
 
-	// if bundle is specified, just do that and ignore the rest
-	if ko.RFC3161TimestampPath != "" {
-		signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(sig)
-
-		contents, err := json.Marshal(signedPayload)
+	if ko.RFC3161TimestampPath != "" && rfc3161Timestamp != nil {
+		ts, err := json.Marshal(rfc3161Timestamp)
 		if err != nil {
 			return nil, err
 		}
-		if err := os.WriteFile(ko.RFC3161TimestampPath, contents, 0600); err != nil {
+		if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
 			return nil, fmt.Errorf("create rfc3161 timestamp file: %w", err)
 		}
-		fmt.Printf("RF3161 timestamp bundle wrote in the file %s\n", ko.RFC3161TimestampPath)
+		fmt.Printf("RFC3161 timestamp bundle written to file %s\n", ko.RFC3161TimestampPath)
 	}
 
 	// if bundle is specified, just do that and ignore the rest
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -74,6 +74,10 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 
 	var rfc3161Timestamp *cbundle.RFC3161Timestamp
 	if ko.TSAServerURL != "" {
+		if ko.RFC3161TimestampPath == "" {
+			return nil, fmt.Errorf("timestamp output path must be set")
+		}
+
 		clientTSA, err := tsaclient.GetTimestampClient(ko.TSAServerURL)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create TSA client: %w", err)
@@ -86,6 +90,18 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 
 		rfc3161Timestamp = cbundle.TimestampToRFC3161Timestamp(respBytes)
 		// TODO: Consider uploading RFC3161 TS to Rekor
+
+		if rfc3161Timestamp == nil {
+			return nil, fmt.Errorf("rfc3161 timestamp is nil")
+		}
+		ts, err := json.Marshal(rfc3161Timestamp)
+		if err != nil {
+			return nil, err
+		}
+		if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
+			return nil, fmt.Errorf("create rfc3161 timestamp file: %w", err)
+		}
+		fmt.Printf("RFC3161 timestamp bundle written to file %s\n", ko.RFC3161TimestampPath)
 	}
 	if ShouldUploadToTlog(ctx, ko, nil, tlogUpload) {
 		rekorBytes, err = sv.Bytes(ctx)
@@ -102,17 +118,6 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 		}
 		fmt.Fprintln(os.Stderr, "tlog entry created with index:", *entry.LogIndex)
 		signedPayload.Bundle = cbundle.EntryToBundle(entry)
-	}
-
-	if ko.RFC3161TimestampPath != "" && rfc3161Timestamp != nil {
-		ts, err := json.Marshal(rfc3161Timestamp)
-		if err != nil {
-			return nil, err
-		}
-		if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
-			return nil, fmt.Errorf("create rfc3161 timestamp file: %w", err)
-		}
-		fmt.Printf("RFC3161 timestamp bundle written to file %s\n", ko.RFC3161TimestampPath)
 	}
 
 	// if bundle is specified, just do that and ignore the rest
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -99,9 +99,9 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 			return nil, err
 		}
 		if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
-			return nil, fmt.Errorf("create rfc3161 timestamp file: %w", err)
+			return nil, fmt.Errorf("create RFC3161 timestamp file: %w", err)
 		}
-		fmt.Printf("RFC3161 timestamp bundle written to file %s\n", ko.RFC3161TimestampPath)
+		fmt.Fprintf(os.Stderr, "RFC3161 timestamp written to file %s\n", ko.RFC3161TimestampPath)
 	}
 	if ShouldUploadToTlog(ctx, ko, nil, tlogUpload) {
 		rekorBytes, err = sv.Bytes(ctx)
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -33,6 +33,7 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
 	"github.com/sigstore/cosign/pkg/blob"
 	"github.com/sigstore/cosign/pkg/cosign"
+	"github.com/sigstore/cosign/pkg/cosign/bundle"
 	"github.com/sigstore/cosign/pkg/cosign/pivkey"
 	"github.com/sigstore/cosign/pkg/cosign/pkcs11key"
 	"github.com/sigstore/cosign/pkg/oci/static"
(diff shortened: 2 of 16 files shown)


haydentherapper:
Still working on e2e testing and some more unit tests, but just wanted to get this up early for feedback

haydentherapper:
Now it should be go, would like to add some more tests in verify-blob for completeness, but might do that in a follow up.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -67,19 +67,20 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
...
+	var rfc3161Timestamp *cbundle.RFC3161Timestamp
 	if ko.TSAServerURL != "" {
 		clientTSA, err := tsaclient.GetTimestampClient(ko.TSAServerURL)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create TSA client: %w", err)
 		}
-		b64Sig := []byte(base64.StdEncoding.EncodeToString(sig))
 
-		respBytes, err := tsa.GetTimestampedSignature(b64Sig, clientTSA)
+		respBytes, err := tsa.GetTimestampedSignature(sig, clientTSA)
```
hectorj2f, line 77: Q: why are we using b64sig ? I remember we agreed on using base64 for the blob.

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -552,6 +525,10 @@ func TestVerifyBlob(t *testing.T) {
 				expiredLeafPem, true)},
 			shouldErr: true,
 		},
+		// TODO: Add tests for TSA:
```
hectorj2f, line 528: This basically removes tests that we had in place. I am not sure we want to change functionality and remove tests. How does it work if I only sign/verify a blob without using tlog ?

internal/pkg/cosign/tsa/signer.go
```diff
@@ -58,7 +60,7 @@ func GetTimestampedSignature(sigBytes []byte, tsaClient *tsaclient.TimestampAuth
 		return nil, err
 	}
 
-	fmt.Fprintln(os.Stderr, "Timestamp fetched with time:", ts.Time)
+	fmt.Fprintln(os.Stderr, "Timestamp fetched with time: ", ts.Time)
```
hectorj2f, line 63: This is how it looks today: `Timestamp fetched with time: 2022-12-01 17:44:21 +0100 CET` This will add an extra whitespace here.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -73,16 +74,16 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 	opts := make([]static.Option, 0)
 
 	// Require a certificate/key OR a local bundle file that has the cert.
-	if options.NOf(c.KeyRef, c.CertRef, c.Sk, c.BundlePath, c.RFC3161TimestampPath) == 0 {
-		return fmt.Errorf("please provide a cert to verify against via --certificate or a bundle via --bundle or --rfc3161-timestamp-bundle")
```
hectorj2f, line 77: This changes the functionality forcing the usage of a bundle path instead of making it optional to passing a rfc3161 timestamp.

cmd/cosign/cli/sign/sign_blob.go
haydentherapper: We discussed this a bit in https://github.com/sigstore/timestamp-authority/issues/116, but the only reason base64 was used was because the interface exposed the signature as base64. I'd say it's better to use the raw bytes because you don't have any ambiguity (for example, some base64 encodings might allow variable numbers of `=`).

FWIW, every client can make their own decision on what's timestamped as long as the client's verification layer knows what was timestamped (like for example what you implemented for understanding attestations without signatures). But I think Cosign should stick with raw bytes (and this aligns nicely with the future bundle format that doesn't use base64).

cc @znewman01 and @asraa who I've chatted with about this

internal/pkg/cosign/tsa/signer.go
haydentherapper: If this is meant to be debug output, I think it's more readable with whitespace, but I'm not super picky if y'all think otherwise.

cmd/cosign/cli/verify/verify_blob_test.go
haydentherapper: I plan to add some more tests in, but the presence of the bundle is just a way to get the signature and/or other timestamp. If you don't use the tlog, then you must provide the cert/sig the same way you currently do, via --certificate/--signature flags.

internal/pkg/cosign/tsa/signer.go
hectorj2f: It is fine, I wanted to be sure we are aware that there is a whitespace between the time and the text.

cmd/cosign/cli/verify/verify_blob.go
haydentherapper: This is intentional. I don't think we should conflate the 3161 timestamp with the on-disk bundle.

To verify an artifact, you need:
* A signature, which can be provided in the "--bundle" or via a "--sig" flag
* A verification key, provided by "--certificate", "--sk", or "--bundle"

I would prefer to not change the on-disk bundle format and treat the timestamp as separate. I think it's a little confusing to bundle the artifact certificate with the timestamp, since those are unrelated. I would prefer that each element be provided separately.

So tl;dr, you would either:
* provide a --certificate flag, --signature flag, and --3161timestamp
* provide a --bundle flag and --3161timestamp

haydentherapper: Another reason I wanted to do this is that `--bundle` == `LocalSignedPayload`, and I didn't want to make changes to that struct. To me, it seems only like an optimization to include the 3161 timestamp in `LocalSignedPayload`, and I think simply tacking on a timestamp flag when you want to use the timestamp is cleaner and easier to explain.

Another option would be a different struct for storing the ts locally, like `Local3161Bundle`, but I think this is confusing to add another bundle format especially as we're working on another bundle format for the specs.

cmd/cosign/cli/verify/verify_blob_test.go
haydentherapper: With these changes, it actually requires less tests since this is only an additional flag that affects verification, rather than modifying codepaths that read in the cert/signature.

haydentherapper:
Verified with a local TSA instance:

```
# Sign blob and upload to Rekor
cosign sign-blob --timestamp-server-url http://localhost:3000 --rfc3161-timestamp tsaresponse --tlog-upload=true --output-certificate=blob.cert --output-signature=blob.sig --bundle blob.bundle blob

# Verify with local Rekor bundle
cosign verify-blob --rfc3161-timestamp tsaresponse --bundle blob.bundle --timestamp-cert-chain tsacertchain blob

# Verify with skipping checking Rekor
cosign verify-blob --rfc3161-timestamp tsaresponse --timestamp-cert-chain tsacertchain --certificate blob.cert --signature blob.sig --insecure-skip-tlog-verify --offline blob

# Sign blob without Rekor upload
cosign sign-blob --timestamp-server-url http://localhost:3000 --rfc3161-timestamp tsaresponse --tlog-upload=false --output-certificate=blob.cert --output-signature=blob.sig  blob
```

An unrelated issue is if I skip tlog-upload, then I can't output a certificate (`cosign sign-blob --timestamp-server-url http://localhost:3000 --rfc3161-timestamp tsaresponse --tlog-upload=false --output-certificate=blob.cert --output-signature=blob.sig blob` fails to output a cert), so I can't verify. I'll fix this separately.

specs/SIGNATURE_SPEC.md
```diff
@@ -129,6 +129,11 @@ Gyp4apdU7AXEwysEQIb034aPrTlpmxh90SnTZFs2DHOvCjCPPAmoWfuQUwPhSPRb
 
 For instructions on using the `bundle` for verification, see [USAGE.md](../USAGE.md#verify-a-signature-was-added-to-the-transparency-log).
 
+* `rfc3161timestamp` string
+
+  This OPTIONAL property contains a JSON formatted `RFC3161Timestamp` bundle containing the timestamp response from a
```
hectorj2f, line 134: I suggest to avoid using the term bundle here, as you mentioned it can be confusing.

cmd/cosign/cli/verify/verify_blob.go
hectorj2f: 👍🏻 It sounds good to me.

hectorj2f (commented):
LGTM, I added some nit comments. Ideally, I'd like to see some tests to validate the usage of the rfc3161timestamp with blobs.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -98,18 +99,15 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
 		signedPayload.Bundle = cbundle.EntryToBundle(entry)
 	}
 
-	// if bundle is specified, just do that and ignore the rest
-	if ko.RFC3161TimestampPath != "" {
-		signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(sig)
-
-		contents, err := json.Marshal(signedPayload)
+	if ko.RFC3161TimestampPath != "" && rfc3161Timestamp != nil {
```
znewman01, line 102: Should we handle the case where I ask for the timestamp to be written, but I don't actually get a timestamp?

Feels worth a warning at least.

internal/pkg/cosign/tsa/signer.go
znewman01: +1 to adding the whitespace

internal/pkg/cosign/tsa/signer.go
```diff
@@ -84,8 +86,14 @@ func (rs *signerWrapper) Sign(ctx context.Context, payload io.Reader) (oci.Signa
 		return nil, nil, err
 	}
 
-	// Here we get the response from the timestamped authority server
-	responseBytes, err := GetTimestampedSignature([]byte(b64Sig), rs.tsaClient)
+	// create timestamp over raw bytes of signature
+	rawSig, err := base64.StdEncoding.DecodeString(b64Sig)
```
znewman01, line 90: Feels a little gross to ask for the base64 signature then immediately decode it. I really want to push *all* encoding to the application boundaries as much as possible.

Can we add a TODO to add a `Signature()` method to `oci.Signature`?

pkg/cosign/verify.go
```diff
@@ -986,52 +986,59 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, co *CheckOpts, rekorCl
 	return true, nil
 }
 
-func VerifyRFC3161Timestamp(ctx context.Context, sig oci.Signature, tsaCerts *x509.CertPool) (bool, error) {
```
znewman01, line 989: Thank you for this change 🙏  I never figured out what `bundleVerified` means

znewman01 (commented):
Some nits but overall great changes

cmd/cosign/cli/sign/sign_blob.go
asraa: It seems like this drops the TSA timestamp when there's no timestamp path specified? it seems like there should be a default one then -- otherwise I'd think to only set TSAServerURL (and then my timestamp disappears)

specs/SIGNATURE_SPEC.md
asraa: +1 here it's not a bundle anymore of cert/sig as well, so it's just the RFC 3161 response, I think?

cmd/cosign/cli/sign/sign_blob.go
znewman01: We already do that if you don't specify `--output-signature` or `--output-certificate` 🤷 Maybe a `y/N` prompt would be the right amount of annoying here, you could bypass with `--yes`

I'd also be comfortable just waiting until we cut over to the new bundle format and then making `--bundle` required.

asraa: @znewman01 I think the signature is always output to STDOUT though regardless, so you can always pass it.

I guess it's fair that the certificate is swallowed when you don't specify a output-certificate or bundle even if you use keyless, that seems error-prone to me too.

asraa: (But also the cert is written to STDERR at some point during the command output)

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -98,18 +99,15 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, regOpts options.Re
...
+		ts, err := json.Marshal(rfc3161Timestamp)
 		if err != nil {
 			return nil, err
 		}
-		if err := os.WriteFile(ko.RFC3161TimestampPath, contents, 0600); err != nil {
+		if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
```
haydentherapper, line 107: Not to self, should require that both TSAServerURL and RFC3161TimestampPath is set

cmd/cosign/cli/sign/sign_blob.go
haydentherapper: I'm going to require that you set both TSAServerURL and RFC3161TimestampPath, to always write the timestamp. If we add in writing to Rekor at some point, we can change the requirement.

specs/SIGNATURE_SPEC.md
haydentherapper: Correct, removed.

znewman01 (dismissed):

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -73,16 +74,16 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
+	if options.NOf(c.KeyRef, c.CertRef, c.Sk, c.BundlePath) == 0 {
+		return fmt.Errorf("please provide a cert to verify against via --certificate or a bundle via --bundle")
```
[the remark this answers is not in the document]
haydentherapper: done

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -72,19 +72,36 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
...
+			return nil, fmt.Errorf("rfc3161 timestamp is nil")
+		}
+		ts, err := json.Marshal(rfc3161Timestamp)
+		if err != nil {
+			return nil, err
+		}
+		if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
+			return nil, fmt.Errorf("create rfc3161 timestamp file: %w", err)
+		}
+		fmt.Printf("RFC3161 timestamp bundle written to file %s\n", ko.RFC3161TimestampPath)
```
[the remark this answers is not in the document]
haydentherapper: done

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -72,19 +72,36 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
...
 		if err != nil {
 			return nil, err
 		}
 
-		signedPayload.RFC3161Timestamp = cbundle.TimestampToRFC3161Timestamp(respBytes)
+		rfc3161Timestamp = cbundle.TimestampToRFC3161Timestamp(respBytes)
+		// TODO: Consider uploading RFC3161 TS to Rekor
+
+		if rfc3161Timestamp == nil {
+			return nil, fmt.Errorf("rfc3161 timestamp is nil")
```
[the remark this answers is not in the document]
haydentherapper: This is a just-in-case check, since `TimestampToRFC3161Timestamp` might return nil if `respBytes` is empty for some reason.

znewman01 (dismissed):

znewman01 (approved):

hectorj2f (approved):
lgtm

PR #2500, opened 2022-12-01
cmd/cosign/cli: stop loading the entire blob into memory

### LINKED ISSUE #2097: sign-blob attemps to read the whole file into memory
**Description**

At the moment `sign-blob` logic attempts to read the whole file into memory and then sign it. It can lead to problems with signing large blobs and leads to increased memory requirements on build agents perform signing. 

I suggest refactoring that place and using more "traditional" steps like
- digest calculation which can be done via sequential reading of small potrions of file
- signing the digest

This will allow to avoid reading the whole blob into memory.

AnishShah:
sign-blob always loads the entire blob into memory. This can lead to problems when signing large blob files and increased memory usage. Instead of loading the whole blob, we plumb down `io.Reader` for signing.

#### Summary

Fixes #2097

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -25,6 +25,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash"
 	"os"
 	"strconv"
 	"strings"
@@ -138,8 +139,8 @@ func GetRekorPubs(ctx context.Context, _ *client.Rekor) (map[string]RekorPubKey,
 }
 
 // TLogUpload will upload the signature, public key and payload to the transparency log.
-func TLogUpload(ctx context.Context, rekorClient *client.Rekor, signature, payload []byte, pemBytes []byte) (*models.LogEntryAnon, error) {
-	re := rekorEntry(payload, signature, pemBytes)
+func TLogUpload(ctx context.Context, rekorClient *client.Rekor, signature []byte, sha256CheckSum hash.Hash, pemBytes []byte) (*models.LogEntryAnon, error) {
+	re := rekorEntry(sha256CheckSum, signature, pemBytes)
 	returnVal := models.Hashedrekord{
 		APIVersion: swag.String(re.APIVersion()),
 		Spec:       re.HashedRekordObj,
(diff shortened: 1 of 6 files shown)


cpanato:
@AnishShah thanks for your PR, please sign the DCO

pkg/cosign/tlog.go
```diff
@@ -184,17 +184,13 @@ func doUpload(ctx context.Context, rekorClient *client.Rekor, pe models.Proposed
 	return nil, errors.New("bad response from server")
 }
 
-func rekorEntry(payload, signature, pubKey []byte) hashedrekord_v001.V001Entry {
-	// TODO: Signatures created on a digest using a hash algorithm other than SHA256 will fail
```
znewman01, line 188: You should probably leave this TODO here.

pkg/cosign/tlog.go
```diff
@@ -184,17 +184,13 @@ func doUpload(ctx context.Context, rekorClient *client.Rekor, pe models.Proposed
...
-	// upload right now. Plumb information on the hash algorithm used when signing from the
-	// SignerVerifier to use for the HashedRekordObj.Data.Hash.Algorithm.
-	h := sha256.Sum256(payload)
+func rekorEntry(sha256CheckSum, signature, pubKey []byte) hashedrekord_v001.V001Entry {
```
znewman01, line 187: I wonder if you want to change `sha256CheckSum []byte` to `checksum hash.Hash` just to make it harder to pass in the full payload (and not the hash output) by accident.

znewman01: I'm a littler nervous that you didn't fix *every* `TLogUpload` call, and doing that would help!

pkg/cosign/tlog.go
AnishShah: Done.

pkg/cosign/tlog.go
AnishShah: Done.

znewman01 (dismissed):
This looks great! Thank you for an excellent first contribution!

znewman01 (approved):

AnishShah:
Is this ready to be merged?

PR #2504, opened 2022-12-02
Consolidate certificate expiry logic

mtrmac:
#### Summary
This is a subset of #2482 as requested in https://github.com/sigstore/cosign/pull/2482#issuecomment-1329505333 (a bit larger than that):
- User-observable: verify RFC 3161 timestamps, and Rekor data, _first_, before processing certificates. Conceptually we need to do that because certificate processing requires a timestamp (although that’s not yet what this PR finishes doing). This can affect user-observable error messages, which can now complain about untrusted signed Rekor data instead of untrusted signed certificates.
- Behavior change: If there is a certificate, always verify it against _some_ timestamp (in particular, the current time, if there is neither a RFC 3161 timestamp nor any Rekor data. (This is more than was asked — I included it because the current behavior is such a surprise for callers, and because _always_ verifying actually makes the code simpler. I can split that into some future PR, and add an explicit “are we dealing with Rekor” condition instead, to preserve the current behavior.)
- Code consolidation: `verifyInternal` now has one place (although not yet the _correct_ place) to do certificate expiry checks, instead of that happening in `VerifyRFC3161Timestamp`

See #2482 for previous discussion, and the general rationale of only using data after it is verified (which is not what this PR really does, but it’s a prerequisite).

#### Release Note

`cosign verify-blob --cert-chain … --certificate … --skip-tlog-verify` now requires the leaf certificate to not be expired.


pkg/cosign/verify.go
```diff
@@ -620,20 +620,71 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
...
 	bundleVerified bool, err error) {
+	var acceptableRFC3161Time, acceptableRekorBundleTime *time.Time // Timestamps for the signature we accept, or nil if not applicable.
+
+	if co.TSACerts != nil {
+		acceptableRFC3161Timestamp, err := VerifyRFC3161Timestamp(ctx, sig, co.TSACerts)
+		if err != nil {
+			return false, fmt.Errorf("unable to verify RFC3161 timestamp bundle: %w", err)
+		}
+		if acceptableRFC3161Timestamp != nil {
+			bundleVerified = true
```
haydentherapper, line 631:
```suggestion
```

haydentherapper: This is only used for printing a verification message, and means that the Rekor bundle was verified.

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -452,13 +453,22 @@ func TestVerifyBlob(t *testing.T) {
...
-			name:         "valid signature with expired certificate",
+			name:         "valid signature with expired certificate + Rekor",
 			blob:         blobBytes,
 			signature:    blobSignature,
 			cert:         expiredLeafCert,
 			experimental: false,
 			shouldErr:    true,
 		},
+		{
+			name:           "valid signature with expired certificatem, no Rekor",
```
haydentherapper, line 464:
```suggestion
			name:           "valid signature with expired certificate, no Rekor",
```

pkg/cosign/verify.go
```diff
@@ -650,76 +701,43 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
...
-		if err != nil {
-			return bundleVerified, err
+	if cert != nil {
+		// 2. Check the validity time of the signature.
+		// This is the signature creation time. As a default upper bound, use the current
+		// time.
+		validityTime := time.Now()
+
+		if acceptableRFC3161Time != nil {
+			// Verify the cert against the integrated time.
```
haydentherapper, line 732:
```suggestion
			// Verify the cert against the timestamp time.
```

pkg/cosign/verify.go
```diff
@@ -986,52 +1004,44 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, co *CheckOpts, rekorCl
...
-		ts, err := timestamp.ParseResponse(bundle.SignedRFC3161Timestamp)
-		if err != nil {
-			return false, fmt.Errorf("error parsing response into timestamp: %w", err)
-		}
-		// Verify the cert against the integrated time.
-		// Note that if the caller requires the certificate to be present, it has to ensure that itself.
-		if err := CheckExpiry(cert, ts.Time); err != nil {
-			return false, fmt.Errorf("checking expiry on cert: %w", err)
-		}
+	// FIXME: tsaverification.VerifyTimestampResponse has done this parsing; we shouldn’t parse again.
```
haydentherapper, line 1039: please file feedback in sigstore/timestamp-authority for updating the verifier

haydentherapper (commented):
Thanks so much for splitting this up! This looks great.

pkg/cosign/verify.go
asraa: sounds good, right now the verification header will result in it saying tlog existence offline.
there probably should be a better verification struct output.

pkg/cosign/verify.go
```diff
@@ -620,20 +620,71 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
...
+		} else {
+			// If the --offline flag was specified, fail here. bundleVerified returns false with
+			// no error when there was no bundle provided.
+			// TODO: You can be offline when you are verifying with a public key. This shouldn't
+			// error out, but maybe should be a log message.
+			if co.Offline {
+				return false, fmt.Errorf("offline verification failed")
+			}
+
+			if co.RekorClient != nil {
```
asraa, line 658: moving my comment over here, but I think this might need to fail, since this will implicitly allow tlog verification to be skipped if there is no RekorClient. I think it would be an error if this were nil at this stage

haydentherapper: +1

mtrmac: Combined with https://github.com/sigstore/cosign/pull/2504#discussion_r1038508344 , if this function succeeds, `bundleVerified == !co.SkipTlogVerify`.

I think that’s actually good; the `bundleVerified` return value should be removed entirely. “verify signature” should give a yes/no answer based on a submitted policy, not a menu for the poor external caller to make a choice (which different callers making different choices).

asraa: > Combined with https://github.com/sigstore/cosign/pull/2504#discussion_r1038508344 , if this function succeeds, bundleVerified == !co.SkipTlogVerify.

That's a nice invariant!

pkg/cosign/verify.go
```diff
@@ -620,20 +620,72 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
...
+			// Update with the verified bundle's integrated time.
+			t, err := getBundleIntegratedTime(sig)
+			if err != nil {
+				return false, fmt.Errorf("error getting bundle integrated time: %w", err)
+			}
+			acceptableRekorBundleTime = &t
+		} else {
+			// If the --offline flag was specified, fail here. bundleVerified returns false with
+			// no error when there was no bundle provided.
+			// TODO: You can be offline when you are verifying with a public key. This shouldn't
```
asraa, line 651: @haydentherapper already mentioned this to me, this can be removed at this point (use the --skip flag if you want to skip)

pkg/cosign/verify.go
mtrmac: https://github.com/sigstore/timestamp-authority/issues/169

pkg/cosign/verify.go
haydentherapper: Removed

cmd/cosign/cli/verify/verify.go
```diff
@@ -142,7 +142,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		co.TSACerts = tsaCertPool
 	}
 
-	if keylessVerification(c.KeyRef, c.Sk) {
+	if keylessVerification(c.KeyRef, c.Sk) || !c.SkipTlogVerify {
```
asraa, line 145: I *think* these || statements could be replaced by just !c.SkipTlogVerify. I think you know that and will do in a follow up :)

asraa (approved):

cmd/cosign/cli/verify/verify.go
haydentherapper: Done in https://github.com/sigstore/cosign/pull/2505

haydentherapper:
@cpanato Any idea about why the windows test is failing? Not sure how best to debug this.

haydentherapper:
Anyone have experience with Windows shell? This doesn't parse correctly:

```
Write-Output $pass | .\cosign.exe sign --key $signing_key --output-signature interactive.sig --tlog-upload $false $test_img
.\cosign.exe verify --key $verification_key --signature interactive.sig --insecure-skip-tlog-verify $true $t
```

I originally tried `false` and `true`, but it seems like neither is parsing correctly for CLI options. If you look at the output, Cosign thinks that the image path is false/true rather than `$test_img`.

pkg/cosign/verify.go
```diff
@@ -986,52 +1003,44 @@ func VerifyBundle(ctx context.Context, sig oci.Signature, co *CheckOpts, rekorCl
...
-		if err != nil {
-			return false, fmt.Errorf("error parsing response into timestamp: %w", err)
-		}
-		// Verify the cert against the integrated time.
-		// Note that if the caller requires the certificate to be present, it has to ensure that itself.
-		if err := CheckExpiry(cert, ts.Time); err != nil {
-			return false, fmt.Errorf("checking expiry on cert: %w", err)
-		}
+	// FIXME: tsaverification.VerifyTimestampResponse has done this parsing; we shouldn’t parse again.
+	acceptableTS, err := timestamp.ParseResponse(acceptedTimestamp.SignedRFC3161Timestamp)
```
asraa, line 1039: can you just return this line?

haydentherapper: done

pkg/cosign/verify.go
```diff
@@ -620,20 +620,70 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
...
+				}
+
+				e, err := tlogValidateEntry(ctx, co.RekorClient, sig, pemBytes)
+				if err != nil {
+					return false, err
+				}
+				t := time.Unix(*e.IntegratedTime, 0)
+				acceptableRekorBundleTime = &t
+			} else {
+				return false, fmt.Errorf("rekor client not provided for online verification")
```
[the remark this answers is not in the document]
haydentherapper: Agreed! Less nested code looks better too.

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -1198,8 +1207,8 @@ func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
 			IgnoreSCT:      true,
 		}
 		err = cmd.Exec(context.Background(), blobPath)
-		if err == nil || !strings.Contains(err.Error(), "certificate signed by unknown authority") {
-			t.Fatalf("expected error with invalid root CA, got %v", err)
+		if err == nil || !strings.Contains(err.Error(), "rekor log public key not found for payload") {
```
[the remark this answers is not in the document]
asraa: OH got it -- the test changes the keyless stack / root, and this PR changes the order of verification so that the bundle needed to be verified before the cert chain was verified.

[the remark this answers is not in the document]
asraa: Just to ensure the test still makes sense, can you apply this diff?
```
--- a/cmd/cosign/cli/verify/verify_blob_test.go
+++ b/cmd/cosign/cli/verify/verify_blob_test.go
@@ -1173,7 +1173,7 @@ func TestVerifyBlobCmdWithBundle(t *testing.T) {
 func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
        keyless := newKeylessStack(t)
        // Change the keyless stack.
-       _ = newKeylessStack(t)
+       newKeyless := newKeylessStack(t)
        t.Run("Invalid certificate root when specifying cert via certRef", func(t *testing.T) {
                identity := "hello@foo.com"
                issuer := "issuer"
@@ -1190,8 +1190,8 @@ func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
 
                // Create bundle
                entry := genRekorEntry(t, hashedrekord.KIND, hashedrekord.New().DefaultVersion(), []byte(blob), leafPemCert, sig)
-               b := createBundle(t, sig, leafPemCert, keyless.rekorLogID, leafCert.NotBefore.Unix()+1, entry)
-               b.Bundle.SignedEntryTimestamp = keyless.rekorSignPayload(t, b.Bundle.Payload)
+               b := createBundle(t, sig, leafPemCert, newKeyless.rekorLogID, leafCert.NotBefore.Unix()+1, entry)
+               b.Bundle.SignedEntryTimestamp = newKeyless.rekorSignPayload(t, b.Bundle.Payload)
                bundlePath := writeBundleFile(t, keyless.td, b, "bundle.json")
                blobPath := writeBlobFile(t, keyless.td, blob, "blob.txt")
                certPath := writeBlobFile(t, keyless.td, string(leafPemCert), "cert.pem")
@@ -1207,8 +1207,8 @@ func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
                        IgnoreSCT:      true,
                }
                err = cmd.Exec(context.Background(), blobPath)
-               if err == nil || !strings.Contains(err.Error(), "rekor log public key not found for payload") {
-                       t.Fatalf("expected error with rekor public key mismatch, got %v", err)
+               if err == nil || !strings.Contains(err.Error(), "certificate signed by unknown authority") {
+                       t.Fatalf("expected error with certificate signed by unknown authority, got %v", err)
                }
        })
        t.Run("Invalid certificate root when specifying cert in bundle", func(t *testing.T) {
@@ -1227,8 +1227,8 @@ func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
 
                // Create bundle
                entry := genRekorEntry(t, hashedrekord.KIND, hashedrekord.New().DefaultVersion(), []byte(blob), leafPemCert, sig)
-               b := createBundle(t, sig, leafPemCert, keyless.rekorLogID, leafCert.NotBefore.Unix()+1, entry)
-               b.Bundle.SignedEntryTimestamp = keyless.rekorSignPayload(t, b.Bundle.Payload)
+               b := createBundle(t, sig, leafPemCert, newKeyless.rekorLogID, leafCert.NotBefore.Unix()+1, entry)
+               b.Bundle.SignedEntryTimestamp = newKeyless.rekorSignPayload(t, b.Bundle.Payload)
                bundlePath := writeBundleFile(t, keyless.td, b, "bundle.json")
                blobPath := writeBlobFile(t, keyless.td, blob, "blob.txt")
 
@@ -1243,8 +1243,8 @@ func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
                        IgnoreSCT:      true,
                }
                err = cmd.Exec(context.Background(), blobPath)
-               if err == nil || !strings.Contains(err.Error(), "rekor log public key not found for payload") {
-                       t.Fatalf("expected error with rekor public key mismatch, got %v", err)
+               if err == nil || !strings.Contains(err.Error(), "certificate signed by unknown authority") {
+                       t.Fatalf("expected error with certificate signed by unknown authority, got %v", err)
                }
        })
 }
```

This uses the correct Rekor keyless stack signer, but just changes the certs, so you can revert the error check change.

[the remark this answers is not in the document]
haydentherapper: Done!

pkg/cosign/verify_test.go
```diff
@@ -1377,30 +1379,8 @@ func TestVerifyRFC3161Timestamp(t *testing.T) {
 		base64.StdEncoding.EncodeToString(signature),
 		static.WithCertChain(pemLeaf, appendSlices([][]byte{pemRoot})),
 		static.WithRFC3161Timestamp(&rfc3161TS))
-	err = VerifyRFC3161Timestamp(ociSig, pool)
+	_, err = VerifyRFC3161Timestamp(ociSig, pool)
 	if err == nil || !strings.Contains(err.Error(), "hashed messages don't match") {
 		t.Fatalf("expected error verifying mismatched signatures, got: %v", err)
 	}
-
-	// failure with old certificate
```
asraa, line 1385: FYI asked offline: this is removed because VerifyRFC3161Timestamp doesn't check cert expiration time. it only checks the TS

asraa (approved):

PR #2515, opened 2022-12-06
attest-blob: add functionality for keyless signing

asraa:

I realized that the sign path didn't exist for keyless here, so it didn't make sense to implement https://github.com/sigstore/cosign/pull/2389 before I did this one.

#### Release Note

* feat: adds support for keyless signing for attestations on blob through `attest-blob`

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -214,17 +214,18 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		// signer is a certificate
 		if err != nil {
 			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Was a certificate used? ", err)
+			return nil
+
 		}
 		if len(cert) != 1 {
 			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Expected a single certificate")
+			return nil
 		}
-		if err == nil && len(cert) == 1 {
-			bts := signer
-			if err := os.WriteFile(c.OutputCertificate, bts, 0600); err != nil {
-				return fmt.Errorf("create certificate file: %w", err)
-			}
-			fmt.Fprintln(os.Stderr, "Certificate written to file ", c.OutputCertificate)
+		bts := signer
+		if err := os.WriteFile(c.OutputCertificate, bts, 0600); err != nil {
+			return fmt.Errorf("create certificate file: %w", err)
 		}
+		fmt.Fprintln(os.Stderr, "Certificate written to file ", c.OutputCertificate)
 	}
 
 	return nil
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"crypto"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -126,6 +140,54 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+		entry, err := cosign.TLogUploadInTotoAttestation(ctx, rekorClient, sig, rekorBytes)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(os.Stderr, "tlog entry created with index:", *entry.LogIndex)
+		signedPayload.Bundle = cbundle.EntryToBundle(entry)
+	}
+
+	if c.BundlePath != "" {
+		signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(sig)
```
asraa, line 178: @znewman01 @haydentherapper would like some feedback here: for an intoto attestation, the bundle would contain the "signature" (i.e. the whole dsse envelope). that lines up with the bundle in the protobuf-specs. just FYI. this bundle is fairly large.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -126,6 +140,54 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+	if c.TSAServerURL != "" {
+		clientTSA, err := tsaclient.GetTimestampClient(c.TSAServerURL)
+		if err != nil {
+			return fmt.Errorf("failed to create TSA client: %w", err)
+		}
+		rfc3161Timestamp, err := tsa.GetTimestampedSignature(sig, clientTSA)
+		if err != nil {
+			return err
+		}
+		if err := os.WriteFile(c.RFC3161TimestampPath, rfc3161Timestamp, 0600); err != nil {
```
asraa, line 153: @haydentherapper I'll follow along on your PR for this, but to me it feels like if we do upload to a TSA we should always write the response out.

haydentherapper: Yea, I agree. You need to check that `RFC3161TimestampPath` is set. One case where we could relax this constraint is if we start uploading TSs to Rekor, then you don't need to write them, but that's something for the future. 

I'll make the same change to require both TSAServerUrl and RFC3161TimestampPath.

cmd/cosign/cli/attest/attest_blob.go
haydentherapper: Is the concern that it's too large to store in OCI?

asraa: This is about blobs, but are you asking generally? Nope -- OCI already stored the whole attestation.

cmd/cosign/cli/attest/attest_blob.go
asraa: sg!

asraa:
@znewman01 @priyawadhwa @haydentherapper PTAL! I added preliminary tests. I'll be going back to the verification piece of this too.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -126,6 +142,52 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+			return fmt.Errorf("create rfc3161 timestamp file: %w", err)
+		}
+		fmt.Printf("RF3161 timestamp bundle wrote in the file %s\n", c.RFC3161TimestampPath)
```
znewman01, line 158: STDERR?

Also nit: "RF**C**33161"

Also another nit: "written to file:"

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -142,5 +204,21 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+			return fmt.Errorf("error getting signer: %w", err)
+		}
+		cert, err := cryptoutils.UnmarshalCertificatesFromPEM(signer)
+		// signer is a certificate
+		if err == nil && len(cert) == 1 {
+			bts := signer
+			if err := os.WriteFile(c.OutputCertificate, bts, 0600); err != nil {
+				return fmt.Errorf("create certificate file: %w", err)
+			}
+			fmt.Printf("Certificate wrote in the file %s\n", c.OutputCertificate)
```
znewman01, line 219: STDERR?

cmd/cosign/cli/options/attest_blob.go
```diff
@@ -44,6 +75,27 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
...
+		"skip confirmation prompts for non-destructive operations")
+
+	cmd.Flags().BoolVar(&o.TlogUpload, "tlog-upload", true,
+		"whether or not to upload to the tlog")
+
+	cmd.Flags().StringVar(&o.TSAServerURL, "timestamp-server-url", "",
+		"url to the Timestamp RFC3161 server, default none")
+
+	cmd.Flags().StringVar(&o.RFC3161TimestampPath, "rfc3161-timestamp-bundle", "",
+		"write everything required to verify the blob to a FILE")
```
znewman01, line 99: Nit:

> path to timestamp bundle FILE

znewman01 (approved):
LGTM modulo nits

Feels like we could use a refactor to pull out the "output" flags into one place (both the flags and actually writing out to a file) but happy to have that happen later.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -142,5 +204,21 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
 	}
 
+	if c.OutputCertificate != "" {
+		signer, err := sv.Bytes(ctx)
+		if err != nil {
+			return fmt.Errorf("error getting signer: %w", err)
+		}
+		cert, err := cryptoutils.UnmarshalCertificatesFromPEM(signer)
+		// signer is a certificate
+		if err == nil && len(cert) == 1 {
```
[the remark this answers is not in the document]
asraa: A warn might be good -- this CAN happen if someone passes --output-certificate when using a key (so `signer` is a public key).

asraa:
Sorry for the delay, but rebased and comments addressed.

znewman01 (dismissed):

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -142,5 +205,27 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
 
+	if c.OutputCertificate != "" {
+		signer, err := sv.Bytes(ctx)
+		if err != nil {
+			return fmt.Errorf("error getting signer: %w", err)
+		}
+		cert, err := cryptoutils.UnmarshalCertificatesFromPEM(signer)
+		// signer is a certificate
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Was a certificate used? ", err)
```
znewman01, line 216: Should this return early?  Then you can drop the `err == nil` check below

znewman01 (dismissed):

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -126,6 +143,52 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+	if c.TSAServerURL != "" {
+		clientTSA, err := tsaclient.GetTimestampClient(c.TSAServerURL)
+		if err != nil {
+			return fmt.Errorf("failed to create TSA client: %w", err)
+		}
+		rfc3161Timestamp, err := tsa.GetTimestampedSignature(sig, clientTSA)
+		if err != nil {
+			return err
+		}
+		if err := os.WriteFile(c.RFC3161TimestampPath, rfc3161Timestamp, 0600); err != nil {
```
haydentherapper, line 156: We'll need to transform the response into the expected JSON - https://github.com/sigstore/cosign/blob/631a26fb92347230b423bedfbaac84ef177a6fd0/cmd/cosign/cli/sign/sign_blob.go#L91-L104

That's what Verify expects - https://github.com/sigstore/cosign/blob/631a26fb92347230b423bedfbaac84ef177a6fd0/cmd/cosign/cli/verify/verify_blob.go#LL231C14-L231C14

haydentherapper: I can also be convinced that we don't need this intermediate representation, but then we should also update sign-blob/verify-blob

haydentherapper (commented):
LGTM, just one comment about ondisk format for the ts

cmd/cosign/cli/attest/attest_blob.go
asraa: ah thanks for the catch! i'll make sure it's fixed on the verify-blob-attestation path too.

haydentherapper (approved):

PR #2527, opened 2022-12-08
Fix timestamp verification, add verify-blob tests

haydentherapper:
While adding unit tests for verify-blob with timestamping, I found that we try to verify with the current time if no bundle is provided, even if we have a timestamp. I cleaned up this function to check the timestamp time and bundle time if either are provided, and if neither is provided, check against the current time.

I also added a mock client to clean up tests. I'll move this mock client over to the timestamp-authority repo.

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob_test.go
+++ b/cmd/cosign/cli/verify/verify_blob_test.go
@@ -268,7 +268,6 @@ func TestVerifyBlob(t *testing.T) {
 	}
 	rfc3161Timestamp = &bundle.RFC3161Timestamp{SignedRFC3161Timestamp: tsRespBytes.Bytes()}
 	unexpiredTSPath := writeTimestampFile(t, td, rfc3161Timestamp, "unexpiredrfc3161TS.json")
-	fmt.Println(unexpiredTSPath)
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -239,6 +235,49 @@ func TestVerifyBlob(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	tsRespBytes.Reset()
+	_, err = tsaClient.Timestamp.GetTimestampResponse(&timestamp.GetTimestampResponseParams{}, &tsRespBytes)
+	if err != nil {
+		t.Fatalf("unable to generate a timestamp response: %v", err)
+	}
+	rfc3161Timestamp = &bundle.RFC3161Timestamp{SignedRFC3161Timestamp: tsRespBytes.Bytes()}
+	unexpiredTSPath := writeTimestampFile(t, td, rfc3161Timestamp, "unexpiredrfc3161TS.json")
+	fmt.Println(unexpiredTSPath)
```
cpanato, line 271: Is this intentional?

cpanato (commented):
lgtm, a small nit

dlorenc (approved):

hectorj2f (approved):
Lgtm

PR #2538, opened 2022-12-12
Allow users to pass in a path for the --identity-token flag

priyawadhwa:
Seems cleaner to pass in a file name than to pass in an entire token.

#### Release Note
Allow users to pass in a path for the --identity-token flag


haydentherapper:
Any concerns with this encouraging users to write ID tokens to files and risk token theft?

priyawadhwa:
@haydentherapper I can't think of anything compelling, doesn't seem much worse than passing in the token on the command line.

dlorenc (approved):

PR #2539, opened 2022-12-13
Update timestamp dependency to use updated verifier

haydentherapper:
Verification now properly verifies the full certificate chain in addition to supporting other verification, like nonce (useful for verifying on signing).

Also updated the flag name to use "certificate" rather than "cert" to be in line with other flag names.


haydentherapper:
Bump on review

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -112,14 +111,19 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
 			return fmt.Errorf("error reading certification chain path file: %w", err)
 		}
-		// TODO: Update this logic once https://github.com/sigstore/timestamp-authority/issues/121 gets merged.
-		// This relies on untrusted leaf certificate.
-		tsaCertPool := x509.NewCertPool()
-		ok := tsaCertPool.AppendCertsFromPEM(pemBytes)
-		if !ok {
-			return fmt.Errorf("error parsing response into Timestamp while appending certs from PEM")
+
+		leaves, intermediates, roots, err := tsa.SplitPEMCertificateChain(pemBytes)
```
malancas, line 115: Not necessary to address in this pull request but it looks like this `Exec` function and the `Exec` function in `cmd/cosign/cli/verify/verify.go` have a lot of repeating logic?

malancas (approved):
lgtm

cmd/cosign/cli/options/verify.go
```diff
@@ -29,11 +29,13 @@ func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
...
 
-	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-cert-chain", "",
-		"path to certificate chain PEM file for the Timestamp Authority")
+	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-certificate-chain", "",
+		"path to PEM-encoded certificate chain file for the RFC3161 timestamp authority. Must contain the root CA certificate. "+
+			"Optionally may contain intermediate CA certificates, and may contain the leaf TSA certificate if not present in the timestamp")
 
 	cmd.Flags().BoolVar(&o.SkipTlogVerify, "insecure-skip-tlog-verify", false,
-		"skip tlog verification")
+		"skip transparency log verification. used when an artifact signature has not been uploaded to the transparency log. Artifacts "+
```
hectorj2f, line 37: extra dot here `verification. used`

cmd/cosign/cli/verify/verify_attestation.go
hectorj2f: +1 I'll create a function and call it from the different places.

pkg/cosign/verify.go
```diff
@@ -133,8 +133,13 @@ type CheckOpts struct {
 	// Force offline verification of the signature
 	Offline bool
 
-	// TSACerts are the intermediate CA certs used to verify a time-stamping data.
-	TSACerts *x509.CertPool
+	// Set of flags to verify an RFC3161 timestamp used for trusted timestamping
+	// TSACertificate is the leaf TSA certificate. Optional, if provided in the timestamp
+	TSACertificate *x509.Certificate
```
hectorj2f, line 138: Shouldn't this be an array too ?

hectorj2f: Suggestion: I'd rename it to TSALeaveCertificates, or avoid using the term leaves to refer to the tsa certificate. I can be confusing.

pkg/cosign/verify.go
```diff
@@ -1049,14 +1054,12 @@ func VerifyRFC3161Timestamp(sig oci.Signature, tsaCerts *x509.CertPool) (*timest
...
 
-	err = tsaverification.VerifyTimestampResponse(ts.SignedRFC3161Timestamp, bytes.NewReader(tsBytes), tsaCerts)
-	if err != nil {
-		return nil, fmt.Errorf("unable to verify TimestampResponse: %w", err)
-	}
-	acceptedTimestamp := ts
-
-	// FIXME: tsaverification.VerifyTimestampResponse has done this parsing; we shouldn’t parse again.
-	return timestamp.ParseResponse(acceptedTimestamp.SignedRFC3161Timestamp)
+	return tsaverification.VerifyTimestampResponse(ts.SignedRFC3161Timestamp, bytes.NewReader(tsBytes),
```
hectorj2f, line 1057: Do we expect users to pass the three certificates or it could be alright with a single leave certificate.

cmd/cosign/cli/verify/verify_attestation.go
haydentherapper: Yea, all are looking similar, I think this was brought up in another PR recently. Let's do that as a follow up.

pkg/cosign/verify.go
haydentherapper: Yea, I'll make that update. Like with rekor pub keys, we just need to find the matching TSA certificate to verify against

pkg/cosign/verify.go
haydentherapper: If you only pass a leaf, then we can't verify the chain. You must pass at least a trust root - The TSACertificate is optional since it may come from the timestamp itself, and intermediates are optional depending on your PKI.

hectorj2f: True! We mentioned to verify the chain 👍🏻 .

pkg/cosign/verify.go
haydentherapper: Thinking a bit more, I think it might be cleaner to just accept a single one for now.

There's a couple of issues:
* The expectation is that the TSA certificate is embedded in the timestamp, so it's likely you won't need to specify the TSA certificate
* If you specify a list of TSA certificates, what if none of them are the same as the cert embedded in the timestamp? Does this mean you don't trust what's in the timestamp? I think the only reason you should provide a TSA certificate is if it's missing from the timestamp, and that the root of trust for a TSA is the root certificates.
* There's no easy way to know which is the correct certificate to verify the timestamp without first parsing it, which I'd rather just let the verification library handle parsing after a successful verification. Otherwise, we need to run through all of the possible TSA certificates to try to find a successful verification

I think if we need to change the configuration in the future, it shouldn't be too hard, but I think we should just stick with one cert for now. Does that sound good?

haydentherapper:
@hectorj2f @asraa should be ready for review

hectorj2f (approved):
lgtm

haydentherapper:
I can't don't have merge perms, so feel free to merge whenever. The failing test is failing for all PRs, unrelated to this.

PR #2544, opened 2022-12-13
Change go module name to github.com/sigstore/cosign/v2 for Cosign 2.0

priyawadhwa:

This is required to build a 2.0 version by go modules, the original rc release failed with this error:

```
/workspace/go/src/sigstore/cosign/dist/proxy/linux/go.mod:4: require github.com/sigstore/cosign: version "v2.0.0-rc.0" invalid: should be v0 or v1, not v2
```

#### Release Note
Change go module name to github.com/sigstore/cosgn/v2 for Cosign 2.0


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/sigstore/cosign
+module github.com/sigstore/cosign/v2
 
 go 1.18
```
cpanato, line 3: should we bump this to 1.19?  1.20 is almost out and 1.18 will be deprecated

znewman01 (approved):
Gosh I love how "simple" Go is, good work Golang team

go.mod
znewman01: I think that's smart but let's do it in a different PR

PR #2548, opened 2022-12-14
Add CTLogPubKeys into checkopts so that they can be specified like rekor pubkeys.

### LINKED ISSUE #2537: Need to expose CTLog public key in cosign.CheckOpts
**Description**

Support was added to specify things like RekorPub keys as part of the CheckOpts structure so that we can decouple the TUF / Verify boundaries.

What's missing from CheckOpts however is the ability to specify the CTLog public key:
https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L119

This should be added.

@asraa or am I missing something?

vaikas:

#### Summary
* Fixes #2537 
* Add CTLogPubKeys into cosign.CheckOpts so that they can be specified by callers without using env variables. Refactor the code to decouple fetching the said keys and from using them. 
* Add e2e test for cosign sign-blob/verify-blob

Couple of implementation notes:
* RekorPubKeys and LogIDMetadata are extremely similar but the Rekor ones are restricted to ecdsa keys only. We should consider unifying the types, and maybe just check/restrict the public keys we accept at the Rekor paths, or maybe in the future we allow other keys in Rekor and we could just use LogIDMetadata. Seemed like a bigger change than necessary now, and we can do that switch in a follow up PR if we want to.
* Because of the above, I changed the key from hex => hex2string so that in the future the conversion is easier. This is what RekorPubKeys already does under the covers, so just unified them.


vaikas:
Oops, some tests do not run by default...

pkg/cosign/verify_sct_test.go
```diff
@@ -40,10 +43,18 @@ func TestValidateAndUnpackCertWithSCT(t *testing.T) {
 
 	rootPool := x509.NewCertPool()
 	rootPool.AddCert(chain[1])
+
+	// Grab the CTLog public keys
+	pubKeys, err := ctl.GetCTLogPubs(context.Background())
```
haydentherapper, line 48: There's no keys at this point set in the environment, so I think you only need the call that's already done below. You can just not set them in CheckOpts yet

pkg/cosign/fulcioverifier/ctl/verify.go
```diff
@@ -190,5 +180,60 @@ func VerifyEmbeddedSCT(ctx context.Context, chain []*x509.Certificate) error {
...
-	return VerifySCT(ctx, certPEM, chainPEM, []byte{})
+	return VerifySCT(ctx, certPEM, chainPEM, []byte{}, pubKeys)
+}
+
+func PemToECDSAKey(pemBytes []byte) (*ecdsa.PublicKey, error) {
+	pub, err := cryptoutils.UnmarshalPEMToPublicKey(pemBytes)
+	if err != nil {
+		return nil, err
+	}
+	ecdsaPub, ok := pub.(*ecdsa.PublicKey)
```
haydentherapper, line 191: Does this have to be an ECDSA key?

pkg/cosign/verify_sct_test.go
vaikas: Previous code would have fetched them from the TUF root deep in the bowels, so the error messages were differing (CTFE keys not found vs. what's expected here). Tried to keep the behaviour consistent by not changing that. But, I can change it if you feel strongly?

pkg/cosign/fulcioverifier/ctl/verify.go
vaikas: lol, sure doesn't :) Hang on.

pkg/cosign/verify_sct_test.go
haydentherapper: I'm just wondering what this is testing for. This is fetching the default set of keys from the embedded TUF root, and then we override this set of keys by fetching them again, but from the disk. Given we just want to set up the correct set of keys at this point, I think we can just rely on the call below to initialize the keyset from the disk keys.

vaikas: nm, you're totes right. sorry, too many of these areound :)

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -270,6 +271,11 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 		opts = append(opts, static.WithCertChain(certPEM, chainPEM))
 	}
 
+	co.CTLogPubKeys, err = ctl.GetCTLogPubs(ctx)
```
haydentherapper, line 274: Do you also need to check IgnoreSCT here?

vaikas: I had it there originally, but figured the only harm in fetching the keys here would be in failure cases. I can add a check for it however if you want :)

.github/workflows/kind-verify-attestation.yaml
```diff
@@ -112,6 +112,10 @@ jobs:
         echo -n 'foobar e2e test' > ./predicate-file
         ./cosign attest --predicate ./predicate-file --fulcio-url ${{ env.FULCIO_URL }} --rekor-url ${{ env.REKOR_URL }} --allow-insecure-registry --yes ${{ env.demoimage }} --identity-token ${{ env.OIDC_TOKEN }}
 
+    - name: Sign a blob
+      run: |
+        ./cosign sign-blob README.md --fulcio-url ${{ env.FULCIO_URL }} --rekor-url ${{ env.REKOR_URL }} --output-certificate cert.pem --output-signature sig --yes --identity-token ${{ env.OIDC_TOKEN }}
```
asraa, line 117: thanks! does this belong here instead? https://github.com/sigstore/cosign/blob/48e6404789cc34addfe15d0d6df01c23e529e034/.github/workflows/e2e-with-binary.yml

cmd/cosign/cli/verify/verify.go
```diff
@@ -172,6 +173,9 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 	keyRef := c.KeyRef
 	certRef := c.CertRef
 
+	if !c.IgnoreSCT {
+		co.CTLogPubKeys, err = ctl.GetCTLogPubs(ctx)
```
asraa, line 177: check err here?

cmd/cosign/cli/verify/verify_blob.go
asraa: Either way is fine with me too. Technically you could gate under whether there's a cert present for all of these, but I think the logic gets complicated

pkg/cosign/fulcioverifier/ctl/verify.go
```diff
@@ -27,20 +27,43 @@ import (
...
+	"github.com/sigstore/cosign/v2/pkg/cosign/fulcioverifier/ctutil"
 
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/tuf"
 )
 
-// This is the CT log public key target name
-var ctPublicKeyStr = `ctfe.pub`
+// LogIDMetadata holds information for mapping a key ID hash (log ID) to associated data.
+type LogIDMetadata struct {
```
asraa, line 38: could we make this the same as the rekor key map? maybe the rekor one uses ecdsa keys, but both could use a public key, or even signature.Verifier

pkg/cosign/fulcioverifier/ctl/verify.go
```diff
@@ -27,20 +27,43 @@ import (
...
-	pubKey crypto.PublicKey
-	status tuf.StatusKind
+// AddCTLogPubKey adds a public key with the proper index (constructed) from the
+// Public Key representing the PEM-encoded Rekor key and a status.
+func (t *TrustedCTLogPubKeys) AddCTLogPubKey(pemBytes []byte, status tuf.StatusKind) error {
+	pubKey, err := cryptoutils.UnmarshalPEMToPublicKey(pemBytes)
+	if err != nil {
+		return err
+	}
+	keyID, err := ctutil.GetCTLogID(pubKey)
```
asraa, line 61: I think this might be the same alg as the rekor log ID function too

asraa (commented):
thank you for getting to this!

cmd/cosign/cli/verify/verify_blob.go
haydentherapper: What if you're using a custom TUF root that doesn't specify CT keys (because you're running a private instance for example)? This would err out then I think (unless it always falls back to the embedded keys even if you've configured it to use a specific root? I forget)

pkg/cosign/fulcioverifier/ctl/verify.go
vaikas: Yeah, thought about that :) See the 'implementation notes' in the PR description. It does make the change a bit larger (not sure what 'bit' here means yet :) ).

When you say same, I think it would be wonky to use it here and still call it TrustedRekorPubKeys for users. But, if we change the CheckOpts for RekorPubKeys to use a more generic (whatever name we should?) type that accepts crypto.PublicKey then that's a breaking change.
My preference would be to use PublicKey here instead of Verifier as a user of these methods.

pkg/cosign/fulcioverifier/ctl/verify.go
vaikas: Yes, ideally we'd get rid of all the snowflake things from the Rekor because I think it's clearer they are generic CT related methods and keep all the things that are Rekor specific contained in the Rekor packages.

cmd/cosign/cli/verify/verify.go
vaikas: oops, I had forgot to push the changes when I took out the c.IgnoreSCT (See the comment above about the need for it).

.github/workflows/kind-verify-attestation.yaml
vaikas: Added there as well.

cmd/cosign/cli/verify/verify_blob.go
vaikas: done :)

haydentherapper (approved):

pkg/cosign/fulcioverifier/ctl/verify.go
asraa: >  then that's a breaking change.

fortunately, this hasn't been in a release yet, so we're welcome to break it before 2.0

vaikas: @asraa 
https://github.com/chainguard-dev/mono/pull/7143

PR #2569, opened 2022-12-21
Unify RekorPubKeys and CTLogPubKeys, refactor.

vaikas:

#### Summary

This is a more invasive follow up change as discussed in #2548 

This unifies both Rekor / CTLog public key handling, and in particular the map / data structures that deal with mapping from LogID (in CT Log v1 this is a hash of the Public Key) to share implementations / definitions.

As part of the work, moved most of the code from the pkg/cosign/fulcioverifier => pkg/cosign for couple of reasons:
 * Well it seemed like it makes sense for the verify functions to be next to each other
 * without it, there would have been dependency hell

#### Release Note
* This is a breaking change if you are using cosign golang packages as they have moved around and data definitions have changed. And in particular if you are using the cosign.CheckOpts for Rekor / CTLog public keys.


pkg/cosign/tlog.go
```diff
@@ -446,7 +448,7 @@ func VerifyTLogEntryOffline(e *models.LogEntryAnon, rekorPubKeys *TrustedRekorPu
 	if !ok {
 		return errors.New("rekor log public key not found for payload")
 	}
-	err = VerifySET(payload, []byte(e.Verification.SignedEntryTimestamp), pubKey.PubKey)
+	err = VerifySET(payload, []byte(e.Verification.SignedEntryTimestamp), pubKey.PubKey.(*ecdsa.PublicKey))
```
haydentherapper, line 451: Should we be enforcing that all Rekor keys are ECDSA keys earlier on, maybe in GetRekorPubs? Otherwise, it fails during verifications. 
L

vaikas: Yeah, would be better to have up the call stack, agreed. I don't think it can be in GetRekorPubs cause people can populate their own CheckOpts values for them. let me see if I can find a better place.

vaikas: Ok, added them right after we check if the keys are nil or there are none, so added check for them right after that which seems like a good place for it.

asraa: Wondering if we only need to check the `pubKey` retrieved a couple lines earlier. It probably doesn't matter much, but just wondering if someone's TUF root contains our ECDSA keys plus more Rekor keys

asraa (dismissed):

hectorj2f (dismissed):

hectorj2f (approved):

PR #2580, opened 2022-12-29
Only print warning about uploading to tlog if we're uploading to the tlog

priyawadhwa:
Otherwise this warning can be concerning or confusing for users if they know they've passed in --upload-tlog=false

fixes https://github.com/sigstore/cosign/issues/2581

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -80,6 +80,7 @@ func Attest() *cobra.Command {
 				OIDCProvider:             o.OIDC.Provider,
 				SkipConfirmation:         o.SkipConfirmation,
 				TSAServerURL:             o.TSAServerURL,
+				TlogUpload:               o.TlogUpload,
 			}
 			attestCommand := attest.AttestCommand{
 				KeyOpts:         ko,
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -80,7 +80,6 @@ func Attest() *cobra.Command {
 				OIDCProvider:             o.OIDC.Provider,
 				SkipConfirmation:         o.SkipConfirmation,
 				TSAServerURL:             o.TSAServerURL,
-				TlogUpload:               o.TlogUpload,
 			}
 			attestCommand := attest.AttestCommand{
 				KeyOpts:         ko,
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -22,7 +22,6 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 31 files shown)


haydentherapper (commented):
There is a case where you provide your own key and don’t use Fulcio too. Can we file an issue to print the privacy statement for that case or update it here?

priyawadhwa:
Sure, opened https://github.com/sigstore/cosign/issues/2581

cmd/cosign/cli/attest.go
```diff
@@ -80,6 +80,7 @@ func Attest() *cobra.Command {
 				OIDCProvider:             o.OIDC.Provider,
 				SkipConfirmation:         o.SkipConfirmation,
 				TSAServerURL:             o.TSAServerURL,
+				TlogUpload:               o.TlogUpload,
```
asraa, line 83: It worries me that this flag is duped between KeyOpts and AttestCommand. Instead of adding it to KeyOpts, can we print the skip confirmation message sometime during the flow the TLOG upload is actually happening (instead of being inside a signer)?

A pretty clean place to put it would probably be `ShouldUploadToTlog` -- we already check for image being public there, so makes sense to put the human confirmation. If the person denies the confirmation return false.

It probably addresses Hayden's issue of using your own key and tlog upload all in one go

asraa (changes requested):

cmd/cosign/cli/attest.go
priyawadhwa: SGTM, updated!

asraa (dismissed):
thanks!

cmd/cosign/cli/sign/sign.go
```diff
@@ -67,7 +68,36 @@ const TagReferenceMessage string = `Image reference %s uses a tag, not a digest,
...
 `
 
-func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, tlogUpload bool) bool {
+const (
+	// spacing is intentional to have this indented
+	privacyStatement = `
+        Note that there may be personally identifiable information associated with this signed artifact.
+        This may include the email address associated with the account with which you authenticate.
+        This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later.`
+	privacyStatementConfirmation = "By typing 'y', you attest that you grant (or have permission to grant) and agree to have this information stored permanently in transparency logs."
```
haydentherapper, line 77: The extra space got deleted

priyawadhwa: Yep, with the new UI package confirmation the space was creating an indent that didn't look right to me.

priyawadhwa:
@asraa I fixed a merge conflict, could you please re-approve? Thank you!!

cmd/cosign/cli/sign/sign.go
haydentherapper: SG, I haven't tried out cosign since that was added so if it looks good, great!

haydentherapper (approved):

priyawadhwa:
@haydentherapper thanks! had to fix up some tests, could i get another lgtm?

haydentherapper (approved):

Issue #2586, opened 2023-01-02
feat: custom name for imported key pairs for `import-key-pair` command


ChrisJBurns:
**Description**
Currently when importing key pairs using `import-key-pair` command, it currently hardcodes the outputted key pair names to [`import-key.key` & `import-key.pub`](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/importkeypair/import_key_pair.go#L54-L62). We want to offer the user the ability to specify a custom name if they want - otherwise, default to `import-key`.

This to be seen as the same nature of change as implemented in https://github.com/sigstore/cosign/pull/2561 for `generate-key-pair`.

PR #2561, opened 2022-12-20
feat: allows custom name for keypair generation

ChrisJBurns:
Closes #1076

#### Summary
Adds `--name` option for `cosign generate-key-pair` command that allows to specify a custom name for the key pair that's generated. Also added some updates to the `README.md` that details how to run `lint` and `test` for contributors.

#### Release Note
- adds `--name` option for `cosign generate-key-pair` command that allows to specify a custom name for the key pair that's generated.
- 
Signed-off-by: ChrisJBurns <29541485+ChrisJBurns@users.noreply.github.com>

PR #2587, opened 2023-01-02
feat: allows custom key names for import-key-pair

ChrisJBurns:
- the outputted private and public keys that where created when importing key pairs with `--import-key-pair` command where hardcoded to `imported-cosign`. this commit allows for a custom name to be specified using the `--output-key-prefix` or `-o` flag to `--import-key-pair`. example: `--output-key-prefix my-test` leads to `my-test.key` and `my-test.pub` files being outputted. if no `--output-key-prefix` flag is specified, the default `imported-cosign`, so backwards compatibility allowed.
- also changes the `--name` flag in `generate-key-pair` command to `--output-key-prefix` to be more consistent with `import-key-pair`

Closes https://github.com/sigstore/cosign/issues/2586

Signed-off-by: ChrisJBurns <29541485+ChrisJBurns@users.noreply.github.com>


ChrisJBurns:
Not entirely sure what the `e2e-tests/e2e-kms` are failing? Not sure if it's a change I've made or if it sometimes trips over itself?

Error:
```log
cut: write error: Broken pipe
Error: Process completed with exit code 1.
```

znewman01:
> Not entirely sure what the `e2e-tests/e2e-kms` are failing? Not sure if it's a change I've made or if it sometimes trips over itself?
> 
> Error:
> 
> ```
> cut: write error: Broken pipe
> Error: Process completed with exit code 1.
> ```

Not your fault, don't worry

znewman01 (approved):

znewman01 (changes requested):
Is there a reason this doesn't match the `--name` from https://github.com/sigstore/cosign/pull/2561 ? I think consistency would be nice here

ChrisJBurns:
> 

Mainly because thought that it may confuse users with the `--key` flag, `--out` describes that it's the output rather than the input that `--name` may suggest. Although, that's purely the way my mind worked when I thought of it, if it's easily understandable to the average user without confusing them, then am happy to change. I guess we can always make the description clear about the differences between `--key` and `--name`?

With a possible `doc` change also? Just to demonstrate the use of both flags? Or do you think the website documentation should just be the `--key` field?

znewman01:
I'd strongly prefer that they match.

Maybe `--key-prefix` across the board would be good? Feel free to modify `generate-key-pair` in this PR too

ChrisJBurns:
Just to ensure I've got the correct understanding, are you saying to change the `--name` in `generate-key-pair` and the `--out` in `import-key-pair` to `--key-prefix`? Leaving the `--key` flag in `import-key-pair` as is?

znewman01:
Thanks for checking!

I'd like `cosign import-key-pair --key-prefix` **and** `cosign generate-key-pair --key-prefix` (so change `generate-key-pair`, and update this PR for `import-key-pair`)

ChrisJBurns:
So:
`cosign generate-key-pair --key-prefix new-keys`
`cosign import-key-pair --key key-to-import.pem --key-prefix new-keys` ?

Apologies, my brains a bit slow this morning 😆

znewman01:
No, my brain's a bit slow. I forgot that `import-key-pair` had a `--key` flag already.

I see the conundrum...let me think

znewman01:
How about `--output-key-prefix`? It's wordy, but it's not a flag you'll use most of the time. So we have:

- `cosign generate-key-pair --output-key-prefix my-cosign` -> `my-cosign.pub`, `my-cosign.key`
- `cosign import-key-pair --key to-import.pem --output-key-prefix my-cosign` -> `my-cosign.pub`, `my-cosign.key`

WDYT? I'm open to other ideas as well, though I'd *really* like to keep the flags identical if possible.

ChrisJBurns:
@znewman01  Yep, I agree with the consistency part. I do wonder if we make `--key` in `cosign import-key-pair --key to-import.pem` to `--key-to-import` ? A bit more wordy but removes a user thinking "What key?" (although they should read the docs, but that assumes devs read docs 😂 ). The only issue with that is the fact that it's a breaking change to the exist `import-key-pair` command as we're changing an existing flag. I'm happy either way though to be honest, I like the `--output-key-prefix` on both import and generate :+1:

znewman01 (approved):

PR #2591, opened 2023-01-04
chore: Remove COSIGN_EXPERIMENTAL uses from few lingering places.

vaikas:

#### Summary
Remove some lingering uses of COSIGN_EXPERIMENTAL. Drop the warning about public-good-instance not being GA since it is.

### DIFF
--- a/KEYLESS.md
+++ b/KEYLESS.md
@@ -111,7 +111,6 @@ Signature timestamps are checked in the [rekor](https://github.com/sigstore/reko
 
 ## Upcoming work
 
-* Root CA hardening: We should use intermediate certs rather than the root, and support chained verification.
 * Other timestamps: We should allow for other timestamp attestations, including attached [RFC3161](https://www.ietf.org/rfc/rfc3161.txt) signatures.
 * Probably a lot more: This is very experimental.
 * More OIDC providers: Obvious.
--- a/KEYLESS.md
+++ b/KEYLESS.md
@@ -5,8 +5,8 @@ The full design document for this can be found [here](https://docs.google.com/do
 This document explains how the `keyless` signatures work in `cosign`.
 Try it out!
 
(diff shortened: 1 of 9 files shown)


vaikas:
e2e-kms failed (looks like a flake):
```
INFO: Hashicorp Vault version '1.12.2' requested
INFO: Downloading platform-specific version '1.12.2' of vault...
      https://releases.hashicorp.com/vault/1.12.2/vault_1.12.2_linux_amd64.zip
vault_1.12.2_linux_amd64.zip: OK
cut: write error: Broken pipe
Error: Process completed with exit code 1.
```

KEYLESS.md
```diff
@@ -114,14 +112,12 @@ Signature timestamps are checked in the [rekor](https://github.com/sigstore/reko
 ## Upcoming work
 
 * Root CA hardening: We should use intermediate certs rather than the root, and support chained verification.
-* Root CA configuration: We should allow users to change the roots and add their own.
```
asraa, line 117: I think the bullet point above is also fixed? @haydentherapper now that we use Fulcio intermediates for the public good instance & support cert chains (I think this is what this meant)

README.md
```diff
@@ -163,9 +163,6 @@ The following feature set is not considered stable yet, but we are committed to
 
 #### Anything under the `COSIGN_EXPERIMENTAL` environment variable
 
-* Integration with the `Rekor` transparency log
```
asraa, line 166: should we just remove the COSIGN_EXPERIMENTAL var reference? I think it's not used at all

cpanato (approved):

README.md
vaikas: I'm fine with removing it, but my thinking was that we may want to keep it in case we release other features as experimental in the future, so precedent has been set.

znewman01:
I'm a little concerned about merging this before the 2.0 release (not RC) is out. If folks are looking here for docs, they'll be mislead if they're installing using homebrew or similar

PR #2606, opened 2023-01-09
    cli: Handle `-` as a CLI argument

znewman01:
#### Summary

Before:

```
$ cat /dev/null | COSIGN_EXPERIMENTAL=1 ./cosign attest caphill4/test:latest --type custom --predicate -
WARNING: the - flag is deprecated and will be removed in a future release. Please use the -- flag instead.
```

After there's no warning, which is good. `-output` or similar continue to warn.

#### Release Note

NONE

#### Documentation

N/A this makes behavior consistent with docs


spiffcs (approved):
Confirmed that this fixes the issue reported here. Thanks @znewman01 for the quick fix!
https://sigstore.slack.com/archives/C01PZKDL4DP/p1673282474743979

dlorenc (approved):

Issue #2510, opened 2022-12-05
Inappropriate printing to STDOUT


znewman01:
Generally, in Cosign we try to make STDOUT have the *output* of the command: the (often machine-readable) result of execution. We put status messages, errors, etc. on STDERR.

This is a common convention for CLI tools and a nice thing to do for scriptability.

However, we're pretty sloppy sometimes.

---------------------------------------------------------------------
 
I'm looking mostly at:

- https://github.com/sigstore/cosign/blob/fa8a799784cafb0779849e69cb939c8dc6132a5e/pkg/cosign/tlog.go#L170
-  https://github.com/sigstore/cosign/blob/381ba64a577505d378b49482ee34d39eeab6ea50/cmd/cosign/cli/pivcli/commands.go#L152
- https://github.com/sigstore/cosign/blob/381ba64a577505d378b49482ee34d39eeab6ea50/cmd/cosign/cli/initialize/init.go#L52
- https://github.com/sigstore/cosign/blob/d720f040bf493861c1ea8280f462c14e705990e0/cmd/cosign/cli/sign/sign.go#L250

But even some of the ones that are actually correct just work by spitting out to STDOUT in the middle of command execution:
 - https://github.com/sigstore/cosign/blob/381ba64a577505d378b49482ee34d39eeab6ea50/cmd/cosign/cli/pivcli/commands.go#L265
- https://github.com/sigstore/cosign/blob/d720f040bf493861c1ea8280f462c14e705990e0/cmd/cosign/cli/tree.go#L152

-------------------------------------------------------------------

My proposal is the following:

1. Have the CLI top-level entry point *print* the output of the command. That is, something like [`SaveCmd`](https://github.com/sigstore/cosign/blob/30bf1c09c6fde849245648ff294725d86094fe28/cmd/cosign/cli/save.go#L50) would be implemented as a function with a return value, and then we'd hook into Cobra to print the output (if we need to stream output, we could use channels or something).
1. Ban (using [Forbidigo](https://github.com/sigstore/cosign/blob/381ba64a577505d378b49482ee34d39eeab6ea50/.golangci.yml#L38)) any calls to `fmt.Print{f,ln}` in the Cosign codebase.
1. Implement a "Cosign output" package. Instead of `fmt.Fprintln(os.Stderr, "warning: foo")` directly, Cosign code would call `Warn()`. Same for any other common output patterns. (That is, calling code can assign *semantics* to its output which helps make sure it goes to the right place).
[...]

PR #2607, opened 2023-01-09
refactor: use internal.ui lib instead of printf directly

znewman01:
In some of these cases, we're moving output from STDOUT to STDERR but it always should have been on STDERR.

Signed-off-by: Zachary Newman <zjn@chainguard.dev>


znewman01:
e2e test failure looks unrelated.

Addresses #2510

hectorj2f (approved):

PR #2617, opened 2023-01-10
Add schemes to predicate URIs.

### LINKED ISSUE #2612: Predicate URIs aren't RFC 3986 compliant
**Description**

[In-toto spec says](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#TypeURI) that predicate URIs should be [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3) compliant, which means that we should include a scheme (i.e. `https://` in the URIs we're producing.

https://github.com/sigstore/cosign/blob/04fb9c0a5bdebf7580d772dfd3e439ae66ac2b8a/pkg/cosign/attestation/attestation.go#L31-L37

The change to add the scheme seems simple enough, but could be construed as a breaking change if clients are switching on these values, so not sure if we need to take extra steps to change this. Thoughts? 👀 

**Version**

[04fb9c0](https://github.com/sigstore/cosign/commit/04fb9c0a5bdebf7580d772dfd3e439ae66ac2b8a) (latest main)

wlynch:
#### Summary

According to
https://github.com/in-toto/attestation/blob/main/spec/field_types.md#TypeURI this should be RFC 3986 compliant.

Fixes #2612

#### Release Note

**BREAKING CHANGE**: Predicate types now include https:// scheme to be compliant with In-toto spec.


haydentherapper:
Does this break verification of older attestations?

wlynch:
Closing in favor of #2717

PR #2622, opened 2023-01-12
Fix missing privacy statement, print in multiple locations

haydentherapper:
The statement wasn't being printed, only the confirmation, so updated that and its spacing.

Additionally, I added the statement to be printed before using Fulcio in addition to Rekor. It'll only be printed once since we use sync.Once. The motivation for this is that Fulcio includes the certificate in its own CT log, so users should be prompted before. Otherwise, by the time they would say no to the Rekor prompt, the email would already be publicly logged.

Updates to #2580


asraa (approved):
AH gotcha, same privacy statement for Fulcio and for rekor upload.

znewman01 (approved):

PR #2631, opened 2023-01-16
Support keyless verification without Fulcio roots

### LINKED ISSUE #2630: Feature: Support "keyless" verification non-Fulcio certificate authorities
**Description**

Currently we have some support for signing with x509 certificates using haven't come from a Fulcio CA and and don't chain up to a TUF root. For example

```
cosign sign --certificate cert.pem --key key.pem --certificate-chain bundle.pem example.com/foo@sha256:.....
```

> Aside: the key format needs to be cosign's format to make this work so you often need to run e.g `cosign import-key-pair` to make that the case

The current invocation to verify this kind of signature is a little inconvenient as you need to specify the leaf certificate like so:

```
cosign verify --certificate cert.pem --certificate-chain bundle.pem example.com/foo@sha256:....
```

This means grabbing that leaf certificate off the OCI manifest or from the Rekor bundle or something.  Would be great to be able to do the same kind of invocation as Fulcio's verification like

```
cosign verify --certificate-chain bundle.pem --certificate-identity foo@example.com --certificate-oidc-issuer https://id.example.com example.com/foo@sha256:.....
```

This convenient keyless verification would be awesome for folks using a custom CA, provided they're using certificates in a similar shape that Fulcio does (Efforts to standardize these extensions like https://github.com/sigstore/fulcio/pull/945 would make it easy for folks to create a CA that does this)

nsmith5:
Adds support for verifying keyless signatures from a specificied CA bundle instead of assuming Fulcio and loading its roots.

Fixes #2630

#### Summary

This changes allows folks to more easily verify signatures from a custom CA (see #2630 for additional discussion)

#### Release Note

* Allow verifying signatures from a custom CA without needing leaf certificate

#### Documentation

Will attach an PR for docs here if folks agree on the parent issue that its worth doing

#### TODO

The TODO list as this idea is still in discussion etc.
- [ ] Add PR for docs with examples of this kind of verification
- [ ] Add some end to end tests of this (where are these now?)

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -168,15 +168,30 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		}
 	}
 	if keylessVerification(c.KeyRef, c.Sk) {
-		// This performs an online fetch of the Fulcio roots. This is needed
-		// for verifying keyless certificates (both online and offline).
-		co.RootCerts, err = fulcio.GetRoots()
-		if err != nil {
-			return fmt.Errorf("getting Fulcio roots: %w", err)
-		}
-		co.IntermediateCerts, err = fulcio.GetIntermediates()
-		if err != nil {
-			return fmt.Errorf("getting Fulcio intermediates: %w", err)
+		if c.CertChain != "" {
+			chain, err := loadCertChainFromFileOrURL(c.CertChain)
+			if err != nil {
+				return err
+			}
+			co.RootCerts = x509.NewCertPool()
+			co.RootCerts.AddCert(chain[len(chain)-1])
+			if len(chain) > 1 {
+				co.IntermediateCerts = x509.NewCertPool()
+				for _, cert := range chain[:len(chain)-1] {
+					co.IntermediateCerts.AddCert(cert)
+				}
+			}
+		} else {
+			// This performs an online fetch of the Fulcio roots. This is needed
+			// for verifying keyless certificates (both online and offline).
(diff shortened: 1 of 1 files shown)


haydentherapper (commented):
LGTM. Do we also need to update the other verify- commands?

haydentherapper:
For tests, verify_blob tests are really thorough and should be pretty easy to add this case. You could try to add them in e2e_tests too, but it’s not always clear what’s being tested.

nsmith5:
Yeah limme take a look at the other verify commands and see if they work. If this sound reasonable I'll add e2e tests and some example docs and open this up for review :)

dmitris:
could we reopen this? (Or would need to open a new PR?).  Currently we are using an internal copy of `cosign` with the code of this PR, would be great to get it merged 😄  Let me know if I can help to move this forward (do some testing etc.)

haydentherapper:
@dmitris Ive reopened it, would you like to work on it? The remaining changes are adding support for all verify commands and testing

dmitris:
> @dmitris Ive reopened it, would you like to work on it? The remaining changes are adding support for all verify commands and testing

thanks @haydentherapper I will take a look.  Since this PR is off @nsmith5's branch, I'd need to create a new PR to make additional changes, right?

haydentherapper:
I believe so

Note you should be able to work around the lack of this feature in a number of ways. You can either set up a TUF repository to distribute the chain, or provide both the leaf and the chain.

haydentherapper:
Superseded by https://github.com/sigstore/cosign/pull/2845

Issue #2632, opened 2023-01-16
bug: x509 verification broken


nsmith5:
**Description**

The [documented](https://docs.sigstore.dev/cosign/verify/#local-verifications) x509 certificate verification isn't working as expected. This is broken in two different ways at HEAD (29360f6a3390d44dd8faef636dd0c3449a213c88) and v2.0.0-rc0

HEAD
```
$cosign verify --certificate-chain bundle.pem --certificate cert.pem  $(cat image)
Error: --certificate-identity or --certificate-identity-regexp is required for verification in keyless mode
main.go:63: error during command execution: --certificate-identity or --certificate-identity-regexp is required for verification in keyless mode
```

v2.0.0-rc0
```
$ cosign verify --certificate-chain bundle.pem --certificate cert.pem  $(cat image) --insecure-ignore-sct                                     
Error: no matching signatures:                                                                                                                                               
error verifying bundle: comparing public key PEMs, expected -----BEGIN PUBLIC KEY-----                                                                                       
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEgC9YnGfLG1oNUg7qbVoI9RaCYNmU                                                                                                             
SC8QYw9JCIEdkn+ySEfwRPMVwd8ljiSljFSvw9TWuqxj5WvOMU43wmO9jQ==                                                                                                                 
-----END PUBLIC KEY-----                                                                                                                                                     
, got -----BEGIN CERTIFICATE-----                                                                                                                                            
MIICEzCCAbqgAwIBAgIRAKHhIO8ezAumM0UvolewscEwCgYIKoZIzj0EAwIwJDEi                                                                                                             
MCAGA1UEAxMZTm90RnVsY2lvIEludGVybWVkaWF0ZSBDQTAeFw0yMzAxMTUyMjUz                                                                                                             
NThaFw0yMzAxMTUyMjU5NThaMAwxCjAIBgNVBAMTATEwWTATBgcqhkjOPQIBBggq                                                                                                             
hkjOPQMBBwNCAAQDOAZjN5VZ7wARVO7hoYvf4Ra/UROo/Img1bPIOk5jF1ha+sEp                                                                                                             
duoZ3pKuw7Xv3QCqWWPNNYKr4X5OJoYAbubBo4HkMIHhMA4GA1UdDwEB/wQEAwIH                                                                                                             
[...]

PR #2633, opened 2023-01-16
Fix verifying when a certificate is provided

haydentherapper:
Keys were preferred over certificates. This caused an issue, because in the CLI functions that call verify, if you pass a certificate to verify with, we extract the public key from it and set that to co.SigVerifier while also setting sig.Cert. Checking if a certificate was provided first should mitigate this.

Fixes: https://github.com/sigstore/cosign/issues/2632

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -725,15 +725,20 @@ func keyBytes(sig oci.Signature, co *CheckOpts) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	// We have a public key.
-	if co.SigVerifier != nil {
+	switch {
+	// certificate provided from the OCI image, or via flag (--certificate or --bundle)
+	case cert != nil:
+		return cryptoutils.MarshalCertificateToPEM(cert)
+	// no certificate provided, verifier must be a public key
+	case co.SigVerifier != nil:
 		pub, err := co.SigVerifier.PublicKey(co.PKOpts...)
 		if err != nil {
 			return nil, err
 		}
 		return cryptoutils.MarshalPublicKeyToPEM(pub)
+	default:
+		return nil, errors.New("no verifier found, must provide either certificate or public key")
 	}
-	return cryptoutils.MarshalCertificateToPEM(cert)
 }
 
 // VerifyBlobSignature verifies a blob signature.


haydentherapper:
@asraa, can you check this is correct, since you did the refactor for this?

cc @nsmith5

haydentherapper:
Looking into unit test failures now

haydentherapper:
The failure is when you provide both a key (which you want to verify with) and a bundle with a certificate. I think the fix is just to make this an error when taking in input. We don't allow you to specify more than one of (cert, security key, public key), so we shouldn't allow you to specify a bundle with a verifier (pub key or cert) if you also specified the verifier. (Note we would still allow a bundle + verifier if the bundle only contains the Rekor entry)

Will follow up on Tuesday.

pkg/cosign/verify.go
```diff
@@ -725,15 +725,20 @@ func keyBytes(sig oci.Signature, co *CheckOpts) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	// We have a public key.
-	if co.SigVerifier != nil {
+	switch {
+	// certificate provided from the OCI image, or via flag (--certificate or --bundle)
```
nsmith5, line 729: I'm not 100% sure I understand the logic here, but I'm worried that this bit might have a subtle verification bug. Let me know if this makes sense:

If the user has specified `--certificate` at the command line my understanding is that it get parsed and its public key is used in initialize `co.SigVerifier`. This function appears to then return a certificate PEM if found on the OCI signature (which may be different than `--certificate` 😬 ). The function calling this one https://github.com/sigstore/cosign/blob/29360f6a3390d44dd8faef636dd0c3449a213c88/pkg/cosign/verify.go#L1078-L1081 then appears to compare the returned certificate to the one it finds in the rekor bundle.

My worry here is that the certificate compared to the rekor bundle might not be the one the user actually specified at the command line, unless that bit of logic is lurking in a different place?

haydentherapper: I think we have this issue now also, although it doesn’t surface any error. If you provide a certificate that certified the same public key as a certificate on the OCI imagine, but the rest of the cert differs, we wouldn’t notice. 

Note this isn’t an issue for verify-blob since we construct a static OCI struct from input. For verify on OCI, theres no way to pass the certificate passed in from input  

Sigh…maybe we could add a check that a provided cert matches what’s on the image annotations? Or disallow providing a certificate when the image annotations contain a certificate already?

haydentherapper:
Oof my brain

I was confused why we hadn't caught this for `verify-blob`. Turns out, it's because we don't set `co.SigVerifier` in verify-blob:

https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L193-L197

Contrast this to verify:

https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify.go#L214-L243

Not sure if this was intentional or accidental. I think it wasn't intentional because if a bundle is provided with a certificate in the bundle, we do set the verifier: https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L220

I think we should pick one place to set the verifier. Either require a verifier always provided (and figure out how to pass a certificate value into `VerifyImageSignature`, or always leverage https://github.com/sigstore/cosign/blob/29360f6a3390d44dd8faef636dd0c3449a213c88/pkg/cosign/verify.go#L638-L670 to set the verifier.

pkg/cosign/verify.go
asraa, line 729: is this comment accurate / misleading? If it's OCI it will always take it in from the OCI image, NOT from the `--certificate` flag. I think @nsmith5 issue shows a "Red herring" like you mentioned. I think the issue should have happened if the `--certificate` was specified or not since it's comparing `sig.Cert()` form the OCI annotation. The OCI verify CLI never sets that. So maybe

```	// certificate provided from the OCI image, or, in the case of a blob, via flag (--certificate or --bundle)```

asraa: > Sigh…maybe we could add a check that a provided cert matches what’s on the image annotations? Or disallow providing a certificate when the image annotations contain a certificate already?

+1, see my comment below. I think `--certificate` is a little misleading here. The `--certificate` flag for OCI image never "makes" it into the pkg verification code (just the pub key does). In the CLI layer there's some verification of it, but once we get into `pkg/verify`, we rely on `sig.Cert()`.

asraa (commented):
Thanks! Yes this should have been the correct behavior -- chose cert, or choose public key PEM for the verifier. 

One comment -- I think @nsmith5 identified the problem about the comparison in `pkg/cosign/verify.go` linked in the comment. The more weird problem is in my comment, but `--certificate` is never actually used besides public key extraction and verification BEFORE hitting the pkg. I think we either want to set the `sig.Cert()` with this / override it, or disallow this flag when a `sig.Cert()` is already specified on the OCI layer. Flag confusion..

haydentherapper:
After trying a few different approaches, the cleanest is to override sig.Cert() (if not set). We don't have any issues in verify-blob because we are setting sig.Cert() based on `--certificate`. We should be doing the same for the OCI case, just with an additional check if there's any difference. Then the logic in `verifyInternal` doesn't have to change. Adding a field to `CheckOpts` like Nathan suggested for passing the certificate. (This struct is TOO BIG)

haydentherapper:
OK, so after repeatedly typing lines and deleting them, here's what I think I'll do:

* Add `VerifierCert` to `CheckOpts` (`co`)
* We need to predictably set `co.SigVerifier`. I'm going to remove setting SigVerifier when a cert is passed, so that `SigVerifier == nil` means verify with a certificate. Effectively now `co.SigVerifier` is just a provided public key. (maybe even rename it to PublicKey)
* We will pull the cert from either `sig.Cert` or `co.VerifierCert`, and pull chain from either `sig.Chain` or `co.Roots`/`co.Intermediates` (which will have been initialized with either a provided chain or the Fulcio/TUF certs). It will prefer `co` over `sig` for chain (current behavior), and check for equality for cert.
* We will make the change here to read either SigVerifier or the parsed cert (What if Rekor has the full chain uploaded? Currently Cosign uploads only the cert, not chain, so I think this is out of scope)

@asraa @nsmith5 Lemme know if you see any issues

nsmith5:
Yup, this sounds like the right approach to me!

PR #2636, opened 2023-01-16
feature: support sign-blob with certificate

### LINKED ISSUE #2635: feature: `cosign sign-blob` should accept `--certificate` and `--certificate-chain`
**Description**

While `cosign verify-blob` will accept a certificate and CA chain to verify against, it appears the signing command won't accept them. Feels reasonable to make these two commands symmetric.

nsmith5:
#### Summary

The `verify-blob` and every other signing command all allow folks to specify a custom x509 cert and chain to sign with. This work adds support for specifying the cert and CA chain for the `sign-blob` command as it was missing.

As specifying a certificate and cert chain is supported in all signing and verifying commands this work also refactors our `options.KeyOpt` struct to move the certificate details there. This makes the passing of these details to initialize a signer the same for all commands.

Fixes #2635

#### Release Note

* Adds support for specifying a custom certificate and certificate chain to `sign-blob` with the `--certificate` and `--certificate-chain` flags

#### Documentation

Added docs in `/docs` and in CLI options. Can follow up with a PR to docs.sigstore.dev if folks feel we need an example over there.

### DIFF
--- a/cmd/cosign/cli/signblob.go
+++ b/cmd/cosign/cli/signblob.go
@@ -41,6 +41,9 @@ func SignBlob() *cobra.Command {
   # sign a blob with a local key pair file
   cosign sign-blob --key cosign.key <FILE>
 
+  # sign a blob with local certificate and CA bundle
+  cosign sign-blob --key cosign.key --certificate <FILE> --certificate-chain <FILE> --bundle <FILE> <FILE>
+
   # sign a blob with a key pair stored in Azure Key Vault
   cosign sign-blob --key azurekms://[VAULT_NAME][VAULT_URI]/[KEY] <FILE>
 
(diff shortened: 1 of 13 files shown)


nsmith5:
@asraa and @haydentherapper I _think_ the small refactor here to add the Cert and CertChain to `KeyOpts` might be the first step towards that `--certificate` issue we identified in https://github.com/sigstore/cosign/pull/2633. I think if we just go a step further and plumb that information into `cosign.CheckOpts` (https://github.com/sigstore/cosign/blob/e69e19027327c45057a0d3636d65f0d4c1374638/pkg/cosign/verify.go#L73) we might be able to distinguish these different cases of public key versus certificate verification

cmd/cosign/cli/signblob.go
```diff
@@ -41,6 +41,9 @@ func SignBlob() *cobra.Command {
   # sign a blob with a local key pair file
   cosign sign-blob --key cosign.key <FILE>
 
+  # sign a blob with local certificate and CA bundle
+  cosign sign-blob --key cosign.key --certificate <FILE> --certificate-chain <FILE> <FILE>
```
haydentherapper, line 45: Should this example show outputting the bundle too, since that is the main use case for adding these?

haydentherapper (commented):
LGTM, I'll let @asraa approve since she's mostly worked in this part of the codebase

cmd/cosign/cli/signblob.go
nsmith5: Ah yes good point!

PR #2652, opened 2023-01-25
Make sure a cert passed in via --cert matches the bundle cert

priyawadhwa:
This is part of https://github.com/sigstore/cosign/issues/2632, and is one of the smaller action items listed in https://github.com/sigstore/cosign/issues/2632#issuecomment-1400977569.

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
@@ -214,7 +215,7 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 			if isb64(certBytes) {
 				certBytes, _ = base64.StdEncoding.DecodeString(b.Cert)
 			}
-			cert, err = loadCertFromPEM(certBytes)
+			bundleCert, err := loadCertFromPEM(certBytes)
 			if err != nil {
 				// check if cert is actually a public key
 				co.SigVerifier, err = sigs.LoadPublicKeyRaw(certBytes, crypto.SHA256)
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -214,14 +215,21 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
 			if err != nil {
 				// check if cert is actually a public key
 				co.SigVerifier, err = sigs.LoadPublicKeyRaw(certBytes, crypto.SHA256)
 				if err != nil {
 					return fmt.Errorf("loading verifier from bundle: %w", err)
 				}
 			}
+			// if a cert was passed in, make sure it matches the cert in the bundle
+			if cert != nil {
+				if !reflect.DeepEqual(cert, bundleCert) {
```
haydentherapper, line 228: You can also use cert.Equal()

priyawadhwa: thanks, updated

haydentherapper (approved):

asraa (approved):

znewman01 (approved):

znewman01:
> This doesn't handle the OCI image case, right? Since in that case the bundle is on-disk and only gets read during the internal verification call.

That's a good catch. I think long-run, we're going to want:

1. An inner `verify` method that doesn't know anything about the bundle
2. An outer `verifyBundle` method that takes in certificate, bundle, etc. and unpacks them into a `verify` call

Then, all of `verify`, `verify-blob`, `verify-blob-attestation`, etc. would go through `verifyBundle`; this check would happen in (2).

As it stands, I haven't been able to figure out whether this actually happens in `verify` (and if so, where). Given that users probably shouldn't be doing this anyway, I'm comfortable with just fixing it here.

asraa:
> As it stands, I haven't been able to figure out whether this actually happens in verify (and if so, where). Given that users probably shouldn't be doing this anyway, I'm comfortable with just fixing it here.

I think this was the source of the original issue though, right? From https://github.com/sigstore/cosign/issues/2632 is an OCI image.

znewman01:
I think https://github.com/sigstore/cosign/pull/2633 is more important for that, this was just a side issue and I think it's more obviously a problem for `verify-blob`

PR #2653, opened 2023-01-25
Add flags for timestamp verification on signing

haydentherapper:


haydentherapper:
@priyawadhwa Here's the flag changes in all relevant commands

haydentherapper:
The plan was to add a NewSigner that supports verification, like what we have for SCTs (so you can either verify or not verify while signing)

cmd/cosign/cli/options/timestamp.go
```diff
@@ -0,0 +1,41 @@
...
+// AddFlags implements Interface
+func (o *TimestampAuthorityOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.TSAServerURL, "timestamp-server-url", "",
+		"url to the Timestamp RFC3161 server, default none")
+
+	cmd.Flags().StringVar(&o.TimestampCertChainPath, "timestamp-certificate-chain", "",
+		"path to PEM-encoded certificate chain file for the RFC3161 timestamp authority. Must contain the root CA certificate. "+
+			"Optionally may contain intermediate CA certificates, and may contain the leaf TSA certificate if not present in the timestamp")
+
+	cmd.Flags().BoolVar(&o.InsecureSkipTSResponseVerify, "insecure-skip-ts-verify", false,
```
hectorj2f, line 39: should we remain consistent and use `timestamp` instead of `ts` ?

hectorj2f (approved):
lgtm

haydentherapper:
Will pick up again later.

Issue #2661, opened 2023-01-25
Can cosign add support for VSAs?


TomHennen:
**Description**

I'd like the ability to verify an artifact against a [VSA](https://slsa.dev/verification_summary/v0.2) about that artifact.  As I understand it cosign only supports specific types, can cosign add support for VSAs?

asraa:
Most of the Go type definitions are imported from github.com/in-toto/in-toto-golang/in_toto -- maybe the first step is supporting the struct definitions there? After that, the change is fairly simple, adding it to the supported `PredicateTypeMap` and adding it here: https://github.com/sigstore/cosign/blob/5990090352bd31817295759e93f4e359ecd9e274/pkg/policy/attestation.go#L84

TomHennen:
Looking around a bit I think there are a few tasks needed to get this done.

1. Add the VSA struct and verification as @asraa pointed out.  This should make VSA verification using the `verify-attestation` command possible.
2. Add _policy_ support to verify-blob-attestation to match what verify-attestation does here https://github.com/sigstore/cosign/blob/5990090352bd31817295759e93f4e359ecd9e274/cmd/cosign/cli/verify/verify_attestation.go#L282-L301

Maybe that's it?

TomHennen:
Before I add the VSA type to that in-toto library I'm checking to see if the SLSA folks are expecting anything to happen with the proto definitions they have for provenance v1.  I could imagine it would be nicer if the proto were compiled into a go library and then in-toto/in-toto-golang could just use those definitions instead of their own structs.

asraa:
> I could imagine it would be nicer if the proto were compiled into a go library and then in-toto/in-toto-golang could just use those definitions instead of their own structs.

+1, can you also chime in here? I was not sure what's the situation for that. It seemed like in-toto-golang was the source of truth for v0.2/v0.1, but maybe that can change.
https://github.com/in-toto/in-toto-golang/issues/197

PR #2670, opened 2023-01-27
feat: add policy verification to blob attestations

asraa:

* Adds CUE/Rego policy verification to `verify-blob-attestation`
* Fix: fixes an issue where predicate types were not checked properly in `verify-blob-attestation`. I don't think this command was released so not a security fix....
* Adds tests for policies in verify attestation (which didn't exist??)

#### Release Note

* feat: adds CUE/Rego policy verification in `verify-blob-attestation`

#### Documentation

Part of https://github.com/sigstore/cosign/issues/2661#issuecomment-1405104771

cc @TomHennen

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob_attestation_test.go
+++ b/cmd/cosign/cli/verify/verify_blob_attestation_test.go
@@ -118,3 +118,109 @@ func TestVerifyBlobAttestation(t *testing.T) {
 		})
 	}
 }
+
+func TestVerifyBlobAttestationPolicy(t *testing.T) {
+	ctx := context.Background()
+	td := t.TempDir()
+	defer os.RemoveAll(td)
+
+	blobPath := writeBlobFile(t, td, blobContents, "blob")
+	keyRef := writeBlobFile(t, td, pubkey, "cosign.pub")
+
+	validPolicy1 := writeBlobFile(t, td, "predicateType: \"https://slsa.dev/provenance/v0.2\"", "valid-policy1.cue")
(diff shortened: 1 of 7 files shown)


cmd/cosign/cli/verify/verify_blob_attestation_test.go
```diff
@@ -118,3 +118,78 @@ func TestVerifyBlobAttestation(t *testing.T) {
 		})
 	}
 }
+
+func TestVerifyBlobAttestationPolicy(t *testing.T) {
```
TomHennen, line 122: Would it make sense to include a rego policy in here too?  Probably something like

```
package signature
allow {
  input.predicateType == "https://slsa.dev/provenance/v0.2"
}
```

(borrowed from rego_test.go)

TomHennen (commented):
Thanks Asra, this looks great!

cmd/cosign/cli/verify/verify_blob_attestation_test.go
asraa: yes, i'll do one w/ rego and one with both rego + cue. you caught me that I didn't know rego :P

TomHennen: Ha, me neither. :)

asraa: done

TomHennen (approved):
Sorry for the late review, this fell off my radar!

PR #2676, opened 2023-01-30
add generate-key-pair GitHub Enterprise server support

### LINKED ISSUE #2675: Feature: GitHub Enterprise Server (GHES) support for 'cosign generate-key-pair github:'
**Description**

`cosign generate-key-pair github:` currently only supports `github.com` repositories.
It would be nice if [GitHub Enterprise Server](https://docs.github.com/en/enterprise-server@3.7/admin/overview/about-github-enterprise-server) (GHES) would be also supported by the above command.

netsandbox:
#### Summary

GitHub Enterprise Server (GHES) support for 'cosign generate-key-pair github:'

Fixes #2675

#### Release Note

GitHub Enterprise Server (GHES) support for 'cosign generate-key-pair github:'

### DIFF
--- a/pkg/cosign/env/env.go
+++ b/pkg/cosign/env/env.go
@@ -57,6 +57,7 @@ const (
 	VariableSigstoreRekorPublicKey     Variable = "SIGSTORE_REKOR_PUBLIC_KEY"
 
 	// Other external environment variables
+	VariableGitHubHost               Variable = "GITHUB_HOST"
 	VariableGitHubToken              Variable = "GITHUB_TOKEN" //nolint:gosec
 	VariableGitHubRequestToken       Variable = "ACTIONS_ID_TOKEN_REQUEST_TOKEN"
 	VariableGitHubRequestURL         Variable = "ACTIONS_ID_TOKEN_REQUEST_URL"
@@ -120,6 +121,12 @@ var (
 			External:    true,
 		},
 
+		VariableGitHubHost: {
+			Description: "is URL of the GitHub Enterprise instance",
+			Expects:     "string with the URL of GitHub Enterprise instance",
+			Sensitive:   false,
+			External:    true,
+		},
 		VariableGitHubToken: {
 			Description: "is a token used to authenticate with GitHub",
 			Expects:     "token generated on GitHub",
--- a/pkg/cosign/git/github/github.go
+++ b/pkg/cosign/git/github/github.go
@@ -51,9 +51,19 @@ func (g *Gh) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
 		)
 		httpClient = oauth2.NewClient(ctx, ts)
 	} else {
-		return fmt.Errorf("could not find %q environment variable", env.VariableGitHubRequestToken.String())
+		return fmt.Errorf("could not find %q environment variable", env.VariableGitHubToken.String())
 	}
-	client := github.NewClient(httpClient)
+
+	var client *github.Client
+	if host, ok := env.LookupEnv(env.VariableGitHubHost); ok {
+		var err error
+		client, err = github.NewEnterpriseClient(host, host, httpClient)
+		if err != nil {
+			return fmt.Errorf("could not create github enterprise client: %w", err)
+		}
+	} else {
+		client = github.NewClient(httpClient)
+	}
 
 	keys, err := cosign.GenerateKeyPair(pf)
 	if err != nil {


cpanato:
please sign the DCO

cpanato (approved):
cool! thanks
lgtm

pkg/cosign/env/env.go
```diff
@@ -57,6 +57,7 @@ const (
 	VariableSigstoreRekorPublicKey     Variable = "SIGSTORE_REKOR_PUBLIC_KEY"
 
 	// Other external environment variables
+	VariableGitHubHost               Variable = "GITHUB_HOST"
```
znewman01, line 60: Just want to check that this is a standard env var recognized by e.g. official GitHub tooling. We're trying to define as few env vars that *don't* begin with `COSIGN_` as possible 🙂  The other GitHub env vars here are populated by GH Actions.

A cursory Google search doesn't reveal anything obvious but I could be missing something.

Otherwise, can you rename `COSIGN_GITHUB_HOST` and mark `External: false`?

znewman01 (changes requested):
Thanks for this, LGTM except the one comment!

(Thanks also for your patience with the review, it's been a few days.)

pkg/cosign/env/env.go
netsandbox: As mentioned in https://github.com/sigstore/cosign/issues/2675#issuecomment-1408708548 I choose `GITHUB_HOST` here to make it similar to the existing `GITLAB_HOST` environment variable.
Also as `GITHUB_TOKEN` is already defined, the common prefix would make users aware that these two environment variables belong together.

We don't have a dependency to any GitHub tooling here.
But if you mean to reuse an environment variable that is also used in GitHub tooling, the `GH_HOST` environment variable from GitHub CLI would be the right one (also mentioned in https://github.com/sigstore/cosign/issues/2675#issuecomment-1408708548).

Or do you mean that someone will use `cosign generate-key-pair github: ...` in a GitHub action?
Then `GITHUB_SERVER_URL` will be the right one, because GitHub actions runner (Cloud and self-hosted) expose this environment variable (see https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

znewman01: Thanks, that clarifies a lot. (I missed your earlier comment, sorry.)

*Strictly speaking*, `External` is supposed to mean "this is *provided/specified/set* by an external tool that we have no control over, and therefore we're going to exempt it from the `COSIGN_` naming convention.

That said, following this rule is less important than a good experience for users, so we can easily make an exception.

>  I choose `GITHUB_HOST` here to make it similar to the existing `GITLAB_HOST` environment variable.

To be honest, I think consistency *between* CI/CD providers is less useful than consistency *within* CI/CD providers.

> Also as `GITHUB_TOKEN` is already defined, the common prefix would make users aware that these two environment variables belong together.
>
> But if you mean to reuse an environment variable that is also used in GitHub tooling, the `GH_HOST` environment variable from GitHub CLI would be the right one (also mentioned in [#2675 (comment)](https://github.com/sigstore/cosign/issues/2675#issuecomment-1408708548)).

Annoying that the "official" (GitHub CLI) way to set this is inconsistent between `GH_HOST` and `GITHUB_TOKEN` (I assume historical accident).

> Or do you mean that someone will use `cosign generate-key-pair github: ...` in a GitHub action?
Then `GITHUB_SERVER_URL` will be the right one, because GitHub actions runner (Cloud and self-hosted) expose this environment variable (see https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

That's the reason for e.g. `ACTIONS_ID_TOKEN_REQUEST_TOKEN`.

------------------------------------------------------------------------------------------------

The "maximalist" approach here is probably to do *all* of the above, with a well-defined priority order (this order is up for debate):

1. `GH_HOST`
2. `GITHUB_HOST`
3. `GITHUB_SERVER_URL`

Then, all are marked `External: true`. This would support whatever methods users are already using to set these variables.

I'd be okay with merging what you have for now and doing this in a follow-up, so I'll approve this PR

znewman01 (approved):

pkg/cosign/env/env.go
netsandbox: I'm not sure if the "maximalist approach" makes sense here, but as you sad, we can create a follow-up issue where we gather opinions regarding this.

Issue #2704, opened 2023-02-09
Support non-Sigstore TSAs


haydentherapper:
The [sigstore TSA](https://github.com/sigstore/timestamp-authority/) implements an API to fetch timestamps, as does any other public TSA. Every API uses different paths to request timestamps, but all should conform to the same API structure, accepting timestamp queries and returning timestamp responses. We should support providing a path (`timestamp-url`) not to the sigstore TSA instance, but instead to the complete path to fetch a TSR (e.g. `<sigstoretsapath.com>/api/v1/timestamp`, or `https://freetsa.org/tsr`, or `http://timestamp.digicert.com`)

This means using a different TSA client than the sigstore TSA client, which is only useful for fetching timestamps from a service that has the same API paths as the sigstore TSA.

**Ignore everything below, left as an unnecessary alternative**

For users that want to leverage a TSA that uses a different API than the [sigstore TSA](https://github.com/sigstore/timestamp-authority/), we should support attaching a signed timestamp after artifact signing. 

For container signing, UX would look something like:

1. User signs container with Cosign
2. User fetches signature from OCI annotation
3. User crafts a timestamp query, for example using openssl (where `--data` should be the signature)
4. User requests a timestamp from their trusted TSA, for example [FreeTSA](https://www.freetsa.org/index_en.php#tcp)
5. User attaches the signed timestamp to OCI

Verification with a signed timestamp should then work without any modification, but this should be confirmed.

For blob signing, during verification, simply pass the signed timestamp you fetched from a trusted TSA after signing with `--rfc3161-timestamp`. The one caveat is it must be crafted as a [struct]https://github.com/sigstore/cosign/blob/a84600e3f403c16e3cfc4bcf90a483bd34392a69/pkg/cosign/bundle/tsa.go#L18-L24). I propose we be more lax in accepting a raw or base64 encoded TSR in addition to the struct (I don't think we should add a command to convert a TSR into this struct)

asraa:
RFC 3161 TSA's have a required API - shouldn't cosign be using that? That way it'd be compatible with all RFC 3161 TSA's.

I think there were some specifics of the sigstore's TSA API, but maybe cosign should implement a TSA client from the base API in the spec?

haydentherapper:
[...]

PR #2708, opened 2023-02-10
Support non-Sigstore TSA requests

haydentherapper:
This changes the client to instead simply exchange a TSQ for a TSR given a URL to the API. This will mean that for clients using timestamping currently, they will need to update their Cosign calls to use the full path (e.g. tsa.sigstore/api/v1/timestamp)

Fixes https://github.com/sigstore/cosign/issues/2704

### DIFF
--- a/internal/pkg/cosign/tsa/client/client.go
+++ b/internal/pkg/cosign/tsa/client/client.go
@@ -16,7 +16,6 @@ package client
 
 import (
 	"bytes"
-	"crypto/x509"
 	"fmt"
 	"io"
 	"net/http"
@@ -27,27 +26,27 @@ import (
 	"github.com/pkg/errors"
 )
 
+// TimestampAuthorityClient should be implemented by clients that want to request timestamp responses
+type TimestampAuthorityClient interface {
+	GetTimestampResponse(tsq []byte) ([]byte, error)
+}
+
 // TimestampAuthorityClient provides a client to query RFC 3161 timestamp authorities
-type TimestampAuthorityClient struct {
-	Timestamp TimestampAuthorityService
+type TimestampAuthorityClientImpl struct {
+	TimestampAuthorityClient
 
 	// URL is the path to the API to request timestamp responses
 	URL string
 
-	// CertificateChain contains an optional certificate chain used to verify timestamp responses
-	CertificateChain []*x509.Certificate
-}
-
-// TimestampAuthorityService should be implemented by clients that want to request timestamp responses
-type TimestampAuthorityService interface {
-	GetTimestampResponse(tsq []byte) ([]byte, error)
+	// Timeout is the request timeout
+	Timeout time.Duration
 }
 
 // GetTimestampResponse sends a timestamp query to a timestamp authority, returning a timestamp response.
 // The query and response are defined by RFC 3161.
-func (t *TimestampAuthorityClient) GetTimestampResponse(tsq []byte) ([]byte, error) {
+func (t *TimestampAuthorityClientImpl) GetTimestampResponse(tsq []byte) ([]byte, error) {
 	client := http.Client{
-		Timeout: time.Second * 5,
+		Timeout: t.Timeout,
 	}
 	req, err := http.NewRequest("POST", t.URL, bytes.NewReader(tsq))
 	if err != nil {
@@ -59,7 +58,7 @@ func (t *TimestampAuthorityClient) GetTimestampResponse(tsq []byte) ([]byte, err
 	if err != nil {
 		return nil, errors.Wrap(err, "error making request to timestamp authority")
 	}
-	if tsr.StatusCode != 200 {
+	if tsr.StatusCode != 200 && tsr.StatusCode != 201 {
 		return nil, fmt.Errorf("request to timestamp authority failed with status code %d", tsr.StatusCode)
 	}
 
@@ -77,4 +76,8 @@ func (t *TimestampAuthorityClient) GetTimestampResponse(tsq []byte) ([]byte, err
 	fmt.Fprintln(os.Stderr, "Timestamp fetched with time: ", ts.Time)
 
 	return resp, nil
+}
+
+func NewTSAClient(url string) *TimestampAuthorityClientImpl {
+	return &TimestampAuthorityClientImpl{URL: url, Timeout: 10 * time.Second}
 }
--- a/internal/pkg/cosign/tsa/client/client.go
+++ b/internal/pkg/cosign/tsa/client/client.go
@@ -1,0 +1,80 @@
+// Copyright 2023 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"bytes"
+	"crypto/x509"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/digitorus/timestamp"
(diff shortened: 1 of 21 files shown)


haydentherapper:
Will test PR in the morning

hectorj2f (commented):
Can we add tests against these non-sigstore supported TSAs ?

internal/pkg/cosign/tsa/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+
+// TimestampAuthorityClient provides a client to query RFC 3161 timestamp authorities
+type TimestampAuthorityClient struct {
+	Timestamp TimestampAuthorityService
+
+	// URL is the path to the API to request timestamp responses
+	URL string
+
+	// CertificateChain contains an optional certificate chain used to verify timestamp responses
+	CertificateChain []*x509.Certificate
```
asraa, line 38: where does this get populated? it seems like it's not used outside of tests and the flags aren't used to populate it

Is this worth creating a `NewTimestampAuthorityClient` that can populate it?

internal/pkg/cosign/tsa/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+	"os"
+	"time"
+
+	"github.com/digitorus/timestamp"
+	"github.com/pkg/errors"
+)
+
+// TimestampAuthorityClient provides a client to query RFC 3161 timestamp authorities
+type TimestampAuthorityClient struct {
+	Timestamp TimestampAuthorityService
```
asraa, line 32: You don't need this here - the `TimestampAuthorityClient` becomes an impl of that interface as soon as it implements the interface.

internal/pkg/cosign/tsa/client/client.go
hectorj2f: In my opinion, this should be populated when signing to verify the response.

internal/pkg/cosign/tsa/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+// TimestampAuthorityService should be implemented by clients that want to request timestamp responses
+type TimestampAuthorityService interface {
+	GetTimestampResponse(tsq []byte) ([]byte, error)
+}
+
+// GetTimestampResponse sends a timestamp query to a timestamp authority, returning a timestamp response.
+// The query and response are defined by RFC 3161.
+func (t *TimestampAuthorityClient) GetTimestampResponse(tsq []byte) ([]byte, error) {
+	client := http.Client{
+		Timeout: time.Second * 5,
```
hectorj2f, line 50: I believe we had a 10 second timeout, if we change it. We might also think about exposing this value to be configurable.

internal/pkg/cosign/tsa/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	req, err := http.NewRequest("POST", t.URL, bytes.NewReader(tsq))
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating HTTP request")
+	}
+	req.Header.Set("Content-Type", "application/timestamp-query")
```
hectorj2f, line 56: Q: Where is this header verified ? We need to add it.

internal/pkg/cosign/tsa/client/client.go
```diff
@@ -0,0 +1,80 @@
...
+
+	tsr, err := client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "error making request to timestamp authority")
```
hectorj2f, line 60: let's log both the status code and the error.

internal/pkg/cosign/tsa/client/client.go
haydentherapper: With the current design, I need a named interface, otherwise I can't inject in a mocked interface. But thinking more, this design is overly complicated, instead I'll just make `TimestampAuthorityClient` an interface and have the mock implement `GetTimestampResponse` and add a method for `GetCertChain`.

internal/pkg/cosign/tsa/client/client.go
haydentherapper: Done and done!

internal/pkg/cosign/tsa/client/client.go
haydentherapper: This is not verified client side, this is a required header for the server.

internal/pkg/cosign/tsa/client/client.go
haydentherapper: That's done on the next line. If err != nil here, this would be if the client failed to call the server for some reason (say, network issue or TCP dial failure). There would be no status code.

internal/pkg/cosign/tsa/client/client.go
haydentherapper: Reworked, it looks a lot cleaner now (and we dont expose CertChain for the real implementation)

internal/pkg/cosign/tsa/client/client.go
haydentherapper: This is gone now! Only needed for mocks

haydentherapper:
Tested with Sigstore TSA:

```
<started up local TSA, copied cert chain>

cosign sign-blob --timestamp-server-url http://localhost:3000/api/v1/timestamp --rfc3161-timestamp tsaresponse --tlog-upload=false --bundle=blob.bundle --output-certificate=blob.cert  blob --yes

cosign verify-blob --rfc3161-timestamp tsaresponse --timestamp-certificate-chain tsacertchain  --certificate-identity <email> --certificate-oidc-issuer https://accounts.google.com --insecure-ignore-tlog --bundle blob.bundle --offline --certificate blob.cert blob
```

Tested with FreeTSA:

```
cosign sign-blob --timestamp-server-url https://freetsa.org/tsr --rfc3161-timestamp tsaresponse --tlog-upload=false --bundle=blob.bundle --output-certificate=blob.cert  blob --yes

# Grabbed the tsa chain from the website

cosign verify-blob --rfc3161-timestamp tsaresponse --timestamp-certificate-chain freetsachain  --certificate-identity <email> --certificate-oidc-issuer https://accounts.google.com --insecure-ignore-tlog --bundle blob.bundle --offline --certificate blob.cert blob
```

@hectorj2f I don't think we should add a non-sigstore TSA to the e2e tests because that's adding an external dependency.

internal/pkg/cosign/tsa/client/client.go
hectorj2f: But we are returning here, there is not status code in the error. Right ?

hectorj2f:
> @hectorj2f I don't think we should add a non-sigstore TSA to the e2e tests because that's adding an external dependency.

How would we know we didn't break it ? 

Also, should we document the list of supported non-TSAs ?

haydentherapper:
We just send the timestamp query to the timestamp authority, any TSA should work. We are no longer treating the sigstore TSA as different, so if we break something, we'll notice it in the e2e tests.

internal/pkg/cosign/tsa/client/client.go
haydentherapper: Correct, there's only a status code if err == nil

haydentherapper:
I also want to make sure we're not describing the sigstore implementation of a TSA as something special. It's just an server implementation of rfc3161, there should be no differences in the accepted API requests and responses to any TSA implementation. This PR just changes the client so that you have to pass the path to fetch a TSR, whereas previously the sigstore tsa client knew the path.

hectorj2f (approved):

hectorj2f:
Sgtm

haydentherapper:
This should be good to merge.

hectorj2f:
Yes, let's merge it. We need to change the clients to adapt to these changes :).

haydentherapper:
Do we know who's using this? This is just in the 2.0 RC so I assume we only need to update the blog post you published?

hectorj2f:
@haydentherapper I was thinking the policy-controller, but the verification isn't affected by this. On the other hand, we use the signing piece in a different component, so whenever we release a new RC, I'll change our client calls.

haydentherapper:
Ah, didn't realize it was used there. Would this only affect deployments of policy-controller? I assume there's no hardcoded TSA APIs, no other code should need to be updated

haydentherapper:
Bump, should be ready for merging.

Issue #2714, opened 2023-02-10
sign-blob: Certificate is not included in bundle output if entry is not uploaded to Rekor


haydentherapper:
[rekorBytes](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign_blob.go#L130) will be empty if not uploaded, even if there is a certificate. Easy fix, we should just read the certificate from Fulcio's response and not Rekor's.

PR #2715, opened 2023-02-10
Output certificate in bundle when entry is not uploaded to Rekor

haydentherapper:
The issue was that we read the certificate from a variable set only if an entry was uploaded.

Fixes https://github.com/sigstore/cosign/issues/2714


haydentherapper:
Verified with `cosign sign-blob  --tlog-upload=false --bundle=blob.bundle --output-certificate=blob.cert  blob --yes` and `cosign verify-blob  --certificate-identity <email> --certificate-oidc-issuer https://accounts.google.com --insecure-ignore-tlog --bundle blob.bundle --offline  blob`, the latter of which previously failed because it said no verifier was presented.

znewman01 (approved):

Issue #2126, opened 2022-08-04
Why is an SBOM attestation predicate wrapped in CosignPredicate?


otms61:
**Question**

When Cosign creates an SBOM attestation, the SBOM predicate is wrapped in [CosignPredicate](https://github.com/sigstore/cosign/blob/0f017f82d1f6201799548dc1d60dfbe61426e562/specs/COSIGN_PREDICATE_SPEC.md
). ([spdx](https://github.com/sigstore/cosign/blob/5aa17b98ec2cb16721b314d48d5599df8d4387d9/pkg/cosign/attestation/attestation.go#L244-L246), [cyclonedx](https://github.com/sigstore/cosign/blob/5aa17b98ec2cb16721b314d48d5599df8d4387d9/pkg/cosign/attestation/attestation.go#L257-L259))
Therefore, SBOM locates under `predicate.Data`.

The photo below is an example of spdx.
![sbom_attestation](https://user-images.githubusercontent.com/1850144/182736647-5e81a1cb-b904-4f60-90e7-0632d8d8e76c.png)

This differs from the [in-toto spdx specification](https://github.com/in-toto/attestation/blob/bfd304734d26e5a10e16609ca812d605407d24c5/spec/predicates/spdx.md
). It defines that the SBOM is located under the `predicate`.

```json
{
  // Standard attestation fields:
  "_type": "https://in-toto.io/Statement/v0.1",
  "subject": [{ ... }],

  // Predicate:
  "predicateType": "https://spdx.dev/Document",
  "predicate": {
    "SPDXID" : "SPDXRef-DOCUMENT",
    "spdxVersion" : "SPDX-2.2",
    ...
  }
}
```

This difference makes it difficult to tell whether the SBOM is in `predicate` or `predicate.Data`, when we get an SBOM attestation.

So, why is an SBOM attestation predicate wrapped in CosignPredicate?

developer-guy:
Hi @otms61, thanks for the question. You don't need to use _CosignPredicate_ unless you specify a type for your attestation. AFAIK, cosign will use the _custom_ type and put the given predicate under the _Data_ section if you don't specify any type. Still, of course, there is a chance to define your attestation type by using a URL format like this _--type https://spdx.dev/Document_; if that happens, cosign will put the data under the _predicate_ section like the following screenshot:
[...]

Issue #2307, opened 2022-10-04
Feature: Provide an easier way to retrieve SBOM from In-Toto attestation


ChristianCiach:
Since https://github.com/sigstore/cosign/pull/1756 cosign issues a warning when using `cosign attach sbom` and `cosign download sbom`, heavily encouraging users to use `cosign attest` and `cosign verify-attestment` instead.

While I don't really like this change, because all our sboms are signed and there is no need to scare the users like that, we are trying to migrate to use said attestations.

Unfortunately, there doesn't seem to be an easy way to retrieve the original SBOM document from an attestation. The `cosign download sbom` could be used to retrieve the original SBOM (and nothing else), this is a lot harder when using attestations: 

```
$ cosign verify-attestation --key cosign.pub --type cyclonedx -o text my-registry/my-image:1.0.0 | jq '.payload | @base64d | fromjson | .predicate.Data'
```

This is the easiest way I could come up. This also requires the user to have additional tools available (`jq` in this case).

I think there should be a command to retrieve the SBOM with the same output as we were able by using `cosign download sbom` before. Currently, SBOMs attached as attestation-predicates feel like second-class citizens.

ChristianCiach:
Slight off-topic: I think there should be a way to avoid the warnings when using `cosign download sbom` and `cosign attach sbom`.  Maybe like this:  We could add a `--key` option to both commands. If present, `cosign download sbom` automatically verifies the signature of the downloaded sbom-image against the given public key and `cosign attach bom` automatically signs the attached sbom using the given private key. This would be consistent with the rest of the CLI commands.

Would this idea even have a chance to be implemented? Should I create an issue for this?

ChaosInTheCRD:
In relation to your first question, do you think that a flag such as `--view-predicate` would suffice here? This flag would ensure that if called it can strip out and decode the predicate from everything wrapped around it.

ChaosInTheCRD:
on the second question - in order to sign the SBOMs, have you up until now been using `cosign sign-blob`?

ChristianCiach:
Being able to combine these two commands into one command would actually be *double nice*, because the same image tag is used in both commands, but the tag could resolve to difference digests between both invocations, as described here: https://github.com/sigstore/cosign/pull/2313

znewman01:
#2526 points out that have docs such that when you follow them, Cosign emits warnings: https://docs.sigstore.dev/cosign/other_types/#sboms-software-bill-of-materials

[...]

PR #2718, opened 2023-02-11
Fix: Drop the `CosignPredicate` wrapper around SBOM attestations.

mattmoor:
:bug: This change drops the `CosignPredicate` that `cosign` wraps around SPDX/CycloneDX attestations.

Currently `cosign` wraps SPDX and CycloneDX attestations produced via their shortnames (`cosign attest --type {spdxjson|cyclonedx}`) in a `CosignPredicate` envelope.

However, the whole point of the in-toto `predicateType` is to specify the schema of the `predicate`, and despite using the SPDX and Cyclone predicate type URIs, this envelope violates their schema with the extra layer.

Moreover, if users were to attest these SBOMs with the explicit predicate type URI:
```
cosign attest --type https://spdx.dev/Document ...
```

Then `cosign` will NOT add this additional envelope, which makes it effectively impossible to know the schema to use for policy validation based strictly on the `predicateType` because even `cosign` will produce these attestations both ways.

Fixes: https://github.com/sigstore/cosign/issues/2126

/kind bug

#### Release Note

BREAKING: This change removes the `CosignPredicate` envelope that wraps the predicates of SPDX and CycloneDX attestations, which was a violation of the schema specified via their `predicateType` field.

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -241,9 +241,7 @@ func generateSPDXStatement(rawPayload []byte, digest string, repo string, parseJ
 	}
 	return in_toto.SPDXStatement{
 		StatementHeader: generateStatementHeader(digest, repo, in_toto.PredicateSPDX),
-		Predicate: CosignPredicate{
-			Data: data,
-		},
+		Predicate:       data,
 	}, nil
 }
 
@@ -254,9 +252,7 @@ func generateCycloneDXStatement(rawPayload []byte, digest string, repo string) (
 	}
 	return in_toto.SPDXStatement{
 		StatementHeader: generateStatementHeader(digest, repo, in_toto.PredicateCycloneDX),
-		Predicate: CosignPredicate{
-			Data: data,
-		},
+		Predicate:       data,
 	}, nil
 }
 
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -260,8 +260,8 @@ func TestAttestVerifySPDXJSON(t *testing.T) {
 	attestVerify(t,
 		"spdxjson",
 		string(attestationBytes),
-		`predicate: Data: spdxVersion: "SPDX-2.2"`,
-		`predicate: Data: spdxVersion: "SPDX-9.9"`,
+		`predicate: spdxVersion: "SPDX-2.2"`,
+		`predicate: spdxVersion: "SPDX-9.9"`,
 	)
 }
 
@@ -273,8 +273,8 @@ func TestAttestVerifyCycloneDXJSON(t *testing.T) {
 	attestVerify(t,
 		"cyclonedx",
 		string(attestationBytes),
-		`predicate: Data: specVersion: "1.4"`,
-		`predicate: Data: specVersion: "7.7"`,
+		`predicate: specVersion: "1.4"`,
+		`predicate: specVersion: "7.7"`,
 	)
 }
 


hectorj2f:
Please, sign off your commits.

hectorj2f (approved):

mattmoor:
@hectorj2f yeah, I was just letting CI run to tell me what all I broke as I juggle this and the other commit.  I already signed it locally, I was just wondering whether I could save MS some money by waiting to update things 😁

mattmoor:
🚨 🚨 Let's not merge this until (at least) Monday 🚨 🚨 

... since I understand I'm sending breaking changes on the weekend, and I'd like to get more 👀 on this.

hectorj2f:
Lgtm

priyawadhwa (approved):

hectorj2f (approved):

znewman01:
I think this is a good fix and I'm okay with a breaking change here *but* I want to understand who this would break before I approve.

AFAICT there're no Sigstore tools (other than policy-controller) that we'd be breaking, and they already know about both versions of this.

But from #2126:

> But changing this now can lead to issues elsewhere. For example, Trivy explicitly supports the current attestation format: https://aquasecurity.github.io/trivy/v0.31.3/docs/attestation/sbom/

So the question is: does `trivy sbom` work with proper SBOM attestations (e.g., no `CosignPredicate`)? If not, do we just accept the breakage with Cosign 2.0-produced SBOMs? (It is technically a bug since folks should have been handling the `--type https://spdx.dev/...` case but I wouldn't be shocked otherwise.)

Same question for tools other than trivy that might be depending on this.

The alternative might be:

1. Deprecate the old behavior. Allow folks to opt-in to the new behavior with a flag.
2. Proactively try to find consumers and get them to fix.
3. Wait.
4. Switch over to the new behavior. Probably still cause a lot of breakage.

I'm not sure it's worth that much effort so I think I'd be okay with just merging this. Just wanted to think it through.

developer-guy (approved):
Totally agree with you

mattmoor:
@znewman01 So I believe that everything in the Trivy example you link will continue to work just fine because this doesn't change anything about the signatures, and these are just signing and attesting.

What breaks is when the user has a Cue/Rego policy over the attestation content, then they previously had an extra `Data:` layer, which is now gone.

znewman01:
> @znewman01 So I believe that everything in the Trivy example you link will continue to work just fine because this doesn't change anything about the signatures, and these are just signing and attesting.

They have an example at the bottom where they parse the output of `verify-attestation`. Does `verify-attestation` strip the `Data:` wrapper?

```
$ cosign verify-attestation --key /path/to/cosign.pub --type cyclonedx <IMAGE> > sbom.cdx.intoto.jsonl
$ trivy sbom ./sbom.cdx.intoto.jsonl
```
> 
> What breaks is when the user has a Cue/Rego policy over the attestation content, then they previously had an extra `Data:` layer, which is now gone.

Yeah, that's a good point. I think I've been talked into just ripping off the band-aid, though, so I'm okay if you want to merge.

mattmoor:
@znewman01 Thanks, I missed that bit.

> Does verify-attestation strip the Data: wrapper?

I'm pretty sure it doesn't, but I could be wrong.

🤔  Now you have me wondering what `trivy sbom FILE` actually does because (as a naive user) I'd assume that it would support taking the output of `trivy image --format cyclonedx -o sbom.cdx.json <IMAGE>`, which is missing `Data` (but also all of the attestation wrapping).

If Trivy doesn't support both ways today, then it is already broken today for `cosign attest --type https://cyclonedx.com/...` unfortunately.  Put a bit more generally...  any tooling that is broken by this change is already partially broken in the case where users are attesting with an explicit predicate type.


I dislike that this could break some user policies, but at the same time because of this strange duality we also can't really advise users on how to write policies against a schema because the current mechanism violates the schema things should have.  My hope is that if we "stop the bleeding" now, we can get ahead of things before this stuff goes too mainstream to walk back 🤞

ChristianCiach:
> > Does verify-attestation strip the Data: wrapper?
> 
> I'm pretty sure it doesn't, but I could be wrong.

It does not, which is exactly why I created the issue:
* https://github.com/sigstore/cosign/issues/2307

As commented in https://github.com/sigstore/cosign/issues/2307#issuecomment-1415670359, this would also make the newly added attestation-support in Trivy obsolete.

Issue #2376, opened 2022-10-24
Cosign 2.0 Tracking


haydentherapper:
**Description**

This is an issue to track Cosign 2.0, with the primary change being the removal of the experimental flag. I've added some initial issues I think would be in scope, feel free to discuss and edit. We should also triage open bugs.

- [ ] Cosign experimental and offline mode - https://github.com/sigstore/cosign/issues/2255
- [ ] Cosign inspect - https://github.com/sigstore/cosign/issues/2210
- [ ] Require verification flags - https://github.com/sigstore/cosign/issues/2056

There's also the open question of sigstore-go - Should Cosign 2.0 also involve the redesign of the Cosign API? I'm thinking no, let's narrowly scope 2.0 to focus on client work to support GA, and a 3.0 release could be focused on library redesign.

cc @znewman01 @priyawadhwa

asraa:
I'd like to propose another thing: conglomerating all the env vars we use for pinning rekor/ct/fulcio keys and creating a Root provider interfact, that either uses TUF or populates with any env vars that are enabled.

This would be a MAJOR cleanup and help people use cosign in special environments.

znewman01:
See also: https://github.com/sigstore/cosign/discussions/2365 especially the proposal to separate the CLI versioning and the library versioning, and the proposal to provide no guarantees about API stability. 

I'm in favor of a quick, relatively narrow 2.0 (scoped to CLI only) because dropping COSIGN_EXPERIMENTAL is a big win and because a long-lived `v2-staging` branch is a recipe for headache. While we're in there, we should skim for deprecated functionality and things we've been planning to yank, too. So a quick cleanup around env vars SGTM, but I'm a little skeptical of a total refactor of the root provider. I think that would fold nicely into the sigstore-go work, though.

priyawadhwa:
Thanks for opening this tracking issue @haydentherapper! +1 to a narrow-scoped 2.0, I think all the tagged issues are in-scope. For the root provider refactor, I'm ok not making it a requirement for 2.0 but if it gets in by the time we're ready to release that would be awesome :)

asraa:
> I'm ok not making it a requirement for 2.0 but if it gets in by the time we're ready to release that would be awesome :)

I think that's fair: if we can make the requirement of ripping out any TUF calls from the internal verification functions I think that would suffice! Doing it quickly makes sense. @vaikas

asraa:
This would be great: https://github.com/sigstore/cosign/issues/2222

[...]

PR #2735, opened 2023-02-22
Default generated PEM labels to SIGSTORE

### LINKED ISSUE #2471: Generated PEM labels should say "SIGSTORE" rather than "COSIGN"
```shell
$ cd $(mktemp -d) && COSIGN_PASSWORD="" cosign generate-key-pair && head -n 1 cosign.key
-----BEGIN ENCRYPTED COSIGN PRIVATE KEY-----
```

This means that any Sigstore client implementation must:

1. Accept the "COSIGN PRIVATE KEY" format, or else they won't be able to use Cosign keys
2. Generate keys with "COSIGN PRIVATE KEY", or else Cosign won't be able to understand those keys. This is a little weird if the client has nothing to do with cosign.

I'd prefer "BEGIN ENCRYPTED SIGSTORE PRIVATE KEY" if this is going to be a shared format. (That's what sigstore-rs emits.)

I propose that Cosign:

- *Immediately* start accepting "SIGSTORE" keys (in addition to "COSIGN" keys)
- *Eventually* start emitting "SIGSTORE" keys.
   - I think we could definitely do this right away if we get it in before 2.0 (https://github.com/sigstore/cosign/issues/2376)
   - Otherwise, we *may* want to transition the behavior subject to a deprecation period (as in [Proposal: Cosign Versioning](https://docs.google.com/document/d/1urWUPhtzXKWqL9CoaEw4Z35v5IDl9yrTRQ40XlYekOo/edit#heading=h.r9yrcczhn25h)). I can see an argument for skipping deprecation too.
 - *Very* long-term could stop accepting "COSIGN" public keys, probably doesn't need to do this though.
 - We could consider an option to emit "COSIGN" keys for backwards compat.

Other Sigstore clients:

- MUST emit "SIGSTORE" keys by default
- SHOULD accept "COSIGN" keys
- MAY have an option to emit "COSIGN" keys

Context: https://github.com/sigstore/sigstore-rs/pull/165

ivanayov:
#### Summary

This change adds "SIGSTORE" PEM label and accepts both the old "COSIGN" label and the new one.
Newly generated keys are now created with "SIGSTORE" label

Resolves #2471

As #2376 is not closed yet, this change updates to generating private keys with "SIGSTORE" PEM label by default.

Did not update the CI (in particular .github/workflows/cosign-test.key) for backwards compatibility safety (i.e. still available "COSIGN" label support).

#### Release Note

Generated PEM labels were updated to "SIGSTORE" rather than the previous "COSIGN". Both are still supported and accepted, but newly generated will be with the updated label.

#### Documentation

PEM labels were not previously referenced in the docs, so nothing to update.


znewman01 (changes requested):
Sorry for the slow review! I should be quicker in the future.

Can we add a test case to make sure that both are read successfully?

One for writing too would be good.

ivanayov:
Thanks @znewman01  !

I added tests for reading both the old and the new type, and writing the new type, as the current change defaults to "SIGSTORE".

You mentioned an option to emit the old "COSIGN" type for backwards compatibility and I have some questions:
* As both are accepted successfully now, do we consider it's breaking the backwards compatibility for not emitting "COSIGN"? (To me it looks like No, but you might know some use-cases where people would still need generating the old type.)
* Supporting both would mean giving a cmd or config options to select from and I have some concerns here:
  -  Generating key pair at the moment is pretty simple and straightforward and adding one more step for something that's usually auto-generated might look confusing
  - Allowing an option to select a PEM label looks a bit strange from user perspective and I think if we do that, there should be some strong reasoning behind
  - I had a look and it requires some refactoring - not big, but touching generic functionality. I'm ok with addressing that if you think it's worth and not risky for a temporary change (considering it's going to be removed in several months)
  
  
This is why the current change emits only "SIGSTORE", but I'm happy to add support for both if you think that would be the right approach.

Thank you for your input in advance.

znewman01:
> As both are accepted successfully now, do we consider it's breaking the backwards compatibility for not emitting "COSIGN"? (To me it looks like No, but you might know some use-cases where people would still need generating the old type.)

The question is: are we worried about scenarios where a *new* client generates a key and an older client uses that key?

I can see an argument for "no" because this is mostly a client-side concern, but I worry about public keys generated by new clients.

That said--I'd be okay omitting the option for now for the other reasons you mention.

znewman01 (approved):

znewman01:
If you fix lint I'm happy with this!

ivanayov:
Thanka @znewman01, I pushed an update

PR #2737, opened 2023-02-23
Loosen verification predicate type + better error messages

vaikas:
#### Summary
Loosen the policy validation restrictions. With the move to strict RFC3986 and the fact that some of the existing attestations out there are still using the non conformant predicate types, allow validating them with non-conformant predicate types.
Also, to aid folks with the 'guess-the-existing-predicate-type' game :) if attestations are found, but they do not match the predicate type that is being looked for, print out the predicates that were found. Hope is that this will make the migration from non conformant predicate types little easier.

### DIFF
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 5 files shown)


hectorj2f (approved):
lgtm, thanks. I truly believe we need this change in before releasing v2.0.0.

mattmoor (approved):
Awesome, I think this will be super helpful for folks.

haydentherapper:
+1 to this, breaking verification for existing attestations is not ideal. This should go in 2.0

hectorj2f:
Okay. I just talked to @priyawadhwa ✅ . I'm merging this change as it seems everyone approved it.

PR #2746, opened 2023-02-24
verify-blob-attestation: Loosen arg requirements if --check-claims=false

wlynch:
#### Summary

1. Update --check-claims documentation to better describe the behavior of the flag.
2. Modify verify-blob-attestation to not require a blob argument if --check-claims=false. This file is not used if we're not checking claims - it can be set to anything and still pass.

See https://sigstore.slack.com/archives/C01PZKDL4DP/p1677006107531799 for more discussion.

#### Release Note

* verify-blob-attestation will now accept 0 arguments if `--check-claims=false` is set.

#### Documentation

✅


wlynch:
CC @priyawadhwa @asraa

priyawadhwa (approved):

hectorj2f (approved):
lgtm

Issue #2131, opened 2022-08-04
Cosign signing bundles


patflynn:
Details of this proposal available in this [Google Doc.](https://docs.google.com/document/d/1gucjOA_bGyRjK6TeaOI-X5GIUv8WsPzeMDMkq25Kv4Y/edit#heading=h.we5fqok7jai5)

### Summary

The current `sign-blob` command and documentation steer the user towards producing two files (raw sig and cert) for each signing operation.  This is not ideal because:

- the two files do not contain enough information to enable offline verification.
- more files to store and pass as arguments is worse from a usability stand point.
- there is a lack of symmetry with the `cosign sign` command for OCI images which produces just one signing artifact.

These issues are likely to exacerbate inconsistencies across package manager and language ecosystems as each platform could produce a different combination of sig|cert|bundle.  

[Our proposal](https://docs.google.com/document/d/1gucjOA_bGyRjK6TeaOI-X5GIUv8WsPzeMDMkq25Kv4Y/edit#heading=h.we5fqok7jai5) produces a default blob signing and verification path that produces just one complete artifact supporting all verification use-cases. 

At this stage mainly looking for feedback/sanity check. Some details left to be worked out. Very happy to take suggestions for improvements.

kommendorkapten:
This initiative is great, and as I'm working on a similar concept but for attestations, where the actual payload is captured in the signature file I created this proposal: https://github.com/sigstore/cosign/issues/2147.
As they are so similar, there came an immediate comment https://github.com/sigstore/cosign/issues/2147#issuecomment-1210803817 that we could actually compose these two proposals into one! 
I would be very interesting to hear your thoughts on that, I think it's very compelling.

patflynn:
Fantastic! Attestations was a distant blip on my radar for blobs but I was getting an uneasy feeling that they applied here. I'm very happy to see you connect the dots with this proposal. It makes complete sense to me that simple signatures couple be folded into some kind of attestation type so that we can use the same format across these very similar use-cases.  Merging these proposals seem trivial at first glance assuming that the binary signature as an attestation makes sense to the sigstore experts.

@znewman01 
@asraa 
@haydentherapper

kommendorkapten:
[...]

PR #2785, opened 2023-03-10
Require a payload to be provided with a signature

mtrmac:
#### Summary
In near future, I plan to propose changing the contents of the “signature payload” (the blob that is signed by the raw cryptographic signature.

That’s currently not _cleanly_ possible because various code paths assume that the payload can generated afresh just from the image name/digest. Because the cryptographic signature signs that payload, that essentially forces us to freeze the payload contents forever, and assumes that Golang’s JSON marshaler will always produce byte-for-byte identical output to its current version. That’s not a reasonable thing to assume long-term.

So, make it explicit that that’s not how signing workflows should work, and wherever the user is providing a signature, force the user to also provide the payload signed by that signature. This, I think, _mostly_ affects low-level users that manipulate the contents of the signature bundle as individual components, but not most users that operate with the signature bundle as a whole (or don’t think about the bundle at all).

(Alternatively, we could not _require_ the payload to be provided now, and assume the current content for compatibility, with a warning. But that risks breaking the users at some inopportune time not under anyone’s control, maybe with a Go security update, instead of being able to manage that transition within Cosign. Based on the unquantified guess that this is not likely to affect most users, I think just breaking these uses might be acceptable.)

See just below for the CLI changes.

See individual commit messages for a bit more detail.

#### Release Note
- Added `--output-payload` option to `cosign sign`, to be used by users of `--output-signature`.
- `cosign attach signature` now warns if the `--payload` option is not used along with `--signature`.
- Added `--payload` option to `cosign verify`, `cosign dockerfile verify` and `cosign manifest verify`; these commands now warn if `--payload` is not used along with `--signature`.

#### Documentation
https://github.com/sigstore/docs/pull/84


mtrmac:
Updated: The generation of implied payload is now restored, consolidated into `pkg/cosign.ObsoletePayload`, and triggers a warning when used.

haydentherapper:
How is the blob or container used for verification? I'm still not sold on users providing the hash instead of the container/blob, because you get into cases where a user wants to verify X but are given a hash/sig to Y. If the issue is that additional metadata/annotations are not bundled alongside the blob/container, should we instead be finding ways to persist those? https://github.com/sigstore/protobuf-specs might be the answer to that, attaching the additional metadata alongside the [artifact](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_verification.proto#L115).

mtrmac:
> How is the blob or container used for verification?

This PR changes absolutely nothing in that respect.
- For verify-blob, there is no “payload”; the signed bytes are the raw blob.
- For verifying signatures of container images, https://github.com/sigstore/cosign/blob/8cb4c38ec3d354b1e776f59abe71576af8a2bdaf/pkg/cosign/verifiers.go#L45 . In fact, this this is also an essential required component of authenticating with the “implied payload” mechanism that I’m trying to deprecate (conceptually, with implied payload, the registry is queried for a digest, that is used to form the payload, the payload is then cryptographically verified; with an explicit payload, the explicit payload is cryptographically verified, and then used to read the signed digest, and that’s matched with the image being verified). Again, nothing changes.

> I'm still not sold on users providing the hash instead of the container/blob, because you get into cases where a user wants to verify X but are given a hash/sig to Y.

I’m not sure what this refers to; nothing like that happens, and I’m not proposing adding nothing like that either, AFAICS. Where exactly are users “providing the hash”?  Sure, I guess, users running `cosign verify quay.io/foo/bar@sha256…` are “providing the hash”, but a successful verification authenticates the hash (other than the identity concerns I’ll be trying to resolve next), so that’s not a risk.

(It _is_ a risk that users verify bar@sha256… and then carelessly run maybe bar:v1.2.3, but that’s not a risk that can be fixed with the `cosign verify` command structure; that’s exactly why Podman does not provide a `podman verify` command at all, and instead verification and enforcement is implied by a `podman pull`. It’s not my ambition improve _that_ in this PR.)

> If the issue is that additional metadata/annotations are not bundled alongside the blob/container, should we instead be finding ways to persist those?

That’s _exactly_ what this PR is doing! The data is already been generated during signing, so this adds a new option to persist the data on signing, and to use that data on verification, where those options were missing.

(And, as a reminder, nothing of this is user-relevant in the ~typical scenarios of `cosign sign` and `cosign verify` that store data on the registry, because the payload is _always_ saved on the registry, transparently to users. It affects only users who manage and move around the individual components of the signature bundle as local files.)

haydentherapper:
Ah, I misunderstood, thanks for your explanation. I had thought payload was referring to just a hash that’s signed over, hence my reservation. Now I see this is meant to be a set of values as additional metadata. 

I’ll take another look tomorrow!

patricklawsongoogle:
I'd just like to +1 the general notion of preferring to include the signed payload with the signature instead of relying on implicit payload generation, particularly if canonicalization is involved.  The fewer dependencies needed for the authentication step of consuming attestations, the better.  This was one of the design motivations for DSSE when we were comparing it against existing signature enveloping schemes, like In-Toto's.  Here's one relevant comment in the broader thread: https://github.com/secure-systems-lab/dsse/issues/1#issuecomment-701335606

mtrmac:
Rebased, PTAL.

znewman01:
Thanks @mtrmac for the PR and the detailed explanation. I'm also in favor of avoiding implicit payload generation.

A minor point:

> We can put a deprecation warning in, but shouldn't break users until we have a roadmap for a new major release.

FWIW we can break users [after at least 6 months](https://github.com/sigstore/cosign/blob/main/VERSIONING.md#deprecation). So I think a warning if --payload is omitted for verify is okay for now.

haydentherapper:
Given that we know we want to integrate with the sigstore bundle format defined in https://github.com/sigstore/protobuf-specs/, should we hold off in this PR and instead add payload to the protobuf (if it's not already there, sorry, I don't recall)? I'd prefer we don't add another option to Cosign, especially given the feedback we've received so far with other flags that we added.

mtrmac:
> Given that we know we want to integrate with the sigstore bundle format defined in https://github.com/sigstore/protobuf-specs/,

I don’t think that’s directly relevant: The existing “signature bundle” format already contains a payload, and this PR changes nothing in the use cases and code paths where a “signature bundle” is used.

It’s the non-bundle commands and option combinations that currently make a deterministic payload assumption; and that’s what this PR is proposing to change.

If the project were moving towards new data structure formats[1], and isn’t interested in keeping feature parity in the non-bundle use cases, fine; I can work on the payload change proposals just as well without this PR. In that case, though, I think it could be useful to at least warn users if they invoke the old non-bundle commands with the deterministic payload assumption, even if we don’t add the new options to make payloads explicit.

Would you be interested in a variant of this PR that adds the warnings, but not the new options?

---

> should we hold off in this PR and instead add payload to the protobuf (if it's not already there, sorry, I don't recall)?

I don’t really feel comfortable reverse-engineering a design just from a protobuf; with that caveat, the DSSE format does seem to have a payload of _some kind_, the `dev.sigstore.common.v1.MessageSignature` clearly doesn’t.

I'd prefer we don't add another option to Cosign, especially given the feedback we've received so far with other flags that we added.

---

[1]  (I’m a bit unhappy about the extra work keeping up with changes to the data representation would cause us…)

znewman01:
Taking some time to sort through this finally, appreciate your patience.

I think that the right thing to do long-term is definitely to fold this into the protobuf Signature bundle format. This is intended to be an opaque blob that gets passed around containing all of the material needed for verification. We might do that by adding a "PayloadAndSignature" `content` type or something.

That said, in the short term this does need to get done. I agree generally that more flags is unpleasant, but we can get rid of it [after we fold the bundles in](https://github.com/sigstore/cosign/issues/2131) (since we have to get rid of a bunch of flags anyway). I'm okay with the PR as-is (and talking to @haydentherapper offline he is too).

znewman01 (approved):

Issue #2730, opened 2023-02-17
cosign is not working with Gitlab self hosted 


pcsgithubid:
I am trying to sign and push images to GitLab self-hosted environment using the below command, but the signing process trying to get project info using  gitlab.com API instead of  GitLab self-hosted API.

 cosign generate-key-pair gitlab://373
Enter password for private key:Enter password for private key again:
Error: could not create "COSIGN_PASSWORD" variable: POST https://gitlab.com/api/v4/projects/373/variables: 401 {message: 401 Unauthorized}
main.go:52: error during command execution: could not create "COSIGN_PASSWORD" variable: POST https://gitlab.com/api/v4/projects/373/variables: 401 {message: 401 Unauthorized}

how to work with self-hosted GitLab environments, please provide info

cpanato:
you need to set the following environment variables and then you can run the cosign command

```
export GITLAB_TOKEN=<YOUR_TOKEN>
export GITLAB_HOST=<YOUR_GITLAB_HOST>

cosign generate-key-pair gitlab://<YOUR_PROJECT_ID>
```

pcsgithubid:
That works. Thanks for the information.

znewman01:
It wouldn't be a terrible idea to suggest this as a fix when we see that error.

paolomainardi:
Hi :) if this one is still available, it seems a very small one to start contributing the project :)

znewman01:
Thanks @paolomainardi ! Let me know if you're running into any trouble or have any questions.

paolomainardi:
Wow, thanks @znewman01, I’ll get back to you with some questions and hopefully a PR very soon :)

paolomainardi:
@znewman01 not very sure about the text and the format I should use and neither if there are tests for these files, asking you for a quick review.

manuschillerdev:
I'd also be interested in a full example for a private gitlab instance.

> you need to set the following environment variables and then you can run the cosign command
> 
> ```
> export GITLAB_TOKEN=<YOUR_TOKEN>
> export GITLAB_HOST=<YOUR_GITLAB_HOST>
> 

`GITLAB_HOST` should be equal to `$CI_SERVER_HOST:$CI_SERVER_PORT`, right?
What would the GITLAB_TOKEN be? Which scopes should be allowed?

As a starting point for discussion: a basic pipeline that builds an image and tries to sign it:

```yaml
variables:
[...]

PR #2816, opened 2023-03-16
Add a warning to remember how to configure a custom Gitlab host

paolomainardi:
Resolves sigstore/cosign#2730

#### Summary

Add a warning to explain how to use a custom Gitlab host.
I've also changed `fmt.Fprintln`with `ui.Infof` where applicable. 

#### Release Note

NONE

#### Documentation

NO

### DIFF
--- a/pkg/cosign/git/gitlab/gitlab.go
+++ b/pkg/cosign/git/gitlab/gitlab.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 
+	"github.com/sigstore/cosign/v2/internal/ui"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
 	"github.com/sigstore/cosign/v2/pkg/cosign/env"
 	"github.com/xanzy/go-gitlab"
@@ -70,7 +71,7 @@ func (g *Gl) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
 		EnvironmentScope: gitlab.String("*"),
 	})
 	if err != nil {
-		fmt.Fprintln(os.Stderr, "WARNING: if you want to use a custom Gitlab host, set the \"GITLAB_HOST\" environment variable.")
+		ui.Infof(ctx, "WARNING: if you want to use a custom Gitlab host, set the \"GITLAB_HOST\" environment variable.")
 		return fmt.Errorf("could not create \"COSIGN_PASSWORD\" variable: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cosign/git/gitlab/gitlab.go
```diff
@@ -70,6 +71,7 @@ func (g *Gl) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
 		EnvironmentScope: gitlab.String("*"),
 	})
 	if err != nil {
+		ui.Infof(ctx, "WARNING: if you want to use a custom Gitlab host, set the \"GITLAB_HOST\" environment variable.")
```
cpanato, line 74: maybe, we can say something like: "if you are using a self-hosted gitlab please set the GITLAB_HOST your server name"

thanks for this PR!

paolomainardi: thanks for suggestion, changed!

pkg/cosign/git/gitlab/gitlab.go
```diff
@@ -70,6 +71,7 @@ func (g *Gl) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
 		EnvironmentScope: gitlab.String("*"),
 	})
 	if err != nil {
+		ui.Infof(ctx, "WARNING: if you are using a self-hosted gitlab please set the \"GITLAB_HOST\" your server name.")
```
hectorj2f, line 74: If you want to prompt a warning, use `Warnf(` instead.

paolomainardi: @hectorj2f thanks for your comment! do you think that this kind of message should be a warning? I have doubts about that, basically, here we are suggesting to the user to configure that variable if the intention is to deploy the secret on a self-hosted GitLab instance.

hectorj2f: I agree the message fits more into `info` here, but we are trying to call the attention of the user. Otherwise they might not pay attention to this. 

Other than that, the `ui.Warnf` function is just a `fmt.Fprintf(w.Stderr, "WARNING: %s\n", msg)` :), that is why I suggested to change it.

hectorj2f (dismissed):

znewman01 (dismissed):

pkg/cosign/git/gitlab/gitlab.go
```diff
@@ -70,6 +71,7 @@ func (g *Gl) PutSecret(ctx context.Context, ref string, pf cosign.PassFunc) erro
 		EnvironmentScope: gitlab.String("*"),
 	})
 	if err != nil {
+		ui.Warnf(ctx, "if you are using a self-hosted gitlab please set the \"GITLAB_HOST\" your server name.")
```
cpanato, line 74: nit: if in capital letter `"If you are using a self-hosted ...`

paolomainardi: done

cpanato (approved):
thank you

PR #2840, opened 2023-03-24
feat: consistent tlog warnings during verification

### LINKED ISSUE #2839: Inconsistent warning for ignoring tlog verification
**Description**

When the flag `--insecure-ignore-tlog` is used in the `verify*` commands, `cosign verify` emits a warning about this being an insecure practice. `verify-attesatation`, `verify-blob`, and `verify-blob-attestation` do not.

**Version**

v2.0.0

lcarva:
Closes #2839

#### Summary

This adds consistent warnings when using the `--insecure-ignore-tlog` flag of the verification commands. The motivation is that if this is deemed an insecure practice, it should emit a warning whenever it is used.

#### Release Note

NONE

#### Documentation

I don't think a documentation change is needed, but I'm happy to make one if deemed useful/necessary.

### DIFF
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -27,6 +27,8 @@ import (
 	"github.com/sigstore/cosign/v2/internal/ui"
 )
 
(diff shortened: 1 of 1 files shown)


znewman01 (approved):
Thank you!

PR #2841, opened 2023-03-24
allow customizing 'go' executable with GOEXE var

dmitris:
#### Summary
Allow customising the `go` executable path with the `GOEXE` variable setting - for example, If one needs to build a debugging version of `go` etc.  `go` is the default value which is used if `GOEXE` is not set.

#### Release Note
NONE

#### Documentation
No documentation changes are needed.


Makefile
```diff
@@ -148,9 +149,9 @@ ko-sget:
 .PHONY: ko-local
 ko-local:
 	$(create_kocache_path)
-	KO_DOCKER_REPO=ko.local LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
+	LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
 	KOCACHE=$(KOCACHE_PATH) ko build --base-import-paths \
-		--tags $(GIT_VERSION) --tags $(GIT_HASH) \
+		--tags $(GIT_VERSION) --tags $(GIT_HASH) --local \
```
cpanato, line 154: Please revert this change, we export the `KO_DOCKER_REPO` in the top of the makefile and then we need to set this here to build locally, the `KO_DOCKER_REPO` have more priority over the `--local`

cpanato (changes requested):
all lgtm except for a small change request

thanks for this PR

Makefile
dmitris: done

dmitris:
> all lgtm except for a small change request
addressed your feedback, please take a look @cpanato

hectorj2f (approved):

cpanato (approved):
Thank you

Issue #2472, opened 2022-11-18
"Chain" for verification is confusingly named


znewman01:
See helpful context: https://github.com/sigstore/cosign/pull/2461#discussion_r1024644134_

The `verify-*` commands have this argument:

> `--certificate-chain string`: path to a list of CA certificates in PEM format which will be needed when building the certificate chain for the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and end with the root certificate
```

We **automatically trust any provided chain** here. That's a reasonable behavior, _if_ you're trying to specify a trust root.

However, when I see a flag named `--certificate-chain` I assume that it's just something I provide to link the cert back up to a default trust root! It feels eminently reasonable to me to download a certificate and a chain alongside a signature, and want to pass that in.

Can we rename this to something like `--ca-trust-root` or `--root-ca` and then just specify in the help text that accepts a certificate or a chain of certs?

znewman01:
There are two possible changes here:

1. Just rename the current `--certificate-chain` flag to something more descriptive (possibly keeping an alias from the old name for backwards compatibility)
2. Shuffle around how users provide certs. So you'd pass something like `--cert-and-intermediates` and `--root-pool` (names very much TBD). This would probably require a transition period.

This means that it's no longer a 2.0 blocker.

Our APIs should also reflect this.

haydentherapper:
Something to learn from is how openssl implements this - You pass a set of trusted certificates (typically root certs) and "untrusted" certs used for chain building (typically intermediates). I'd avoid using "untrusted", but splitting up a chain makes sense. It also lets us easily pass sets of roots rather than a single root in a chain.

We also chatted about how TUF doesn't differentiate between trusted roots and chain building intermediates. This is something we should address.

dmitris:
> It also lets us easily pass sets of roots rather than a single root in a chain.

being able to pass a set of roots (a bundle) would be very convenient! If there are no big "conceptual" hurdles / objections etc, I'd volunteer to help this happen.

dmitris:
> Something to learn from is how openssl implements this - You pass a set of trusted certificates (typically root certs) and "untrusted" certs used for chain building (typically intermediates). I'd avoid using "untrusted", but splitting up a chain makes sense. It also lets us easily pass sets of roots rather than a single root in a chain.

[...]

PR #2845, opened 2023-03-27
verify command: support keyless verification using only a provided certificate chain with non-fulcio roots

dmitris:
This is a continuation of the work started by @nsmith5 in #2631 and uses the code of that PR rebased on the current trunk.  

Fixes https://github.com/sigstore/cosign/issues/2630

Summary

This change adds support for verifying keyless signatures from the specified `--certificate-chain` instead of assuming Fulcio and loading its roots and without requiring to pass the certificate in addition to the certificate chain.

As @haydentherapper noted in the comment below, the keyless verification without the Fulcio certificate roots is already possible but requires passing both certificate-chain and certificate.
This would allow folks to more easily verify signatures from a custom CA (see https://github.com/sigstore/cosign/issues/2630 for additional discussion).

Release Note

Allow verifying signatures from a custom CA without needing leaf certificate
Documentation

Will attach an PR for docs here if folks agree on the parent issue that its worth doing

TODO

The TODO list as this idea is still in discussion etc. 

- [x] Add PR for docs with examples of this kind of verification (done: https://github.com/sigstore/docs/pull/153)

- [x]  Add some end to end tests of this (done: https://github.com/dmitris/cosign-keyless)

### DIFF
--- a/cmd/cosign/cli/verify/verify_test.go
+++ b/cmd/cosign/cli/verify/verify_test.go
@@ -198,3 +198,20 @@ func TestVerifyCertMissingIssuer(t *testing.T) {
 		t.Fatal("verify expected 'need --certificate-oidc-issuer'")
 	}
 }
+
+func TestVerifyKeylessVerification(t *testing.T) {
+	ctx := context.Background()
+	verifyCommand := VerifyCommand{
+		CertRef: "cert.pem",
+		CertVerifyOptions: options.CertVerifyOptions{
+			CertIdentity:         "identity",
+			CertOidcIssuerRegexp: ".*",
+		},
+	}
+
+	err := verifyCommand.Exec(ctx, []string{"foo", "bar", "baz"})
+	t.Logf("INFO: verifyCommand error: %v", err)
+	if err == nil {
+		t.Fatal("verify expected 'need --certificate-oidc-issuer'")
+	}
+}
--- a/cmd/cosign/cli/verify/verify_test.go
+++ b/cmd/cosign/cli/verify/verify_test.go
@@ -198,20 +198,3 @@ func TestVerifyCertMissingIssuer(t *testing.T) {
 		t.Fatal("verify expected 'need --certificate-oidc-issuer'")
 	}
 }
-
-func TestVerifyKeylessVerification(t *testing.T) {
-	ctx := context.Background()
-	verifyCommand := VerifyCommand{
-		CertRef: "cert.pem",
-		CertVerifyOptions: options.CertVerifyOptions{
-			CertIdentity:         "identity",
-			CertOidcIssuerRegexp: ".*",
-		},
-	}
-
-	err := verifyCommand.Exec(ctx, []string{"foo", "bar", "baz"})
-	t.Logf("INFO: verifyCommand error: %v", err)
-	if err == nil {
-		t.Fatal("verify expected 'need --certificate-oidc-issuer'")
-	}
-}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -358,7 +358,7 @@ Critical breaking changes include:
 # v1.12.1
 
 > # Highlights
-> * Pulls Fulcio root and intermediate when `--certificate-chain` is not passed into `verify-blob`. The v1.12.0 release introduced a regression: when `COSIGN_EXPERIMENTAL` was not set, cosign `verify-blob` would check a` --certificate` (without a `--certificate-chain` provided) against the operating system root CA bundle. In this release, Cosign checks the certificate against Fulcio's CA root instead (restoring the earlier behavior).
+> * Pulls Fulcio root and intermediate when `--certificate-chain` is not passed into `verify-blob`. The v1.12.0 release introduced a regression: when `COSIGN_EXPERIMENTAL` was not set, cosign `verify-blob` would check a `--certificate` (without a `--certificate-chain` provided) against the operating system root CA bundle. In this release, Cosign checks the certificate against Fulcio's CA root instead (restoring the earlier behavior).
 
 ## Bug Fixes
 
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -62,7 +62,7 @@ against the transparency log.`,
   cosign verify --cert cosign.crt --cert-chain chain.crt <IMAGE>
 
   # verify image using keyless verification with the given certificate
-  # chain and identity parameters, without Fulcio roots (fro BYO PKI):
+  # chain and identity parameters, without Fulcio roots (for BYO PKI):
   cosign verify --cert-chain chain.crt --certificate-oidc-issuer https://issuer.example.com --certificate-identity foo@example.com <IMAGE>
 
   # verify image with public key provided by URL
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -61,6 +61,10 @@ against the transparency log.`,
   # verify image with local certificate and certificate chain
   cosign verify --cert cosign.crt --cert-chain chain.crt <IMAGE>
 
+  # verify image using keyless verification with the given certificate
+  # chain and identity parameters, without Fulcio roots (fro BYO PKI):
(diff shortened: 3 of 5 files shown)


nsmith5:
Oh lovely! Thanks for picking this up @dmitris

dmitris:
(planning to work on this the weak after Easter - after April 11)

haydentherapper:
cc @znewman01 @woodruffw This relates to the conversation we've been having. If we go through with this, it would assume that the chain in the bundle contains the root. Note that Cosign already does this, with `cosign sign --certificate cert.pem --certificate-chain chain.pem <artifact>`, so this PR is mostly an optimization that would allow you to provide `--bundle` instead of `--certificate` & `--certificate-chain`. We could either say that this is no different than what we're doing currently, so it's fine, or argue that this behavior actually should be removed and we should be explicit about the root.

nit for the title, can we update to reflect this, something like "Support providing a bundle to verify with non-Fulcio roots", because identity-based verification is already suppported

dmitris:
> so this PR is mostly an optimization that would allow you to provide `--bundle` instead of `--certificate` & `--certificate-chain`. 

We are running for an internal PoC a version of cosign with this modification, and I think we do not change the command-line parameters - we call it like this:
```
./cosign verify --insecure-ignore-tlog --insecure-ignore-sct --check-claims=true --certificate-identity-regexp spiffe://<ci.internal.identity> --certificate-oidc-issuer-regexp '.*' --certificate-chain path/to/internal-cert-bundle.pem 
```

@haydentherapper - do you mean that we should introduce the `--bundle` parameter for `verify` and use it instead of the `--certificate-chain` one?

Looking at the commands, I see that `verify-blob` has the `--bundle` parameter but `verify` doesn't. 
Do you know what's the reason for that?  Do you want `--bundle` to replace `--certificate-chain` for the keyless use case for consistency of those two commands?


> 
> nit for the title, can we update to reflect this, something like "Support providing a bundle to verify with non-Fulcio roots", because identity-based verification is already suppported

identity-based verification is, but you cannot verify the image without the Fulcio roots, and I believe this Is the essence of this change (please correct me if I'm wrong!).   I think with this we can fix the issue on the "Bad News" slide of @nsmith5's Seattle talk https://youtu.be/d5vROySf1I8?t=1726 28:45.
So I wonder if we could agree on the title like "verify command: support keyless verification using provided certificate chain with non-fulvio roots"?

haydentherapper:
Sorry @dmitris, I got my wires crossed, I was thinking this was about `verify-blob` and a bundle instead of `verify`. We don't need to add support for `--bundle`, as per your question on Slack.

Your PR title suggestion sounds good. For a small change, can we add "only", "verify command: support keyless verification using only a provided certificate chain with non-fulcio roots", because it is possible to do keyless verification with non-fulcio roots now, you just have to specify the leaf cert with `--certificate` in addition to the chain

woodruffw:
Thanks for pinging me in here!

> We could either say that this is no different than what we're doing currently, so it's fine, or argue that this behavior actually should be removed and we should be explicit about the root.

Making sure I understand: would the changes in this PR mean that including a root in the bundle's chain would be an intended use case? If that's correct, then I'd argue this should be removed entirely rather than expanded upon: including (and trusting) the bundled root is no different from accepting a self-signed signature. 

(It's possible I completely misunderstood, in which case please ignore me!)

haydentherapper:
The current state is that you can specify the certificate chain to verify the provided certificate. The root is not provided out of band - It's been noted on another issue that we need to correct this, I think we'll probably just tackle that when we implement support for the new bundle format.

This proposed change, for only `cosign verify` (containers) lets you specify _only_ the certificate chain without specifying the leaf, because the leaf is in the OCI annotations.`verify-blob` is unaffected because you always have to specify the certificate (or the `--bundle` flag, which is more of an optimization). This is no worse than what we've got right now, but still has the issue of providing the root in-band. 

So the question is, would we rather make the current, suboptimal behavior more uniform, or stop making any changes to the chain until we mandate provided roots out of band?

woodruffw:
Got it, thanks for explaining!

That sounds like a logistical choice for `cosign`, but my outsider's preference would be for unifying the different Sigstore clients on enforcing OOB trust materials 🙂

znewman01:
IMO this all goes back to https://github.com/sigstore/cosign/issues/2472 — the "chain" concept is really confusing, and the `--certificate-chain` flag should be called something like `--trusted-root-chain`.

Given that, I think it's reasonable to make the behavior be (1) configure the trusted certificates, then (2) get the cert from wherever you can find it.

tuminoid:
> Full output of `bash -x test/e2e_test.sh`: https://gist.github.com/dmitris/44b78d3de43c645fa55ca4c5ae324fb3
> 
> [1] https://sigstore.slack.com/archives/C01PZKDL4DP/p1681925607818049

Yeah, this e2e is not really documented, but looking at the GH workflows `.github/workflows/tests.yaml` it seems to require kind cluster and GODEBUG env set up:

```
      - name: setup kind cluster
        run: |
          # Used to test: cosign generate-key-pair k8s://...
          go install sigs.k8s.io/kind@v0.17.0
          kind create cluster

      - name: Run end-to-end tests
        env:
          # See #2091 for the issue describing this temp workaround.
          GODEBUG: x509sha1=1
        run: ./test/e2e_test.sh
```
You also need `ko` installed: https://github.com/ko-build/ko/releases

dmitris:
I have put together a "one-shot" / automated script that goes through the keyless verification with non-fulcio root flow: https://github.com/dmitris/cosign-keyless/blob/initial/run.sh 
https://github.com/dmitris/cosign-keyless/blob/initial/sample.out contains sample output of a test run.

As expected, it fails with the trunk version of `cosign` and works OK using the branch of this PR.  Maybe we could use it as the test evidence to merge? I would follow up with the unit tests afterwards. 
/cc @haydentherapper

test/e2e_test.go
```diff
@@ -176,6 +176,51 @@ func TestSignVerify(t *testing.T) {
 	mustErr(verify(pubKeyPath, imgName, true, map[string]interface{}{"foo": "bar", "baz": "bat"}, ""), t)
 }
 
+func TestSignVerifyKeylessCertChain(t *testing.T) {
```
haydentherapper, line 179: We can remove this, because this test doesn't exercise the newly added code.

cmd/cosign/cli/verify/verify_test.go
```diff
@@ -198,3 +198,20 @@ func TestVerifyCertMissingIssuer(t *testing.T) {
 		t.Fatal("verify expected 'need --certificate-oidc-issuer'")
 	}
 }
+
+func TestVerifyKeylessVerification(t *testing.T) {
```
haydentherapper, line 202: Is this any different than the above?

haydentherapper (commented):
Thanks for trying to get tests working, but I think it's clear that there's no great way to test this out besides a manual test. Given you've shown it works, I'm fine with this change. Can you just clean up the duplicated tests?

test/e2e_test.go
dmitris: removed, sorry - was a placeholder for the future work

cmd/cosign/cli/verify/verify_test.go
dmitris: deleted

dmitris:
@haydentherapper removed the `TestVerifyKeylessVerification` test and also squashed some commits. Thanks for your help! 👋

test/e2e_test.go
haydentherapper, line 179: It looks like this wasn't deleted?

haydentherapper (commented):
@znewman01 Any other thoughts? I'm fine with this since we already do support providing the trust root via --cert-chain, this is just an optimization to also not require --certificate.

znewman01:
> @znewman01 Any other thoughts? I'm fine with this since we already do support providing the trust root via --cert-chain, this is just an optimization to also not require --certificate.

:shipit:

test/e2e_test.go
dmitris: deleted now + squashed 😄

haydentherapper (approved):
Thanks for fixing this!

Issue #2697, opened 2023-02-08
GitLab provider


adanil:
**Description**

Need a gitlab provider to automate the build and signing of artifacts. It can be implemented using the JWT token automatically generated by GitLab CI

znewman01:
See also https://github.com/sigstore/cosign/issues/1254

znewman01:
Does the token have an `aud` claim now?

haydentherapper:
There's an open PR in Fulcio, but we might hold off on merging it until their OIDC issuer stabilizes (they said it's likely to change in the near future).
I don't think we should merge this into Cosign yet until there's support in Fulcio, or else you're just going to get errors on GitLab CI.

haydentherapper:
@wlynch Is this now complete with the above PR submitted?

wlynch:
Yeah I think we can.

There's some room for follow up work once https://gitlab.com/gitlab-org/gitlab/-/issues/390711 lands to request the JWT on demand like we do for github actions, but the current flow works well enough.

Issue #1254, opened 2021-12-27
Add GitLab Support for Ambient Credential Detection


developer-guy:
**Description**

There are some environment variables provided by GitLab to the runner environment which we can use to authenticate.

```shell

CI_JOB_JWT: A RS256 JSON web token to authenticate with third party systems that support JWT authentication, for example HashiCorp’s Vault.

CI_JOB_JWT_V2: alpha: A newly formatted RS256 JSON web token to increase compatibility. Similar to CI_JOB_JWT, except the issuer (iss) claim is changed from gitlab.com to https://gitlab.com, sub has changed from job_id to a string that contains the project path, and an aud claim is added. Format is subject to change.

CI_JOB_TOKEN: A token to authenticate with certain API endpoints. The token is valid as long as the job is running.
```
👉 https://docs.gitlab.com/ee/ci/variables/predefined_variables.html 

cc: @dlorenc @dentrax @orhun

dlorenc:
There's some discussion over here: https://github.com/sigstore/fulcio/issues/243

The tokens are currently missing an audience so we can't use them directly. It sounds like that might be in progress though!

mattmoor:
Yeah, we should NOT do this before the public Fulcio supports gitlab via `--identity-token`, otherwise it'll create annoying failure modes.

marshall007:
If possible, I think using `CI_JOB_TOKEN` is probably the best medium-term approach.

If Fulcio is passed our `CI_JOB_TOKEN` it can make an API call to `GET <gitlab-server-url>/api/v4/job` to tretrieve the job definition for the given token. See docs:

https://docs.gitlab.com/ee/api/jobs.html#get-job-tokens-job

One desirable property of the `CI_JOB_TOKEN` is that we automatically revoke its access as soon as the job is terminated. Its access is also limited to API endpoints that are specifically relevant to the CI job it is created for.

I understand this is still not as ideal as crafting a JWT with a specific `aud` claim, just noting that we have some other mechanisms in play that limit the replay-ability of `CI_JOB_TOKEN`s.

---

I'm pushing on adding support for custom `aud` claims in `CI_JOB_JWT_V2`, but it's going to take some time as it will involve extending our CI job YAML definitions and deprecating some old behavior (both user-facing and potentially breaking changes). Our tentative plan is to formalize `CI_JOB_JWT_V2` and deprecate `CI_JOB_JWT` in v15.0.

[...]

PR #2864, opened 2023-04-04
Add environment variable token provider.

### LINKED ISSUE #2863: Environment variable ID token detection
**Description**

I'd like to propose adding environment variable OIDC token detection for cosign.

Main use case is to support [GitLab CI OIDC tokens](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html) - see https://github.com/sigstore/fulcio/pull/983 for more details.

## Proposal

Today, users can already use arbitrary OIDC tokens by using the `--identity-token` flag.

```sh
cosign sign --identity-token=${TOKEN} ...
```

To make this easier, I want to allow cosign to detect a token from environment variables.
To help avoid collisions with existing token variables, we should use a prefixed value such as `SIGSTORE_ID_TOKEN`.

This could also be used to allow other CI providers to hook in without needing additional changes to cosign (though I'm still in support for allowing custom providers per host).

Example: GitLab

```
build:
  stage: build
  id_tokens:
    SIGSTORE_ID_TOKEN:
      aud: sigstore
  script:
     - cosign sign registry.gitlab.com/my-image
```

I'm happy to implement this.

Previous GitLab related issues:
- https://github.com/sigstore/cosign/issues/2697
- https://github.com/sigstore/cosign/issues/1254

wlynch:
#### Summary

This adds a new token provider that looks for tokens located at `SIGSTORE_ID_TOKEN`.

This is an alternative to `COSIGN_IDENTITY_TOKEN`, which is [implemented as a Viper binding](https://github.com/sigstore/cosign/pull/2346).

`ID_TOKEN` was used instead of `IDENTITY_TOKEN`, since this seems to be consistent with other providers (e.g. `ACTIONS_ID_TOKEN_REQUEST_TOKEN`). If this ends up being confusing we can add support for both variants later if needed.

Fixes #2863, #2697, #1254 

#### Release Note

* Added new token provider for looking up OIDC tokens in the `SIGSTORE_ID_TOKEN` environment variable.

#### Documentation

Should docs be added before this is actually released and available to people? 🤔 
(it would be great if we could base the docs on cosign versions)

### DIFF
--- a/pkg/cosign/env/env.go
+++ b/pkg/cosign/env/env.go
@@ -55,6 +55,7 @@ const (
 	VariableSigstoreCTLogPublicKeyFile Variable = "SIGSTORE_CT_LOG_PUBLIC_KEY_FILE"
 	VariableSigstoreRootFile           Variable = "SIGSTORE_ROOT_FILE"
 	VariableSigstoreRekorPublicKey     Variable = "SIGSTORE_REKOR_PUBLIC_KEY"
+	VariableSigstoreIDToken            Variable = "SIGSTORE_ID_TOKEN" //nolint:gosec
 
 	// Other external environment variables
 	VariableGitHubHost                Variable = "GITHUB_HOST"
@@ -195,6 +196,12 @@ var (
 			Description: "the log level for the Buildkite agent",
 			Expects:     "string with log level, either debug, notice, info, error, warn, fatal (default: notice)",
 			Sensitive:   false,
+			External:    true,
+		},
+		VariableSigstoreIDToken: {
+			Description: "is a OIDC token used to authenticate to Fulcio",
+			Expects:     "string with a OIDC token",
+			Sensitive:   true,
 			External:    true,
 		},
 	}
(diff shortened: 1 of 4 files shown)


haydentherapper:
Why does this solve the GitLab provider issue? Does GitLab let you control which environment variable its ID token is written to, in contrast to the other CI providers which have predefined, fixed env vars?

wlynch:
> Why does this solve the GitLab provider issue? Does GitLab let you control which environment variable its ID token is written to, in contrast to the other CI providers which have predefined, fixed env vars?

Yup! Exactly. So you should be able to do something like this:

```yaml
build:
  stage: build
  id_tokens:
    SIGSTORE_ID_TOKEN:
      aud: sigstore
  script:
     - cosign sign ...
```

haydentherapper:
Sweet, thanks for clarifying! 

Can you also update pkg/providers/all/all.go?

haydentherapper:
After this is in, we'll cut Cosign 2.0.1

hectorj2f (approved):

hectorj2f:
Yes, I'd say we need a section to describe our available environment variables.

hectorj2f (approved):

cpanato (approved):

PR #2876, opened 2023-04-06
cmd: Change error message from KeyParseError to PubKeyParseError for …

eiffel-fl:
Hi.

I was using `cosign verify-blob` and just realized a bad error message was used when using both `--key` and `--cert`.
So, this commit corrects this problem by using the corresponding error message.

Best regards.


cpanato (approved):
thanks

Issue #2002, opened 2022-06-17
copy: introduce new `--sbom-only`, `--att-only` flags


Dentrax:
**Description**

Currently, we only have `--sig-only` flag to copy only image signature. I thought it would be nice to have also `--sbom-only` for copy _SBOM_ and `--sbom-only` for copy _ATT_.

But using both these flags in the same time might be mutually exclusive since flags have `-only` suffix means _do not copy anything but only .._.

UX design:

```bash
# 1. mutually exclusive, throw an error
$ cosign copy --sbom-only --sig-only --att-only <SRC> <DST>

# 2. remove -only suffix to do both operations in the same time
# this would also mean: _also copy the image itself_, but actually does not
$ cosign copy --sbom --sig --att <SRC> <DST>

# 3. if user want to copy sig, sbom and att, have to run the followings
$ cosign copy --sig-only <SRC> <DST>
$ cosign copy --sbom-only <SRC> <DST>
$ cosign copy --att-only <SRC> <DST>

# 4. introduce a new --copy-only flag that accept custom string array
# no need to call cosign twice
$ cosign copy --copy-only=sbom,sign,att,etc <SRC> <DST>

# 5. happiest path
$ cosign copy --sig-only <SRC> <DST>
$ cosign copy --sbom-only <SRC> <DST>
$ cosign copy --att-only <SRC> <DST>
```

PTAL @developer-guy

znewman01:
I strongly prefer (4) (`--copy-only`), and deprecating the `--sig-only` flag; I think that's very clear as a CLI user.

joshuabezaleel:
Hi @Dentrax @znewman01 ! I've just recently gotten interested in the supply chain security and Chainguard with its set of tools is definitely something I came to really appreciate for what you folks contribute to the community, thank you!  I've just scoured on the Chainguard Academy and looking to find my way to contribute to Chainguard's repos which I think would be good to start with either sigstore or cosign. 

Is this repo open to public contribution and would this issue be a good candidate for first-time contributors like me to have a shot on?

Thank you lots in advance! 🙂

znewman01:
Give me a few minutes to give you a detailed answer but just want to point out that, while Chainguard is proud to contribute to Sigstore and Cosign, the projects have been successful because of contributors from a huge number of companies and the support of the OpenSSF :)

joshuabezaleel:
[...]

PR #2901, opened 2023-04-13
feat: Make cosign copy faster

jonjohnsonjr:
#### Summary

I noticed that `cosign copy` was remarkably slow, so I bumped the go-containerregistry dependency to pick up some work I've been doing recently to simplify fixing that slowness.

I tested this by running `cosign copy` twice to copy `gcr.io/projectsigstore/cosign:v1.13.0` to Artifact Registry in us-central1. The second push is slower because it doesn't actually have to copy anything.

We're roughly 3x faster after this change.

|           | First Push | Second Push |
|-----------|------------|-------------|
| Before    | 69s        | 23s         |
| After     | **21s**    | **8s**      |
| Reduction | 70%        | 65%         |

There's more we can do, but this is an easy place to start.

#### Release Note

`cosign copy` was optimized to copy images faster.

### DIFF
--- a/cmd/cosign/cli/copy/copy.go
+++ b/cmd/cosign/cli/copy/copy.go
@@ -72,7 +72,7 @@ func CopyCmd(ctx context.Context, regOpts options.RegistryOptions, srcImg, dstIm
 		}
 		srcDigest := srcRepoRef.Digest(h.String())
 
-		for i, tm := range []tagMap{ociremote.SignatureTag, ociremote.AttestationTag, ociremote.SBOMTag} {
+		copyTag := func(tm tagMap) error {
 			src, err := tm(srcDigest, ociremote.WithRemoteOptions(remoteOpts...))
 			if err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


cmd/cosign/cli/copy/copy.go
```diff
@@ -109,15 +124,7 @@ func descriptorsEqual(a, b *v1.Descriptor) bool {
...
-func copyTagImage(tm tagMap, srcDigest name.Digest, dstRepo name.Repository, overwrite bool, opts ...remote.Option) error {
-	src, err := tm(srcDigest, ociremote.WithRemoteOptions(opts...))
-	if err != nil {
-		return err
-	}
-	return copyImage(src, dstRepo.Tag(src.Identifier()), overwrite, opts...)
-}
-
-func copyImage(src, dest name.Reference, overwrite bool, opts ...remote.Option) error {
+func copyImage(ctx context.Context, pusher *remote.Pusher, src, dest name.Reference, overwrite bool, opts ...remote.Option) error {
```
znewman01, line 127: Nit: seems like this copies more than "images." Should it have a new name?

cmd/cosign/cli/copy/copy.go
```diff
@@ -48,41 +50,49 @@ func CopyCmd(ctx context.Context, regOpts options.RegistryOptions, srcImg, dstIm
...
-		}
+			dst := dstRepoRef.Tag(src.Identifier())
+			g.Go(func() error {
+				return copyImage(ctx, pusher, src, dst, force, remoteOpts...)
+			})
 
-		// Copy SBOMs
-		if err := copyTagImage(ociremote.SBOMTag, srcDigest, dstRepoRef, force, remoteOpts...); err != nil {
-			return err
+			if i == 0 && sigOnly {
```
znewman01, line 86: Nit: I don't love that this depends on the order in the `tagMap`.

My preference might be to have a small helper function like:

```go
func whichTagsToCopyFromArgs(sigOnly bool) []tagMap {
  if sigOnly {
    return []tagMap{ociremote.SignatureTag}
  }
  return []tagMap{ociremote.SignatureTag, ociremote.AttestationTag, ociremote.SBOMTag}
}
```

This is extremely forwards-compatible with https://github.com/sigstore/cosign/issues/2002 and IMO properly separates parsing args from "what does this command do?"

znewman01 (approved):
\m/

Left some nitpicks, feel free to leave as TODOs or ignore entirely!

cmd/cosign/cli/copy/copy.go
jonjohnsonjr: Not quite as simple as your helper (we avoid pushing the entity itself as well), but added something like that.

cmd/cosign/cli/copy/copy.go
jonjohnsonjr: I propose `copy` :)

znewman01 (approved):
LGTM thanks!

znewman01 (approved):

PR #2920, opened 2023-04-24
Add zot as a supported registry

rchincha:


rchincha:
cosign is included in zot's CI/CD tests. Pls see:
https://github.com/project-zot/zot/blob/main/test/blackbox/annotations.bats#L99

znewman01 (approved):

PR #2959, opened 2023-05-07
Feature: Allow cosign to sign digests before they are uploaded.

mattmoor:
:gift: This feature allows `cosign` to sign a digest that doesn't exist yet, to support scenarios such as signing an image prior to pushing it.

This adapts the ideas from the two prior approaches, which were closed by stale-bot.

Fixes: https://github.com/sigstore/cosign/issues/1905

/kind feature

#### Release Note

 * `cosign` can now be used to sign digests before the image has been pushed to a registry (this only supports shallow signing of digests, so tag-based signing and recursive signing are not supported because both need the image structure available)

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -214,10 +214,9 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		return err
 	}
 
-	se, err := ociremote.SignedEntity(digest, ociremoteOpts...)
-	if err != nil {
-		return err
-	}
+	// We don't actually need to access the remote entity to attach things to it
+	// so we use a placeholder here.
+	se := ociremote.SignedUnknown(digest)
 
 	signOpts := []mutate.SignOption{
 		mutate.WithDupeDetector(dd),
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -179,7 +179,9 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 
 		if digest, ok := ref.(name.Digest); ok && !signOpts.Recursive {
 			se, err := ociremote.SignedEntity(ref, opts...)
-			if err != nil {
+			if err == ociremote.ErrEntityNotFound {
+				se = ociremote.SignedUnknown(digest)
+			} else if err != nil {
 				return fmt.Errorf("accessing image: %w", err)
 			}
 			err = signDigest(ctx, digest, staticPayload, ko, signOpts, annotations, dd, sv, se)
--- a/pkg/oci/remote/unknown.go
+++ b/pkg/oci/remote/unknown.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 The Sigstore Authors.
+// Copyright 2023 The Sigstore Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/oci/remote/unknown.go
+++ b/pkg/oci/remote/unknown.go
@@ -1,0 +1,60 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package remote
+
+import (
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/sigstore/cosign/v2/pkg/oci"
+)
+
+// SignedUnknown provides access to signed metadata without directly accessing
+// the underlying entity.  This can be used to access signature metadata for
+// digests that have not been published (yet).
+func SignedUnknown(digest name.Digest, options ...Option) oci.SignedEntity {
+	o := makeOptions(digest.Context(), options...)
+	return &unknown{
+		digest: digest,
+		opt:    o,
+	}
+}
+
+type unknown struct {
+	digest name.Digest
+	opt    *options
+}
+
+var _ oci.SignedEntity = (*unknown)(nil)
+
+// Digest implements digestable
+func (i *unknown) Digest() (v1.Hash, error) {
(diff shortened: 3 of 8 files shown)


hectorj2f (dismissed):
lgtm

mattmoor:
I changed it to a hardcoded digest since the `macos` leg doesn't have `sha256sum`.

mattmoor:
I'm going to fix `cosign attest` too while I'm in here, since it has the same problem.

mattmoor:
Ok, so the CI failure is a legit failure, and I'm glad it caught this because there's a subtle bug with this approach!

Essentially the `ociempty` approach is subtly flawed in that it doesn't allow the signing/attesting logic to access pre-existing signatures and attestations, so `sign` and `attest` always overwrite the existing signatures instead of augmenting them.

I think this will require some changes more integrated into `ociremote.SignedEntity` itself, which needs some thought.

mattmoor:
Ok, this no longer uses `ociempty` and I introduced an `ociremote.SignedUnknown` that we can explicitly fallback on (e.g. when we see the new `ErrEntityNotFound`) to support these cases.

It felt wrong to automatically do this in `SignedEntity` since it would break existing walks that assume a SignedEntity must be one of `oci.Signed{Entity,Image}`, but I actually like this a lot more than the prior approach because this is `digestable` (we know the digest), but it doesn't pretend to be a `v1.Image` (or index).

mattmoor:
Blah, it still needs some work to pass tests.

mattmoor:
This is RFAL!

hectorj2f (dismissed):

test/e2e_test_secrets.sh
```diff
@@ -73,6 +74,21 @@ do
...
+# We don't just run this here because the macos leg doesn't
+# have sha256sum
+./cosign sign --key ${signing_key} "$img3@sha256:17b14220441083f55dfa21e1deb3720457d3c2d571219801d629b43c53b99627"
+PREDICATE_FILE=$(mktemp)
+cat > "${PREDICATE_FILE}" <<EOF
+{
+  "foo": "bar"
+}
+EOF
+./cosign attest --key ${signing_key} --type custom --predicate "${PREDICATE_FILE}" "$img3@sha256:17b14220441083f55dfa21e1deb3720457d3c2d571219801d629b43c53b99627"
```
imjasonh, line 90: Can you attest something else and check that both are present, to guard against the overwriting behavior you discovered?

mattmoor: There's already a Go test that checks exactly that (which is what caught the bug!), and I'm not sure doing it for the "unknown" case (or in bash) adds much value 😅

pkg/oci/remote/unknown.go
```diff
@@ -0,0 +1,60 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
jonjohnsonjr, line 2:
```suggestion
// Copyright 2023 The Sigstore Authors.
```

pkg/oci/remote/unknown_test.go
```diff
@@ -0,0 +1,101 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
jonjohnsonjr, line 2:
```suggestion
// Copyright 2023 The Sigstore Authors.
```

cmd/cosign/cli/sign/sign.go
```diff
@@ -179,7 +179,9 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 
 		if digest, ok := ref.(name.Digest); ok && !signOpts.Recursive {
 			se, err := ociremote.SignedEntity(ref, opts...)
-			if err != nil {
+			if err == ociremote.ErrEntityNotFound {
+				se = ociremote.SignedUnknown(digest)
```
jonjohnsonjr, line 183: Are you sure we always want to do this on 404?

cmd/cosign/cli/attest/attest.go
```diff
@@ -214,10 +214,9 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		return err
 	}
 
-	se, err := ociremote.SignedEntity(digest, ociremoteOpts...)
-	if err != nil {
-		return err
-	}
+	// We don't actually need to access the remote entity to attach things to it
+	// so we use a placeholder here.
+	se := ociremote.SignedUnknown(digest)
```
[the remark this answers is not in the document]
mattmoor: I worry that Digest isn't clear enough, since it is effectively what `SignedEntity` is (especially after my PR to add `Digest` to `oci.SignedEntity`).

That said, I don't feel particularly strongly about the naming here `:bikeshed:`

cmd/cosign/cli/sign/sign.go
mattmoor: I think yes, but do you have an alternative in mind?

imjasonh:
Ping. I think this looks good (I have no power to approve), and I'm not sure any of the remaining comments are blocking. I don't want to lose momentum on this long-standing feature request. 🤞

cpanato (approved):
This lgtm
Will wait for @jonjohnsonjr because made some comments here

jonjohnsonjr:
This is not great because OCI referrers stuff expects a whole descriptor (digest+size+mediaType). It would be better if this took a Descriptor, assuming we want to adopt the OCI referrers stuff down the road.

imjasonh:
> This is not great because OCI referrers stuff expects a whole descriptor (digest+size+mediaType). It would be better if this took a Descriptor, assuming we want to adopt the OCI referrers stuff down the road.

Good point. However, in that case, I think we can just set the `subject` without a size and maybe with some default mediaType, to denote "I'm attaching to something I don't know anything else about".

```
"subject": {
  "digest": "sha256:abcdef...",
  "size": 0,
  "mediaType": "application/octet-stream"
}
```

We can either establish this convention and use it, or try to specify this better in OCI image-spec. Supporting this case has been a goal of OCI referrers, and if implementers like cosign don't know how to achieve it then OCI should help them. Or if they want to solve it with this convention, that works too, we just need registries not to validate that the size+mediaType matches reality.

mattmoor:
> This is not great because OCI referrers stuff expects a whole descriptor (digest+size+mediaType). It would be better if this took a Descriptor, assuming we want to adopt the OCI referrers stuff down the road.

Digests need to be fully self-describing, the absence of this property was (as you know) a CVE.

jonjohnsonjr:
> Digests need to be fully self-describing, the absence of this property was (as you know) a CVE.

Sure, but it's nice to know the size prior to decoding the payload, which you can do once you have an entrypoint into the dag. If you want to go against the grain, that's fine, but I'd file an issue with OCI.

imjasonh:
> Ping. I think this looks good (I have no power to approve), and I'm not sure any of the remaining comments are blocking. I don't want to lose momentum on this long-standing feature request. 🤞

priyawadhwa:
I'll go ahead and merge since it seems like next steps are to file an issue with OCI (sorry if I'm reading this thread wrong!).

PR #2960, opened 2023-05-08
Cleanup: Add `Digest` to the `SignedEntity` interface.

mattmoor:
:broom: We have worked around the absence of this in the core interface (while implemented in both `Signed{Entity,Image}`) a variety of places.

I will start cleaning those up after confirming adding it doesn't break anything in-tree.

/kind cleanup

#### Release Note

 * `oci.SignedEntity` now requires a `Digest` method (which was already largely assumed because of the two implementations.


cpanato (approved):
Thank you looks great to me

hectorj2f (approved):

PR #2976, opened 2023-05-13
chore(refactor) refactors error handling with exit codes

ChrisJBurns:
This PR refactors the way we throw errors and map them to exit codes.

Previously we had a lookup table `exitCodeLookup` that used `errorType`'s in order to lookup the correct exit code mapping to the error. This worked fine, but we found that it was slightly hacky in it's working as difficult to work with when we wanted to customise error types.

Before, to create a new error type and exit code mapping, you needed to create the `errorType`, `errorMessage`, make the explicit throw using `VerificationError`, add the lookup entry for it with the associated exit code. This was difficult to work with because we found that we had to force/shove information into the `VerificationError` (now `VerificationFailure` and is deprecated) type which become awkward and hacky. The new code hasn't change a whole lot but it is more flexible when it comes to the error types. If a user wants to add a new error type, they add the error type and associated `Error()` function, throw it using the new `ThrowError` and then add the lookup for it by using the common pattern `errors.As()`. This allows for greater flexibility in what data the errors hold and what you want to return as part of the associated `Error()` function. 

There was a difficult decision that we had to make when dealing with the trade-offs in the design of the code for exit code functionality. This was namely between:
* 1) doing the "best practice" thing and making `cosign/pkg` clueless to the exit codes which were a concern of `cosign/cmd` in combination with the awkward/inconvenient method of having to keep a map of known errors (throw from `cosign/pkg` and their associated error codes (held in `cosign/cmd`)
* 2) keeping code footprint small and doing the exit code mapping in `cosign/pkg` and throwing them to `cosign/cmd` to return back to the user. 

Although option 2 was less code involved, we did have to accept the leaking of concerns between layers and also the exit code automatic documentation would have been a bit trickier to make work.

We opt'd to go with option 1 as, although there is more code all around, we keep concerns separated.

### DIFF
--- a/cmd/cosign/errors/error_wrap_test.go
+++ b/cmd/cosign/errors/error_wrap_test.go
@@ -18,23 +18,7 @@ package errors
 import (
 	"errors"
 	"testing"
-
-	verificationError "github.com/sigstore/cosign/v2/pkg/cosign"
 )
-
-func TestWrapWithVerificationError(t *testing.T) {
-	ve := &verificationError.VerificationError{}
-	ve.SetErrorType(verificationError.ErrNoMatchingSignaturesType)
-	err := WrapError(ve)
-
-	var cosignError *CosignError
-	if errors.As(err, &cosignError) {
-		if cosignError.ExitCode() != NoMatchingSignature {
-			t.Fatalf("verification error unsuccessfully wrapped")
-		}
-		t.Logf("verification error successfully wrapped and exit code returned")
-	}
-}
 
 func TestWrapWithGenericCosignError(t *testing.T) {
 	errorText := "i am a generic cosign error"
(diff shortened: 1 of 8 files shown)


ChrisJBurns:
@znewman01 As spoke about on Slack. Give it a look whenever you got time :)

hectorj2f (dismissed):

znewman01 (changes requested):
There are a few lint issues but otherwise LGTM!

Sorry this took so long, thanks for your patience and your contribution 😄

ChrisJBurns:
@znewman01 No problems! Only now had time to fix! Checks seemed to have passed nicely as well :+1:

ChrisJBurns:
@cpanato Have fixed the lint errors :+1: are you happy to approve this for merge?

cpanato (approved):
thanks

PR #2994, opened 2023-05-18
Feature: adds rekor support for cosign attach command

### LINKED ISSUE #2904: Support attaching a Rekor bundle to a container
**Description**

`cosign attach` supports attaching signatures, but not Rekor bundles. Should be straightforward to add support.

mritunjaysharma394:
#### Summary

This PR adds support for Rekor bundles to be attached in containers. Currently `cosign attach` supports attaching signatures, but not Rekor bundles. 

Fixes #2904  

/kind feature

#### Release Note

`cosign attach` got added with the support of attaching Rekor Bundles to the container. 

#### Documentation

Yes, it requires an update. Updated docs here with `make docgen` for now.

### DIFF
--- a/cmd/cosign/cli/attach/rekor.go
+++ b/cmd/cosign/cli/attach/rekor.go
@@ -1,0 +1,152 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attach
+
+import (
+	"context"
+	"crypto/sha256"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
+	"github.com/sigstore/cosign/v2/cmd/cosign/cli/rekor"
+	"github.com/sigstore/cosign/v2/cmd/cosign/cli/sign"
+	"github.com/sigstore/cosign/v2/pkg/cosign"
+	"github.com/sigstore/cosign/v2/pkg/cosign/bundle"
+	"github.com/sigstore/cosign/v2/pkg/oci/mutate"
+	ociremote "github.com/sigstore/cosign/v2/pkg/oci/remote"
+	"github.com/sigstore/cosign/v2/pkg/oci/static"
+	"github.com/sigstore/rekor/pkg/generated/client"
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+type tlogUploadFn func(*client.Rekor, []byte) (*models.LogEntryAnon, error)
+
+func uploadToTlog(ctx context.Context, sv *sign.SignerVerifier, rekorURL string, upload tlogUploadFn) (*bundle.RekorBundle, error) {
+	rekorBytes, err := sv.Bytes(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	rekorClient, err := rekor.NewClient(rekorURL)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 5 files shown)


cmd/cosign/cli/attach/rekor.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+
+	sv, err := sign.SignerFromKeyOpts(ctx, certRef, certChainRef, options.KeyOpts{})
+	if err != nil {
+		return fmt.Errorf("getting signer: %w", err)
+	}
+	defer sv.Close()
+
+	bundle, err := uploadToTlog(ctx, sv, rekorURL, func(r *client.Rekor, b []byte) (*models.LogEntryAnon, error) {
```
haydentherapper, line 123: A bundle should already exist, correct? This command, like attach signature, should take an existing bundle and attach it to the container metadata.

haydentherapper (changes requested):

haydentherapper:
I would take a look at https://github.com/sigstore/cosign/blob/92df24c7d548560c8d70a63190abd05e6b84abe1/cmd/cosign/cli/attach/sig.go#L34 for inspiration - the bundle should be provided as part of the command.

cmd/cosign/cli/attach/rekor.go
mritunjaysharma394: Thanks for the review, Yes even I felt the same initially but I was skeptical and had little less idea being new to the code base on how we will get rekor Bundle struct back? What should the user pass as a flag? Should rekor-url be passed (like it is passed with attest command?) or should we straight up use `sig.Bundle()`to see if it exists or not and if it does then write/mutate signature using withBundle() opts ?

mritunjaysharma394:
Thanks a lot @haydentherapper, I tried to took my inspiration from there only but I was confused on how we will take this bundle? As []bytes (would be difficult for user to pass directly if they want to) or as struct or as url? Or we don't need to take it at all and just use `sig.Bundle()` or use our `certRef`, `certChainRef` for it?

cmd/cosign/cli/attach/rekor.go
mritunjaysharma394: A rekor bundle is usually in the format: 
```json
{"Bundle":{"SignedEntryTimestamp":"MEUCIG...yoIY=","Payload":{"body":"...","integratedTime":1643917737,"logIndex":1,"logID":"4d2e4...97291"}
```

My only doubt is how will we allow the user to pass it to be attached? Or it hasn't has to be passed at all? And we can leverage `sig.Bundle()` because it already exists but was not attached in containers manifest (sounds contradicting to me as I am unsure about the passing thing)

haydentherapper: I would accept a path to a file with the bundle. `sig.Bundle` is where it's stored, and if it's set, it will be attached to the container.

mritunjaysharma394: Got it so we will read that file and unmarshal it in the form of `*bundle.RekorBundle` and do something like this, right?:
```golang
	b, err := rekorBytes(rekorRef)
	if err != nil {
		return err
	}

	remoteOpts, err := regOpts.ClientOpts(ctx)
	if err != nil {
		return err
	}

	dstRef, err := ociremote.RekorTag(ref, remoteOpts...) //will have to create new func for RekorTag in ociremote 

	if err != nil {
		return err
	}

        //rough pseudo code only to tell the logic
	var bundle *bundle.RekorBundle
	json.Unmarshal(b, bundle)
	img, err := static.NewFile(b, static.WithBundle(bundle))
	if err != nil {
		return err
	}
```

haydentherapper:
@mritunjaysharma394 Sorry for the delayed response. Are you interested in completing this PR?

See https://github.com/sigstore/cosign/pull/3001/files for an example. You should already have an existing bundle on disk (for example, output by https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign_blob.go#L123-L139), and simply call `mutate.Signature` `WithBundle` (https://github.com/sigstore/cosign/blob/731adeb0ad7ce08e86f38417e7f70833db1bee15/pkg/oci/mutate/options.go#L81) after reading the bundle from disk. And we don't need another command, we can reuse `cosign attach signature` like in PR 3001.

haydentherapper:
One thing to note - "bundle" is not the same as "rekor bundle" or "rekor response" confusingly. A bundle will include the signature, certificate, and rekor response. I would recommend we implement a way to output only the rekor response (see https://github.com/sigstore/cosign/issues/3110), which is JSON marshalled https://github.com/sigstore/cosign/blob/main/pkg/cosign/bundle/rekor.go#L21, and attach that.

PR #3000, opened 2023-05-22
Feature: adds '--allow-insecure-registry' for cosign load

### LINKED ISSUE #2986: cosign load '--allow-insecure-registry'
**Description**
There is a need for being able to 'load' an image to a registry that is had either NO tls cert, or a self signed one. Having the ability to tell cosign to ignore the cert, or use http instead is important. 

Use case. Tactical edge deployments for the Government. ;)

mritunjaysharma394:
#### Summary
According to the issue #2986, we need in certain cases like that of "Tactical edge deployments for the Government",  to 'load' an image to a registry that is had either NO tls cert, for the ability to tell cosign to ignore the cert, or use http instead. 
This PR fixes #2986 

#### Release Note

`cosign load` was added with `--allow-insecure-registry` to disable TLS verification when interacting with insecure (e.g. self-signed) container registries

#### Documentation
Yes, added it in the github docs

### DIFF
--- a/cmd/cosign/cli/load.go
+++ b/cmd/cosign/cli/load.go
@@ -17,16 +17,14 @@ package cli
 
 import (
 	"context"
-	"crypto/tls"
 	"fmt"
-	"net/http"
 
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
+
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v2/pkg/oci/layout"
+	"github.com/sigstore/cosign/v2/pkg/oci/remote"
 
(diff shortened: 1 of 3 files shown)


znewman01 (changes requested):
Can we use the RegistryOpts directly? https://github.com/sigstore/cosign/blob/f21081a1820939cddc41c1c119632a6c489d57fd/cmd/cosign/cli/options/registry.go#LL41C1-L41C1

This means that you don't have to duplicate the code to create the `Opts`.

mritunjaysharma394:
Thanks @znewman01 , just fixed that, I (wrongly) thought that we add opts for each command separately :sweat_smile: but yes we should try to be as DRY as we can be, thanks for the review! Let me know if it needs further improvements

znewman01 (approved):

PR #3001, opened 2023-05-23
Added support for attaching Time stamp authority Response in attach command

Mukuls77:
#### Summary
The PR provide the support to attach Timestamp authority response for signed signature time stamp in attach command.
using this functionality now we can attach a TSR response for the generated signature and attach that along with the associated certificates and certificate chain. This PR will help specifically in private CA cases where now Private CA can issue short lived certificates and attach the Time stamp of the signature. This will help in checking the validity of the associated certificates using the signature time stamp 
#### Release Note
Following files have been modified for the PR.
cmd/cosign/cli/attach.go
cmd/cosign/cli/attach/sig.go
cmd/cosign/cli/options/attach.go
doc/cosign_attach_signature.md
test/e2e_test.go
test/e2e_test_attach.sh
test/testdata/test_attach_freetsacacert.pem
Add a release note for each of the following conditions:

#### Documentation
This PR introduce a new parameter that can be used in the Attach command to pass the path of the Time stamp Authority response.
The updated command help is as below

$ ./cosign attach signature --help
Attach signatures to the supplied container image

Usage:
  cosign attach signature [flags]

Examples:
  cosign attach signature <image uri>

Flags:
      --allow-http-registry                                                                      whether to allow using HTTP protocol while connecting to registries. Don't use this for anything but testing
      --allow-insecure-registry                                                                  whether to allow insecure connections to registries (e.g., with expired or self-signed TLS certificates). Don't use this for anything but testing
      --attachment-tag-prefix [AttachmentTagPrefix]sha256-[TargetImageDigest].[AttachmentName]   optional custom prefix to use for attached image tags. Attachment images are tagged as: [AttachmentTagPrefix]sha256-[TargetImageDigest].[AttachmentName]
      --certificate string                                                                       path to the X.509 certificate in PEM format to include in the OCI Signature
      --certificate-chain string                                                                 path to a list of CA X.509 certificates in PEM format which will be needed when building the certificate chain for the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and end with the root certificate. Included in the OCI Signature
  -h, --help                                                                                     help for signature
      --k8s-keychain                                                                             whether to use the kubernetes keychain instead of the default keychain (supports workload identity).
      --payload string                                                                           path to the payload covered by the signature
      --signature string                                                                         path to the signature, or {-} for stdin
      --timeStampedSignatureResponse string                                                      path to the Time Stamped Signature Response from RFC3161 compliant TSA

Global Flags:
      --output-file string   log output to a file
  -t, --timeout duration     timeout for commands (default 3m0s)
  -d, --verbose              log debug output

#### Testing
Make , lint , Docgen test were done. also e2e test related to attach command were done. attached are the logs
[Build_Lint_Docgen_test_logs.txt](https://github.com/sigstore/cosign/files/11543725/Build_Lint_Docgen_test_logs.txt)

[e2e_attach_test_logs.txt](https://github.com/sigstore/cosign/files/11543728/e2e_attach_test_logs.txt)


test/e2e_test_attach.sh
```diff
@@ -46,6 +47,10 @@ IMAGE_URI_DIGEST=$IMAGE_URI@$SRC_DIGEST
 
 ## Sign with Leafcert Private Key
 openssl dgst -sha256 -sign ./private_key -out payload.sig payload.json
+## Generate TSR for the signature
+openssl ts -query -data payload.sig -sha256 -cert -out payload.tsq
+curl -H "Content-Type: application/timestamp-query" --data-binary '@payload.tsq' https://freetsa.org/tsr > payload.tsr
```
haydentherapper, line 52: Can we use the libraries in the sigstore/timestamp-authority repo to generate a timestamp (see unit tests, there should be some examples), or stand up an instance of the tsa for an e2e test (also there should be examples of such). We shouldn’t call out to a third party service as part of our tests.

cmd/cosign/cli/options/attach.go
```diff
@@ -54,6 +55,8 @@ func (o *AttachSignatureOptions) AddFlags(cmd *cobra.Command) {
 			"when building the certificate chain for the signing certificate. "+
 			"Must start with the parent intermediate CA certificate of the "+
 			"signing certificate and end with the root certificate. Included in the OCI Signature")
+	cmd.Flags().StringVar(&o.TimeStampedSig, "timeStampedSignatureResponse", "",
```
haydentherapper, line 58: Can we use an acronym “—tsr” for the flag name instead?

haydentherapper (commented):
LGTM on the feature, just a comment on flag name and testing.

cmd/cosign/cli/options/attach.go
Mukuls77: Agree i will change the flag name from --timeStampedSignatureResponse  to --tsr

test/e2e_test_attach.sh
Mukuls77: I tried to check the instance in the existing e2e test cases where we have used a separate instance of the tsa, but was not able to find any example. in the unit test i could see there are many instance of test cases covering the TSA testing and verification. so i think adding a new unit test case for attach case will not add up much functionality as in unit test case we directly modify the signature structure to include cert, cert-chain and tsa.
i was thinking if it is ok with you i can cerate a static TSR using freeTSA and use that instead of calling freeTSA in my test case. the certificates used  have a expiry time of 5 years so i can write a note in Readme.txt which suggest how to generate a fresh TSR when the existing certificates expires. please suggest what if this approach is ok

haydentherapper: This test has an example of spinning up a local server - https://github.com/sigstore/cosign/blob/main/test/e2e_test.go#L724

Mukuls77: Hi Hayden i have created a new test cases for using TSR in attach command in e2e_test.go file also i have reverted the changes in e2e_test_attach.sh file which were using freetsa for getting TSR. kindly review the changes,

malancas (commented):
This LGTM now that the test uses sigstore/timestamp-authority libraries but I'll wait for @haydentherapper to give this another look as well.

Mukuls77:
Thanks @malancas for the review @haydentherapper  can you pls also review the PR changes

haydentherapper (commented):
Looks great! Just two small comments, thanks for updating the tests.

Mukuls77:
Thanks @haydentherapper for the review i have done the required changes pls check

haydentherapper (approved):

PR #3020, opened 2023-05-30
bump sigstore/sigstore dep to @v1.6.5

imjasonh:
This picks up https://github.com/sigstore/sigstore/pull/1115 and https://github.com/sigstore/sigstore/pull/1178 to consume split Go modules for each of the KMS providers.

This has no direct effect on the cosign CLI since it [depends on these packages-now-modules directly](https://github.com/sigstore/cosign/blob/ec4c01ee046ecd3c5ec7c329a668f4f57d82f112/cmd/cosign/main.go#L29-L33), but consumers of cosign packages that aren't `cmd/cosign` should be able to drop these dependencies.

cc @haydentherapper


haydentherapper:
Do we want to cut a new release of sigstore/sigstore first so we can use the versioned modules? @cpanato or I can cut a new release

imjasonh:
That would be fine too. I was opening this mainly to get an idea if there was anything blocking it, and it doesn't look like it.

Maybe let's wait to see CI pass on this PR, then if you cut a s/s release I'll update this to use that, how's that sound?

haydentherapper:
@imjasonh looks good, ill go ahead and cut the release now

haydentherapper:
tags look good! https://github.com/sigstore/sigstore/tags

haydentherapper (approved):

cpanato (approved):

Issue #2937, opened 2023-04-27
`cosign attest` requires `--predicate` to be passed explicitly


znewman01:
From Slack:

@anderssonw writes:

> https://github.com/sigstore/cosign/blob/17cc13812d8a7bfbb068ffc10fdbbdb3a9416f11/cmd/cosign/cli/attest/attest.go#L135
>
> Is this working as intended? It seems an empty predicate path returns the following error:
>
> ```
> Error: signing ghcr.io/ORG/IMAGE@sha256:DIGEST: open : no such file or directory
> main.go:74: error during command execution: signing ghcr.io/ORG/IMAGE@sha256:DIGEST: open : no such file or directory
> ```
>
> This is running a keyless attest, btw:
`cosign attest --identity-token $IDENTITY_TOKEN ghcr.io/ORG/IMAGE@sha256:DIGEST`

> piping like this: `echo 'foo' | cosign attest --identity-token $IDENTITY_TOKEN ghcr.io/ORG/IMAGE@sha256:DIGEST`
does not change the error
>
> if a predicate file is required, why is it not a required flag?
>
> if i explicitly set `--predicate -` using stdin worked fine

@mattmoor:

> I was looking at this code the other day and iirc the absence of the flag causes it to use os.Stdin, so not sure why the behavior doesn’t change without a bit of digging
> 
> I wonder if there is a bad default value
>
> The default should probably be - in that case
> The flags are mostly defined in an options package, so I’d check for attest options there

lcarva:
> The default should probably be - in that case

I believe this is the right fix.

A couple of caveats:

1. If the default value is `-`, then not providing standard input redirection causes the command to halt with this message:     `Using payload from: standard input`. We can probably detect when interactive mode is being used to provide a meaningful error message. Similar to https://github.com/sigstore/cosign/blob/83c08ce4f35116da7c74350a67d5740ca5e28ca5/cmd/cosign/cli/fulcio/fulcio.go#L148-L149
2. When using the syntax `echo 'foo' | cosign attest ...`, either `--yes` or `--no-upload` must also be provided. Otherwise, the error `user declined the prompt` is raised. This is probably ok.

lcarva:
Alternatively, we could mark the flag as required (must not be an empty string) and let the user decide which behavior better suits their needs. This actually seems like the least surprising option.

PR #3033, opened 2023-06-02
Make predicate a required flag in attest commands

lcarva:
#### Summary

The default value for the `predicate` parameter for the `attest` and `attest-blob` commands is an empty string. However, an empty string is not a valid default value for this parameter resulting in an error.

In https://github.com/sigstore/cosign/issues/2937, it was suggested that the default value for this parameter should be `-` (read from stdin). But this results in somewhat odd behavior when stdin is not used.

This commit retains the default value of `""`, but adds the following guardrails:
1. Makes the `--predicate` parameter required.
2. Performs an early check for empty string, and reports an error accordingly.

I also ported the new behavior to the `attest-blob` command.

#### Release Note

* `attest-blob` can now read predicate payload from standard input.

#### Documentation

Updated the CLI docs accordingly.

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -55,7 +55,10 @@ func Attest() *cobra.Command {
   cosign attest --predicate <FILE> --type <TYPE> --key cosign.key --cert cosign.crt --cert-chain chain.crt <IMAGE>
 
   # attach an attestation to a container image which does not fully support OCI media types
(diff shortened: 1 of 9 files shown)


znewman01 (approved):

znewman01:
lgtm modulo docs

lcarva:
@znewman01, fixed the docs and a minor issue on the tests :pray:

znewman01 (approved):

PR #3049, opened 2023-06-12
Allow overriding remote options

jonjohnsonjr:
RegistryOptions.GetRegistryClientOpts is used in a bunch of places, and it always instantiates a new set of options. This is ~fine for callers that only use the options once (e.g. the cosign CLI), but is unfortunate for callers that would like to reuse those options.

It's useful to reuse these options because they represent a shared context that eliminates a lot of duplicate work across calls (e.g. authenticating with the registry and doing existence checks).


hectorj2f (approved):
lgtm

cpanato (approved):
thanks

PR #3050, opened 2023-06-12
Fix pkg/cosign/errors

jonjohnsonjr:
1. Remove ThrowError

This function is the identity function, so it doesn't appear to actually
do anything? Tests pass after deleting it.

2. Unwrap

Errors should implement Unwrap when they wrap another error so that
errors.Is/As will work with it.

3. Don't break policy-controller

This adds back some public symbols that were removed that prevent
policy-controller from bumping its cosign dependency. They are marked
deprecated.

https://github.com/sigstore/policy-controller/blob/b59a2bf7e4f95e2f9126ce8b4d9dd5bcfbd11c9f/pkg/webhook/validator.go#L518
https://github.com/sigstore/policy-controller/blob/b59a2bf7e4f95e2f9126ce8b4d9dd5bcfbd11c9f/pkg/webhook/validator.go#L836
https://github.com/sigstore/policy-controller/blob/b59a2bf7e4f95e2f9126ce8b4d9dd5bcfbd11c9f/pkg/webhook/validator.go#L912

### DIFF
--- a/pkg/cosign/errors.go
+++ b/pkg/cosign/errors.go
@@ -14,26 +14,21 @@
 
 package cosign
 
+import "fmt"
+
 // VerificationFailure is the type of Go error that is used by cosign to surface
 // errors actually related to verification (vs. transient, misconfiguration,
 // transport, or authentication related issues).
-// It is now marked as deprecated and will be removed in favour of defined
-// error types with use of the ThrowError function.
 type VerificationFailure struct {
 	err error
 }
 
-// ThrowError returns the error type that is passed. It acts as a
-// single consistent way of throwing errors from the pkg level.
-// As long as the error type is defined before hand, this can be
-// used to throw errors up to the calling code without discrimination
-// around the error type.
-func ThrowError(err interface{ error }) error {
-	return err
-}
-
 func (e *VerificationFailure) Error() string {
 	return e.err.Error()
+}
+
+func (e *VerificationFailure) Unwrap() error {
+	return e.err
 }
 
 type ErrNoMatchingSignatures struct {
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
@ChrisJBurns I think I'm missing some context for some of this stuff, do these changes seem ~reasonable to you?

ChrisJBurns:
Hi @jonjohnsonjr , changes seem good with me! :+1:

 The `ThrowError` was mainly to have a common way of throwing errors with a bit of syntactic sugar, there was no functional benefit for it being there. The unwrapping functions seem good to me also.  On the Policy Controller side of things, I didn't actually realise that we were using functions in Cosign to throw errors in other applications, so for backwards compatibility that seems ok with me to put it back in. Although I'd definitely raise the question of whether we should be using Cosign as a library for throwing errors in other applications? As we seem to be coupling the two a bit? Would it be easier to copy the VerificationFailure error code we're putting back in this PR and just put it in policy controller? It doesn't really seem to be anything Cosign specific? Feel free to correct me as I've not got any experience with the policy controller thus far!

jonjohnsonjr:
> Although I'd definitely raise the question of whether we should be using Cosign as a library for throwing errors in other applications? As we seem to be coupling the two a bit? Would it be easier to copy the VerificationFailure error code we're putting back in this PR and just put it in policy controller? It doesn't really seem to be anything Cosign specific? Feel free to correct me as I've not got any experience with the policy controller thus far!

I think that would have been better to do originally, but for the sake of the dependency graph, I think it's easier to just add these back (as deprecated) and then have policy-controller deal with that problem when it bumps cosign.

imjasonh (approved):

ChrisJBurns:
@jonjohnsonjr  Yep that's fair enough. Out of curiosity, do you know whereas Cosign is used like it is in the policy controller? Just so I can bare it in mind when doing future changes? As I'll be adding a bunch of exit codes when certain errors happen from here onwards (and encourage others to do the same - perfect for good first issues)

jonjohnsonjr:
> Out of curiosity, do you know where else Cosign is used like it is in the policy controller?

I only happened upon this because I tried to bump a cosign dependency and it broke policy-controller :P

I think a lot of these are forks, but if you wanted to dig into it, I'd start here: https://pkg.go.dev/github.com/sigstore/cosign/pkg/cosign?tab=importedby

cpanato (approved):
lgtm, thanks sorry for taking to long to review

PR #3113, opened 2023-07-16
switch to uploading DSSE types to rekor instead of intoto

bobcallaway:
#### Summary
This switches cosign from uploading `intoto` Rekor types when dealing with intoto attestations to using `dsse` Rekor types. This will result in attestations no longer being stored by Rekor, and enforces a more correct behavior in that attestations should be stored along side the artifact in a registry and/or in a bundle file.

#### Release Note
* cosign now uploads entries to Rekor instances for intoto attestations using the `dsse` Rekor type. This means that Rekor will no longer return the attestation content as part of an entry response. The entry response includes a digest over both the full DSSE envelope as well as the payload content for verifiers to compare against local copies of the DSSE envelope.


bobcallaway:
Rekor instance launched by scaffolding is quite old, https://github.com/sigstore/scaffolding/pull/701 needs to merge and a new release cut before that will pass.

haydentherapper (approved):
Only question is if you think we should do a minor version bump of cosign instead of a patch. This isn’t a breaking change, unless clients are currently relying on this storage.

haydentherapper (approved):

Issue #3120, opened 2023-07-19
Artifacts are non-deterministic


jonjohnsonjr:
**Description**

This effectively reverts https://github.com/sigstore/cosign/pull/2108

Determinism is a feature, not a bug. Libraries should not depend on external, global, mutable state (`time.Now()`).

Once this is fixed, I will reopen https://github.com/sigstore/cosign/issues/522, and (if someone wants) they can plumb an option through that sets the time.

**Version**

Since https://github.com/sigstore/cosign/issues/522

Akaame:
It seems there were arguments regarding

- The importance of reproducibility (@dlorenc @dekkagaijin)
- Some providers using timestamps to perform housekeeping, thus not depending on `time.Now()` biting you in the back (again @dekkagaijin @imjasonh)

PR #3121, opened 2023-07-19
Fix nondeterminsitic timestamps

jonjohnsonjr:
We should not (by default) be making the artifacts we produce nondeterministic.

#### Summary

Fixes https://github.com/sigstore/cosign/issues/3120


cpanato (approved):
thank you

bobcallaway (approved):

PR #3138, opened 2023-07-26
bundle support for sign-blob

cdris:
#### Summary

Adds support for the new sigstore bundle format defined in github.com/sigstore/protobuf-specs to `sign-blob`. When the `--bundle` flag is provided with an output file, users can now specify `--use-new-bundle-format=true` to use the new common format. This flag defaults to false.

https://github.com/sigstore/cosign/issues/3139


pkg/protobundle/bundle.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/sigstore/rekor/pkg/generated/models"
+
+	"github.com/sigstore/rekor/pkg/tle"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+// GenerateTransparencyLogEntry returns a sigstore/protobuf-specs compliant message containing a
+// TransparencyLogEntry as defined at https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_rekor.proto
+func GenerateTransparencyLogEntry(anon models.LogEntryAnon) (*pbrekor.TransparencyLogEntry, error) {
+	return tle.GenerateTransparencyLogEntry(anon)
```
cdris, line 31: @haydentherapper currently this still has the log index bug fixed in https://github.com/sigstore/rekor/pull/1599 because of the pinned version of rekor imported. Do we need to do a rekor release to update the version? Or can I just pin the most recent commit? Or should I just wrap this call in the fix as well for the time being?

haydentherapper: We can just update the version of Rekor in go.mod - `go get github.com/sigstore/rekor@96dad3c1f9e210514858fcb2473268aab3aff69a` for the latest commit. When we cut a new release, dependabot will update the rekor dependency to the latest release

haydentherapper:
Ref https://github.com/sigstore/cosign/issues/3139

cmd/cosign/cli/options/signblob.go
```diff
@@ -71,6 +72,9 @@ func (o *SignBlobOptions) AddFlags(cmd *cobra.Command) {
 		"write everything required to verify the blob to a FILE")
 	_ = cmd.Flags().SetAnnotation("bundle", cobra.BashCompFilenameExt, []string{})
 
+	cmd.Flags().BoolVar(&o.UsePBBundleFormat, "use-new-bundle-format", false,
```
haydentherapper, line 75: Can we add a TODO to eventually make this flag a no-op and add a deprecated-format flag?

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -117,26 +124,83 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
...
-		certBytes, err := extractCertificate(ctx, sv)
-		if err != nil {
-			return nil, err
-		}
-		signedPayload.Cert = base64.StdEncoding.EncodeToString(certBytes)
+			certBytes, err := sv.Bytes(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("error getting signer: %w", err)
+			}
+			certChain, err := protobundle.GenerateX509CertificateChain(certBytes)
```
haydentherapper, line 139: A recent decision was to have this only be the non-trusted certificates, and any CA certs that appear in the root of trust should not be included - https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L171-L188

We can either:
* Load in just the leaf certificate and add a TODO to handle private deployments
* Just cut out the root and include leaves and intermediates
* Cut out any CA certs that are returned from these two funcs - https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/fulcio/fulcioroots/fulcioroots.go#L37-L57

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -117,26 +124,83 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
...
-		signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(sig)
+		if ko.UsePBBundleFormat {
+			vm := pbbundle.VerificationMaterial{}
 
-		certBytes, err := extractCertificate(ctx, sv)
-		if err != nil {
-			return nil, err
-		}
-		signedPayload.Cert = base64.StdEncoding.EncodeToString(certBytes)
+			certBytes, err := sv.Bytes(ctx)
```
haydentherapper, line 135: We'll also need to handle where `sv.Bytes()` returns a key, since that goes into a different message. We could change `sv.Bytes()` to return whether it's a key or cert, or try to parse the return value to determine if it's a key or cert.

PR #3146, opened 2023-07-31
improve error messages around bundle != payload hash

imjasonh:
Before this change, when the bundle hash doesn't match the payload hash, the error message is `error verifying bundle: matching bundle to payload: %!w(<nil>)` which isn't terribly helpful.

This hopes to improve the error experience in this area, by:

- clearly failing ASAP when bundle hash computation fails
- clearly failing separately if the bundle algorithm is somehow not sha256
- clearly failing with both hashes when they don't match, hopefully making debugging easier

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1105,11 +1105,16 @@ func VerifyBundle(sig oci.Signature, co *CheckOpts) (bool, error) {
 	}
 
 	alg, bundlehash, err := bundleHash(bundle.Payload.Body.(string), signature)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):
lgtm, thanks @imjasonh

haydentherapper (approved):

imjasonh:
```
google.golang.org/grpc/internal/googlecloud: mkdir /tmp/go-build89590620/b1593/: no space left on device
```

I think I need an adult to retry to workflow.

haydentherapper:
image

Issue #3151, opened 2023-08-01
Document use of sign-blob with key from environment variable


pjbgf:
**Description**

The [cosign_sign-blob](doc/cosign_sign-blob.md) doc does not have references or examples for the usage of loading keys from environment variables.

PR #3152, opened 2023-08-01
doc: Add example of sign-blob with key in env var

pjbgf:
#### Summary

Adds example on how to load a key from an environment variable when using `sign-blob`.

Closes https://github.com/sigstore/cosign/issues/3151.

#### Release Note

NONE

#### Documentation

N/A


hectorj2f (dismissed):

hectorj2f:
@pjbgf you need to run `make docgen`

Issue #3149, opened 2023-08-01
CI jobs: validate-release-job failing with 'no space left on device'


hectorj2f:
**Description**

It seems we are now facing a new problem with one of our CI jobs `validate-release-job` throwing errors caused by a lack of space on the device. 

When I run `df -h` on the runner, I got the following results:
```
Filesystem      Size  Used Avail Use% Mounted on
overlay          84G   71G   14G  85% /
tmpfs            64M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/root        84G   71G   14G  85% /__t
tmpfs           1.4G  1.2M  1.4G   1% /run/docker.sock
tmpfs           3.4G     0  3.4G   0% /proc/acpi
tmpfs           3.4G     0  3.4G   0% /proc/scsi
tmpfs           3.4G     0  3.4G   0% /sys/firmware
```

**Version**

`main`

hectorj2f:
I can see once goreleaser runs it fills up the whole overlay available space 😱 1

```
**overlay          84G   83G  737M 100% /**
tmpfs            [6](https://github.com/sigstore/cosign/actions/runs/5726129049/job/15515969205?pr=3150#step:7:7)4M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/root        84G   83G  [7](https://github.com/sigstore/cosign/actions/runs/5726129049/job/15515969205?pr=3150#step:7:8)37M 100% /__w
tmpfs           1.4G  1.2M  1.4G   1% /run/docker.sock
tmpfs           3.4G     0  3.4G   0% /proc/acpi
tmpfs           3.4G     0  3.4G   0% /proc/scsi
tmpfs           3.4G     0  3.4G   0% /sys/firmware
```

hectorj2f:
It seems to only happen when building `windows_amd64_v1`.

hectorj2f:
It works now! It seems as once we managed to successfully execute the job by changing some build settings, things got cleanup at a runner level.

hectorj2f:
@haydentherapper I'm gonna remove the execution of the validate release ci job for every PR. 

I tried to only build a single binary (linux/arm64) but the rest of our building scripts use `--platform=all` which will require many more changes. After all this script only checks that the build of the amd64 binary was successful. So I propose to only running for the `release-` branches. Perhaps @cpanato has other ideas to solve the problem.

haydentherapper:
[...]

PR #3164, opened 2023-08-07
ci: remove validate release ci step

hectorj2f:
#### Summary

This CI job is using all the available space for the runner when building the windows binary (and all the others). I tried couple of configuration changes.  I even attempted to build a binary for a single platform but the rest of our release scripts are configured to build all the platforms which will require changes to all the script if we just want to release a single binary (instead of all). Even if I set an env variable `PLATFORM=linux`, this is conflicting with any GOOS or GOARCH set with goreleaser to only build a single target. 

So I suggest to remove it to run on every single PR.

closes https://github.com/sigstore/cosign/issues/3149

### DIFF
--- a/.github/workflows/validate-release.yml
+++ b/.github/workflows/validate-release.yml
@@ -20,7 +20,6 @@ on:
     branches:
       - main
       - release-*
-  pull_request:
 
 jobs:
   check-signature:


hectorj2f:
TODO: Remove `validate-ci-job` from the list of required ci jobs.

hectorj2f:
I found a better solution in https://github.com/sigstore/cosign/pull/3169. Closing this PR now!

PR #3200, opened 2023-08-23
add mTLS to TSA for sign-blob command

### LINKED ISSUE #3006: add support for mTLS connection to the TSA (timestamp server)
**Description**

Currently there is no way to:
* pass client TLS certificates when establishing https connection to the timestamp server (TSA)
* specify custom CA certificate chain for that connection except the `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables which have a number of disadvantages such as 1) not having any effect on Darwin or Windows 2) to overwrite instead of extending the system certificate store you normally need to set _both_ `SSL_CERT_FILE` and `SSL_CERT_DIR`, possibly to some "fake" locations (overwriting instead of extending is necessary if you want to make sure that only custom CA certs can be used for the connection to the TSA)
* specify SAN values that the server's certificates must have (SAN validation of the TSA server's TLS certificates)

These missing features make it currently impossible to use a TSA server in an environment where every connection must be done using mTLS (for example cloud setup where mTLS is required for authentication and authorisation).

Currently we have the `--timestamp-server-url` parameter for `cosign sign`.  
We can add additional command-line (or possibly configuration) parameters to add those features. We can borrow some names from the corresponding `curl` parameters:
```bash
$ curl --help all | egrep 'cert|key'
     --cacert <file>      CA certificate to verify peer against
-E, --cert <certificate[:password]> Client certificate file and password
      --key <key>          Private key file name
     --key-type <type>    Private key file type (DER/PEM/ENG)
     --pass <phrase>      Pass phrase for the private key
```
     
Some possible new parameter names:     
* `--timestamp-server-cacert`
* `--timestamp-server-cert`
* `--timestamp-server-key`
* `--timestamp-server-san`   # for SAN validation of the TSA server's certificates

Eventually, for the "full blown" mTLS, it would be good to have a parameter on the server's side to specify the allowed values in the SAN values of the client's TLS certificate.

References:

1. https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_default_verify_paths.html
2. https://go.googlesource.com/go/+/refs/heads/master/src/crypto/x509/cert_pool.go#102
> 
```
On Unix systems other than macOS the environment variables SSL_CERT_FILE and
SSL_CERT_DIR can be used to override the system default locations for the SSL
certificate file and SSL certificate files directory, respectively. The
latter can be a colon-separated list.
```
3. https://go-review.googlesource.com/c/go/+/36093 crypto/x509: load certs from env vars + extra locations
4. [Discussion](https://gophers.slack.com/archives/C6CKB7AA2/p1685095252459469?thread_ts=1685051992.319459&cid=C6CKB7AA2) about `SSL_CERT_FILE` and `SSL_CERT_DIR` in #crypto channel of the Gophers slack (highlights that both needs to be set to achieve overwrite of the system roots)

dmitris:
#### Summary
Clone of https://github.com/sigstore/cosign/pull/3052 but for the `cosign sign-blob` command.

Add command-line parameters for key/cert/cacert used for the connection to the TSA server to the `sign-blob` command.

Fixes #3006.

#### Release Note
* New features and improvements, including behavioural changes, UI changes and CLI changes
** Additional command-line options for `cosign sign-blob` command for mTLS connection to the TSA server

#### Documentation
https://github.com/sigstore/docs/pull/239


haydentherapper (approved):
Looks great!

znewman01 (approved):

PR #3219, opened 2023-09-05
Add SLSA 1.0 attestation support to cosign. Closes #2860

### LINKED ISSUE #2860: Add support for SLSA 1.0 predicate
Cosign supports the v0.2 SLSA predicate through the `--type slsaprovenance` flag on the `attest` and `attest-blob` commands. SLSA 1.0 Release Candidate 1 (RC1) is out and RC2 should be coming out today. By the end of this month (April 2023) it is expected that SLSA v1.0 will be officially released. We don't expect any new major changes to the SLSA provenance predicate.

For reference, here's the specification for the predicate: https://slsa.dev/provenance/v1-rc1

ziel:
#### Summary
This change adds support for the SLSA 1.0 provenance predicate (https://slsa.dev/spec/v1.0/provenance).

`cosign attest` and related commands cannot validate it currently, and the `custom` type must be used for it instead. After this change, it will be possible to specify the new provenance attestation type as `slsaprovenance1`.

Related issue: https://github.com/sigstore/cosign/issues/2860

#### Release Note
Added a new type `slsaprovenance1` to support slsa 1.0 provenance in attest and verify commands.

#### Documentation
The attestation types do not appear to be documented at https://docs.sigstore.dev so there is no related change there.

### DIFF
--- a/pkg/cosign/attestation/attestation.go
+++ b/pkg/cosign/attestation/attestation.go
@@ -23,7 +23,8 @@ import (
 	"strings"
 	"time"
 
-	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+	slsa02 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
 
 	"github.com/in-toto/in-toto-golang/in_toto"
 )
@@ -99,7 +100,7 @@ type GenerateOpts struct {
 }
 
 // GenerateStatement returns an in-toto statement based on the provided
-// predicate type (custom|slsaprovenance|spdx|spdxjson|cyclonedx|link).
+// predicate type (custom|slsaprovenance|slsaprovenance1.0|spdx|spdxjson|cyclonedx|link).
 func GenerateStatement(opts GenerateOpts) (interface{}, error) {
 	predicate, err := io.ReadAll(opts.Predicate)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


haydentherapper:
I think this is a good way to approach this, and we can plan to change the behavior of the flags in a later release as a breaking change, where `slsaprovenance` will generate a v1.0 attestation and we'll drop support for v0.2.

cc @znewman01

pkg/cosign/attestation/attestation.go
```diff
@@ -99,7 +100,7 @@ type GenerateOpts struct {
 }
 
 // GenerateStatement returns an in-toto statement based on the provided
-// predicate type (custom|slsaprovenance|spdx|spdxjson|cyclonedx|link).
+// predicate type (custom|slsaprovenance|slsaprovenance1.0|spdx|spdxjson|cyclonedx|link).
```
znewman01, line 103: nit: `slsaprovenance1`.

znewman01 (changes requested):
Maybe to ease the transition can we add a `slsaprovenance0.2` option that's just an alias for the current behavior?

Then we could at some future point switch over `slsaprovenance` to point to `slsaprovenance1`.

Otherwise looks great, thanks for the change!

pkg/cosign/attestation/attestation.go
ziel: oh oops! will fix

haydentherapper (approved):

ziel:
hi @znewman01 -- i added a patch related to your change request, if you have a moment to peek at it. i called the type `slasprovenance02` because there's places in the code base which don't like the `.` -- though we could have it diverge from the those if that'd be better

znewman01 (approved):

PR #3224, opened 2023-09-06
Fail timestamp verification if no root is provided

haydentherapper:
The bug was that we would conditionally check a timestamp if a root was provided. If no root was provided even if a timestamp was provided, then signature verification would succeed.

The good news is this will not show a successful signature if the transparency log does not contain the entry too, for timestamp verification.

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1101,11 +1101,12 @@ func VerifyBundle(sig oci.Signature, co *CheckOpts) (bool, error) {
 // no root is provided with a timestamp.
 func VerifyRFC3161Timestamp(sig oci.Signature, co *CheckOpts) (*timestamp.Timestamp, error) {
 	ts, err := sig.RFC3161Timestamp()
-	if err != nil {
+	switch {
+	case err != nil:
 		return nil, err
-	} else if ts == nil {
+	case ts == nil:
 		return nil, nil
-	} else if co.TSARootCertificates == nil {
+	case co.TSARootCertificates == nil:
 		return nil, errors.New("no TSA root certificate(s) provided to verify timestamp")
 	}
 
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -648,14 +648,12 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 	bundleVerified bool, err error) {
 	var acceptableRFC3161Time, acceptableRekorBundleTime *time.Time // Timestamps for the signature we accept, or nil if not applicable.
 
-	if co.TSARootCertificates != nil {
-		acceptableRFC3161Timestamp, err := VerifyRFC3161Timestamp(sig, co)
-		if err != nil {
-			return false, fmt.Errorf("unable to verify RFC3161 timestamp bundle: %w", err)
-		}
-		if acceptableRFC3161Timestamp != nil {
-			acceptableRFC3161Time = &acceptableRFC3161Timestamp.Time
-		}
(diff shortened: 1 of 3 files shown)


zhaoyonghe:
Hi @haydentherapper, just to clarify, is this fix handling the scenario when we verify the keyless signature within the validity of the signing cert? Because if we verify the keyless signature after the signing cert expired and we do not provide the TSA root, the verification will fail like this:
```
Error: no matching signatures: expected a signed timestamp to verify an expired certificate
main.go:69: error during command execution: no matching signatures: expected a signed timestamp to verify an expired certificate
```
Thank you!

haydentherapper:
@zhaoyonghe, no, that's working as intended. If no signed entry timestamp (from the transparency log) or timestamp is provided, certificate verification should fail.

This is fixing a bug reported on https://sigstore.slack.com/archives/C01PZKDL4DP/p1693930315937879, where if no TSA root is provided, timestamp verification is skipped.

znewman01 (approved):

zhaoyonghe:
> @zhaoyonghe, no, that's working as intended. If no signed entry timestamp (from the transparency log) or timestamp is provided, certificate verification should fail.
> 
> This is fixing a bug reported on https://sigstore.slack.com/archives/C01PZKDL4DP/p1693930315937879, where if no TSA root is provided, timestamp verification is skipped.

@haydentherapper Thank you for the reply! 

I asked that question because I could not reproduce the same bug based on the PR title and description. After studying the original bug report, I finally realized that this bug is only triggered when users use the public key (**not the cert**, and that is what I used when I tried to reproduce it) in the verification. The bug reporter did not use a cert in verification, so `cosign` will not execute [these code](https://github.com/sigstore/cosign/blob/29396d66f0d5f78a98ae52f24bc5da4ec78a3ea9/pkg/cosign/verify.go#L753-L784) and raise the error ["expected a signed timestamp to verify an expired certificate"](https://github.com/sigstore/cosign/blob/29396d66f0d5f78a98ae52f24bc5da4ec78a3ea9/pkg/cosign/verify.go#L778).

mtrmac:
> This is fixing a bug reported on https://sigstore.slack.com/archives/C01PZKDL4DP/p1693930315937879, where if no TSA root is provided, timestamp verification is skipped.

Why is that a problem?

The way I think about security, the _client_ decides what the policy is, and what gets verified:
- If the client wants a timestamp, the client decides what the root of thrust is, and potentially whether to require a timestamp.
- If the client doesn't provide a trust root for a timestamp, the client _doesn’t care_ and the timestamp, even if present, is ignored.

With the changes in this PR, the _image_ decides whether the timestamp is validated. Why is it up to the image to do that?

The practical impact is, AFACS, that _adding_ a timestamp the client is not configured to trust triggers a “no TSA root certificate(s) provided to verify timestamp” verification failure. To me, it is _deeply surprising_ that clients can accept an image with no timestamp (i.e. they clearly don’t need the timestamp to exist) but they will refuse the same image with a timestamp added (even if the timestamp is completely correct).

haydentherapper:
> If the client doesn't provide a trust root for a timestamp, the client doesn’t care and the timestamp, even if present, is ignored.

I would guess that most of the time, if a user provides something that needs to be verified, they expect the client to verify it. If the user doesn't want to verify time, then a user should not provide signed time. We do the same thing with other signed material - if a certificate is provided without a trust root, we err out.

mtrmac:
In the context of signature verification, I think the natural threat model implies that the image is not coming from the trusted user: it is coming from a potentially-malicious registry. So a paranoid signature verifier should, I think, _minimize_ the number of decisions, or code patch choices, that potentially-malicious registry can affect.

PR #3225, opened 2023-09-06
attest: pass OCI remote opts to att resolver.

wlynch:
#### Summary

We need cred helpers and other info set in the remote opts to resolve the upstream attesations, otherwise this will fail in automated environments.

/cc @priyawadhwa @strongjz 

#### Release Note

- Fixed bug in `cosign attest` where OCI remote configuration wasn't properly set.

#### Documentation

n/a


cpanato (approved):

cpanato:
thanks!

PR #3233, opened 2023-09-09
fix: update error in `SignedEntity` to be more descriptive

### LINKED ISSUE #2930: Better error messages for unauthorized PATs to private packages
**Description**

*disclaimer: im not sure if this is possible / makes sense considering how PATs work*

Trying to sign an image using a PAT gave me this error:
```
Error: signing [ghcr.io/ORG/IMAGE@sha256:DIGEST]: accessing image: entity not found in registry
main.go:74: error during command execution: signing [ghcr.io/ORG/IMAGE@sha256:DIGEST]: accessing image: entity not found in registry
```

I was sort of stumped as to why it happened, but it simply turned out I had not allowed the PAT access to my organization via the SAML login.

On a public package, this error was handled quite well, telling me that the PAT was not "logged in" using SAML. However, only a 404-esque response was given for private packages.

vishal-chdhry:
#### Summary

Closes #2930 

The exact error described in the issue cannot happen anymore as [that behavior was changed in v2.1.0](https://github.com/sigstore/cosign/pull/2959/files#diff-72911b6572099462e5f2e8ff920f4f9228587ab100820a26d3e90377c36bd73fR182-R184). 

PR https://github.com/sigstore/cosign/pull/2959 updated the old behaviour and added an if case where the [error message is used in an if statement](https://github.com/sigstore/cosign/pull/2959/files#diff-72911b6572099462e5f2e8ff920f4f9228587ab100820a26d3e90377c36bd73fR182-R183). The issue with this is that the error value is lost when `ErrEntityNotFound` error is created. This creates a problem where the error in `SignedEntity` are not descriptive enough when `remoteGet` function fails. 

This PR will:
- Add a function called `NewEntityNotFoundError` which uses the error received by `remoteGet` to create a new error.
- Add a function called `IsEntityNotFoundError` which is used in the if statement that checks if the error message contains an `ErrEntityNotFound` string.

#### Release Note

Updated error in `SignedEntity` to be more descriptive

cc @znewman01

### DIFF
--- a/pkg/oci/remote/remote.go
+++ b/pkg/oci/remote/remote.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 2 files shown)


pkg/oci/remote/remote.go
```diff
@@ -39,9 +40,17 @@ var (
 
 	// ErrEntityNotFound is the error that SignedEntity returns when the
 	// provided ref does not exist.
-	ErrEntityNotFound = errors.New("entity not found in registry")
+	ErrEntityNotFound = "cosign remoteGet: entity not found in registry"
 )
 
+func NewEntityNotFoundError(err error) error {
```
znewman01, line 46: At this point, it probably makes sense to make a first-class error type. Then we don't need `IsEntityNotFoundError`, we can use reflection.

pkg/oci/remote/remote.go
```diff
@@ -39,9 +40,17 @@ var (
 
 	// ErrEntityNotFound is the error that SignedEntity returns when the
 	// provided ref does not exist.
-	ErrEntityNotFound = errors.New("entity not found in registry")
+	ErrEntityNotFound = "cosign remoteGet: entity not found in registry"
```
znewman01, line 43: It's a little weird to have `ErrEntityNotFound` to just be a string. I also think you don't actually need this anymore if you make it a real error.

znewman01 (changes requested):
Thanks for this change! Overall looks good, I have a couple of Go-specific requests.

znewman01 (approved):

PR #3237, opened 2023-09-12
Fixes #3236, disable SCT checking for a cosign verification when usin…

### LINKED ISSUE #3236: Signed Certificate Timestamp with Long-lived Keys
**Question**
When signing an image with the command below:
`cosign sign --key cosign.key <IMAGE>  --tlog-upload=false`

it's required to use this command to verify it (in a restricted environment, i.e. no access to TUF root CDN):
`cosign verify --key cosign.pub <IMAGE> --insecure-ignore-tlog --insecure-ignore-sct`

even though no timestamp will be added.  This is related to [this](https://sigstore.slack.com/archives/C01PZKDL4DP/p1694462110654769) Slack thread.

Should it be necessary to ignore the "Signed Certificate Timestamp" when not signing with a certificate? Are their valid use-cases for a timestamp with a long-lived key?

jkjell:
…g a public key

- Closes #3236 

#### Summary
This changes remove the Signed Certificate Timestamp checking when providing a public key to `cosign verify`.  One of the issues SCT verification causes is the retrieval of its public keys from the TUF CDN by default. This breaks verification in disconnected or restricted environment. SCT is not necessary to validate a key pair.

#### Release Note
<!--
Add a release note for each of the following conditions:

* In private sigstore environments, if you were previously using the `insecure-ignore-sct` flag, that is no longer required.
* Fixed #3236 

#### Documentation
None required

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -202,7 +202,8 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 	keyRef := c.KeyRef
 	certRef := c.CertRef
 
-	if !c.IgnoreSCT {
+	// Ignore Signed Certificate Timestamp if the flag is set or a key is provided
+	if !c.IgnoreSCT || keyRef != "" {
 		co.CTLogPubKeys, err = cosign.GetCTLogPubs(ctx)
 		if err != nil {
 			return fmt.Errorf("getting ctlog public keys: %w", err)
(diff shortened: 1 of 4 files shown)


haydentherapper (changes requested):
Thanks! Could you also make this change in the other verify* files, like https://github.com/sigstore/cosign/blob/86252aaf9e6682f42daefe371806f1bd67890ec9/cmd/cosign/cli/verify/verify_attestation.go#L113, https://github.com/sigstore/cosign/blob/86252aaf9e6682f42daefe371806f1bd67890ec9/cmd/cosign/cli/verify/verify_blob_attestation.go#L192, https://github.com/sigstore/cosign/blob/86252aaf9e6682f42daefe371806f1bd67890ec9/cmd/cosign/cli/verify/verify_blob.go#L288

haydentherapper (approved):

PR #3239, opened 2023-09-13
Allow `cosign save` to reuse save directory to help dedupe shared layers.

amartin120:
#### Summary

When using `cosign save` with dozens, or in my case, hundreds of images, the storage consumption can really add up with each image requiring their own directory.  There's potentially a case where images can share layers so this change allows for the `cosign save` function to be able to reuse a directory from a previous save and just build on the `index.json` along with an additional annotation for each entry to aide with the corresponding `cosign load`.

From there, I've modified `cosign load` to have a new optional flag `--registry` where you can specify a registry to load all of the images to from a single directory.  In cases where you are only dealing with a single image, the load command would function exactly how it does today where you just provide the remote reference via the cli.

#### Release Note

* Updated `cosign save` to append the index.json if pointed at a `--dir` used in a previous save.
* Added the `org.opencontainers.image.ref.name` annotation to the index.json for clarity and to help with `cosign load` in cases where the directory is reused.
* Updated `cosign load` to have an optional flag called `--registry` that can be used in cases where the index.json from a save contains multiple images.  This handles being able to control where the images get loaded to in the case where providing a single ref via the CLI no longer makes sense.  If you don't provide the `--registry` flag, the `cosign load` command will continue to function as it does today.

#### Documentation

Example usage:
`cosign save image1:v1.0 --dir ~/saves`
`cosign save image2:v1.0 --dir ~/saves`

`cat ~/saves/index.json`

```
{
   "schemaVersion": 2,
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1581,
         "digest": "sha256:c5b6da2ffbc9f63e8d250d5985f722830dfb5cc8958e8f64f7e14edc18b584fa",
         "annotations": {
            "kind": "dev.cosignproject.cosign/image",
            "org.opencontainers.image.ref.name": "image1:v1.0"
         }
      },
      {
         "size": 558,
         "digest": "sha256:077bc8e8cfe1577f89959c749da1fbe941bde8b35b7ccdfe4f5832fb5ade2ec1",
         "annotations": {
            "kind": "dev.cosignproject.cosign/sigs",
            "org.opencontainers.image.ref.name": "image1:v1.0"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 468,
         "digest": "sha256:7e6fb78b8f35c859c9200567229a4350c17e057d4d13c80121a3f603748a1cb0",
         "annotations": {
            "kind": "dev.cosignproject.cosign/atts",
            "org.opencontainers.image.ref.name": "image1:v1.0"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
         "size": 967,
         "digest": "sha256:7e42b1a631859c6740835659a022ed96097955dde2707f32e7a089f5b5f64c06",
         "annotations": {
            "kind": "dev.cosignproject.cosign/imageIndex",
            "org.opencontainers.image.ref.name": "image2:v1.0"
         }
      },
      {
         "size": 558,
         "digest": "sha256:88145843e5dcd7a0d515cf874997b1d4eaa742a1594dc75c6bd111f109cf4455",
         "annotations": {
            "kind": "dev.cosignproject.cosign/sigs",
            "org.opencontainers.image.ref.name": "image2:v1.0"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 469,
         "digest": "sha256:6fa5a413e06af7babaeae39bb06c92b563463080fd46003b078050787c3ef6f0",
         "annotations": {
            "kind": "dev.cosignproject.cosign/atts",
            "org.opencontainers.image.ref.name": "image2:v1.0"
         }
      }
   ]
}
```

`cosign load  --dir ~/saves --registry your.registry.com`

### DIFF
--- a/cmd/cosign/cli/load.go
+++ b/cmd/cosign/cli/load.go
@@ -32,13 +32,19 @@ func Load() *cobra.Command {
 	o := &options.LoadOptions{}
 
 	cmd := &cobra.Command{
-		Use:              "load",
-		Short:            "Load a signed image on disk to a remote registry",
-		Long:             "Load a signed image on disk to a remote registry",
-		Example:          `  cosign load --dir <path to directory> <IMAGE>`,
-		Args:             cobra.ExactArgs(1),
+		Use:     "load",
+		Short:   "Load a signed image on disk to a remote registry",
+		Long:    "Load a signed image on disk to a remote registry",
+		Example: `  cosign load --dir <path to directory> <IMAGE> OR cosign load --dir <path to directory> --registry <REGISTRY>`,
+		//Args:             cobra.ExactArgs(1),
 		PersistentPreRun: options.BindViper,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if o.Registry.Name != "" && len(args) == 0 {
+				return LoadCmd(cmd.Context(), *o, "")
+			}
+			if len(args) != 1 {
+				return fmt.Errorf("image argument is required")
+			}
 			return LoadCmd(cmd.Context(), *o, args[0])
 		},
 	}
@@ -48,12 +54,22 @@ func Load() *cobra.Command {
 }
 
 func LoadCmd(ctx context.Context, opts options.LoadOptions, imageRef string) error {
-	ref, err := name.ParseReference(imageRef)
-	if err != nil {
-		return fmt.Errorf("parsing image name %s: %w", imageRef, err)
+
+	if opts.Registry.Name != "" && imageRef != "" {
+		return fmt.Errorf("both --registry and image argument provided, only one should be used")
 	}
 
-	// get the signed image from disk
+	var ref name.Reference
+	var err error
+	if opts.Registry.Name == "" {
+		// Use the provided image reference
+		ref, err = name.ParseReference(imageRef)
+		if err != nil {
+			return fmt.Errorf("parsing image name %s: %w", imageRef, err)
+		}
+	}
+
+	// get the signed image(s) from disk
 	sii, err := layout.SignedImageIndex(opts.Directory)
 	if err != nil {
 		return fmt.Errorf("signed image index: %w", err)
@@ -64,5 +80,9 @@ func LoadCmd(ctx context.Context, opts options.LoadOptions, imageRef string) err
 		return err
 	}
 
-	return remote.WriteSignedImageIndexImages(ref, sii, ociremoteOpts...)
+	if opts.Registry.Name == "" {
+		return remote.WriteSignedImageIndexImages(ref, sii, ociremoteOpts...)
+	} else {
+		return remote.WriteSignedImageIndexImagesBulk(opts.Registry.Name, sii, ociremoteOpts...)
+	}
 }
--- a/cmd/cosign/cli/load.go
+++ b/cmd/cosign/cli/load.go
@@ -39,6 +39,9 @@ func Load() *cobra.Command {
 		//Args:             cobra.ExactArgs(1),
 		PersistentPreRun: options.BindViper,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if o.Registry.Name != "" && len(args) != 0 {
+				return fmt.Errorf("both --registry and image argument provided, only one should be used")
+			}
 			if o.Registry.Name != "" && len(args) == 0 {
 				return LoadCmd(cmd.Context(), *o, "")
 			}
@@ -54,10 +57,6 @@ func Load() *cobra.Command {
 }
 
 func LoadCmd(ctx context.Context, opts options.LoadOptions, imageRef string) error {
-	if opts.Registry.Name != "" && imageRef != "" {
-		return fmt.Errorf("both --registry and image argument provided, only one should be used")
-	}
-
 	var ref name.Reference
 	var err error
 	if opts.Registry.Name == "" {
--- a/cmd/cosign/cli/load.go
+++ b/cmd/cosign/cli/load.go
@@ -54,7 +54,6 @@ func Load() *cobra.Command {
 }
 
 func LoadCmd(ctx context.Context, opts options.LoadOptions, imageRef string) error {
-
 	if opts.Registry.Name != "" && imageRef != "" {
 		return fmt.Errorf("both --registry and image argument provided, only one should be used")
 	}
@@ -82,7 +81,6 @@ func LoadCmd(ctx context.Context, opts options.LoadOptions, imageRef string) err
 
 	if opts.Registry.Name == "" {
 		return remote.WriteSignedImageIndexImages(ref, sii, ociremoteOpts...)
-	} else {
-		return remote.WriteSignedImageIndexImagesBulk(opts.Registry.Name, sii, ociremoteOpts...)
 	}
+	return remote.WriteSignedImageIndexImagesBulk(opts.Registry.Name, sii, ociremoteOpts...)
 }
--- a/pkg/oci/layout/write.go
+++ b/pkg/oci/layout/write.go
@@ -16,10 +16,10 @@
 package layout
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
-	"errors"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -31,14 +31,14 @@ import (
 // WriteSignedImage writes the image and all related signatures, attestations and attachments
 func WriteSignedImage(path string, si oci.SignedImage, ref name.Reference) error {
 	layoutPath, err := layout.FromPath(path)
-    if os.IsNotExist(err) {
-        // If the layout doesn't exist, create a new one
-        layoutPath, err = layout.Write(path, empty.Index)
-    }
-    if err != nil {
-        return err
-    }
-	
+	if os.IsNotExist(err) {
+		// If the layout doesn't exist, create a new one
+		layoutPath, err = layout.Write(path, empty.Index)
+	}
+	if err != nil {
+		return err
+	}
+
 	// write the image
 	if err := appendImage(layoutPath, si, ref, ImageAnnotation); err != nil {
 		return fmt.Errorf("appending signed image: %w", err)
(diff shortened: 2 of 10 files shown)


pkg/oci/layout/write.go
```diff
@@ -17,51 +17,71 @@ package layout
...
+    if err != nil {
+        if os.IsNotExist(err) {
+            // If the layout doesn't exist, create a new one
+            layoutPath, err = layout.Write(path, empty.Index)
+            if err != nil {
+                return err
+            }
+        } else {
+            return err
+        }
```
dweomer, line 42:
```suggestion
    if os.IsNotExist(err) {
        // If the layout doesn't exist, create a new one
        layoutPath, err = layout.Write(path, empty.Index)
    }
    if err != nil {
        return err
    }
```

pkg/oci/layout/write.go
```diff
@@ -17,51 +17,71 @@ package layout
...
+    }
```
dweomer, line 65:
```suggestion
	layoutPath, err := layout.FromPath(path)
    if os.IsNotExist(err) {
        // If the layout doesn't exist, create a new one
        layoutPath, err = layout.Write(path, empty.Index)
    }
    if err != nil {
        return err
    }
```

pkg/oci/layout/write.go
```diff
@@ -86,8 +106,17 @@ func isEmpty(s oci.Signatures) bool {
...
+func appendImage(path layout.Path, img v1.Image, ref name.Reference, annotation string) error {
 	return path.AppendImage(img, layout.WithAnnotations(
-		map[string]string{kindAnnotation: annotation},
-	))
+        map[string]string{KindAnnotation: annotation, ImageTitleAnnotation: getImageRef(ref)},
+    ))
 }
+
+func getImageRef(ref name.Reference) string {
+	registry := ref.Context().RegistryStr() + "/"
```
dweomer, line 116: panic on nil

pkg/oci/layout/write.go
```diff
@@ -17,51 +17,64 @@ package layout
...
+func WriteSignedImage(path string, si oci.SignedImage, ref name.Reference) error {
+	layoutPath, err := layout.FromPath(path)
+    if os.IsNotExist(err) {
+        // If the layout doesn't exist, create a new one
+        layoutPath, err = layout.Write(path, empty.Index)
+    }
+    if err != nil {
+        return err
+    }
+    }
```
dweomer, line 40: oops, bracket spacing. i blame my suggestion!

dweomer, line 40:
```suggestion
    return err
}
```

pkg/oci/layout/write.go
amartin120: Adjusted.  Thanks.

pkg/oci/layout/write.go
amartin120: Adjusted.  Thanks.

priyawadhwa (commented):
Hey @amartin120 thanks for contributing this! I had a couple thoughts:

1) Instead of a `--registry` flag on load, maybe it would make sense to take a regex on the image name? That way you could say something like "load all the images" or "load all the images from a specific registry"
2) Would you mind adding an e2e test similar to [this one](https://github.com/sigstore/cosign/blob/main/test/e2e_test.go#L1645) to test this new functionality? 

I wasn't able to find anything specific in the oci-layout spec which says that storing multiple images in `index.json` is valid. If you have any pointers for that it would be helpful. Otherwise though, I read through this [github thread](https://github.com/opencontainers/image-spec/issues/811) and it seems reasonable to me.

Let me know if you have any questions. Thank you!

amartin120:
Hello @priyawadhwa.  Thanks for the feedback!

> 1. Instead of a `--registry` flag on load, maybe it would make sense to take a regex on the image name? That way you could say something like "load all the images" or "load all the images from a specific registry"

I'm looking at this from a scenario where I'm ...

- saving images from "Registry A"
- tar/compressing the directory used as the output
- walking it over to an air-gapped environment
- loading it all to "Registry B" and maybe even a subsequent call to load "Registry C".

So with that I kind of need to specify the registry when the load command is called.  Does that help?

I'll get to work on those e2e test a.s.a.p.

priyawadhwa:
@amartin120 thanks for the explanation, I think I misunderstood before what the `--registry` flag was for!

We have this environment variable we use to signal pushing signatures to a different registry called `COSIGN_REPOSITORY` -- https://github.com/sigstore/cosign#specifying-registry

WDYT of reusing that instead of the flag?

amartin120:
I'll take a look at `COSIGN_REPOSITORY `.  Thanks for the info!

amartin120:
@priyawadhwa I apologize for the delay.  `COSIGN_REPOSITORY` seems like a good option for a single repository.  However, this PR allows for loading multiple images in a single `cosign load` execution which could load to multiple repositories, just in another registry.

We could still reuse `COSIGN_REPOSITORY` if we could just put a registry name in there instead of a full repository designation.  Would this still make sense over a separate registry flag?  

Hopefully I'm understanding things correctly here.  Thanks for your patience.

priyawadhwa:
@amartin120 no worries, I think we can just go ah we can keep the flag since it seems more accurately named and more intuitive. I can take a look at this once CI is green!

amartin120:
@priyawadhwa Thank you.  I may need some help with the CI.  Both of the failing stages look like timeouts as opposed to true failures.  Am I reading the logs right?

priyawadhwa:
@amartin120 I think the linter is picking up some things that need to be fixed! If you go through the files on this PR it should be visible in the Github UI:

```
Error: S1017: should replace this `if` statement with an unconditional `strings.TrimPrefix` (gosimple)
  Error: ineffectual assignment to err (ineffassign)
  Error: ineffectual assignment to err (ineffassign)
  Error: ineffectual assignment to err (ineffassign)
  Error: SA[40](https://github.com/sigstore/cosign/actions/runs/6551101939/job/18094664470?pr=3239#step:4:42)06: this value of `err` is never used (staticcheck)
  Error: SA4006: this value of `err` is never used (staticcheck)
  Error: SA4006: this value of `err` is never used (staticcheck)
```

amartin120:
Strange.  Those were not there when I responded originally.  Maybe a Github hiccup at the time?  Maybe I was just blind?  The logs just showed timeouts.  Oh well.  I'll work on this today now that I can see them.  Thanks again for the patience.

dweomer:
Oh, hey! All checks are all passing. Any objections to merging this @priyawadhwa?

cmd/cosign/cli/load.go
```diff
@@ -32,13 +32,19 @@ func Load() *cobra.Command {
...
-		Use:              "load",
-		Short:            "Load a signed image on disk to a remote registry",
-		Long:             "Load a signed image on disk to a remote registry",
-		Example:          `  cosign load --dir <path to directory> <IMAGE>`,
-		Args:             cobra.ExactArgs(1),
+		Use:     "load",
+		Short:   "Load a signed image on disk to a remote registry",
+		Long:    "Load a signed image on disk to a remote registry",
+		Example: `  cosign load --dir <path to directory> <IMAGE> OR cosign load --dir <path to directory> --registry <REGISTRY>`,
+		//Args:             cobra.ExactArgs(1),
```
priyawadhwa, line 39: nit: we should probably delete this

cmd/cosign/cli/load.go
```diff
@@ -48,12 +54,21 @@ func Load() *cobra.Command {
 }
 
 func LoadCmd(ctx context.Context, opts options.LoadOptions, imageRef string) error {
-	ref, err := name.ParseReference(imageRef)
-	if err != nil {
-		return fmt.Errorf("parsing image name %s: %w", imageRef, err)
+	if opts.Registry.Name != "" && imageRef != "" {
```
priyawadhwa, line 57: Let's move this up into `RunE` where we do the rest of the flag/arg validation

pkg/oci/remote/write.go
```diff
@@ -91,6 +92,109 @@ func WriteSignedImageIndexImages(ref name.Reference, sii oci.SignedImageIndex, o
 	return nil
 }
 
+// WriteSignedImageIndexImagesBulk writes the images within the image index.
+// Bulk version.  Uses targetRegistry for multiple images/sigs/atts.
+// This includes the signed image and associated signatures in the image index
+func WriteSignedImageIndexImagesBulk(targetRegistry string, sii oci.SignedImageIndex, opts ...Option) error {
```
priyawadhwa, line 98: I wonder if we could refactor the [copy command](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/copy/copy.go#L41) here, which will handle copying over the image and any signatures/attestations associated with it, rather than having to do it manually here.

Right now the `copy` takes a source image and a destination image as a string. We could look into pulling out the copy logic into a reusable function which takes a `SignedEntity` and passing through to that here?

pkg/oci/remote/write.go
```diff
@@ -91,6 +92,109 @@ func WriteSignedImageIndexImages(ref name.Reference, sii oci.SignedImageIndex, o
...
+	// loop through all of the items in the manifest
+	manifest, err := sii.IndexManifest()
+	if err != nil {
+		return err
+	}
+	for _, m := range manifest.Manifests {
+		// write image index if exists
+		if val, ok := m.Annotations[layout.KindAnnotation]; ok && val == layout.ImageIndexAnnotation {
+			imgTitle := m.Annotations[layout.ImageRefAnnotation]
+			fmt.Println(imgTitle)
```
priyawadhwa, line 108: nit: leftover debugging

pkg/oci/remote/write.go
```diff
@@ -91,6 +92,109 @@ func WriteSignedImageIndexImages(ref name.Reference, sii oci.SignedImageIndex, o
...
+				if err := remote.WriteIndex(ref, si, o.ROpt...); err != nil {
+					return fmt.Errorf("writing index: %w", err)
+				}
+			}
+		}
+
+		// write any images
+		if val, ok := m.Annotations[layout.KindAnnotation]; ok && val == layout.ImageAnnotation {
+			imgTitle := m.Annotations[layout.ImageRefAnnotation]
+			fmt.Println(imgTitle)
```
priyawadhwa, line 129: nit: leftover debugging

priyawadhwa (commented):
Thanks for your patience, I left a few comments!

pkg/oci/remote/write.go
amartin120: Good thoughts.  Let me look into this more.

amartin120: My only concern for the copy command logic is the usabiltiy in an air-gapped situation where cosign may not be able to reach the source image in its original registry/repo.  Will the `copy` mechanics work if the `SignedEnitity` is what's on local disk via `cosign save`?

3pings:
This PR is almost a year old with no activity.  Bumping for visibility to see if we can get some traction.  This is a much needed feature to help bloat when using multiple images.

Hayden-IO:
Moving this to draft.

Hayden-IO:
Going to close this as we work towards a v3 release. Feel free to reopen at a later point. I'll also note that signatures and other verification metadata will now be stored in referring artifacts by default, which means we need to update `save`.

PR #3240, opened 2023-09-14
Update .ko.yaml

loosebazooka:
Use newer debian12 images explicitly

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # We need a shell for a lot of redirection/piping to work
-defaultBaseImage: gcr.io/distroless/base:debug-nonroot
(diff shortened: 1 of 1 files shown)


loosebazooka:
cc: @jku

haydentherapper:
@cpanato Do you know if we need a shell? the debug option for distroless adds that, I'm thinking we can remove that for the base images

cpanato:
> @cpanato Do you know if we need a shell? the debug option for distroless adds that, I'm thinking we can remove that for the base images

we dont, only if the user need to have a long running thing to run several commands
i think it is ok to remove

trevrosen (approved):

loosebazooka:
uh so should this not use `:debug...` ?

haydentherapper:
Let's open another PR. Since we aren't using libc or libssl, do we want to just switch to static?

PR #3243, opened 2023-09-15
Stop using deprecated in_toto.ProvenanceStatement

lcarva:
#### Summary
The struct `in_toto.ProvenanceStatement` is marked as deprecated. Instead, it is recommened to use a version-specific struct. The `in_toto.ProvenanceStatementSLSA02` is the equivalent.

#### Release Note
* When generating a SLSA v0.2 statement, `attestation.GenerateStatement` now returns an object of type `in_toto.ProvenanceStatementSLSA02` instead of the deprecated `in_toto.ProvenanceStatement`.

#### Documentation

NONE


haydentherapper (approved):
Thanks!

Issue #3110, opened 2023-07-13
Support outputting Rekor response on signing 


haydentherapper:
**Description**

https://github.com/sigstore/cosign/issues/2904 proposed support for attaching a Rekor response/bundle to a container. However, there's no support for outputting a bundle with `cosign sign` ([code](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign.go#L266-L298)), only `sign-blob` ([code](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign_blob.go#L123-L140)). This supports use-cases where a signer does not want to attach metadata immediately to the container.

PR #3248, opened 2023-09-17
feat: add support outputting rekor response on signing

vishal-chdhry:
#### Summary

Closes: #3110

Previously, there's no support for outputting a bundle with `cosign sign` ([code](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign.go#L266-L298)), only `sign-blob` ([code](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign_blob.go#L135-L152)). This PR adds support for outputting rekor response on signing. This supports use-cases where a signer does not want to attach metadata immediately to the container.

#### Release Note

Added support outputting rekor response on signing with `cosign sign`

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -620,7 +620,7 @@ func (c *SignerVerifier) Bytes(ctx context.Context) ([]byte, error) {
 }
 
 func fetchLocalSignedPayload(sig oci.Signature) (*cosign.LocalSignedPayload, error) {
-	var signedPayload *cosign.LocalSignedPayload
+	signedPayload := &cosign.LocalSignedPayload{}
 	var err error
 	signedPayload.Base64Signature, err = sig.Base64Signature()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/sign/sign.go
```diff
@@ -591,3 +617,33 @@ func (c *SignerVerifier) Bytes(ctx context.Context) ([]byte, error) {
 	}
 	return pemBytes, nil
 }
+
+func fetchSignedPayload(sig oci.Signature) (*cosign.SignedPayload, error) {
```
haydentherapper, line 621: Can this return a `LocalSignedPayload` instead to be in sync with sign-blob? It should be loadable with https://github.com/sigstore/cosign/blob/main/pkg/cosign/fetch.go#L199.

haydentherapper (changes requested):
Can you add a test for this too?

vishal-chdhry:
@haydentherapper I have added a test and updated the logic, can you please review it

Besides, I cannot run/debug tests in `test/e2e_test.go`. I think `gopls` is not working for me in that file, autocomplete does not work either. Is there anything I need to configure? I am using VSCode.

test/e2e_test.go
```diff
@@ -951,6 +951,51 @@ func TestRekorBundle(t *testing.T) {
 	must(verify(pubKeyPath, imgName, true, nil, ""), t)
 }
 
+func TestRekorOutput(t *testing.T) {
+	// turn on the tlog
+	defer setenv(t, env.VariableExperimental.String(), "1")()
```
haydentherapper, line 956: Can you remove, the experimental flag is no longer used?

haydentherapper (dismissed):
One tiny comment, looks good!

vishal-chdhry:
@haydentherapper Removed the flag

haydentherapper (approved):

PR #3250, opened 2023-09-19
Update ko.yaml to use different base image

haydentherapper:
Switching to static since libssl and glibc are not needed, and removing debug since a shell is not needed.


haydentherapper:
@cpanato Just wanted to double check, I removed the comment saying we need a shell, this is no longer needed, correct?

cpanato:
yes, but let me run a couple of tests :)

cpanato (approved):

PR #3272, opened 2023-10-03
Add guard flag for experimental OCI 1.1 verify.

vaikas:
#### Summary
Previously the experimental behaviour for verify defaulted to using experimental OCI 1.1 behvaiour. This seems to cause some problems with some registries. So default that behaviour to false, and you have to explicitly enable it with a flag.
Also, by adding it to checkopts, we can now flip this on/off for clients like policy-controller where the problems surfaced.


cpanato (approved):
Thanks

hectorj2f (approved):
LGTM

PR #3292, opened 2023-10-11
Fix ReplaceSignatures

jonjohnsonjr:
This was correctly replacing the base of the signatures but was also
appending those signatures to the Get() response. So the v1.Image
representation was getting out of sync with the oci.Signatures
representation.

Normally you won't see this because cosign doesn't do multiple
attestations in the same command for the same image, so you'd roundtrip
the signatures through the v1.Image being pushed and pulled, which
corrects the discrepancy.

If you (specifically, me) attempt to attach an attestation multiple
times, the Get() will get very out of sync with the v1.Image and Replace
no longer does what you'd expect because it's operating on incorrect
Get() results instead of directly on the v1.Image representation.

### DIFF
--- a/pkg/oci/mutate/mutate_test.go
+++ b/pkg/oci/mutate/mutate_test.go
@@ -351,7 +351,7 @@ func TestSignEntity(t *testing.T) {
 				if al, err := atts.Get(); err != nil {
 					t.Fatalf("Get() = %v", err)
 				} else if len(al) != 1 {
-					t.Errorf("len(Get()) = %d, wanted %d", len(al), i)
+					t.Errorf("len(Get()) = %d, wanted %d", len(al), 1)
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
cc @developer-guy I might be missing something

hectorj2f:
@jonjohnsonjr I assume it is related to this issue https://github.com/sigstore/cosign/issues/923 but it isn't doing what it was meant to do. You have removed the code and the test is still passing 🤔 https://github.com/sigstore/cosign/pull/1366/files#diff-9e762d0df1502ee31a1de1a3bc67be6c2707ec3e6117be3147907a60ef5fb854R260.

hectorj2f:
I can see how the replace tests are failing now.

jonjohnsonjr:
Added a test, without my diff it fails with:

```
--- FAIL: TestSignEntity (0.01s)
    --- FAIL: TestSignEntity/with_replace_op_(attestation) (0.00s)
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
        mutate_test.go:459: len(Get()) = 2, wanted 1
```

Which confirms my assertion about the duplicated `Get()` results.

hectorj2f (approved):
lgtm

PR #3310, opened 2023-10-18
feat: Support basic auth and bearer auth login to registry

tosone:
#### Summary

In some scenarios, user wants to login the registry in oneline command. trivy can be used like this: `trivy image --registry-token xx image`

#### Release Note

Support basic auth and bearer auth login to registry

### DIFF
--- a/cmd/cosign/cli/options/registry.go
+++ b/cmd/cosign/cli/options/registry.go
@@ -44,6 +44,7 @@ type RegistryOptions struct {
 	KubernetesKeychain bool
 	RefOpts            ReferenceOptions
 	Keychain           Keychain
+	AuthConfig         authn.AuthConfig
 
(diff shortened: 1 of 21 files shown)


tosone:
@haydentherapper Could you please review this pr?

cpanato (commented):
it lgtm
but wondering if we can rely on the existing auth like when you do `docker login` ...

tosone:
Yes, `github.com/google/go-containerregistry` can read the auth info from `docker login`, but in some scenarios, I have no the basic auth(username/password) info, only the bearer token, `gcr` also supports bearer token access registry.
I have some code snippets that need this feat. https://github.com/go-sigma/sigma/blob/main/pkg/signing/cosign/sign/sign.go#L82

tosone:
@haydentherapper gently ping.

bobcallaway (approved):

PR #3315, opened 2023-10-21
Support ReplaceOp in Signatures

jonjohnsonjr:
Before this change, calling WithReplaceOp didn't do anything for signatures and only affected attestations. I believe this is because the implementation of Dedupe and Replace was copied to 6 different methods, so they weren't getting updated.

I've extracted the core of the impl into a dedupeAndReplace method on opts, which makes each impl just a one line call with either signatures or attestations passed in.


cpanato (approved):
lgtm

mattmoor (approved):

PR #3355, opened 2023-11-09
DNM: Increase timeout of e2e tests

haydentherapper:

### DIFF
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -50,7 +50,7 @@ echo "running tests"
 
 popd
 go build -o cosign ./cmd/cosign
-go test -tags=e2e -race $(go list ./... | grep -v third_party/)
+go test -timeout 15m -tags=e2e -race $(go list ./... | grep -v third_party/)
 
 # Test `cosign dockerfile verify`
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):

haydentherapper:
I'm thinking what happened is the two new tests, which take 30 seconds each, pushed us over the line of 10 minutes for all tests. Obviously this is not acceptable, we need to revisit not just the e2e tests but the unit tests too.

haydentherapper:
@hectorj2f When I run this locally, each new e2e test takes about 30 seconds, so I would expect to see an extra minute added. Normally the GHA step for running the tests takes about 17 minutes, but it looks like it's taking >25. Any ideas? I had thought the registry might not be faked for e2e tests, but it appears to be.

haydentherapper:
Previous test runs took ~140s, this one took ~890s.

lukehinds (approved):

hectorj2f:
@haydentherapper We are using a fake registry but the size of the machine and the other running tests in this suite might cause this issues.

hectorj2f:
Perhaps we can run the excessive signatures tests separately from the others.

PR #3369, opened 2023-11-17
add --private-infrastructure flag

### LINKED ISSUE #2736: Improvements to verification for private repositories without log upload
**Description**

I received some feedback that it's unexpected you need to explicitly opt out of tlog verification for a container hosted on a private repository when you don't have to explicitly opt out for upload during signing. I'm not sure there is a good way to improve this, but filing this issue so we can discuss more.

Some ideas:
* Alias the `insecure-tlog-skip` flag to something like `private-repository` which would result in the same behavior. I don't really love this though because it is technically "insecure" to skip uploading to a tlog for private artifacts, you aren't getting the auditability Sigstore recommends.
* Detect if the artifact is private during verification, and skip tlog verification if so. However, we then miss the case of an upload occurring on signing. We could require the user to set a flag to force tlog verification, but this isn't secure by default. 
* Do nothing, and just document this.

dylrich:
Partially fixes #2736

#### Summary
This commit adds the `--private-infrastructure` command line flag for verification commands. This flag is an alias for `--insecure-ignore-tlog` with the exception that it also silences the warning message the later flag prints. This flag is intended for users who do not rely on a public transparency log and have their own private infrastructure dedicated to verification.

I decided to make the `--private-infrastructure` flag automatically set the corresponding `--insecure-ignore-tlog` option in order to preserve the warning that is printed only when the later flag is used, but I am open to other approaches if desired!

I added a couple of calls using this flag instead of the `--insecure-ignore-tlog` flag to the e2e tests for `verify` and `verify-blob` commands. This change also affects the `verify-attestation` and `verify-blob-attestation` commands but I couldn't find e2e tests for these, so I left them untested. 

#### Release Note
* Added the `--private-infrastructure` flag for verification commands as an alias for `--insecure-ignore-tlog`

#### Documentation
I will open a PR in the Sigstore docs repo once someone has had a chance to give the initial OK on this change!

### DIFF
--- a/cmd/cosign/cli/options/verify.go
+++ b/cmd/cosign/cli/options/verify.go
@@ -28,7 +28,8 @@ type CommonVerifyOptions struct {
 	MaxWorkers       int
 	// This is added to CommonVerifyOptions to provide a path to support
 	// it for other verify options.
-	ExperimentalOCI11 bool
+	ExperimentalOCI11     bool
+	PrivateInfrastructure bool
 }
 
 func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
@@ -42,6 +43,9 @@ func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.IgnoreTlog, "insecure-ignore-tlog", false,
 		"ignore transparency log verification, to be used when an artifact signature has not been uploaded to the transparency log. Artifacts "+
 			"cannot be publicly verified when not included in a log")
+
+	cmd.Flags().BoolVar(&o.PrivateInfrastructure, "private-infrastructure", false,
+		"skip transparency log verification when verifying artifacts in a privately deployed infrastructure")
 
 	cmd.Flags().BoolVar(&o.ExperimentalOCI11, "experimental-oci11", false,
 		"set to true to enable experimental OCI 1.1 behaviour")
--- a/cmd/cosign/cli/verify.go
+++ b/cmd/cosign/cli/verify.go
@@ -89,6 +89,10 @@ against the transparency log.`,
 		Args:             cobra.MinimumNArgs(1),
 		PersistentPreRun: options.BindViper,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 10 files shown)


haydentherapper (commented):
Thanks for adding this! Can we use pflag aliases described in https://pkg.go.dev/github.com/spf13/pflag#section-readme:~:text=myFlagSet.SetNormalizeFunc(wordSepNormalizeFunc)-,Example%20%232%3A,-You%20want%20to? We've done that for certificate flag names - https://github.com/sigstore/cosign/blob/f57aa2ce0fbbed257e24eaa2cff5f777ff75abd0/cmd/cosign/cli/commands.go#L37-L53

dylrich:
I had tried looking at that originally but couldn't figure out how to differentiate between the two later on for the purpose of hiding the warning in the new flag but not the other. If you think it's OK to just remove the warning altogether I can refactor? Or maybe I am just missing something in the pflag docs that would support what I was trying to do.

cmd/cosign/cli/options/verify.go
```diff
@@ -43,6 +44,9 @@ func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
 		"ignore transparency log verification, to be used when an artifact signature has not been uploaded to the transparency log. Artifacts "+
 			"cannot be publicly verified when not included in a log")
 
+	cmd.Flags().BoolVar(&o.PrivateInfrastructure, "private-infrastructure", false,
+		"an alias for --insecure-ignore-tlog")
```
haydentherapper, line 48: Given we don’t want to refer to this as insecure, can you add a description instead, something like “skip transparency log verification when verifying artifacts in a privately deployed infrastructure” or something like that

haydentherapper (commented):
Ah, no, you’re correct, I don’t think there’d be a way to differentiate. Happy with this then.

cmd/cosign/cli/options/verify.go
dylrich: That's much better! I pushed an update with this instead

haydentherapper (approved):
Thanks!

PR #3381, opened 2023-11-21
feature for issue #3380 to allow for sign/attest against remote artif…

### LINKED ISSUE #3380: Add feature to allow for signing and attestations (attest-blob/sign-blob) against remote blobs/artifacts
**Description**

The ability to sign and attest blobs is a great feature as it allows us to work with objects that are not container images. I would like to be able to sign and attest against remote artifacts that are stored in an artifact repository.

I have a PR raised that addresses this by simply checking to see if the artifactPath passed in through the cli is a uri.

aalsabag:
…acts

This PR closes #3380 . 
It allows us to sign and attest against remote artifacts (i.e. artifacts stored in artifact repositories) while not breaking the existing functionality. More can be added to this to allow for downloads via other protocols if necessary. This is currently a draft PR as I would like to show an example implementation that is functional.

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


haydentherapper:
Closing as per issue discussion

Issue #3379, opened 2023-11-21
`cosign copy` no longer copies signatures by default in 2.2.1


ioanrogers:
**Description**

We recently started using cosign at v2.2.0. We have a CI step that runs `cosign copy temp-image prod-image` which would copy the container image and signatures.
After the `cosign-installer` action got updated in our repo this week, I noticed I could no longer verify signatures on prod images because the signatures were missing from the prod registry.
I tested with `-d` using 2.2.0 and 2.2.1 and confirmed 2.2.1 doesn't copy or look for existing .sig artifacts without setting `-only sign`. 2.2.0 does copy without needing extra params.

I think this change was introduced by #3247. Should there be some default [tags](https://github.com/sigstore/cosign/blob/1a8fa534add62855c50148581b992f7bf030222c/cmd/cosign/cli/copy/copy.go#L82) if nothing is explicitly requested?

**Version**

v2.2.1

marcosbc:
Hi folks, we are also observing the same behavior. Not only does it not copy signatures by default, but it also skips all attestations and sboms.

And according to `cosign copy -h`, it should still copy everything by default if no `-only=` flag is set:

```
$ cosign copy -h
Copy the supplied container image and signatures.

Usage:
cosign copy [flags]

Examples:
  cosign copy <source image> <destination image>

  # copy a container image and its signatures
  cosign copy example.com/src:latest example.com/dest:latest

  # copy the signatures only
  cosign copy --only=sign example.com/src example.com/dest

  # copy the signatures, attestations, sbom only
  cosign copy --only=sign,att,sbom example.com/src example.com/dest

  (...)
```

haydentherapper:
That was unintentional to make any breaking changes, the behavior should be the same as before. I can take a look at a fix later, though feel free to propose a PR with the fix.

haydentherapper:
https://github.com/sigstore/cosign/pull/3409 for the fix

PR #3247, opened 2023-09-17
feat: add `--only` flag in `cosign copy` to copy sign, att & sbom

vishal-chdhry:
#### Summary

Closes: #2002

This PR will add a new `--only` flag that accept custom string array to download sboms, signatures and attestations
```bash
$ cosign copy --only=sbom,sign,att <SRC> <DST>
```
#### Release Note

Added `--only` flag in `cosign copy` to copy sign, att & sbom
#### Documentation

Ran `make docgen` to generate required docs

PR #3409, opened 2023-12-04
Fix copy without any flag set

haydentherapper:
PR #3247 added an --only flag which when set copies only a subset of metadata. By default the flag is set to an empty string, meaning that by default copy would copy nothing. This PR fixes this bug so that if only is unset, copy defaults to the same behavior as before copying everything.

Also update the tag name to be sig rather than sign for the flag.

Fixes https://github.com/sigstore/cosign/issues/3379


hectorj2f (approved):
lgtm

Issue #3462, opened 2024-01-02
feature: 'cosign verify' add flags --ca-roots and --ca-intermediates to allow multiple CA roots


dmitris:
**Description**

**Problem**
Currently you can call `cosign verify --certificate-chain` with a file that contains a single CA root certificate and possibly related intermediate certificates.  However, in a production environment with BYO CA (no Fulcio - as in [1], for example), due to the multiregion cloud redundancy, there may be **multiple** CAs that issue codesigning certificate, and it is necessary to be able to pass a certificate bundle file to `cosign verify` - otherwise the users of `cosign verify` would need to manually implement the "trial-and-error" loop calling `cosign verify` with different CA certificates passed as `--certificate-chain` values until the command succeeds!  Needless to say, this would not be a good user experience 😄 

**Proposed Solution**
add a new `--ca-roots` and `--ca-intermediates` optional flags for `cosign verify` and the related commands (`verify-attestation`, `verify-blob`, `verify-blob-attestation`) which would allow to pass a CA Roots certificate bundle PEM file as well as one for the Intermediate Certificates..  The `--ca-roots` and `--ca-intermediates` would be mutually exclusive with the `--certificate-chain` parameter - should use one or the other but not both. 

```bash
$ cosign verify --help
[...]
    --ca-roots='':
	path to a bundle file of CA Root certificates in PEM format which will be needed when building the certificate
	chains for the signing certificate. Conflicts with --certificate-chain.

    --ca-intermediates='':
	path to a bundle file of CA Intermediate certificates in PEM format which will be needed when building the certificate
	chains for the signing certificate. Conflicts with --certificate-chain.

    --certificate-chain='':
	path to a list of CA certificates in PEM format which will be needed when building the certificate chain for
	the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and
	end with the root certificate. Conflicts with --ca-roots and --ca-intermediates.
```

**Miscellaneous**
https://docs.sigstore.dev/system_config/custom_components/ indicates "a last resort" configuration option with the environment variable `SIGSTORE_ROOT_FILE`:
> This specifies an out of band PEM-encoded X.509 certificate for a custom Fulcio root certificate.

However, this doesn't help to solve the stated problem - relying on the environment variable is inconvenient and fragile in a production deployment, and the documentation talks about the file containing a single certificate, not a certificate bundle with multiple CA roots.

**References**
[...]

PR #3464, opened 2024-01-02
add --certificate-bundle flag to 'cosign verify'

dmitris:
#### Summary
Add a new `--certificate-bundle` flag to allow pass a certificate bundle PEM file with multiple CA roots. Related to issue #3462.  Current commit adds the flag to verify the CLI options.  The new flag doesn't have any effect yet (will add in follow-up PRs).

#### Release Note
* New features and improvements, including behavioural changes, UI changes and CLI changes
added --certificate-bundle flag to take a certificate bundle PEM file with multiple CA roots, exclusive with `--certificate-file`.

-->

#### Documentation
TODO: make a related sigstore/docs PR.


haydentherapper:
@dmitris Is this ready for review? Or were you looking for more initial feedback for moving it out from draft?

dmitris:
> @dmitris Is this ready for review? Or were you looking for more initial feedback for moving it out from draft?

I was waiting for the initial feedback - I think I have it now (in #3462), thanks! - I'll try to add the unit tests shortly and mark it as "ready to review".

cmd/cosign/cli/options/certificate.go
```diff
@@ -75,12 +76,18 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.CertGithubWorkflowRef, "certificate-github-workflow-ref", "",
 		"contains the ref claim from the GitHub OIDC Identity token that contains the git ref that the workflow run was based upon.")
 	// -- Cert extensions end --
+	cmd.Flags().StringVar(&o.CARoots, "ca-roots", "",
```
haydentherapper, line 79: Do we also want an ca-intermediates flag?

I think we should require either (--ca-roots + optionally --ca-intermediates) or --certificate-chain.

dmitris: I wasn't sure if we want both flags right away or to start with `--ca-roots` and follow up with `--ca-intermediates` if/when someone needs such combination.  I can add `--ca-intermediates` flag - should I do it in this PR or in a follow-up?

haydentherapper: With this implementation with only ca-roots, how can you provide chain building CA certificates? I would prefer to add both, since I think it makes documentation clearer - Either use the deprecated --certificate-chain, or use --ca-roots/intermediates

dmitris: ok, will add `--ca-intermediates` shortly.   Working to expand the functional tests for the intermediate case.
(With `--ca-roots` only, you can provide multiple CA roots certificates in a bundle file, but only "one level" roots, cannot have an intermediate certificate (with `CA: true` in the template) signed by the root certificate.)

dmitris: added `--ca-intermediates` and adjusted the functional test as well as the docs PR https://github.com/sigstore/docs/pull/291 - please take a look @haydentherapper

cmd/cosign/cli/verify/verify.go
```diff
@@ -177,29 +179,59 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-			}
-			co.RootCerts = x509.NewCertPool()
-			co.RootCerts.AddCert(chain[len(chain)-1])
-			if len(chain) > 1 {
-				co.IntermediateCerts = x509.NewCertPool()
-				for _, cert := range chain[:len(chain)-1] {
-					co.IntermediateCerts.AddCert(cert)
+		switch {
+		case c.CertChain != "":
+			{
```
haydentherapper, line 184: style nit, you can drop the surrounding `{ }` for each case statement

cmd/cosign/cli/options/certificate.go
```diff
@@ -75,12 +77,23 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.CertGithubWorkflowRef, "certificate-github-workflow-ref", "",
 		"contains the ref claim from the GitHub OIDC Identity token that contains the git ref that the workflow run was based upon.")
 	// -- Cert extensions end --
+	cmd.Flags().StringVar(&o.CAIntermediates, "ca-intermediates", "",
+		"path to a file of intermediate CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. Conflicts with --certificate-chain.")
```
haydentherapper, line 82: Should we specify that `--ca-roots` is needed, and that this is optional?

cmd/cosign/cli/verify/verify.go
```diff
@@ -177,29 +179,59 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
+						}
+					}
+				}
 			}
-			co.IntermediateCerts, err = fulcio.GetIntermediates()
-			if err != nil {
-				return fmt.Errorf("getting Fulcio intermediates: %w", err)
+		default:
+			{
+				// This performs an online fetch of the Fulcio roots. This is needed
```
haydentherapper, line 225:
```suggestion
				// This performs an online fetch of the Fulcio roots from a TUF repository. This is needed
```

haydentherapper: (just since you're changing this, updating the comment to be clearer)

pkg/cosign/verify.go
```diff
@@ -432,6 +432,16 @@ func ValidateAndUnpackCertWithChain(cert *x509.Certificate, chain []*x509.Certif
 	return ValidateAndUnpackCert(cert, co)
 }
 
+// ValidateAndUnpackCertWithCertPools creates a Verifier from a certificate. Verifies that the certificate
+// chains up to the provided root. CheckOpts should contain a pool of CA Roots and optionally the Intermediates.
+// Optionally verifies the subject and issuer of the certificate.
+func ValidateAndUnpackCertWithCertPools(cert *x509.Certificate, co *CheckOpts) (signature.Verifier, error) {
```
haydentherapper, line 438: Given this is only used in one place, can we just call `ValidateAndUnpackCert` directly? I believe if no roots are provided, verification will fail so we don't need an explicit check, though it would be good to have a unit test to confirm.

cmd/cosign/cli/options/certificate.go
```diff
@@ -75,12 +77,23 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
...
+	_ = cmd.Flags().SetAnnotation("ca-intermediates", cobra.BashCompFilenameExt, []string{"cert"})
+	cmd.Flags().StringVar(&o.CARoots, "ca-roots", "",
+		"path to a bundle file of CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. Conflicts with --certificate-chain.")
```
haydentherapper, line 86: Can we add a check that both certificate-chain and ca-roots/ints are not both set?

cmd/cosign/cli/verify/verify.go
```diff
@@ -177,29 +179,59 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		}
 	}
 	if keylessVerification(c.KeyRef, c.Sk) {
```
haydentherapper, line 181: Is there a test we could add to e2e_test.go to exercise this new code?

cmd/cosign/cli/verify/verify.go
```diff
@@ -60,6 +60,8 @@ type VerifyCommand struct {
 	CertGithubWorkflowName       string
 	CertGithubWorkflowRepository string
 	CertGithubWorkflowRef        string
+	CAIntermediates              string
```
haydentherapper, line 63: Unfortunately, we have a lot of duplication in cosign right now. We'll also need to add this to verify-blob, verify-attestation, and verify-blob-attestation. Before you add that, let's wrap up all comments here, and then it should be a straightforward copy-paste.

haydentherapper (changes requested):
Given this is a more sensitive change, cc'ing a few others for thoughts - @woodruffw @kommendorkapten PTAL.

cmd/cosign/cli/options/certificate.go
dmitris: how about this:
```go
	cmd.Flags().StringVar(&o.CAIntermediates, "ca-intermediates", "",
		"path to a file of intermediate CA certificates in PEM format which will be needed "+
			"when building the certificate chains for the signing certificate. "+
			"The flag is optional and must be used together with --ca-roots, conflicts with "+
			"--certificate-chain.")
```

cmd/cosign/cli/verify/verify.go
dmitris: will include ^^ - I wonder if we should say `This is needed for verifying keyless certificates (both online and offline) when no CA roots or certificate chain is provided`?

cmd/cosign/cli/options/certificate.go
dmitris: we already do, don't we? https://github.com/dmitris/cosign/blob/certbundle/cmd/cosign/cli/options/certificate.go#L95-L96
```go
	cmd.MarkFlagsMutuallyExclusive("ca-roots", "certificate-chain")
	cmd.MarkFlagsMutuallyExclusive("ca-intermediates", "certificate-chain")
```

cmd/cosign/cli/verify/verify.go
dmitris: yes, I know - wanted to do it in a follow-up PR to make it easier to review.  Can make an issue not to forget.

cmd/cosign/cli/verify/verify.go
dmitris: will look into this shortly

dmitris:
@haydentherapper addressed your feedback except adding a test to e2e_test.go - will do this next

pkg/cosign/verify.go
dmitris: [FYI] I tried to "combine" the two functions but was getting test failures with cosign attach flows - likely I have missed some code paths.  Backed up to the state that you reviewed for now to get into the green build zone - will try again "with smaller steps."

dmitris: I got rid of the "special" function `ValidateAndUnpackCertWithCertPools` in [44e0ff5](https://github.com/sigstore/cosign/pull/3464/commits/44e0ff5f8a94844a5b94a9d31ba51689241735e1).
Do you think I should remove this check? 
https://github.com/sigstore/cosign/pull/3464/files#diff-fc9a26a4caa1d8684a4bdd9102f187f44e1f7fce7c06a19f4185035f13317cc8R296-R298
```go
                               if co.RootCerts == nil {
                                       return errors.New("no CA roots provided to validate certificate")
                               }
```
I would think it can be good to fail earlier with possibly more explicit error what have gone wrong, to make for easier troubleshooting (if it ever gets triggered).  Let me know if you prefer this yanked @haydentherapper.  I will also look into adding or expanding the unit tests, as you mentioned above.

cmd/cosign/cli/options/certificate.go
haydentherapper: LGTM

cmd/cosign/cli/verify/verify.go
```diff
@@ -256,14 +282,24 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-			// Verify certificate with chain
-			chain, err := loadCertChainFromFileOrURL(c.CertChain)
-			if err != nil {
-				return err
-			}
-			pubKey, err = cosign.ValidateAndUnpackCertWithChain(cert, chain, co)
-			if err != nil {
-				return err
+			if c.CARoots == "" {
+				// Verify certificate with chain
```
haydentherapper, line 286: Do we also want to read in the chain from `ca-roots` and `ca-intermediates` to verify the cert?

cmd/cosign/cli/verify/verify.go
```diff
@@ -256,14 +282,24 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
+				if err != nil {
+					return err
+				}
+				pubKey, err = cosign.ValidateAndUnpackCertWithChain(cert, chain, co)
+				if err != nil {
+					return err
+				}
+			} else {
+				if co.RootCerts == nil {
+					return errors.New("no CA roots provided to validate certificate")
```
haydentherapper, line 297: I would restructure this as a switch statement though, for "ca roots and cert chain == nil", "ca roots provided", "cert chain provided". This last check should not be possible then, correct? Either you've provided no root and we fetch it automatically, or you provide a root.

haydentherapper (commented):
LGTM overall, just two small comments.

So what's left is:
* Add reading the chain from the new flags when a certificate is provided
* Refactor into switch statement if you agree
* Copy-paste this for all other verification functions and add tests

If you can add these as individual commits, that should make reviewing go fast.

The last nice-to-have would be removing the testing script and having these tests as code in e2e_test.go, because we now have fully hermetic testing with its own TSA, Fulcio, Rekor, and registry. I know this might be a bigger change so happy to circle back to this in a later PR.

cmd/cosign/cli/options/certificate.go
```diff
@@ -75,12 +77,25 @@ func (o *CertVerifyOptions) AddFlags(cmd *cobra.Command) {
...
+		"path to a file of intermediate CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. "+
+			"The flag is optional and must be used together with --ca-roots, conflicts with "+
+			"--certificate-chain.")
+	_ = cmd.Flags().SetAnnotation("ca-intermediates", cobra.BashCompFilenameExt, []string{"cert"})
+	cmd.Flags().StringVar(&o.CARoots, "ca-roots", "",
+		"path to a bundle file of CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. Conflicts with --certificate-chain.")
+	_ = cmd.Flags().SetAnnotation("ca-roots", cobra.BashCompFilenameExt, []string{"cert"})
+
```
woodruffw, line 90: I'm less familiar with cosign so please take this with a grain of salt, but: I'm personally -1 on separate state flags for intermediate and root CA certificates.

My rationale:

1. Technically speaking, there is no real distinction between a "root" and an "intermediate" certificate: some people use "root" to mean "self-signed," but it's common to have both non-self-signed and cross-issued roots in real PKIs. In other contexts "intermediate" means "part of the untrusted set", but it's not immediately clear to me whether intermediates are trusted or untrusted in cosign (CC @haydentherapper)
2. Long term, I believe the plan is to have most Sigstore clients use a common `--trusted-root` or similar flag, which will load a shared format for the roots of trust for an "entire" Sigstore deployment (including the Rekor key, CTFE key, etc.). While this doesn't preclude that effort, it does add another set of knobs that will eventually need to be deprecated (or treated as another source of state). 
3. Similarly long-term, I believe the plan is to have most Sigstore clients (including `cosign`) use Sigstore bundles as their interchange format. Bundles in turn support untrusted intermediates for the BYO PKI case, and in an unambiguous fashion (everything in the bundle is presumed untrusted). 

So, TL;DR: if the idea behind `--ca-intermediates` is for it to pass *trusted* intermediates, then IMO it should be removed and collapsed into `--ca-roots` (which IMO could be called `--trusted-certs` to minimize confusion).

On the other hand, if these certificates are meant to be *untrusted* I can see the value of this flag, but I think it might be somewhat short-lived given the trustroot + bundle progression. But maybe that isn't as far along as I think 😅

kommendorkapten: For public good Sigstore, the intermediate is "trusted", as it is provided via the TUF trust root, and it's part of the [trusted_root.json](https://github.com/sigstore/root-signing/blob/main/repository/repository/targets/cec894ad77f79b1cb324150f6363012bcef7492954f3ab9134f932e6aa2e2e20.trusted_root.json#L50), and for public good we do not allow intermediates not provided direct via the trust root.

But if `cosign` is to be used with other deployments, they can be untrusted. So with that in mind, and the bundle format is not yet supported in cosign (to carry untrusted intermediates) I think the current proposal makes sense, but it can feel a bit awkward for the public good instance.

cmd/cosign/cli/verify/verify.go
dmitris: agree - done in dcefda70 (if looks good to you, please let me know or resolve)

cmd/cosign/cli/verify/verify.go
dmitris: still TODO

cmd/cosign/cli/verify/verify.go
dmitris: (TODO - `add this to verify-blob, verify-attestation, and verify-blob-attestation`, after settling on the initial `cosign verify` case)

.github/workflows/e2e-tests.yml
```diff
@@ -81,9 +81,15 @@ jobs:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
-          go-version: '1.21' 
+          go-version: '1.21'
           check-latest: true
 
+      - uses: imjasonh/setup-crane@00c9e93efa4e1138c9a7a5c594acd6c75a2fbf0c # v0.3
```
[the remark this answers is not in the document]
dmitris: > Is this step necessary for the additional e2e test below?

Yes, `crane` is  listed among the requirements for the new e2e test - https://github.com/sigstore/cosign/pull/3464/files#diff-605c1fbbe4548795bd84031f2d7973b208f83528d765fdd07f8d4891d38bd6e3R25, and in the test/e2e_tsa_certbundle.sh test, `crane` called [her](https://github.com/sigstore/cosign/pull/3464/files#diff-605c1fbbe4548795bd84031f2d7973b208f83528d765fdd07f8d4891d38bd6e3R40) and [here](https://github.com/sigstore/cosign/pull/3464/files#diff-605c1fbbe4548795bd84031f2d7973b208f83528d765fdd07f8d4891d38bd6e3R43).

[crane](https://pkg.go.dev/github.com/google/go-containerregistry/cmd/crane) - `a tool for interacting with remote images and registries` (I think of it as a 'docker light' for creating or publishing ISO images) - is used in several functional tests under the [test](https://github.com/sigstore/cosign/tree/main/test) directory ([grep hits](https://gist.github.com/dmitris/4475bf11057260a70dac0472e4b25b37)).  In the same file, that `setup-crane` is already used twice - https://github.com/sigstore/cosign/blob/main/.github/workflows/e2e-tests.yml#L48 and https://github.com/sigstore/cosign/blob/main/.github/workflows/e2e-tests.yml#L67 without additional comments, so  I wonder if the comment would break the consistency? There are no comments for other `uses` lines either so personally, I would prefer not to add it to avoid "visual clutter" 😄 (since the purpose of it would be clear and intuitive to most people working with the functional tests and their configuration).

cmd/cosign/cli/options/certificate.go
dmitris: @woodruffw thanks for your review and suggestions! 

> Technically speaking, there is no real distinction between a "root" and an "intermediate" certificate

Well, for example looking at the Fulcio [Certificate Specification](https://github.com/sigstore/fulcio/blob/main/docs/certificate-specification.md), it has subsections for [Root](https://github.com/sigstore/fulcio/blob/main/docs/certificate-specification.md#root-certificate), [Intermediate](https://github.com/sigstore/fulcio/blob/main/docs/certificate-specification.md#intermediate-certificate), and [Issued](https://github.com/sigstore/fulcio/blob/main/docs/certificate-specification.md#issued-certificate) Certificates, with specifications of what each type of certificate `MUST` / `MUST NOT` / `SHOULD`  have. Also, for better or worth, a search for `intermediate` brings up over 140 hits in the `sigstore/cosign` codebase, including in the existing options for `cosign verify`:

> doc/cosign_verify.md:      --certificate-chain string                                                                 path to a list of CA certificates in PEM format which will be needed when building the certificate chain for the signing certificate. Must start with the parent intermediate CA certificate of the signing certificate and end with the root certificate

also over 60 matches have both `intermediate` and `root` in the same line so these terms seems both well established and useful in the `cosign` repo and code.  Maybe it is due to my 'surrounding environment' but I don't remember any cases where I or people around would have difficulties distinguishing which (CA) certificates are  root ones, and which - intermediates.

> In other contexts "intermediate" means "part of the untrusted set", but it's not immediately clear to me whether intermediates are trusted or untrusted in cosign
> [...]
> But if cosign is to be used with other deployments, they can be untrusted. 

I have to admit having difficulties understanding which certificates are (or should be) considered "trusted" and "untrusted" in cosign (there is only one search hit on `untrusted` in sigstore/cosign and it is to a CHANGELOG in https://github.com/sigstore/cosign/pull/2124; the only [search hit](https://docs.sigstore.dev/about/threat-model/) on `untrusted` in https://docs.sigstore.dev also is not very helpful).  I think the certificate trust is so much context-dependant.  For example, I'm used to think of the certificates issues by our `TheCompanyCA`  as both fully trusted and more trustworthy than those from Digicert & Co. 😄  - while in an environment oriented to the public websites and Certificate Authorities it would likely be considered as "Self-Signed" and untrusted.

Regarding the points 2. and 3.  ("use a common --trusted-root flag" and Sigstore bundles as the interchange format) - in the long term, I have some reservations regarding this in terms of the ease of use and Developer / DevOps Experience.  I worry that they would become custom requirements and make hurdles to using Sigstore for signing artifacts.  PEM certificate bundles are "universal", ubiquitous and extremely familiar to everyone working with them and also deeply interwined with the existing scripts, processes etc.  The TUF roots / Sigstore bundles are (IMO) in comparison relatively new, more complex (PEM -> JSON -> ...),  "special purpose" and require custom handling, adoption steps / timeframe and increase the adoption and operational friction, especially for large users.  Undoubtedly, the change will make life easier for us as the sigstore developers and the codebase probably nicer ("_just one file/format to rule them all!_") but possibly at the expense of placing some extra burden on the end users ("_in addition to distributing the `ca-bundle.pem` to 100,000 endpoints that you are doing now and must continue to do for many years, you also need to create the <new format> file and distribute it to all the endpoints that will or might need to verify cosign signatures_").   So I would recommend and plead 😄  to consider the trade-off carefully and if decided, let the users know well in advance about the upcoming change(s) and provide for an ample transition period (maybe something like one major version during which one could use both _old_ and _new_ ways of supplying certificates) 🙏 /cc @haydentherapper 

But as you said, this would be long-term changes - and currently we have an "acute" issue of not being able to (properly) verify cosign signatures with `cosign verify` and through the [cosign Go library](https://pkg.go.dev/github.com/sigstore/cosign) if there is more than one CA authority issuing certificates - namely  using a certificate bundle PEM file that contains  multiple CA Root certificates.  (And we would like to start issuing and verifying _lots_ of Sigstore/Cosign signatures **in Q1 '24**! 😁).

The suggested `cosign verify --ca-roots <certbundle.pem>` and the expanded [CertVerifyOptions](https://pkg.go.dev/github.com/sigstore/cosign/cmd/cosign/cli/options#CertVerifyOptions) and [VerifyCommand](https://pkg.go.dev/github.com/sigstore/cosign/cmd/cosign/cli/verify#VerifyCommand) in the Go library would solve that problem.  If you have other ideas how we could do this (in the near term), please let me know!  (I thought about expanding the current `--certificate-chain` format or adding `--certificate-chains` [plural], but couldn't figure out how to handle the file format - with the single chain, it is easy to parse the file [`<Intermediate>*<Root>`] but if there are multiple roots, it gets very messy....  I also didn't want to get into the business of building multiple certificate chains from a set of certificates - the Go stdlib already does it very well but it requires providing `Roots` and optionally `Intermediates` in the [VerifyOptions](https://pkg.go.dev/crypto/x509#VerifyOptions) mentioned above - to which the two new suggested flags would correspond.)

cmd/cosign/cli/verify/verify.go
dmitris: [TODO] will check and respond

dmitris: @haydentherapper not sure if this is still relevant - I added a comment to the `co.CARoots != nil` case: `// Verify certificate with root (and if given, intermediate) certificate`.  Both with the chain and the `ca-roots` / `ca-intermediates`, the validation is done with the same function [ValidateAndUnpackCert ](https://pkg.go.dev/github.com/sigstore/cosign/v2/pkg/cosign#ValidateAndUnpackCert), only with the certificate chain there is a step of parsing the file and loading the `CheckOpts`.  Let me know or resolve if this is ok or you would like some changes, please.

cmd/cosign/cli/verify/verify.go
dmitris: added functional tests in https://github.com/sigstore/cosign/pull/3464/commits/a25df3b890a02c1c2e0f081ec220120b211c1aef (as discussed, generating certificates on the fly using https://github.com/sigstore/cosign/blob/main/test/cert_utils.go). @haydentherapper

dmitris: @haydentherapper added functional tests in https://github.com/sigstore/cosign/pull/3464/commits/bd10952e2ecabf0f2ad20c1de98f0887c87a91b2

dmitris:
The rebase on the trunk (after the conflicts with the recent test changes) is WIP (therefore converted it to Draft for now) - the immediate goal is to incorporate the rebasing changes in https://github.com/sigstore/cosign/pull/3730 and get successful tests.  I will work on adapting to the test changes - in particular, to remove the test secret generation in `test/gencert`.

cmd/cosign/cli/verify/verify.go
```diff
@@ -186,9 +189,32 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-		} else {
-			// This performs an online fetch of the Fulcio roots. This is needed
-			// for verifying keyless certificates (both online and offline).
+		case c.CARoots != "":
+			caRoots, err := loadCertChainFromFileOrURL(c.CARoots)
+			if err != nil {
+				return err
+			}
+			co.RootCerts = x509.NewCertPool()
+			if len(caRoots) > 0 {
```
maditya, line 198: nit: should we error out if `len(caRoots)` is `0`?

dmitris: is it a "fatal" condition?  Even if `x509.NewCertPool()` returns no roots, there should can be some added to `co.RootCerts` in the first line of this function.

cmd/cosign/cli/verify/verify.go
```diff
@@ -261,14 +288,26 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
 				return err
 			}
+		case co.RootCerts != nil:
+			// Verify certificate with root (and if given, intermediate) certificate
+			pubKey, err = cosign.ValidateAndUnpackCert(cert, co)
+			if err != nil {
+				return err
+			}
+		default:
+			return errors.New("internal error in handling CertChain and RootCerts - default case should never happen")
```
haydentherapper, line 298: Can the error text be something about "no certificate chain provided to verify certificate"? Even if it's never going to happen, that error text seems more accurate.

test/helpers.go
```diff
@@ -77,6 +87,50 @@ var verify = func(keyRef, imageRef string, checkClaims bool, annotations map[str
...
+			CertChain: certChain,
+		},
+	}
+
+	args := []string{imageRef}
+
+	return cmd.Exec(context.Background(), args)
+}
+
+var verifyCertBundle = func(keyRef, caCertFile, caIntermediates, certFile, imageRef string, checkClaims bool, annotations map[string]interface{}, attachment string, skipTlogVerify bool) error {
```
haydentherapper, line 111: `certFile` is not used, is that intentional?

haydentherapper, line 111: Should we align `caCertFile` and `caIntermediates`? I think they're both files?

test/e2e_test.go
```diff
@@ -791,6 +859,348 @@ func TestAttestationRFC3161Timestamp(t *testing.T) {
...
+			if tt.wantError {
+				t.Errorf("%s - no expected error", tt.name)
+			} else {
+				t.Errorf("%s - unexpected error: %v", tt.name, err)
+			}
+		}
+	}
+}
+
+func TestAttachWithRFC3161Timestamp(t *testing.T) {
```
haydentherapper, line 1059: Is this a dup of https://github.com/sigstore/cosign/blob/7c20052077a81d667526af879ec40168899dde1f/test/e2e_attach_test.go#L169?

test/e2e_test.go
```diff
@@ -791,6 +859,348 @@ func TestAttestationRFC3161Timestamp(t *testing.T) {
...
+	// Upload it!
+	err = attach.SignatureCmd(ctx, options.RegistryOptions{}, sigRef, payloadref, pemleafRef, certchainRef, rfc3161TSRef, "", imgName)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	must(verifyKeylessTSA(imgName, file.Name(), true, true), t)
+}
+
+func TestAttachWithRekorBundle(t *testing.T) {
```
haydentherapper, line 1139: Also, is this a dup?

test/e2e_test.go
```diff
@@ -1422,6 +1832,102 @@ func TestGenerate(t *testing.T) {
 	equals(ss.Optional["foo"], "bar", t)
 }
 
+func TestUploadDownload(t *testing.T) {
```
haydentherapper, line 1835: Dup? https://github.com/sigstore/cosign/blob/7c20052077a81d667526af879ec40168899dde1f/test/e2e_attach_test.go#L310

haydentherapper (commented):
@dmitris, thanks for pushing this through. This looks good, and from comparing to the last review, reusing `ValidateAndUnpackCert` keeps this much cleaner. And the tests look great. I pointed out a few duplicate tests that are in another e2e_test file, can they be removed?

Can you add the duplicated CLI options for verify-blob, verify-attestation, and verify-blob-attestation now?

test/helpers.go
dmitris: done in ec1abdf - renamed to `caIntermediateCertFile`

test/e2e_test.go
dmitris: duplicate removed in 2f930d38

test/e2e_test.go
dmitris: duplicate removed in 2f930d38

test/e2e_test.go
dmitris: duplicate test function removed in 2f930d38, thanks

haydentherapper (approved):

haydentherapper:
Chatted offline with @dmitris who will follow up in a second PR with the updates to the other verify-* commands.

cmd/cosign/cli/verify/verify.go
dmitris: https://github.com/sigstore/cosign/issues/3759

PR #3467, opened 2024-01-03
Remove CertSubject function

haydentherapper:
There is an existing function that handles more certificate subject types.


cpanato (approved):
thanks

PR #3478, opened 2024-01-13
Use local rekor and fulcio instances in e2e tests

cmurphy:
In 70683571 the e2e tests moved from running on the locally-spun-up
rekor instance to the public instance. This means test signatures are
piling up in the public instance, and the tests may be taking longer
than they need to since they are using an external service.

This change moves back to using the local rekor instance, which the e2e
has still been spinning up even though it has been going unused. Also
now do the same for fulcio.

This PR also includes minor cleanups in the e2e tests and test script.

Depends on https://github.com/sigstore/fulcio/pull/1518
Relates to https://github.com/sigstore/sigstore-probers/issues/105

#### Release Note

NONE

#### Documentation

Test changes only.

### DIFF
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -29,6 +29,13 @@ cd rekor
 
 echo "starting services"
 docker-compose up -d
+cleanup_services() {
+    echo "cleaning up"
+    pushd $HOME/rekor
+    docker-compose down
+    popd
+}
+trap cleanup_services EXIT
 
 count=0
 
@@ -50,7 +57,7 @@ echo "running tests"
 
 popd
 go build -o cosign ./cmd/cosign
-go test -tags=e2e -race $(go list ./... | grep -v third_party/)
+go test -tags=e2e -v -race ./test/...
 
 # Test on a private registry
 echo "testing sign/verify/clean on private registry"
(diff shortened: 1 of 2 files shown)


test/e2e_test.sh
```diff
@@ -62,6 +79,8 @@ docker run -d -p 5000:5000 --restart always -e REGISTRY_STORAGE_DELETE_ENABLED=t
 export COSIGN_TEST_REPO=localhost:5000
 go test -tags=e2e -v ./test/... -run TestSignVerifyClean
 
+# Use the public instance to verify existing images and manifests
+unset SIGSTORE_REKOR_PUBLIC_KEY
```
cmurphy, line 83: Thoughts on keeping this? Should we sign local images in a private registry in order to test the below verify commands?

test/e2e_test.sh
```diff
@@ -16,41 +16,58 @@
...
-       let 'count+=1'
+        pushd $repo
+        git pull
+        popd
     fi
 done
 
+echo "starting services"
+export FULCIO_METRICS_PORT=2113
+for repo in rekor fulcio; do
```
haydentherapper, line 34: Using a local instance of Fulcio is great - We don't have any tests currently using it, and we'll need to set up a mock oidc provider, but this will let us have much more rich e2e tests that exercise the certifciate issuance path

test/e2e_test.sh
haydentherapper: Ideally this test would be entirely hermetic, would definitely improve performance, though we can investigate that as a follow up. The signature verification should be offline, though this test depends on ghcr and distroless maintaining signatures on its images.

haydentherapper (approved):

PR #3486, opened 2024-01-18
fix for issue 3476

### LINKED ISSUE #3476: cosign verify fails randomly when we have multiple signatures attached to image
**Description**

Use case impacted: An organization uses cosign to sign the software images using its own PKI infrastructure using its Root Certificate, Intermediate Certificate and Leaf Certificate.
The software is delivered to the client. Now client want to apply its own signatures using client PKI infra using clients Root cert, Intermediate cert and Leaf cert before client can use the software.
So once client apply its signature we have now two signatures attached to the image.
Signature 1: Is signed with cert chain  CompanyRootCA.crt -> CompanyIntermediate.crt -> CompanyLeaf.crt
Signature 2:  Is signed with cert chain ClientRootCA.crt -> ClientIntermediate.crt -> ClientLeaf.crt

Now if we verify the image using cosign verify using  Root certificate in the certChain argument the verification fails.

Like in the above case if we verify the image using Company root cert or Client root cert the verification fails some time and sometime it get success.

cosign verify [ttl.sh/cosign-ci/tarball/5772530a@sha256:1d4b70ca7693a3357edb48d492bee7c4aedeb6e7a74557fd47871a81d507bfdc](http://ttl.sh/cosign-ci/tarball/5772530a@sha256:1d4b70ca7693a3357edb48d492bee7c4aedeb6e7a74557fd47871a81d507bfdc) --certificate-identity-regexp='.*' --certificate-oidc-issuer-regexp='.*' --insecure-ignore-sct --insecure-ignore-tlog --cert-chain=CompanyRootCA.crt

After analysis of the code following root cause was found.
In this case as we are just passing rootCA.crt in the certChain argument following thing will happen.

showing the function stack for the issue flow 

1. cli/verify/verify.go (cosign.VerifyImageSignatures) will be called 

			verified, bundleVerified, err := cosign.VerifyImageSignatures(ctx, ref, co)
			if err != nil {
				return cosignError.WrapError(err)
			}
https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/cmd/cosign/cli/verify/verify.go#L305

2. pkg/cosign/verify.go (VerifyImageSignatures)
this will call verifySignatures function
	return verifySignatures(ctx, sigs, h, co)   
https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L527C2-L527C43     

3. pkg/cosign/verify.go (verifySignatures)
        in this function multiple worker threads are spawned to verify signatures. The same checkOps reference is now available with each worker thread. 
	t := throttler.New(workers, len(sl))
	for i, sig := range sl {
		go func(sig oci.Signature, index int) {
			sig, err := static.Copy(sig)
			if err != nil {
				t.Done(err)
				return
			}

			verified, err := VerifyImageSignature(ctx, sig, h, co)
			bundlesVerified[index] = verified
			if err != nil {
				t.Done(err)
				return
			}
			signatures[index] = sig

			t.Done(nil)
		}(sig, i)

		// wait till workers are available
		t.Throttle()
	}

https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L598C1-L620C3

4. pkg/cosign/verify.go (VerifyImageSignature) 
// VerifyImageSignature verifies a signature
func VerifyImageSignature(ctx context.Context, sig oci.Signature, h v1.Hash, co *CheckOpts) (bundleVerified bool, err error) {
	return verifyInternal(ctx, sig, h, verifyOCISignature, co)
}
https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L817C1-L820C2
5. pkg/cosign/verify.go (verifyInternal)
     in this function if co.IntermediateCerts is NULL as we have only passed the root certificate only. So each thread update  
     co.IntermediateCerts with intermediateCert pool fetched from signature it is verifying. The problem is that this can be set by any
     of the threads which get scheduled and than the remaining thread will not set it again as they will notice the intermediates have
     been set already. this will can corrupt the verification chain if different signatures have different intermediates.

		// If there is no chain annotation present, we preserve the pools set in the CheckOpts.
		if len(chain) > 0 {
			if len(chain) == 1 {
				co.IntermediateCerts = nil
			} else if co.IntermediateCerts == nil {
				// If the intermediate certs have not been loaded in by TUF
				pool := x509.NewCertPool()
				for _, cert := range chain[:len(chain)-1] {
					pool.AddCert(cert)
				}
				co.IntermediateCerts = pool
			}
		}
		verifier, err = ValidateAndUnpackCert(cert, co)
		if err != nil {
			return false, err
		}

https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L723C1-L739C4
6. pkg/cosign/verify.go (ValidateAndUnpackCert)
	chains, err := TrustedCert(cert, co.RootCerts, co.IntermediateCerts)
	if err != nil {
		return nil, err
	}
https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L242C1-L245C3
7. pkg/cosign/verify.go (TrustedCert)
   this function fails some time as the parallel thread which ever get first change to set the co.Intermediatecerts can set it own signature Intermediate certs which may not match to the rootCert passed in the CLI argument and due to which all verification fails.

func TrustedCert(cert *x509.Certificate, roots *x509.CertPool, intermediates *x509.CertPool) ([][]*x509.Certificate, error) {
	chains, err := cert.Verify(x509.VerifyOptions{
		// THIS IS IMPORTANT: WE DO NOT CHECK TIMES HERE
		// THE CERTIFICATE IS TREATED AS TRUSTED FOREVER
		// WE CHECK THAT THE SIGNATURES WERE CREATED DURING THIS WINDOW
		CurrentTime:   cert.NotBefore,
		Roots:         roots,
		Intermediates: intermediates,
		KeyUsages: []x509.ExtKeyUsage{
			x509.ExtKeyUsageCodeSigning,
		},
	})
	if err != nil {
		return nil, fmt.Errorf("cert verification failed: %w. Check your TUF root (see cosign initialize) or set a custom root with env var SIGSTORE_ROOT_FILE", err)
	}
	return chains, nil
}

https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L1319C1-L1335C2

So main issue is that 
function verifySignatures: https://github.com/sigstore/cosign/blob/402820d97997f3a29f06b99c6fc7b7d78067ce62/pkg/cosign/verify.go#L579
invokes parallel thread but passes a common checkOpts pointer to each thread.
Now in the given case where co.IntermediateCerts is Null each thread try to set the Intermediate certs from the signature it is handling. This can corrupt the verification chain.

Solution: I am proposing following solution to the issue. 

changes in func verifyInternal:
instead of modifying the co.intermediatecerts we should use a local variable of *x509.CertPool to fetch the Intermediate certs from Signature and pass this as a separate argument to function ValidateAndUnpackCert

Change in func ValidateAndUnpackCert we should pass Intermediatecerts fetched from Signature as separate argument and use it if co.intermediatecerts is nil.

   

**Version**

Issue is visible in 2.2.2 release and also in previous releases

Mukuls77:
#### Summary
The PR fixes the issue "Closes #3476"
Use case Impacted:
A company sign the artifact using its PKI using
companyRootCA->companyIntermediateCrt->CompanyLeafcert
Client download the software and want to sign it again using client PKI before it can be used using
ClientRootCA->clientIntermediatecrt->Clientleafcrt
so in this case we will have multiple signatures attached to the artifact with different Root->intermediate->leaf chains.
Now if client verify the artifact just passing the client RootCA, the verification fails randomly.

The main issue identified is that 
func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *CheckOpts) (checkedSignatures []oci.Signature, bundleVerified bool, err error) 
function invokes multiple worker thread and call the function VerifyImageSignature() by passing common CheckOpts reference.
now in downstream the function 
func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
        verifyFn signatureVerificationFn, co *CheckOpts)
try to update the co.IntermediateCerts if it is null. ( which will be the case in this scenario as we are just passing root certificate for verification)
since this function is being executed by multiple threads so it can corrupt the co.IntermediateCerts (which ever thread find it Null)
This can corrupt the verification chain , so sometime verification fails.
To solve this issue a new function is introduced 
func ValidateAndUnpackCertWithIntermediates(cert *x509.Certificate, co *CheckOpts, intermediateCerts *x509.CertPool) (signature.Verifier, error)
this has intermediate as separate argument. this new function is now called inside verifyInternal()

#### Release Note
File modified:
cosign/pkg/cosign/verify.go

#### Documentation
no change in existing API


pkg/cosign/verify.go
```diff
@@ -291,6 +291,89 @@ func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Ver
 	return verifier, nil
 }
 
+// ValidateAndUnpackCertWithIntermediates function has same logic as ValidateAndUnpackCert.
+// Main diffrence is that we can not use ValidateAndUnpackCert in multi thread invocation
+// as it uses a common reference of CheckOpts so in multi thread invocation co.intermediateCert object
+// can get overwritten my multiple threads. So to solve this ValidateAndUnpackCertWithIntermediates
+// is created which has a separate argument for Intermediate certs.
+
```
cpanato, line 299: nit: remove this blank line

cpanato (commented):
there is any tests that we can add or the existing ones cover this change?

Mukuls77:
The new function written ValidateAndUnpackCertWithIntermediates()
is covered by already existing verify testing function like
TestVerifyImageSignature()
TestVerifyImageSignatureMultipleSubs()
TestVerifyImageSignatureWithNoChain()
TestVerifyImageSignatureWithOnlyRoot()
TestVerifyImageSignatureWithMissingSub ()

also test function TestVerifyImageSignature() 
create an oci.sig with Root->Intermediate->leaf cert
and verify it using only root cert.

The only logic which existing function can not test is that we need to invoke  verification on Multiple Signature case. i am not sure if we can create a unit test for the same. But if required i can create an end to end test where we can attach  multiple signatures with different root chains and do verification using any of the root certificate. Pls suggest shall we create an end to end test case for this.

Mukuls77:
i propose to enhance existing end to end test case 
cosign/test/e2e_test_attach.sh
to attach another signature to the artifact with different root chain and verify it using root certificate only.

Mukuls77:
@haydentherapper ; @cpanato 
I have done the required changes and added the end to end test case to test multiple signatures with different root chains. kindly review

pkg/cosign/verify.go
```diff
@@ -291,6 +291,88 @@ func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Ver
 	return verifier, nil
 }
 
+// ValidateAndUnpackCertWithIntermediates function has same logic as ValidateAndUnpackCert.
+// Main diffrence is that we can not use ValidateAndUnpackCert in multi thread invocation
+// as it uses a common reference of CheckOpts so in multi thread invocation co.intermediateCert object
+// can get overwritten my multiple threads. So to solve this ValidateAndUnpackCertWithIntermediates
+// is created which has a separate argument for Intermediate certs.
+func ValidateAndUnpackCertWithIntermediates(cert *x509.Certificate, co *CheckOpts, intermediateCerts *x509.CertPool) (signature.Verifier, error) {
```
haydentherapper, line 299: To avoid duplicated code, can we call `ValidateAndUnpackCertWithIntermediates` from `ValidateAndUnpackCert`? Would change `ValidateAndUnpackCert` to just `return ValidateAndUnpackCertWithIntermediates(cert, co, co.IntermediateCerts)`

pkg/cosign/verify.go
```diff
@@ -291,6 +291,88 @@ func ValidateAndUnpackCert(cert *x509.Certificate, co *CheckOpts) (signature.Ver
...
+		for _, oid := range cert.UnhandledCriticalExtensions {
+			if !oid.Equal(cryptoutils.SANOID) {
+				unhandledExts = append(unhandledExts, oid)
+			}
+		}
+		cert.UnhandledCriticalExtensions = unhandledExts
+	}
+
+	// Now verify the cert, then the signature.
+	if intermediateCerts == nil {
```
haydentherapper, line 320: I would remove this conditional, as the purpose of this function is to pass in the intermediates that should be used for chain building.

pkg/cosign/verify.go
```diff
@@ -721,19 +803,19 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 			return false, err
 		}
 		// If there is no chain annotation present, we preserve the pools set in the CheckOpts.
+		var pool = (*x509.CertPool)(nil)
 		if len(chain) > 0 {
 			if len(chain) == 1 {
 				co.IntermediateCerts = nil
```
haydentherapper, line 809: This line needs to be updated too, we can't overwrite `co.IntermediateCerts`

pkg/cosign/verify.go
```diff
@@ -721,19 +803,19 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 			return false, err
 		}
 		// If there is no chain annotation present, we preserve the pools set in the CheckOpts.
+		var pool = (*x509.CertPool)(nil)
```
haydentherapper, line 806:
```suggestion
		var pool *x509.CertPool
```

haydentherapper: This should be sufficient to create a pointer.

pkg/cosign/verify.go
haydentherapper, line 299: We should also write a unit test for this. It just needs one, to demonstrate it uses the passed-in intermediates rather than the ones from `co` like `ValidateAndUnpackCerts`

haydentherapper (changes requested):

pkg/cosign/verify.go
Mukuls77: Agreed i will incorporate the review comment

pkg/cosign/verify.go
Mukuls77: This check is important and should be present.
we will reach the new function ValidateAndUnpackCertWithIntermediates() via two paths
1. ValidateAndUnpackCert()
2. verifyInternal()

verifyInternal is called in some cases where we dont have an intermediate present in signature but we pass that as an argument in verification in those cases co.IntermediateCerts will have a valid intermediate cert but we will not be able to set the intermediatePool argument. one scenario is verification of Blob signatures.
so we need to have this condition if the intermediate Argument is Null that set the value from  co.IntermediateCerts
I tested this by removing this condition and i could see verify blob signature test cases were failing so i put this condition back.

pkg/cosign/verify.go
Mukuls77: Agreed i will incorporate this comment

pkg/cosign/verify.go
Mukuls77: Agreed i will incorporate this comment

pkg/cosign/verify.go
Mukuls77: Agreed i have add a new Unit test case TestValidateAndUnpackCertWithIntermediatesSuccess() in this i dont pass intermediatecert in co.IntermediateCert but just pass it in the argument to the function ValidateAndUnpackCertWithIntermediates() so it pick the value from argument

Mukuls77:
@haydentherapper i have implemented the review comments. kindly review

pkg/cosign/verify.go
```diff
@@ -216,9 +216,16 @@ func verifyOCISignature(ctx context.Context, verifier signature.Verifier, sig pa
 	return verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader(payload), options.WithContext(ctx))
 }
 
-// ValidateAndUnpackCert creates a Verifier from a certificate. Veries that the certificate
-// chains up to a trusted root. Optionally verifies the subject and issuer of the certificate.
+// ValidateAndUnpackCert calls ValidateAndUnpackCertWithIntermediates() by passing intermediate
```
haydentherapper, line 219: I'd keep the previous comment, just adding the detail that the intermediate certificate chain comes from CheckOpts. You can omit the detail that it calls another function.

pkg/cosign/verify.go
haydentherapper: Ah I think I understand the issue. Correct me if I misunderstand, you're saying that co.IntermediateCerts may have been set by one of the functions that calls verifyInternal.

I think the fix would just be to pass `co.IntermediateCerts` to `ValidateAndUnpackCertWithIntermediates` on line 745 if `co.IntermediateCerts != nil`. Then you could remove this check.

Mukuls77:
@haydentherapper  i have incorporated the review comments pls review

haydentherapper (approved):
Thanks for the fix!

pkg/cosign/verify_test.go
```diff
@@ -1228,6 +1228,37 @@ func TestValidateAndUnpackCertWithIdentities(t *testing.T) {
...
+	oidcIssuer := "https://accounts.google.com"
+
+	rootCert, rootKey, _ := test.GenerateRootCa()
+	subCert, subKey, _ := test.GenerateSubordinateCa(rootCert, rootKey)
+	leafCert, _, _ := test.GenerateLeafCert(subject, oidcIssuer, subCert, subKey)
+
+	rootPool := x509.NewCertPool()
+	rootPool.AddCert(rootCert)
+	subPool := x509.NewCertPool()
+	rootPool.AddCert(subCert)
```
haydentherapper, line 1243:
```suggestion
```

haydentherapper: Sorry one little thing, this should be removed, since you're testing that a chain can be built with the intermediate in the subPool to the root in the rootPool

Mukuls77:
@haydentherapper  i have incorporated the review comment. please check

test/e2e_test_attach.sh
```diff
@@ -25,6 +25,10 @@ cp ./test/testdata/test_attach_private_key $tmp/private_key
 cp ./test/testdata/test_attach_leafcert.pem $tmp/leafcert.pem
 cp ./test/testdata/test_attach_certchain.pem $tmp/certchain.pem
 cp ./test/testdata/test_attach_rootcert.pem $tmp/rootcert.pem
+cp ./test/testdata/test_attach_second_private_key $tmp/secondprivate_key
```
haydentherapper, line 28: @Mukuls77 What is the purpose of this extra set of certificates? It looks like this is testing the same behavior, are these e2e test changes needed?

haydentherapper:
LGTM, just one comment about tests

hectorj2f (approved):
lgtm

test/e2e_test_attach.sh
Mukuls77: This test case creates the scenario when multiple signatures are attached to a artifact with different root chains and we want to verify the artifact with any one of the root certificate. because there are more than one signatures attached so multiple threads will be created during verification and each thread will verify the allocated signature. During this operation previously when threads were modifying co.intermediates sometime verification use to fail, now with this fix it will not fail. so in my view this is an important test cases of verifying multiple signatures.

Mukuls77:
@haydentherapper @hectorj2f 
if there are no further comments i request to merge the pull request.
Thanks

test/e2e_test_attach.sh
haydentherapper: Thanks, that makes sense. Can you add a comment about this in the test file?

test/e2e_test_attach.sh
```diff
@@ -59,24 +67,22 @@ echo "Payload: $PAYLOAD"
 
 ## Attach Signature, payload, cert and cert-chain
 ./cosign attach signature --signature ./payloadbase64.sig --payload ./payload.json --cert ./leafcert.pem --cert-chain ./certchain.pem $IMAGE_URI_DIGEST
+./cosign attach signature --signature ./secondpayloadbase64.sig --payload ./payload.json --cert ./secondleafcert.pem --cert-chain ./secondcertchain.pem $IMAGE_URI_DIGEST
```
haydentherapper, line 70: Can we modify this test so that we still test verification with a single attached cert, then after verification, attach another and try verification?

haydentherapper (changes requested):

Mukuls77:
@haydentherapper  i have incorporated the review comments. kindly check

haydentherapper (approved):
Thanks @Mukuls77!

ret2libc:
What is the difference between the new `ValidateAndUnpackCertWithIntermediates` and `ValidateAndUnpackCertWithChain`?

Mukuls77:
Main difference is Function ValidateAndUnpackCertWithChain() uses
intermediate cert from co.Intermediatecerts to construct the verification
chain to verify the certificates. Function
ValidateAndUnpackCertWithIntermediates()
provide the caller to pass intermediate cert as a separate argument and use
that for constructing the certificate chain for cert verification.
Function ValidateAndUnpackCertWithChain() internally calls
ValidateAndUnpackCertWithIntermediates() by passing co.Intermediatecerts in
the Intermediatecert Argument of the function.

haydentherapper:
Purpose was a nonbreaking API change. Long term there's a larger refactor to do to pass a set of read-only CA cert metadata to the verification function.

Mukuls77:
@haydentherapper  thanks for Merging the pull request. I suppose this fix will be available in 2.3.0 release, i wanted to check what is the expected timeframe of next cosign release.
Also i have one query if i see the cosign main page it shows

image

when i checked further i see cross build to macOS is failing in upload artifact step
image

so wanted to check if there is some action i can do to solve this issue,

haydentherapper:
I'll likely cut a release next week or so, there's a few open PRs we should get in first.

I've been noticing that failure too, I'm not sure the root cause, but it's not related to this PR specifically.

PR #3479, opened 2024-01-15
Add support for ED25519ph for sign/verify(-blob) commands

ret2libc:
#### Summary

Integrate https://github.com/sigstore/sigstore/pull/1595 into cosign, allowing sign-blob/verify-blob commmands to use ED25519ph as necessary.

This needs both https://github.com/sigstore/rekor/pull/1959 and https://github.com/sigstore/rekor/pull/1945

How I tested this:
```shell
# regular ecdsa + sha256
~/projects/sigstore/cosign/cosign sign-blob --rekor-url http://127.0.0.1:3000 --bundle cosign.bundle --yes msg.txt
~/projects/sigstore/cosign/cosign verify-blob --rekor-url http://127.0.0.1:3000 --bundle cosign.bundle --certificate-identity riccardo.schirone@trailofbits.com --certificate-oidc-issuer-regexp '.*google.*' msg.txt

# ed25519ph
openssl genpkey -algorithm ed25519 -out private.pem
~/projects/sigstore/cosign/cosign import-key-pair --key private.pem
~/projects/sigstore/cosign/cosign sign-blob --rekor-url http://127.0.0.1:3000 --bundle cosign.bundle --key import-cosign.key --yes msg.txt
~/projects/sigstore/cosign/cosign verify-blob --rekor-url http://127.0.0.1:3000 --bundle cosign.bundle --certificate-identity riccardo.schirone@trailofbits.com --certificate-oidc-issuer-regexp '.*google.*' --key import-cosign.pub msg.txt

# ed25519ph (with custom fulcio from https://github.com/sigstore/fulcio/pull/1517)
openssl genpkey -algorithm ed25519 -out private.pem
~/projects/sigstore/cosign/cosign import-key-pair --key private.pem
~/projects/sigstore/cosign/cosign sign-blob --fulcio-url http://127.0.0.1:5555 --rekor-url http://127.0.0.1:3000 --bundle cosign.bundle --key import-cosign.key --issue-certificate --yes msg.txt
~/projects/sigstore/cosign/cosign verify-blob --rekor-url http://127.0.0.1:3000 --bundle cosign.bundle --certificate-identity riccardo.schirone@trailofbits.com --certificate-oidc-issuer-regexp '.*google.*' --key import-cosign.pub msg.txt
```

#### Release Note

[...]

Issue #3271, opened 2023-10-02
Allow configurable signing algorithms


tetsuo-cpp:
**Description**

_I've filed similar issues under Cosign and Rekor. I realise there's a lot of overlap in maintainers, but wanted to make sure that we discuss each project that we plan to touch. Apologies if this feels a bit spammy._

Hi there! At Trail of Bits, we're looking at potentially implementing part of the [Configurable Crypto Algorithms](https://docs.google.com/document/d/1-wF1t6lmJO37BzStbyz8ZkDhNm86HyvKZSaDl7uyHxI) proposal (specifically Phase 1). I wanted to float this idea to each of the relevant Sigstore sub-projects so we can hash out the details in a more concrete way.

Across the Sigstore stack, we default to using ECDSA for signatures and SHA256 for hashing. There's more detail in the linked proposal but there are a number of motivations for wanting to customise the signatures that are generated, including paving the way for post-quantum signatures. The proposed design includes having a "supported algorithm" registry (perhaps this can go in the Protobuf specs) that outlines enumerates the approved signature/hash algorithm combinations. We specifically don't want to allow arbitrary mixing and matching of signature and hash algorithm to avoid some of the security pitfalls listed in the proposal.

For Cosign, we want to support this set of approved signing algorithms. This can be as simple as a `--signing-algorithm` flag. As a first pass, we'd like to support ECDSA with SHA256 and SHA384 as well as EdDSA, with ECDSA-SHA256 remaining as the default.

I believe that in order to change this, we may have to make changes to the [sigstore-go](https://github.com/sigstore/sigstore-go) and [sigstore](https://github.com/sigstore/sigstore) libs, but I figured that I'd make the issue here since `cosign` is going to be entrypoint for this functionality.

ret2libc:
Some notes. ECDSA is assumed in the following places:
- [cosign.GeneratePrivateKey](https://github.com/sigstore/cosign/blob/main/pkg/cosign/keys.go#L73-L75), used from `GenerateKeyPair`(called when doing `cosign generate-key-pair`) and from `signerFromNewKey` (called when doing `cosign sign-blob` and generating an ephemeral key for getting the certificate)
- [cosign.ImportKeyPair](https://github.com/sigstore/cosign/blob/main/pkg/cosign/keys.go#L78) handles only RSA, ECDSA, and ED25519 keys. This function is used in the `cosign import-key-pair` command
[...]

PR #3497, opened 2024-01-23
Add --signing-algorithm flag

ret2libc:
#### Summary

Give the user the option to choose which signing algorithm to use when generating keypairs (#3271).

Code based on https://github.com/sigstore/cosign/pull/3479 .


cmd/cosign/cli/sign/sign.go
```diff
@@ -138,7 +138,26 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
...
+	}
+
+	if !cosign.ClientAlgorithmsRegistry.HasAlgorithmDetails(ko.SigningAlgorithm) {
+		return fmt.Errorf("unsupported signing algorithm: %s", ko.SigningAlgorithm.String())
+	}
+
+	algorithmDetails, err := cosign.ClientAlgorithmsRegistry.GetAlgorithmDetails(ko.SigningAlgorithm)
+	if err != nil {
+		return err
+	}
```
ret2libc, line 153: This does not make too much sense when the user specifies `--key`. What should be the behaviour in that case? Right now a user can even specifies a e.g. RSA key with `--key` and this check would not detect it (because the signing algorithm defaults to `ECDSA_SHA2_256_NISTP256`).

How do we want cosign to behave? Shall the registry block also --key algorithms? If so, we should probably add RSA to the list of accepted client signing algorithms as that is already valid in the current state.

ret2libc:
Shall we add the `--signing-algorithm` to the verify/verify-blob commands as well?

haydentherapper (commented):
Do we need a client algorithm registry for the sign path? There are roughly three places the client interacts with a key: Generation, signing, and verification.

For generation, it makes sense that the client specify which algorithms are supported for the generated key. This can be for both ephemeral and long-lived key generation.

For the verification path, supported algorithms could be a part of the verification policy, so that makes sense to allow a user to specify a set of trusted algorithms.

For signing, I'm not sure it's needed. When a key is provided, the user is specifying that's the key they want to use (whether it was generated ephemerally or self-managed). The backend (fulcio or rekor) could choose to reject it, which will be surfaced as a response error.

haydentherapper (commented):
Overall LGTM, I'll need to do a deep dive once this is out of draft but overall this seems solid. Can we add e2e tests that exercise generation, signing and verification?

haydentherapper:
Would we want to have this only under the `--new-bundle-format` path, given that I assume we'll want to use the new sigstore-go compatible verifier? It would also decrease the number of places that changes have to be made along the sign path.

ret2libc:
> Would we want to have this only under the `--new-bundle-format` path, given that I assume we'll want to use the new sigstore-go compatible verifier?

I can check `--signing-algorithm` is only used with `--new-bundle-format` and fail with msg if old bundle format is used. However, in terms of code changes I don't think it would really affect anything else because the format of the bundle is just affecting the very last bit in sign_blob.go where we serialize the data.

ret2libc:
Ideally this would be supported for other commands as well and not only for sign-blob, but we need some extra work on those other commands first.

ret2libc:
@haydentherapper i've extended `--signing-algorithm` flag to affect also the signing-config path (which is now default), so it creates an ephemeral key of the right algorithm when used.

haydentherapper (approved):
This looks great, thank you for all the hard work on this!

PR #3499, opened 2024-01-26
Make E2E tests hermetic

cmurphy:
## Use local services for verify tests

(This is the primary change)

Update TestDockerfileVerify and TestManifestVerify to sign ephemeral
images within the tests so that the signatures can be created with and
verified from the locally running Fulcio and Rekor instances instead of
verifying images with the public Rekor instance, so that the tests no
longer depend on external services.

The images are signed using --identity-token to avoid changing the
nature of the verification tests, which were originally written to be
keyless. A mock OIDC server is provisioned to provide the token and
enable verification.

Refactoring/cleanup/supporting changes:

## Set rekor env variable in Go test suite

Move the setting of SIGSTORE_REKOR_PUBLIC_KEY from the e2e shell script
to the Go test suite, so that only the tests that need it have it set
and the shell script is doing less setup. Also remove unnecessary
instances of os.RemoveAll for temporary directories that the Go testing
framework will automatically clean up.

## Move verify tests from shell script to Go suite

Move the `cosign dockerfile verify` and `cosign manifest verify` tests
out of the shell script and into the e2e Go test suite file with all the
other tests. This makes them consistent to manage.

The initialization of fulcio roots in other tests pollutes the trust
root in the new tests, so a reset is added to the fulcioroots package
for testing only.

## Fix cleanup in E2E script

Calling trap multiple times replaces the last signal handler rather than
appending to it. This change ensures that the most recent trap includes
all previous traps so that all cleanups are executed.

## Replace os.Setenv with testing.Setenv in e2e tests

Using os.Setenv pollutes the testing environment with environment
variables with other tests. Deferring a cleanup function to unset the
variable is unnecessary because the Go testing library provides a Setenv
function that localizes the environment setting to the scope of the
test.

Also removes unused variables.

Depends on https://github.com/sigstore/fulcio/pull/1560
Relates to https://github.com/sigstore/sigstore-probers/issues/105 - by making the current e2e tests more self-contained, creating new sigstore-prober scheduled tests run against the production service instances will be more valuable.


test/e2e_test.go
```diff
@@ -1000,7 +1011,7 @@ func TestAttachWithRekorBundle(t *testing.T) {
 
 func TestRekorBundle(t *testing.T) {
 	// turn on the tlog
-	defer setenv(t, env.VariableExperimental.String(), "1")()
+	t.Setenv(env.VariableExperimental.String(), "1")
```
haydentherapper, line 1014: This and any other place in the tests that `VariableExperimental` is set can be removed, it no longer changes behavior (or if it did, that would be a bug)

test/e2e_test.go
```diff
@@ -2524,3 +2550,427 @@ func TestOffline(t *testing.T) {
...
+
+	// unset the roots that were generated for timestamp signing, they won't work here
+	err = fulcioroots.ReInit()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Use the workload github token to exchange for an OIDC token.
+	// To run tests locally, set GITHUB_TOKEN=$(gh auth token).
+	identityToken, err := getGithubOIDCToken()
```
haydentherapper, line 2571: To make this entirely hermetic, would it be possible to create a mock OIDC server (we could use https://github.com/chainguard-dev/justtrustme, or create our own like we did in https://github.com/sigstore/fulcio/blob/main/pkg/server/grpc_server_test.go#L1626) and configure Fulcio to trust that provider? That would make these tests locally executable too without fetching a token out of band.

haydentherapper (commented):
This is fantastic! Nice work!

After this is merged, we can file an issue to track moving other e2e shell script tests over to this.

test/e2e_test.go
cmurphy: I'll look more closely at those options. I had the perception that an arbitrary OIDC provider would be tricky to integrate because I thought it might require registering it in the fulcio code https://github.com/sigstore/sigstore-probers/issues/105#issuecomment-1889123163

haydentherapper: The issue we ran into with conformance tests is that we were testing against the production instance, which couldn't be configured for ephemeral OIDC providers (hence why we use a GitHub Actions token, since that's supported by the prod instance). Since you're spinning up a local instance of Fulcio, we could modify the config (like https://github.com/sigstore/fulcio/blob/main/config/config.jsn) that gets read in when Fulcio starts up so that it trusts the ephemeral OIDC provider.

cmurphy: I put the wrong link in my comment, I meant to link to this doc https://github.com/sigstore/fulcio/blob/main/docs/oidc.md#integration-guide - is that guide up to date?

haydentherapper: Ah, that guide is up to date, but you only need to follow it for adding new providers for the public deployment. You can just modify the config which is passed in via https://github.com/sigstore/fulcio/blob/main/cmd/app/serve.go#L90.

test/e2e_test.go
```diff
@@ -79,6 +84,7 @@ const (
 	serverEnv = "REKOR_SERVER"
```
haydentherapper, line 84: Can this be removed? It doesn't look like this env var is read by Cosign.

If we wanted to confirm the verification is offline, you could pass an invalid URL for Rekor to `verify`

test/e2e_test.go
```diff
@@ -1029,7 +1032,7 @@ func TestRekorBundle(t *testing.T) {
 
 	// Make sure offline verification works with bundling
 	// use rekor prod since we have hardcoded the public key
```
haydentherapper, line 1034: Related to the above comment, it looks like the `verify` function sets `IgnoreTlog` to true, so it's not checking if the entry is in the log. This is fine, we can do a pass over the e2e tests later to make sure we're covering a few key use cases. We should be good to remove this comment and line though in each test.

test/e2e_test.go
```diff
@@ -2524,3 +2541,303 @@ func TestOffline(t *testing.T) {
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			cmd := dockerfile.VerifyDockerfileCommand{
+				VerifyCommand: cliverify.VerifyCommand{
+					CertVerifyOptions: options.CertVerifyOptions{
+						CertOidcIssuer: issuer,
+						CertIdentity:   certID,
+					},
+					RekorURL:   rekorURL,
+					IgnoreTlog: true,
```
haydentherapper, line 2677: Since the tests are now hermetic, can we remove `IgnoreTlog` so it tests inclusion in the log (and for the other new tests)? I would expect the test to still pass.

haydentherapper (commented):
This looks great! Just a small comment about correcting some outdated comments.

test/e2e_test.go
cmurphy: Removed `IgnoreTlog`, but also had to add `TlogUpload` to the sign command.

test/e2e_test.go
```diff
@@ -157,6 +165,24 @@ var verifyLocal = func(keyRef, path string, checkClaims bool, annotations map[st
...
+var verifyOffline = func(keyRef, imageRef string, checkClaims bool, annotations map[string]interface{}, attachment string) error {
+	cmd := cliverify.VerifyCommand{
+		KeyRef:        keyRef,
+		RekorURL:      "notreal",
+		Offline:       true,
+		CheckClaims:   checkClaims,
+		Annotations:   sigs.AnnotationsMap{Annotations: annotations},
+		Attachment:    attachment,
+		HashAlgorithm: crypto.SHA256,
+		IgnoreTlog:    true,
```
haydentherapper, line 177: Can this be removed and the tests still pass? Otherwise we skip rekor verification

cmurphy: This is supposed to be offline only, do we want online rekor verification?

haydentherapper: The Rekor proof _should_ be uploaded to the OCI annotation, and then read in during verification and not require contacting the log. Setting "offline" to "true" and providing the fake rekor URL will force offline verification.

For what it's worth, some of these tests are a bit old so there may be cases where this was turned off incorrectly. I would try to set ignoreTlog to false and see if all tests pass. The only case where it might not is if the test set TlogUpload to false (https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/options/sign.go#L36).

cmurphy: > The Rekor proof should be uploaded to the OCI annotation

It's only uploaded if I set `TlogUpload` to true in the signing command, so I'll have to update all of those.

> I would try to set ignoreTlog to false and see if all tests pass

This is a net-new verify helper function. Should I turn off `IgnoreTlog` for all the other verify helpers as well?

haydentherapper: Ah yes, good point about having to update everything. I was thinking that TlogUpload is true by default, but that's only for the flag, not the CLI struct. Let's get this merged in then as-is and then we can circle back to improving e2e test coverage for Rekor verification later. Does that sound good?

cmurphy: Sounds good :+1:

haydentherapper (approved):

Issue #3525, opened 2024-02-05
ErrNoSignaturesFound should be used when there is no signature attached to an image.


zhaoyonghe:
**Description**

[ErrNoMatchingSignatures here](https://github.com/sigstore/cosign/blob/22fd1dc0348704b7e3eeb1c5ae9901f1e2469d1c/pkg/cosign/verify.go#L595) should be changed to [ErrNoSignaturesFound](https://github.com/sigstore/cosign/blob/v2.2.3/pkg/cosign/errors.go#L58). 

We are using cosign go function [VerifyImageSignatures](https://github.com/sigstore/cosign/blob/22fd1dc0348704b7e3eeb1c5ae9901f1e2469d1c/pkg/cosign/verify.go#L488) and we found it is not possible to differentiate it with the other `ErrNoMatchingSignatures` perfectly.

I think `ErrNoMatchingSignatures` means that there exists some signatures attached to the image but `cosign` cannot verify any of them; `ErrNoSignaturesFound` means there is no signature attached to the image. So it is better to change the first `ErrNoMatchingSignatures` to `ErrNoSignaturesFound`.

**Version**

GitVersion:    v2.2.3-5-g22fd1dc0
GitCommit:     22fd1dc0348704b7e3eeb1c5ae9901f1e2469d1c
GitTreeState:  clean
BuildDate:     2024-02-05T16:48:44Z
GoVersion:     go1.21.6
Compiler:      gc
Platform:      darwin/arm64

PR #3526, opened 2024-02-05
ErrNoSignaturesFound should be used when there is no signature attached to an image.

zhaoyonghe:
#### Summary

This PR closes https://github.com/sigstore/cosign/issues/3525.

#### Release Note

NONE

#### Documentation

No documentation change.

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -592,7 +592,7 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
 	}
 
 	if len(sl) == 0 {
-		return nil, false, &ErrNoMatchingSignatures{
+		return nil, false, &ErrNoSignaturesFound{
 			errors.New("no matching signatures"),
 		}
 	}
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -593,7 +593,7 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
 
 	if len(sl) == 0 {
 		return nil, false, &ErrNoSignaturesFound{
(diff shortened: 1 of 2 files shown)


pkg/cosign/verify.go
```diff
@@ -592,7 +592,7 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
 	}
 
 	if len(sl) == 0 {
-		return nil, false, &ErrNoMatchingSignatures{
+		return nil, false, &ErrNoSignaturesFound{
```
haydentherapper, line 595: Update error message?

haydentherapper (commented):
Any test updates?

pkg/cosign/verify.go
zhaoyonghe: updated

zhaoyonghe:
> Any test updates?

The current tests only check whether an error occurs or not. I do not see them checking the return error type and message. Therefore, this change should not break any current tests.

haydentherapper:
What about adding a test that covers this code  path?

zhaoyonghe:
> What about adding a test that covers this code path?

Yes. I can do that. Do you mean checking the [exit code](https://github.com/sigstore/cosign/blob/v2.2.3/cmd/cosign/errors/exit_codes.go) in the [test scripts](https://github.com/sigstore/cosign/tree/v2.2.3/test)? 

Now we indeed return `NoMatchingSignature` (12) in `ImageWithoutSignature` (10) cases.

haydentherapper:
It can be simpler - Add a unit test that calls `verifySignatures` with signature struct without signatures and verifies the returned error.

haydentherapper (approved):
Thanks!

PR #3528, opened 2024-02-09
Don't ignore transparency log in tests if possible

cmurphy:
Update the e2e tests to default to setting IgnoreTlog to false where possible. In some cases, where the IgnoreTlog functionality is being explicitly tested, continue to set it to true.

Since the transparency log isn't being ignored, the signing commands need to upload it and need the rekor public key and URL in order to do so.

Removes one redundant test.

Followup to https://github.com/sigstore/cosign/pull/3499#discussion_r1482170167

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -92,7 +92,7 @@ var passFunc = func(_ bool) ([]byte, error) {
 	return keyPass, nil
 }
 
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):
Thanks! Very happy to have end to end test coverage for this now.

PR #3541, opened 2024-02-15
fix semgrep issues for dgryski.semgrep-go ruleset

dmitris:
#### Summary
PR fixes the issues flagged by semgrep - `semgrep --config http://semgrep.dev/r/dgryski.semgrep-go`.
Output of `semgrep --config http://semgrep.dev/r/dgryski.semgrep-go` - see the [gist](https://gist.github.com/dmitris/198c82ff7d4c29847d26ee1a3cf3ee41).

#### Release Note
NONE

#### Documentation
n/ap

### DIFF
--- a/cmd/cosign/cli/generate/generate.go
+++ b/cmd/cosign/cli/generate/generate.go
@@ -17,7 +17,6 @@ package generate
 
 import (
 	"context"
-	"fmt"
 	"io"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -49,6 +48,6 @@ func GenerateCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef 
 	if err != nil {
 		return err
 	}
-	fmt.Fprint(w, string(json))
+	w.Write(json)
 	return nil
 }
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/generate/generate.go
```diff
@@ -49,6 +48,6 @@ func GenerateCmd(ctx context.Context, regOpts options.RegistryOptions, imageRef
 	if err != nil {
 		return err
 	}
-	fmt.Fprint(w, string(json))
+	w.Write(json)
```
haydentherapper, line 51: Is this equivalent to `w.Write(string(json))`? I assume that if this were printed to stdout, it needs to be a string?

dmitris: I thought it doesn't matter whether you write out bytes from a byte array, or the string "wrapped around" those bytes - the string conversion is an unnecessary performance degradation (if not eliminated by the compiler).
Writing bytes to stdout seems to be fine (without the `string` conversion): https://go.dev/play/p/tPrMBjXyxvq.

haydentherapper (approved):

PR #3571, opened 2024-03-02
Adds Support for Fulcio Client Credentials Flow, and Argument to Set Flow Explicitly

nkreiger:
#### Summary

Implements this:

https://github.com/sigstore/sigstore/issues/1619

---

It looks like the flag was also missing from the CLI ko.KeyOpts to set the fulcio flow explicitly. Should be 100% backwards compatible because the default was always an empty string.

#### Release Note

* Command Flag Addition - `--fulcio-auth-flow` to explicitly set the fulcio authentication flow
* API Addition, support for `--fulcio-auth-flow=client_credentials` as an authentication flow. Requires the Client ID and Client Secret.

#### Documentation

`cosign sign --fulcio_url=<url> --fulcio-auth-flow=client_credentials --client-secret=<secret file> --client-id=sigstore`


haydentherapper (approved):
Thanks! Were you able to test this out locally to confirm it's working as expected?

haydentherapper (changes requested):
Actually, can you rebase against main? There's a bunch of extra commits in this PR

nkreiger:
> Thanks! Were you able to test this out locally to confirm it's working as expected?

yes, I was!

PR #3578, opened 2024-03-07
Adds Support for Fulcio Client Credentials Flow, and Argument to Set Flow Explicitly

nkreiger:
#### Summary

Implements this:

https://github.com/sigstore/sigstore/issues/1619

---

It looks like the flag was also missing from the CLI ko.KeyOpts to set the fulcio flow explicitly. Should be 100% backwards compatible because the default was always an empty string.

#### Release Note

* Command Flag Addition - `--fulcio-auth-flow` to explicitly set the fulcio authentication flow
* API Addition, support for `--fulcio-auth-flow=client_credentials` as an authentication flow. Requires the Client ID and Client Secret.

#### Documentation

`cosign sign --fulcio_url=<url> --fulcio-auth-flow=client_credentials --client-secret=<secret file> --client-id=sigstore`


cmd/cosign/cli/options/fulcio.go
```diff
@@ -39,6 +40,9 @@ func (o *FulcioOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.IdentityToken, "identity-token", "",
 		"identity token to use for certificate from fulcio. the token or a path to a file containing the token is accepted.")
 
+	cmd.Flags().StringVar(&o.AuthFlow, "fulcio-auth-flow", "",
+		"fulcio interactive oauth2 flow to use for certificate from fulcio. Defaults to determining the flow based on the runtime environment.")
```
bobcallaway, line 44: should we list the options that are valid values in the help text?

bobcallaway:
also, please rebase now that https://github.com/sigstore/cosign/pull/3579 has merged.

cmd/cosign/cli/options/fulcio.go
nkreiger: yeah, great idea

haydentherapper (approved):

cpanato (approved):
nice

bobcallaway (approved):

PR #3583, opened 2024-03-08
bump release to use go 1.21.8

bobcallaway:


cpanato (changes requested):
Need to bump the validate release ci job as well

I have a branch for that just did not have time today

Thanks for doing this

cpanato (approved):
Thanks

The same should go for Fulcio and rekor

PR #3597, opened 2024-03-13
Updated v1 to support latest version of go-tuf

kommendorkapten:
#### Summary
Per the latest [TUF updates](https://github.com/sigstore/root-signing/pull/1164) in Sigstore Public Good instance, the key type is changing for the TUF keys, to keep cosign v1 continue to work I've updated to the latest go-tuf version.

#### Release Note
* Updated go-tuf version to v0.7.0
* Updated sigstore/sigstore to v1.8.0

#### Documentation
N/A

PR #3598, opened 2024-03-14
V1 go tuf update

kommendorkapten:
#### Summary

Per the latest https://github.com/sigstore/root-signing/pull/1164 in Sigstore Public Good instance, the key type is changing for the TUF keys, to keep cosign v1 continue to work I've updated to the latest go-tuf version.

This is the same PR as https://github.com/sigstore/cosign/pull/3597 but against `release-1.13` branch. We should probably push it to a new branch instead, but opening to get all the tests to run.

#### Release Note

* Updated go-tuf version to v0.7.0
* Updated sigstore/sigstore to v1.8.0

#### Documentation
N/A


hectorj2f (approved):

haydentherapper (approved):
@cpanato any concerns with adding this to the release-1.13 branch? and anything special we need to know about triggering a release for 1.13.2?

cpanato:
@haydentherapper nothing, will be similar from the previous one.

but this will be the last for the branch? or how long will we maintain this?

cpanato (approved):

kommendorkapten:
I would hope this would be the last update for v1, i.e we won't continue to maintain it anymore. But of course, I'm not the authoritative person to make such a decision :)

kommendorkapten:
Do we have any data on how many people are still downloading or using cosign v1?

haydentherapper:
I’m still skeptical that we need this. Kyverno was the only major customer I was aware of but they’ve upgraded to v2. We could hold off on merging until after the root is upgraded and see if there’s any complaints from v1 users?

cpanato:
Only kyverno was asking in the past.

I think we can hold this for a while

haydentherapper:
We've had a request for it, so moving forward it this.

Issue #3548, opened 2024-02-20
Upgrade to latest Sigstore TUF client


haydentherapper:
**Description**

Tracking issue for the using the new Sigstore TUF client, https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/client.go. This client adds support for using the new trusted root metadata and improves caching logic.

This removes support for the per-target `custom` metadata, ie https://github.com/sigstore/scaffolding/blob/b0d09de38f7ea4ee5939a52cebadbc7127d4e127/pkg/repo/repo.go#L44-L48, which is used for private deployments. Given this will be a breaking change in Cosign, we can either switch to this client as part of Cosign 3.0, announce deprecation and wait X months, or support both TUF clients via a flag (temporarily, we would still announce deprecation of the previous TUF client).

cc @codysoyland @kommendorkapten

haydentherapper:
A few implementation notes:
* We likely can support both TUF clients concurrently without a flag. We may be able to use the same cache folder too, needs confirmation.
* We'll need to modify how the individual service targets are fetched, as the API differs between the previous and new sigstore TUF clients
* We should update the e2e tests to generate the trusted root file and serve a local TUF repository

haydentherapper:
Another implementation note, as per https://github.com/sigstore/sigstore-go/issues/38, we can now initialize multiple clients for different repositories each with its own local cache, which covers the use case of verifying against multiple trusted roots (eg the public instance + a private instance). We can add this around the same time.

PR #3600, opened 2024-03-15
adds tsa cert chain check for env var or tuf targets.

### LINKED ISSUE #3563: Fetch TSA certificates from TUF targets when available
**Description**

The sigstore/sigstore TUF client has been updated to support the ["TSA" usage type](https://github.com/sigstore/sigstore/blob/364b1acc28de3ea95178e82d0b365036d60c6eb1/pkg/tuf/usage_type.go#L29). We currently require the TSA cert chain to be [provided by flag](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L118-L141). We can also support reading the cert chain from either an environment or the TUF targets, to be more in line with how Fulcio certs or Rekor keys can be provided on verification. See [RekorPubKeys](https://github.com/sigstore/cosign/blob/main/pkg/cosign/tlog.go#L122) as an example for how to support this. Ideally we can refactor all places where we read in the TSA cert chain via flag into a single function that reads from either a flag, env var, or TUF.

Note that as described in https://github.com/sigstore/scaffolding/issues/1001, the usage type will be deprecated in the near future, but given the simplicity of adding this feature, it's worthwhile to do now.

ianhundere:
closes #3563
#### Summary

Creates parity between Cosign / TSA (e.g. TSA values are handled similarly to ctlog, fulcio, and rekor creds now) since sigstore/sigstore TUF client was recently updated to support the ["TSA" usage type](https://github.com/sigstore/sigstore/blob/364b1acc28de3ea95178e82d0b365036d60c6eb1/pkg/tuf/usage_type.go#L29).

Currently, the TSA cert chain is required via Cosign's cli [flag](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L118-L141), though, as per https://github.com/sigstore/cosign/issues/3563, Cosign can support reading the cert chain from either an environment variable or the TUF targets, similar to Fulcio certs, Rekor keys or the CTLog public key that can be provided on verification. I looked at [RekorPubKeys](https://github.com/sigstore/cosign/blob/main/pkg/cosign/tlog.go#L122) and [GetCTLogPubs](https://github.com/sigstore/cosign/blob/main/pkg/cosign/ctlog.go#L35) as an example.

huge thanks to @aalsabag for helping w/ unit tests.

#### Release Note

* Checks for TSA cert-chain in environment variable, `SIGSTORE_TSA_CERTIFICATE_FILE`, and TUF targets

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -76,9 +76,21 @@ type VerifyCommand struct {
 	NameOptions                  []name.Option
 	Offline                      bool
 	TSACertChainPath             string
+	UseSignedTimestamps          bool
 	IgnoreTlog                   bool
 	MaxWorkers                   int
 	ExperimentalOCI11            bool
+}
+
+func (c *VerifyCommand) loadTSACertificates(ctx context.Context) (*cosign.TSACertificates, error) {
+	if c.TSACertChainPath == "" && !c.UseSignedTimestamps {
+		return nil, fmt.Errorf("TSA certificate chain path not provided and use-signed-timestamps not set")
+	}
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		return nil, fmt.Errorf("unable to load TSA certificates: %w", err)
+	}
+	return tsaCertificates, nil
 }
 
 // Exec runs the verification command
@@ -135,13 +147,11 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		co.ClaimVerifier = cosign.SimpleClaimVerifier
 	}
 
-	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
-	if err != nil {
-		if c.TSACertChainPath != "" {
-			return fmt.Errorf("unable to load TSA certificates from specified path '%s': %w", c.TSACertChainPath, err)
+	if c.TSACertChainPath != "" || c.UseSignedTimestamps {
+		tsaCertificates, err := c.loadTSACertificates(ctx)
+		if err != nil {
+			return fmt.Errorf("unable to load TSA certificates: %w", err)
 		}
-		ui.Warnf(ctx, "no TSA certificate chain path provided, or unable to load TSA certificates: %s", err.Error())
-	} else {
 		co.TSACertificate = tsaCertificates.LeafCert
 		co.TSARootCertificates = tsaCertificates.RootCert
 		co.TSAIntermediateCertificates = tsaCertificates.IntermediateCerts
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -34,7 +34,6 @@ import (
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/rekor"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/sign"
 	cosignError "github.com/sigstore/cosign/v2/cmd/cosign/errors"
-	"github.com/sigstore/cosign/v2/internal/pkg/cosign/tsa"
 	"github.com/sigstore/cosign/v2/internal/ui"
 	"github.com/sigstore/cosign/v2/pkg/blob"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
@@ -136,29 +135,16 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		co.ClaimVerifier = cosign.SimpleClaimVerifier
 	}
 
-	if c.TSACertChainPath != "" {
-		_, err := os.Stat(c.TSACertChainPath)
-		if err != nil {
-			return fmt.Errorf("unable to open timestamp certificate chain file: %w", err)
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		if c.TSACertChainPath != "" {
+			return fmt.Errorf("unable to load TSA certificates from specified path '%s': %w", c.TSACertChainPath, err)
 		}
-		// TODO: Add support for TUF certificates.
-		pemBytes, err := os.ReadFile(filepath.Clean(c.TSACertChainPath))
-		if err != nil {
-			return fmt.Errorf("error reading certification chain path file: %w", err)
-		}
-
-		leaves, intermediates, roots, err := tsa.SplitPEMCertificateChain(pemBytes)
-		if err != nil {
-			return fmt.Errorf("error splitting certificates: %w", err)
-		}
-		if len(leaves) > 1 {
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
-		}
-		co.TSAIntermediateCertificates = intermediates
-		co.TSARootCertificates = roots
+		ui.Warnf(ctx, "no TSA certificate chain path provided, or unable to load TSA certificates: %s", err.Error())
+	} else {
+		co.TSACertificate = tsaCertificates.LeafCert
+		co.TSARootCertificates = tsaCertificates.RootCert
+		co.TSAIntermediateCertificates = tsaCertificates.IntermediateCerts
 	}
 
 	if !c.IgnoreTlog {
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -67,6 +67,18 @@ type VerifyAttestationCommand struct {
 	TSACertChainPath             string
 	IgnoreTlog                   bool
 	MaxWorkers                   int
+	UseSignedTimestamps          bool
+}
+
+func (c *VerifyAttestationCommand) loadTSACertificates(ctx context.Context) (*cosign.TSACertificates, error) {
+	if c.TSACertChainPath == "" && !c.UseSignedTimestamps {
+		return nil, fmt.Errorf("TSA certificate chain path not provided and use-signed-timestamps not set")
+	}
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		return nil, fmt.Errorf("unable to load TSA certificates: %w", err)
+	}
+	return tsaCertificates, nil
 }
 
 // Exec runs the verification command
@@ -117,10 +129,11 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		}
 	}
 
-	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
-	if err != nil {
-		ui.Warnf(ctx, fmt.Sprintf("unable to load or get TSA certificates: %s", err.Error()))
-	} else {
+	if c.TSACertChainPath != "" || c.UseSignedTimestamps {
+		tsaCertificates, err := c.loadTSACertificates(ctx)
+		if err != nil {
+			return fmt.Errorf("unable to load TSA certificates: %w", err)
+		}
 		co.TSACertificate = tsaCertificates.LeafCert
 		co.TSARootCertificates = tsaCertificates.RootCert
 		co.TSAIntermediateCertificates = tsaCertificates.IntermediateCerts
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -28,7 +28,6 @@ import (
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/rekor"
-	"github.com/sigstore/cosign/v2/internal/pkg/cosign/tsa"
 	"github.com/sigstore/cosign/v2/internal/ui"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
 	"github.com/sigstore/cosign/v2/pkg/cosign/cue"
@@ -118,30 +117,15 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		}
 	}
 
-	if c.TSACertChainPath != "" {
-		_, err := os.Stat(c.TSACertChainPath)
-		if err != nil {
-			return fmt.Errorf("unable to open timestamp certificate chain file '%s: %w", c.TSACertChainPath, err)
-		}
-		// TODO: Add support for TUF certificates.
-		pemBytes, err := os.ReadFile(filepath.Clean(c.TSACertChainPath))
-		if err != nil {
-			return fmt.Errorf("error reading certification chain path file: %w", err)
-		}
-
-		leaves, intermediates, roots, err := tsa.SplitPEMCertificateChain(pemBytes)
-		if err != nil {
-			return fmt.Errorf("error splitting certificates: %w", err)
-		}
-		if len(leaves) > 1 {
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
-		}
-		co.TSAIntermediateCertificates = intermediates
-		co.TSARootCertificates = roots
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		ui.Warnf(ctx, fmt.Sprintf("unable to load or get TSA certificates: %s", err.Error()))
+	} else {
+		co.TSACertificate = tsaCertificates.LeafCert
+		co.TSARootCertificates = tsaCertificates.RootCert
+		co.TSAIntermediateCertificates = tsaCertificates.IntermediateCerts
 	}
+
 	if !c.IgnoreTlog {
 		if c.RekorURL != "" {
 			rekorClient, err := rekor.NewClient(c.RekorURL)
@@ -157,6 +141,7 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 			return fmt.Errorf("getting Rekor public keys: %w", err)
 		}
 	}
+
 	if keylessVerification(c.KeyRef, c.Sk) {
 		// This performs an online fetch of the Fulcio roots. This is needed
 		// for verifying keyless certificates (both online and offline).
@@ -169,6 +154,7 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 			return fmt.Errorf("getting Fulcio intermediates: %w", err)
 		}
 	}
+
 	keyRef := c.KeyRef
 
 	// Keys are optional!
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -63,7 +63,19 @@ type VerifyBlobCmd struct {
 	IgnoreSCT                    bool
 	SCTRef                       string
 	Offline                      bool
+	UseSignedTimestamps          bool
 	IgnoreTlog                   bool
+}
+
+func (c *VerifyBlobCmd) loadTSACertificates(ctx context.Context) (*cosign.TSACertificates, error) {
+	if c.TSACertChainPath == "" && !c.UseSignedTimestamps {
+		return nil, fmt.Errorf("either TSA certificate chain path must be provided or use-signed-timestamps must be set")
+	}
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		return nil, fmt.Errorf("unable to load TSA certificates: %w", err)
+	}
+	return tsaCertificates, nil
 }
 
 // nolint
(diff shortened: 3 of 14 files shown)


pkg/cosign/tsalog.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/sigstore/cosign/v2/internal/pkg/cosign/tsa"
+	"os"
+
+	"github.com/sigstore/cosign/v2/pkg/cosign/env"
+	"github.com/sigstore/sigstore/pkg/tuf"
+)
+
+var tsaLeafCertStr = `tsa_leaf.crt.pem`
+var tsaRootCertStr = `tsa_root.crt.pem`
+var tsaIntermediateCertStrPattern = `tsa_intermediate_%d.crt.pem`
```
cmurphy, line 17: (nit) these could be consts

pkg/cosign/tsalog.go
```diff
@@ -0,0 +1,75 @@
+package cosign
```
cmurphy, line 1: This will need a copyright header

cmurphy (commented):
This looks pretty good to me, though I'm not a TSA or TUF expert.

It looks like the merge commit accidentally pulled in an unused import so this doesn't build.

This will need some tests.

pkg/cosign/tsalog.go
```diff
@@ -0,0 +1,75 @@
...
+var tsaLeafCertStr = `tsa_leaf.crt.pem`
+var tsaRootCertStr = `tsa_root.crt.pem`
+var tsaIntermediateCertStrPattern = `tsa_intermediate_%d.crt.pem`
+
+// GetTSACerts retrieves trusted TSA certificates from the embedded or cached
+// TUF root. If expired, makes a network call to retrieve the updated targets.
+// By default, the certificates come from TUF, but you can override this for test
+// purposes by using an env variable `SIGSTORE_TSA_CERTIFICATE_FILE`. If using
+// an alternate, the file should be in PEM format.
+func GetTSACerts(ctx context.Context) (leaves [][]byte, intermediates [][]byte, roots [][]byte, err error) {
```
haydentherapper, line 24: Should we return `x509.Certificate` structs instead of byte arrays? This will cut down on repetition in the verify files if you handle unmarshalling from PEM in this function.

haydentherapper (commented):
Thanks, sorry for the delay. Just a couple comments.

+1 to tests.

ianhundere:
thanks for the feedback/comments, will try to implement in the new few weeks.

pkg/cosign/tsalog.go
ianhundere: that's a good point, i'll look into that. thanks for the feedback!

Meeki1l:
@ianhundere Can you tell me the approximate time of making all the improvements?

ianhundere:
today

ianhundere:
@haydentherapper / @cmurphy okay, i think that about does it.

haydentherapper:
Thanks @ianhundere! Will take a closer look tomorrow. Can you take a look at failing tests?

ianhundere:
> Thanks @ianhundere! Will take a closer look tomorrow. Can you take a look at failing tests?

[this](https://github.com/sigstore/cosign/pull/3600/commits/3bf357eea1638a774028f401c49a828046af6326) fixes the `TestGetTSACertsFromTUF` test and `golangci-lint` errors.

ianhundere:
@haydentherapper just a quick note, goin' on a trip for a wk tomorrow / so will be away for a bit, but can make any last minute changes today if need be. thanks again!

ianhundere:
ah, had a couple more issues:
- fixed last lint issue
- copied `splitPEMCertificateChain` from `internal/pkg/cosign/tsa/utils.go` to avoid import cycle.

now running `go test -tags=sct -covermode atomic -coverprofile coverage.txt $(go list ./... | grep -v third_party/)` works other than the current work around, `GODEBUG: x509sha1=1`, for the failing tests.

pkg/cosign/tsalog.go
haydentherapper, line 1: Can we name this and the test `tsa.go`, since this isn't a log?

pkg/cosign/env/env.go
```diff
@@ -138,6 +139,12 @@ var (
 			Sensitive:   false,
 			External:    true,
 		},
+		VariableSigstoreTSACertificateFile: {
+			Description: "path to the TSA certificate file used by Sigstore",
```
haydentherapper, line 143: Can you mention this is a concatenated list of PEM-encoded TSA leaf certificate, intermediates and root?

pkg/cosign/tsalog.go
```diff
@@ -0,0 +1,142 @@
...
+	LeafCert          *x509.Certificate
+	IntermediateCerts []*x509.Certificate
+	RootCert          []*x509.Certificate
+}
+
+type GetTargetStub func(name string) ([]byte, error)
+
+func GetTufTargets(name string) ([]byte, error) {
+	tufClient, _ := tuf.NewFromEnv(context.Background())
+	return tufClient.GetTarget(name)
```
haydentherapper, line 45: Note that this will change when we implement https://github.com/sigstore/cosign/issues/3548, because we will look up targets in a structured [trusted root](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto).

I'm OK with having a set of hardcoded names to fallback to in the meantime, but do you also want to use the `usage` type? Can we do `tufClient.GetTargetsByMeta(tuf.TSA, []string{tsaLeafCertStr, tsaRootCertStr, tsaIntermediateCertStr1, tsaIntermediateCertStr2, ...})`?

You would then call `splitPEMCertificateChain` for all of these found targets to split up the chain.

pkg/cosign/tsalog.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// GetTSACerts retrieves trusted TSA certificates from the embedded or cached
+// TUF root. If expired, makes a network call to retrieve the updated targets.
+// By default, the certificates come from TUF, but you can override this for test
+// purposes by using an env variable `SIGSTORE_TSA_CERTIFICATE_FILE` or a file path
+// specified in `TSACertChainPath`. If using an alternate, the file should be in PEM format.
+func GetTSACerts(_ context.Context, certChainPath string, fn GetTargetStub) (*TSACertificates, error) {
```
haydentherapper, line 53: should this context be plumbed through to `GetTufTargets` where `tuf.NewFromEnv` is called?

haydentherapper (changes requested):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sigstore/cosign/v2
 
-go 1.21
+go 1.22.0
```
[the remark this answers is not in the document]
ianhundere: ah, will do.

pkg/cosign/tsa.go
```diff
@@ -0,0 +1,158 @@
...
+	var raw []byte
+	var err error
+
+	switch {
+	case altTSACert != "":
+		raw, err = os.ReadFile(altTSACert)
+	case certChainPath != "":
+		raw, err = os.ReadFile(certChainPath)
+	default:
+		leafCert, err := fn(ctx, 0, []string{tsaLeafCertStr})
```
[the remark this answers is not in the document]
ianhundere: ah okay / thanks!

i'm on holiday til tues, but should be able to get these changes up before then...thanks again for prompt feedback.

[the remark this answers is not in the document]
ianhundere: @Meeki1l i've got it, but feel free to suggest changes on my incoming commit.

pkg/cosign/tsa.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(leaves) > 1 {
+		return nil, fmt.Errorf("TSA certificate chain must contain at most one TSA certificate")
+	}
+
+	return &TSACertificates{
+		LeafCert:          leaves[0],
+		IntermediateCerts: intermediates,
+		RootCert:          roots,
+	}, nil
+}
```
Meeki1l, line 137: @ianhundere  lines of code 103-116 and lines 123-136 look very similar, maybe delete 103-116 and assign raw = concatenatedPEM?

ianhundere: doh, thanks for that. 🙇

Meeki1l (approved):

ianhundere:
@haydentherapper no rush, but i think we're good / just need to run the pipeline tests once more for ✅.

cmd/cosign/cli/verify/verify.go
```diff
@@ -136,29 +135,13 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
-		}
-		co.TSAIntermediateCertificates = intermediates
-		co.TSARootCertificates = roots
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		ui.Warnf(ctx, fmt.Sprintf("unable to load or get TSA certificates: %s", err.Error()))
```
cmurphy, line 140: Why is this a warning rather than a fatal error?

It's a breaking change because if `c.TSACertChainPath` was set, then what used to be an error is now a warning. I think an error should be treated as fatal, and an empty tsa object should be treated separately, either as a warning or ignored.

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -45,6 +42,10 @@ import (
 	"github.com/sigstore/cosign/v2/pkg/policy"
 	sigs "github.com/sigstore/cosign/v2/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
+
+	"io"
+	"os"
+	"path/filepath"
```
cmurphy, line 48: These should be up with the other stdlib imports

haydentherapper:
Not sure why the tests are stuck, but might want to try to rebase off HEAD.

cmd/cosign/cli/verify/verify.go
ianhundere: something like this?

```
		if c.TSACertChainPath != "" {
			return fmt.Errorf("unable to load TSA certificates from specified path '%s': %w", c.TSACertChainPath, err)
		} else {
			ui.Warnf(ctx, "no TSA certificate chain path provided, or unable to load TSA certificates: %s", err.Error())
```

ianhundere:
> Not sure why the tests are stuck, but might want to try to rebase off HEAD.

i think the runners ran outta space: https://github.com/sigstore/cosign/actions/runs/9415519250/attempts/2

i went ahead and squashed the last commit / force pushed after rebasing again.

haydentherapper:
GitHub Actions outage - https://www.githubstatus.com/incidents/lfrlwdg67fn8

pkg/cosign/tsa.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating TUF client: %w", err)
+	}
+	var allCerts []*x509.Certificate
+	targets, err := tufClient.GetTargetsByMeta(usage, names)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching targets by metadata with usage %v: %w", usage, err)
+	}
+	for _, target := range targets {
+		certs, _, _, err := splitPEMCertificateChain(target.Target)
```
haydentherapper, line 54: I think this logic can be simplified. It looks like you're gathering all targets, unmarshalling each into an object, returning them, marshalling them back into PEM, then unmarshalling then.

Can we instead change this function to return a string which is the concatenation of all targets, split by newlines? This is equivalent to the `raw` variables.

You can remove lines 94-100. Then you'll call `splitPEMCertificateChain(raw)`, which you already do.

pkg/cosign/tsa.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading TSA certificate file: %w", err)
+	}
+
+	leaves, intermediates, roots, err := splitPEMCertificateChain(raw)
+	if err != nil {
+		return nil, fmt.Errorf("error splitting TSA certificates: %w", err)
+	}
+
+	if len(leaves) > 1 {
```
haydentherapper, line 112: Should this be `roots`? A leaf can be embedded in the RFC3161 struct.

haydentherapper: Whoops, ignore this, this is correct.

I think we should add another check that `roots > 0`.

ianhundere: you mean something like `len(roots) == 0` ?

haydentherapper (approved):
Thank you!

ianhundere:
> Thank you!

np, just a heads up that i noticed e2e tests are failing. lemme know if that's something i can fix or not.

haydentherapper:
Yea, this doesn't look like a flake, I see an error printed around TSA verification - https://github.com/sigstore/cosign/actions/runs/9473354802/job/26102612458?pr=3600

ianhundere:
> Yea, this doesn't look like a flake, I see an error printed around TSA verification - https://github.com/sigstore/cosign/actions/runs/9473354802/job/26102612458?pr=3600

ah, i never synced my fork when i rebased / oops.

ianhundere:
this one is failing due to [some infra/network issues](https://github.com/sigstore/cosign/actions/runs/9475242206/job/26106453496?pr=3600#step:8:425):
```
Error from server (parsing address pool config: invalid CIDR "fc00:f853:ccd:e793::/64.255.1-fc00:f853:ccd:e793::/64.255.250" in pool "config": invalid IP range "fc00:f853:ccd:e793::/64.255.1-fc00:f853:ccd:e793::/64.255.250": invalid start IP "fc00:f853:ccd:e793::/64.255.1"): error when creating "./metallb-crds.yaml": admission webhook "ipaddresspoolvalidationwebhook.metallb.io" denied the request
```

cmd/cosign/cli/verify/verify.go
```diff
@@ -136,29 +135,16 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		if c.TSACertChainPath != "" {
+			return fmt.Errorf("unable to load TSA certificates from specified path '%s': %w", c.TSACertChainPath, err)
 		}
-		co.TSAIntermediateCertificates = intermediates
-		co.TSARootCertificates = roots
+		ui.Warnf(ctx, "no TSA certificate chain path provided, or unable to load TSA certificates: %s", err.Error())
```
haydentherapper, line 143: Thinking more about this, I don't think we should add a warning as the vast majority of users aren't using TSA certs and this will just create noise. I'm also not a fan of silently ignoring errors though.

The issue is that we don't know when to look up TSA certs in TUF. For Rekor, we always do unless you provide a flag, and same with Fulcio, we always do unless you provide a key. For TSA certs, it's only going to be for private deployments.

WDYT about adding a flag to control this behavior? If either `--timestamp-certificate-chain=<path>` or `--use-signed-timestamps` is set, then `cosign.GetTSACerts` is invoked. You can then always fail if `err != nil`.

@cmurphy did you have any suggestions?

ianhundere: @haydentherapper should i revert back to the following for this PR ?
```
if err != nil {
	ui.Warnf(ctx, fmt.Sprintf("unable to load or get TSA certificates: %s", err.Error()))
}
```

haydentherapper: FWIW, this is solved in sigstore-go. We load all trust metadata initially, and try to verify signed material with that trust metadata. For timestamps, we only require one to be verifiable.

ianhundere: something like this?
- https://github.com/sigstore/cosign/pull/3600/commits/e02e23ffaa7ee19f30aed5fc79ce7d649ee4a410

haydentherapper: Looks good! Can you do the same for the other verify_*?

ianhundere: okay, "i think" i did that right for the other verify files, but it's late in the day here so would welcome the extra 👀:
- https://github.com/sigstore/cosign/pull/3600/commits/ff6d72b8fe98273b2bf7b4c2c6f987e22a0257fd

haydentherapper:
One last thing, need to run `make docgen`

ianhundere:
> One last thing, need to run `make docgen`

done / done 🙂

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -118,30 +129,15 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
-		}
-		if len(leaves) > 1 {
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
-		}
-		co.TSAIntermediateCertificates = intermediates
-		co.TSARootCertificates = roots
+	tsaCertificates, err := c.loadTSACertificates(ctx)
```
haydentherapper, line 132: Can we have the same behavior as the other verify_* impls? Wrap this in `if c.TSACertChainPath != "" || c.UseSignedTimestamps`, to avoid having to print a warning?

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -140,32 +140,17 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
...
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
-		}
-		co.TSAIntermediateCertificates = intermediates
-		co.TSARootCertificates = roots
+	tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
+	if err != nil {
+		ui.Warnf(ctx, fmt.Sprintf("unable to load or get TSA certificates: %s", err.Error()))
```
haydentherapper, line 149: same here

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -115,29 +126,14 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
 	if c.RFC3161TimestampPath != "" && c.KeyOpts.TSACertChainPath == "" {
```
haydentherapper, line 126: This check needs to be updated since you can use `UseSignedTimestamps` as well with `RFC3161TimestampPath`

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -140,32 +140,17 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
 	}
 
 	// Set up TSA, Fulcio roots and tlog public keys and clients.
-	if c.RFC3161TimestampPath != "" && c.KeyOpts.TSACertChainPath == "" {
+	if c.RFC3161TimestampPath != "" && c.TSACertChainPath == "" {
```
haydentherapper, line 143: same here, needs to be updated with `UseSignedTimestamp`

haydentherapper, line 143: I noticed we inconsistently use `c.KeyOpts.TSACertChainPath` vs `c.TSACertChainPath`. Can we standardize that across the verify_* impls?

ianhundere:
@haydentherapper thanks for the 👀 / lemme know if that satisfies everything.

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -140,32 +141,19 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
...
+		tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
 		if err != nil {
-			return fmt.Errorf("error splitting certificates: %w", err)
-		}
-		if len(leaves) > 1 {
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
+			ui.Warnf(ctx, fmt.Sprintf("unable to load or get TSA certificates: %s", err.Error()))
```
haydentherapper, line 151: We want to remove this warn and `return err` here, correct?

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -140,32 +141,19 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
 	}
 
 	// Set up TSA, Fulcio roots and tlog public keys and clients.
-	if c.RFC3161TimestampPath != "" && c.KeyOpts.TSACertChainPath == "" {
+	if c.RFC3161TimestampPath != "" && c.TSACertChainPath == "" {
```
haydentherapper, line 144: Need to update to use `UseSignedTimestamp`

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -118,30 +129,17 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
+		tsaCertificates, err := c.loadTSACertificates(ctx)
 		if err != nil {
-			return fmt.Errorf("error splitting certificates: %w", err)
-		}
-		if len(leaves) > 1 {
-			return fmt.Errorf("certificate chain must contain at most one TSA certificate")
-		}
-		if len(leaves) == 1 {
-			co.TSACertificate = leaves[0]
+			ui.Warnf(ctx, fmt.Sprintf("unable to load TSA certificates: %s", err.Error()))
```
haydentherapper, line 135: We want to remove this warn and return err here, correct?

haydentherapper (commented):
Just a few more places to update! Sorry for all the duplication across verify_* functions

ianhundere:
> Just a few more places to update! Sorry for all the duplication across verify_* functions

ah, i'll get those / thanks again.

haydentherapper (approved):
Thank you for all of your work on this!

cmurphy (approved):
The error handling lgtm now :+1:

haydentherapper:
@cpanato would you be able to take a look at the test failures? I’m uncertain why they’re failing

cpanato:
> @cpanato, can you take a look at the test failures? I’m uncertain why they’re failing.

I will take a look; just bear with me :) i am traveling back home today

ianhundere:
> @cpanato would you be able to take a look at the test failures? I’m uncertain why they’re failing 

maybe just update metallb as per: https://sigstore.slack.com/archives/C01DGF0G8U9/p1718367949519929?thread_ts=1718204315.549609&cid=C01DGF0G8U9

haydentherapper:
Can you rebase?

cpanato:
sorry for the delay, a rebase might fix

ianhundere:
@haydentherapper done/done 🙂

cpanato (approved):

Issue #2997, opened 2023-05-19
Add Gitpod Provider Interface


Edward13ruf:
**Description**

Gitpod is a remote workspace solution that has the ability (still in BETA) to generate JWT tokens to authenticate users within a workspace against external services like Sigstore, Vault or AWS directly.

Because gitpod can generate these tokens using a command line, and thus does not require a browser to do any auth, it makes sense to create a provider within cosign and gitsign that can automatically call this CLI on the users behalf, thus eliminating the need to do any external auth checks.

The command line in question is `gp idp token --audience sigstore` and is currently still in BETA, but can be run from any gitpod workspace. The devs are currently working to [add the correct email claims](https://github.com/gitpod-io/gitpod/issues/17530) to the resulting token so that it is compatible with sigstore.

mattmoor:
Has support for the Gitpod issuer landed in Fulcio?

Is there a Go library that cosign can use instead of relying on `os.Exec`?

cc @haydentherapper

haydentherapper:
It has not yet landed in Fulcio yet - https://github.com/sigstore/fulcio/pull/1177 

Blocked by an issue in GitPod

PR #2998, opened 2023-05-19
*Draft* Feature/gitpod

ChevronTango:
Closes #2997

#### Summary

This MR adds a provider that, when run from within a Gitpod workspace, retrieves a token automatically from the command line, which means users do not have to do any additional auth checks or config.

PR #3601, opened 2024-03-15
Support Gitpod workspaces

ChevronTango:
Closes https://github.com/sigstore/cosign/issues/2997

Summary
This PR adds a provider that, when run from within a Gitpod workspace, retrieves a token automatically from the command line, which means users do not have to do any additional auth checks or config.

This PR was previously open as https://github.com/sigstore/cosign/pull/2998 however work on Gitpod's end stalled and it took some time before changes were implemented that made this viable. These updates mean some, though not all, tokens can now be used with Sigstore, and Gitpod are correctly populating the `email_verified` field as required by Fulcio. Part of the logic of this PR has been changed to make sure we only send through tokens if they have this field and will be valid.

Release Note
Documentation

### DIFF
--- a/pkg/providers/gitpod/gitpod.go
+++ b/pkg/providers/gitpod/gitpod.go
@@ -16,7 +16,9 @@
 package gitpod
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"os/exec"
 
 	"github.com/sigstore/cosign/v2/pkg/cosign/env"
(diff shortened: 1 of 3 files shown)


pkg/providers/gitpod/gitpod.go
```diff
@@ -0,0 +1,72 @@
...
+
+var _ providers.Interface = (*gitpod)(nil)
+
+// Enabled implements providers.Interface
+func (ga *gitpod) Enabled(_ context.Context) bool {
+	// Check we are in a Gitpod Workspace
+	if env.Getenv(env.VariableGitpodWorkspaceId) != "" {
+
+		//Check we are able to generate tokens with a verified email address
+		output, err := exec.Command("gp", "idp", "token", "--audience", "example.org", "--decode").Output()
```
cmurphy, line 42: How necessary is this check, if the workspace variable is known to be set? The github provider, for example, only bothers to check the variable.

Is there a timeout option for this command?

All the providers are looped through when a new signer is created, so if this command doesn't exit quickly when not enabled, it will stall keyless signing.

haydentherapper:
Closing this as there hasn't been progress. Feel free to reopen once there is.

PR #3628, opened 2024-03-28
Clean up and clarify e2e scripts

cmurphy:
* Removed e2e_test_secrets.sh, it's not used anywhere
* piv_test.go *looks* like it's not used anywhere, but it actually is! Added a comment to make sure no one removes it like I almost did.


cpanato (approved):
thanks

PR #3629, opened 2024-03-29
Remove cross.yaml workflow

cmurphy:
The artifacts uploaded by cross.yaml are not used anywhere. Moreover, the ability to build on all three platforms and use the resulting binary is already tested in e2e-with-binary.yml. This change removes the workflow for the sake of decluttering the workflows and reducing our use of GitHub storage.


haydentherapper (approved):
cc @cpanato in case you're aware of any other usage of these artifacts

cpanato (approved):
LGTM

we need to merge this https://github.com/sigstore/community/pull/426 first

PR #3643, opened 2024-04-04
Fixing issue 3642

### LINKED ISSUE #3642: Memory Leak identified in cosign verify flow
**Description**

We are using cosign as a library in our code. when we execute cosign libraries to verify the images for signature, we identify a memory increase. After memory profiling of the code we found the issue in the file/function
https://github.com/sigstore/cosign/blob/6206f5af392ccefcc6a6cbb167bbccc833b46ed4/pkg/oci/internal/signature/layer.go#L60

The main problem identified is that in this function defer clause is missing to close the file stream.

When we looked similar functionality elsewhere in cosign code , we found that defer clause is present. few examples are
https://github.com/sigstore/cosign/blob/6206f5af392ccefcc6a6cbb167bbccc833b46ed4/pkg/oci/remote/remote.go#L228
 and
https://github.com/sigstore/cosign/blob/6206f5af392ccefcc6a6cbb167bbccc833b46ed4/pkg/oci/static/file.go#L84
in these functions the file stream is getting closed but seems to be missing in 
pkg/oci/internal/signature/layer.go

The main change suggested is to add defer clause which is missing in this function.

// Payload implements oci.Signature
func (s *sigLayer) Payload() ([]byte, error) {
	// Compressed is a misnomer here, we just want the raw bytes from the registry.
	r, err := s.Layer.Compressed()
	if err != nil {
		return nil, err
	}
	payload, err := io.ReadAll(r)
	if err != nil {
		return nil, err
	}
        **defer r.Close()**
	return payload, nil
}

We applied the patch locally and found the memory leak is solved.

**Without Patch**

image

**With Patch**
image

 

**Version**

The issue is identified in release 2.2.3 but it is previous in previous release also

Mukuls77:
This PR Closes #3642 
The issue fixed is related to Memory leak happening in 
https://github.com/sigstore/cosign/blob/6206f5af392ccefcc6a6cbb167bbccc833b46ed4/pkg/oci/internal/signature/layer.go#L60
Function: func (s *sigLayer) Payload() ([]byte, error) 

#### Summary
The cause of the problem as identified is missing Defer call in function
func (s *sigLayer) Payload() ([]byte, error) {
	// Compressed is a misnomer here, we just want the raw bytes from the registry.
	r, err := s.Layer.Compressed()
	if err != nil {
		return nil, err
	}
	payload, err := io.ReadAll(r)
	if err != nil {
		return nil, err
	}
	return payload, nil
}
so the code changes done is to introduce Defer clause before returning from function to close the file stream.
func (s *sigLayer) Payload() ([]byte, error) {
	// Compressed is a misnomer here, we just want the raw bytes from the registry.
	r, err := s.Layer.Compressed()
	if err != nil {
		return nil, err
	}
	payload, err := io.ReadAll(r)
	if err != nil {
		return nil, err
	}
        defer r.Close()
	return payload, nil
}

#### Release Note
Bug fix for Issue #3642 
#### Documentation
No change in Documentation

### DIFF
--- a/pkg/oci/internal/signature/layer.go
+++ b/pkg/oci/internal/signature/layer.go
@@ -67,6 +67,7 @@ func (s *sigLayer) Payload() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer r.Close()
 	return payload, nil
 }
 


cpanato (approved):

hectorj2f (approved):

Mukuls77:
i Moved defer r.Close() before payload, err := io.ReadAll(r) call this will close the stream in case we get issue in io.ReadAll
also same code was present on file 
pkg/oci/signature/layer.go so added the same changes in that file also

cpanato:
@Mukuls77 can you open a new pr?

Mukuls77:
@cpanato  ok i will open a new PR for the delta changes thanks

hectorj2f:
@Mukuls77 sgtm

PR #3670, opened 2024-04-18
Fixing issue 3669

### LINKED ISSUE #3669: Cosign should check Media Type of the layer before download of the signature
**Description**

During signature verification cosign calculates the signature tag by using the artifact digest. ArtifactPath:sha256-Artifact Digest.sig
Using this tag cosign download the signature manifest.
once signature manifest is downloaded cosign  start downloading the layers mentioned in the signature manifest file. 
There should be a check to verify the Media Type of the layers it is downloading is matching to the expected media type of a signature layer "application/vnd.dev.cosign.simplesigning.v1+json"

Why this check is important is that if someone by mistake apply the signature Tag to an artifact which is not really a signature than cosign will start downloading all the layers present in that artifact to which Signature Tag is applied by mistake.
This can account to a huge download of data in cosign memory and eventually cosign will fail verification because the data was not actually a signature.
So proposed enhancement is to check MediaType before download of layers this will stop the erroneous layer download.
This will save from wrong data download and memory consumption.  
This will also make system more robust for a change done by mistake in Registry related to signature tags

Mukuls77:
#### Summary
Closes #3669
Currently while doing the verification of signatures cosign does not check the MediaType of the layers before downloading those. In the case when some one mistakenly put the signature tag on an artifact which is not a signature than cosign will download all the layers present in the Manifest file of the artifact and than we start verification which will eventually fail as the artifact was not really a signature.
The fix provides an additional check to check the Media Type before downloading the layers. this will avoid unnecessary download of stale data and will quickly reject the verification as the artifact is not a signature.

#### Release Note
* Enhanced cosign verification logic to check MediaType before download of signature layers

#### Documentation
No change in documentation

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -608,19 +608,28 @@ func verifySignatures(ctx context.Context, sigs oci.Signatures, h v1.Hash, co *C
 	t := throttler.New(workers, len(sl))
 	for i, sig := range sl {
 		go func(sig oci.Signature, index int) {
-			sig, err := static.Copy(sig)
+
+			gotMT, err := sig.MediaType()
+			if err != nil {
+				t.Done(err)
+			}
+			if gotMT != types.SimpleSigningMediaType {
+				t.Done(errors.New("Incorrect MediaType of Signature"))
+				return
+			}
+			sigNew, err := static.Copy(sig)
 			if err != nil {
 				t.Done(err)
 				return
 			}
 
-			verified, err := VerifyImageSignature(ctx, sig, h, co)
+			verified, err := VerifyImageSignature(ctx, sigNew, h, co)
 			bundlesVerified[index] = verified
 			if err != nil {
 				t.Done(err)
 				return
 			}
-			signatures[index] = sig
+			signatures[index] = sigNew
 
 			t.Done(nil)
 		}(sig, i)
@@ -665,7 +674,6 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 	verifyFn signatureVerificationFn, co *CheckOpts) (
 	bundleVerified bool, err error) {
 	var acceptableRFC3161Time, acceptableRekorBundleTime *time.Time // Timestamps for the signature we accept, or nil if not applicable.
-
 	acceptableRFC3161Timestamp, err := VerifyRFC3161Timestamp(sig, co)
 	if err != nil {
 		return false, fmt.Errorf("unable to verify RFC3161 timestamp bundle: %w", err)


haydentherapper:
I'm not very familiar with the OCI part of the codebase, so this might be incorrect, but isn't the metadata already downloaded by this point?

I'm also not convinced this is needed, as this seems like handling a case where someone has manipulated the contents of the registry object. Exiting out early is fine, but it will fail nonetheless later on.

Mukuls77:
@haydentherapper  yes the MediaType is already known to us when we download the signature manifest. so this check is not really downloading or querying the mediaType from registry. Instead it is fetching it from the downloaded manifest file. So this check does not add any extra Get towards registry.
The check will only insure that we start downloading the layers which are of correct media type. i agree if we don't put the check the verification will staill fail but this will account to download of the layers which can have big memory requirement and unnecessary will consume the runtime memory of the system with the data which is not good.

Mukuls77:
@haydentherapper can you pls comment do you see this change useful as it is not introducing any additional overhead but enhancing the security, as i mentioned we can enhance it further by adding check to not download a signature layer if it is beyond a specific size.

cmurphy:
I'm not sure if @haydentherapper's first question was answered, or if I didn't quite understand it. Isn't the manifest already downloaded at this point? It looks like the download happens [here](https://github.com/sigstore/cosign/blob/e9a3739a8f1140d82688b09ce002ba91b0b8b1ae/pkg/cosign/verify.go#L526), a few lines before the [verifySignatures](https://github.com/sigstore/cosign/blob/e9a3739a8f1140d82688b09ce002ba91b0b8b1ae/pkg/cosign/verify.go#L537) call, so I'm not sure how checking the media type in `verifySignatures` prevents the download?

> If you agree i am extend this check to also check the size of the layer if the size is more than what we define as max permissible size of a signature we can block the download of layer if it exceed that size limit.

We already check the [size of the layer](https://github.com/sigstore/cosign/blob/e9a3739a8f1140d82688b09ce002ba91b0b8b1ae/pkg/oci/internal/signature/layer.go#L66) before reading it into memory - have you found a case where this check is insufficient or incorrect?

Mukuls77:
@cmurphy thanks for your comments.
The problem we want to fix is to stop download of layers of the manifest when someone put a signature tag on a manifest which is not really a signature manifest. so when we download the signatures two things are done by cosign
1. download the manifest file
2. download the layers present in the manifest file
The enhancement i am suggesting is that once we download the manifest file and before we start downloading the layers of the manifest we should check the media Type present in the layer, and if the media type of layer is not that of a signature payload we should not download it. so we just do a check of mediaType of layer from the downloaded manifest file before we trigger download of the manifest layers.

The layers are downloaded
when we invoke 
			sig, err := static.Copy(sig)
https://github.com/sigstore/cosign/blob/fa17fab56bc15a7a86b7e785fabdcf0e469ac802/pkg/cosign/verify.go#L611

I agree we have a env variable COSIGN_MAX_ATTACHMENT_SIZE which has default value of 128 Mb, ideally our signature payload is not that big and it would have been better to have a more specific limit for signatures, but still we can tune this env variable according to our needs.
So as i mentioned if we check the Media Type of layers before we download those layers we would optimize of cosign handling and we would not download stale layers which are actually not signature data.

Hayden-IO:
Closing as we work towards a v3 release. We've significantly changed how verification material is stored, now as referring artifacts. Feel free to review to see if that solves this indirectly.

PR #3693, opened 2024-05-10
Add PayloadProvider interface to decouple AttestationToPayloadJSON from oci.Signature interface

codysoyland:
I'm working on Policy Controller and need to use `AttestationToPayloadJSON` without having an entire `oci.Signature`. Since this func only needs the `Payload`, I added a small `PayloadProvider` interface containing only the `Payload` func, which preserves backwards compatibility while decoupling the `oci.Signature` interface.

Signed-off-by: Cody Soyland <codysoyland@github.com>

### DIFF
--- a/pkg/policy/attestation.go
+++ b/pkg/policy/attestation.go
@@ -29,6 +29,17 @@ import (
 	"github.com/sigstore/cosign/v2/pkg/cosign/attestation"
 )
 
+// PayloadProvider is a subset of oci.Signature that only provides the
+// Payload() method.
+type PayloadProvider interface {
+	// Payload fetches the opaque data that is being signed.
+	// This will always return data when there is no error.
+	Payload() ([]byte, error)
+}
+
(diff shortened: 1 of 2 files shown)


pkg/policy/attestation.go
```diff
@@ -45,7 +56,7 @@ import (
 // or the predicateType is not the one they are looking for. Without returning
 // this, it's hard for users to know which attestations/predicateTypes were
 // inspected.
-func AttestationToPayloadJSON(_ context.Context, predicateType string, verifiedAttestation oci.Signature) ([]byte, string, error) {
+func AttestationToPayloadJSON(_ context.Context, predicateType string, verifiedAttestation PayloadProvider) ([]byte, string, error) {
```
haydentherapper, line 59: Can you add a test for passing in both an `oci.Signature` object and a `PayloadProvider`?

haydentherapper (approved):
Thanks!

PR #3701, opened 2024-05-17
Refactor insecure registry E2E tests

cmurphy:
Move the shell script tests to Go, have them use a local Rekor instance instead of the public one, and clean up the workflow.


haydentherapper (approved):
Thanks so much for cleaning this up!

PR #3707, opened 2024-05-20
Remove sign_blob_test.sh test

cmurphy:
The bug that this test was meant to address[1] applies to a feature that was removed from cosign[2]. The updates made to the script to allow for the breaking changes actually make the test invalid, because while it was once verifying the original artifact signature, the final verification step is now actually verifying against the uploaded rekor entry, so the original signature is inconsequential.

[1] https://github.com/sigstore/cosign/pull/1673
[2] https://github.com/sigstore/cosign/pull/2425


haydentherapper (approved):

haydentherapper:
Thanks for cleaning this up!

PR #3745, opened 2024-06-22
Set `bundleVerified` to true after Rekor verification (Resolves #3740) 

### LINKED ISSUE #3740: `bundleVerified` Flag Not Set After Successful Rekor Online Verification in Cosign
**Description**

The `bundleVerified` flag is not being set to true following a successful online verification using Rekor, as seen in the code snippet below. 

```go
if !co.IgnoreTlog {
    bundleVerified, err = VerifyBundle(sig, co)
    if err != nil {
        return false, fmt.Errorf("error verifying bundle: %w", err)
    }

    if bundleVerified {
        // Update with the verified bundle's integrated time.
        t, err := getBundleIntegratedTime(sig)
        if err != nil {
            return false, fmt.Errorf("error getting bundle integrated time: %w", err)
        }
        acceptableRekorBundleTime = &t
    } else {
        // If the --offline flag was specified, fail here. bundleVerified returns false with
        // no error when there was no bundle provided.
        if co.Offline {
            return false, fmt.Errorf("offline verification failed")
        }

        // no Rekor client provided for an online lookup
        if co.RekorClient == nil {
            return false, fmt.Errorf("rekor client not provided for online verification")
        }

        pemBytes, err := keyBytes(sig, co)
        if err != nil {
            return false, err
        }

        e, err := tlogValidateEntry(ctx, co.RekorClient, co.RekorPubKeys, sig, pemBytes)
        if err != nil {
            return false, err
        }
        t := time.Unix(*e.IntegratedTime, 0)
        acceptableRekorBundleTime = &t
    }
}
```

**Expected Behavior:**
In the else branch of the verification logic, if the Rekor verification succeeds, the `bundleVerified` should be set to true. 

**Version**

`2.2.4`

maxlambrecht:
#### Summary

This PR addresses an issue where the `bundleVerified` flag was not set to true after a successful online Rekor verification. The change ensures that `bundleVerified` accurately reflects the verification status when Rekor lookup is used.

Resolves #3740

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -710,6 +710,7 @@ func verifyInternal(ctx context.Context, sig oci.Signature, h v1.Hash,
 			}
 			t := time.Unix(*e.IntegratedTime, 0)
 			acceptableRekorBundleTime = &t
+			bundleVerified = true
 		}
 	}
 
--- a/pkg/cosign/verify_test.go
+++ b/pkg/cosign/verify_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"crypto"
+	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
@@ -38,6 +39,7 @@ import (
 
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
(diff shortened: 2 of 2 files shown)


haydentherapper:
Thanks! Can you add a test for this? I think you just need to check what's returned by a code block like https://github.com/sigstore/cosign/blob/65969ae877dd2e64520378905c5dda974c98a576/pkg/cosign/verify_test.go#L553-L556. 

If adding this test isn't straightforward, that's fine, this is a simple enough change.

maxlambrecht:
> Thanks! Can you add a test for this? I think you just need to check what's returned by a code block like
> 
> https://github.com/sigstore/cosign/blob/65969ae877dd2e64520378905c5dda974c98a576/pkg/cosign/verify_test.go#L553-L556
> .
> 
> If adding this test isn't straightforward, that's fine, this is a simple enough change.

It wasn't straightforward, but I managed to add a comprehensive test covering several scenarios. I refactored the existing test, which only covered a failing validation, and combined it with other cases to ensure thorough coverage. 

Let me know if you need any further adjustments or additional scenarios covered.

haydentherapper (approved):
Thank you, and thanks for the tests too!

mtrmac:
I have no particular opinion on the _desired_ semantics of the “bundleVerified” return value [my general position is that verification should be provided a policy and, apart from debugging-only output, return a single yes/no outcome, not any details to be further interpreted] — but AFAICS this value is consumed in https://github.com/sigstore/cosign/blob/3d622d1d6d509fdc55d1a4cc84a84559bfd5663c/cmd/cosign/cli/verify/verify.go#L363 , which claims the _off-line_ validation has succeeded. 

With this PR, `bundleVerified` is set to true if off-line validation _did not_ succeed, but an on-line Rekor client check did.

I don’t know which part should be adjusted, but right now the codebase seems to be inconsistent.

haydentherapper:
I agree, unfortunately we have inconsistencies in the codebase between what verification steps were taken and the debug output.

Even `bundleVerified` is misleading because we mean that we've verified Rekor, not the "bundle" aka the sig, cert and inclusion proof.

The plan is to clean all of this up with the adoption of sigstore-go as the internal API for Cosign.

Issue #3139, opened 2023-07-26
Support the protobuf bundle format in Cosign


haydentherapper:
## Background

Sigstore created a common format in [sigstore/protobuf-specs](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto) for the output from Sigstore clients. sigstore-python, sigstore-java and sigstore-js currently support the bundle format. Golang currently does not support the bundle: sigstore-go is under active development, and Cosign has defined its own format.

## Requirements

- [ ] [P0] Output bundle for `cosign sign-blob` or `cosign attest-blob`
- [ ] [P0] Read bundle for `cosign verify-blob` or `cosign verify-blob-attestation`
- [ ] [P0] Verify conformance with other clients

## Secondary goals

- [ ] [P1] Persist bundle in OCI image manifest for `cosign sign` or `cosign attest`
- [ ] [P1] Read bundle from OCI image manifest for `cosign verify` or `cosign verify-attestation`

## Nice to haves

- [ ] [P2] Separation of utilities in a dedicated package to make it easy to move code to sigstore-go
- [ ] [P2] Support for uploading a bundle to OCI with `cosign attach`

## Resources

* [Bundle specification](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto)
* [Rekor output to protobuf library](https://github.com/sigstore/rekor/blob/main/pkg/tle/tle.go)

Justification for sign-blob being a P0 and sign being a P1 is that OCI handles storage already, so it's less critical to have a new format.

wata727:
@haydentherapper I'm interested in working on this. Can we move this forward?
If you don't have any other blockers, I can open a pull request after tweaking the patch https://github.com/sigstore/cosign/commit/d652a110510e7f5c12109a9385acc3620b48847b

A summary of this patch is:

- Introduce `pkg/cosign/bundle.SignedEntity` instead of `cosign.LocalSignedPayload`.
  - In order to cut out a dedicated package to sigstore-go that will generate a Sigstore bundle in the future, I thought we would need a struct that satisfies an interface like [`sigstore-go/pkg/verify.SignedEntity`](https://github.com/sigstore/sigstore-go/blob/v0.1.0/pkg/verify/interface.go#L55-L62).
  - A side effect of introducing this is that we can refactor to use this entity for creating a Cosign bundle.
- Introduce `pkg/bundle` package.
[...]

PR #3752, opened 2024-06-28
Adding protobuf bundle support to sign-blob and attest-blob

steiza:
#### Summary

This pull requests addresses the first part of #3139: adding protobuf bundle support for `cosign sign-blob` and `cosign attest-blob`.

You can test this by generating the new bundles, for example signing a local file with a `cosign` provisioned key (requesting a signed timestamp to corroborate):

```bash
go run cmd/cosign/main.go sign-blob --bundle="test-bundle.sigstore.json" --new-bundle-format=true --key="cosign.key" --tlog-upload=false --timestamp-server-url="https://timestamp.githubapp.com/api/v1/timestamp" ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

Or using Fulcio to get a signing certificate for an attestation:

```bash
go run cmd/cosign/main.go attest-blob --bundle="pgi-bundle.sigstore.json" --new-bundle-format=true --fulcio-url='https://fulcio.sigstore.dev' --rekor-url='https://rekor.sigstore.dev' --identity-token='...' --type=something --predicate="../sigstore-go/examples/sigstore-go-signing/intoto.txt" ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

You can then verify the public good instance bundle using sigstore-go doing something like:
```bash
go run cmd/sigstore-go/main.go -artifact examples/sigstore-go-signing/hello_world.txt -trustedrootJSONpath public-good-trusted-root.json -expectedIssuer "https://token.actions.githubusercontent.com" -expectedSANRegex ".+" ../cosign/pgi-bundle.sigstore.json
```

#### Release Note

NONE - we probably want to finish #3139 (especially the more comprehensive conformance testing!) before we announce this as released.

#### Documentation

N/A - same as above


cmd/cosign/cli/options/attest_blob.go
```diff
@@ -85,6 +86,10 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 		"write everything required to verify the blob to a FILE")
 	_ = cmd.Flags().SetAnnotation("bundle", cobra.BashCompFilenameExt, []string{})
 
+	// TODO: have this default to true as a breaking change
+	cmd.Flags().BoolVar(&o.ProtobufBundleFormat, "protobuf-bundle-format", false,
```
cmurphy, line 90: Suggestion -

"protobuf" is a data format that is used internally to define various sigstore data structures. If cosign was a black box, then "protobuf" is not actually meaningful to the end user, and in fact the output bundle is formatted as JSON, not a protobuf. The fact that we're internally aligning with an abstract data structure that happens to be defined in a `.proto` file doesn't help the user understand the difference. I think it would be more useful to describe it as the "sigstore bundle", or the "v1 bundle" (or v2?), or to maybe have the flag be something like "new-bundle-format" or "experimental-bundle-format", or some other differentiator that doesn't refer to the underlying implementation details that a user shouldn't need to care about.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -189,28 +200,69 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+							Sig: sig,
+						},
+					},
+				},
+			}
+
+			contents, err = protojson.Marshal(bundle)
+			if err != nil {
+				return err
+			}
```
cmurphy, line 255: nit - consider extracting this section into its own function, it's a bit long and the `Exec` function is already extremely long.

cmd/cosign/cli/options/attest_blob.go
steiza: Sounds good to me! I went with `new-bundle-format`, as I think that's more accurate than experimental (it's the bundle format most of the Sigstore client libraries are using).

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -154,25 +159,31 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+			if err := os.WriteFile(c.RFC3161TimestampPath, ts, 0600); err != nil {
+				return fmt.Errorf("create RFC3161 timestamp file: %w", err)
+			}
+			fmt.Fprintln(os.Stderr, "RFC3161 timestamp bundle written to file ", c.RFC3161TimestampPath)
 		}
-		fmt.Fprintln(os.Stderr, "RFC3161 timestamp bundle written to file ", c.RFC3161TimestampPath)
 	}
 
 	rekorBytes, err := sv.Bytes(ctx)
```
haydentherapper, line 189: Unrelated, can you rename this to signer, as `rekorBytes` is somewhat misleading since this is only the cert or key?

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -258,3 +281,41 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
 }
+
+func makeNewBundle(rekorEntry *models.LogEntryAnon, payload, sig, signer, timestampBytes []byte, payloadType string) ([]byte, error) {
+	// Determine if signature is certificate or not
+	var hint string
+	var rawCert []byte
+
+	cert, err := cryptoutils.UnmarshalCertificatesFromPEM(signer)
+	if err != nil || len(cert) == 0 {
+		hashedBytes := sha256.Sum256(signer)
```
haydentherapper, line 292: `signer` is a [PEM-encoded](https://github.com/sigstore/cosign/blob/f7a5725317fbb0ad3317ceddc345d21cc85151e4/cmd/cosign/cli/sign/sign.go#L606) when not a cert. If we want to align to our suggested key hint, which is [`SHA-256(DER(SPKI(key)))`](https://www.rfc-editor.org/rfc/rfc6962#section-3.2:~:text=represented%20as%20SubjectPublicKeyInfo.-,%22issuer_key_hash%22,-is%20the%20SHA), we'll need to do:

```
pkixPubKey, err := x509.MarshalPKIXPublicKey(signer.PublicKey())
if err != nil {
  return nil, err
}
hashedBytes := sha256.Sum256(pkixPubKey)
hint = base64.StdEncoding.EncodeToString(hashedBytes[:])
```

haydentherapper: And same for sign-blob

pkg/cosign/bundle/protobundle_test.go
```diff
@@ -0,0 +1,63 @@
...
+			rekorEntry:     nil,
+			timestampBytes: []byte("timestamp"),
+		},
+		{
+			name:           "only cert",
+			hint:           "",
+			rawCert:        []byte("cert stuff"),
+			rekorEntry:     nil,
+			timestampBytes: []byte{},
+		},
```
haydentherapper, line 44: nit, can we add a test case for rekor?

haydentherapper (commented):
Fantastic!! Just a few tiny comments, also needs a rebase.

pkg/cosign/bundle/protobundle_test.go
steiza: Not easily, although maybe I'm thinking about this the wrong way.

I spent some time trying to mock out a hashedrekord / intoto entry, but the verification that `tle.GenerateTransparencyLogEntry()` performs is pretty substantial. Otherwise I could try to take a production entry and work backwards to populate a `LogEntryAnon`? What do you think?

steiza: I was able to co-opt / adapt a test from https://github.com/sigstore/rekor/blob/main/pkg/tle/tle_test.go

haydentherapper (approved):
Great work on this!

haydentherapper:
@steiza can you rebase? Just to check, anything else here or any other testing, or is this good to merge?

steiza:
> Just to check, anything else here or any other testing, or is this good to merge?

I think this is good to merge!

PR #3796, opened 2024-07-23
Add new bundle support to `verify-blob` and `verify-blob-attestation`

steiza:
#### Summary

This pull request is for the second part of #3139: adding protobuf bundle support for `verify-blob` and `verify-blob-attestation`.

Now that #3752 has landed, you can first generate bundles with something like:

```bash
$ go run cmd/cosign/main.go sign-blob --bundle="test.sigstore.json" --key="cosign.key" --new-bundle-format=true --tlog-upload=false ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

And verify it (with these changes) with something like:

```bash
$ go run cmd/cosign/main.go verify-blob --bundle=test.sigstore.json --new-bundle-format=true --key=cosign.pub --use-signed-timestamps=false --insecure-ignore-tlog=true ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

Or do an attestation with something like:

```bash
$ cat predicate.txt 
{}
$ go run cmd/cosign/main.go attest-blob --bundle="attest-pgi.sigstore.json" --new-bundle-format=true --identity-token="..." --type=something --predicate="predicate.txt" ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

And then verify that attestation with these changes (assuming your identity token came from https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/), with:

```bash
$ go run cmd/cosign/main.go verify-blob-attestation --bundle attest-pgi.sigstore.json --new-bundle-format=true --certificate-oidc-issuer="https://token.actions.githubusercontent.com" --certificate-identity="https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/.github/workflows/extremely-dangerous-oidc-beacon.yml@refs/heads/main" --type=something ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

#### Release Note

NONE - we probably want to finish https://github.com/sigstore/cosign/issues/3139 (especially the more comprehensive conformance testing!) before we announce this as released.

#### Documentation

N/A - same as above


cmd/cosign/cli/verify/verify_bundle.go
```diff
@@ -0,0 +1,163 @@
...
+	keyTrustedMaterial root.TrustedMaterial
+}
+
+func (v *verifyTrustedMaterial) PublicKeyVerifier(hint string) (root.TimeConstrainedVerifier, error) {
+	return v.keyTrustedMaterial.PublicKeyVerifier(hint)
+}
+
+func verifyNewBundle(ctx context.Context, bundlePath, trustedRootPath, keyRef, slot, certOIDCIssuer, certOIDCIssuerRegex, certIdentity, certIdentityRegexp, githubWorkflowTrigger, githubWorkflowSHA, githubWorkflowName, githubWorkflowRepository, githubWorkflowRef, artifactRef string, sk, ignoreTlog, useSignedTimestamps, ignoreSCT bool) error {
+	if certOIDCIssuerRegex != "" {
+		return fmt.Errorf("--new-bundle-format does not support --certificate-oidc-issuer-regexp")
```
steiza, line 44: Although support is coming soon! https://github.com/sigstore/sigstore-go/pull/236

go.mod
```diff
@@ -32,9 +32,10 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0
 	github.com/sigstore/fulcio v1.5.1
-	github.com/sigstore/protobuf-specs v0.3.0
+	github.com/sigstore/protobuf-specs v0.3.2
 	github.com/sigstore/rekor v1.3.6
 	github.com/sigstore/sigstore v1.8.7
+	github.com/sigstore/sigstore-go v0.4.1-0.20240717174219-8554eb6de5ac
```
steiza, line 38: To make use of https://github.com/sigstore/sigstore-go/pull/235. Once https://github.com/sigstore/sigstore-go/pull/236 also lands we should consider cutting a numbered release of sigstore-go to include these two changes.

haydentherapper: Merged - shall we merge all the open dependabot PRs and then cut a new release?

cmd/cosign/cli/verify/verify_bundle.go
```diff
@@ -0,0 +1,163 @@
...
+	trustedmaterial := &verifyTrustedMaterial{TrustedMaterial: trustedroot}
+
+	// See if we need to wrap trusted root with provided key
+	if keyRef != "" {
+		signatureVerifier, err := sigs.PublicKeyFromKeyRef(ctx, keyRef)
+		if err != nil {
+			return err
+		}
+
+		newExpiringKey := root.NewExpiringKey(signatureVerifier, time.Time{}, time.Time{})
```
haydentherapper, line 76: Out of scope for this, should we create a non-expiring key as well?

cmd/cosign/cli/verify/verify_bundle.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	// Perform verification
+	payload, err := payloadBytes(artifactRef)
+	if err != nil {
+		return err
+	}
+	buf := bytes.NewBuffer(payload)
+
+	sev, err := verify.NewSignedEntityVerifier(trustedmaterial, verifierConfig...)
```
haydentherapper, line 156: Originally, what I had in mind was decomposing the bundle into the `CheckOpts` and OCI structs so that we could call [VerifyBlobSignature](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L293). The concern was a lack of parity between the two verification policies.

Given that this is guarded behind a flag, it wouldn't be a breaking change to have a different verification path. And maybe I'm overthinking this, that there isn't actually a significant gap between the two verification policies - keys was one concern, but it seems like you've fully supported that here.

WDYT, between either using sigstore-go's verification library or continuing to use Cosign's with bundle decomposition?

cmd/cosign/cli/verify/verify_bundle.go
steiza: The terminology here is confusing - despite the name `NewExpiringKey`, since the `validityPeriodStart` and `validityPeriodEnd` will return true for `IsZero()`, this key won't actually be treated as expiring.

In the future, if there are cosign CLI flags for specifying the validity period of the key, we could make use of them here!

cmd/cosign/cli/verify/verify_bundle.go
steiza: That was the path I originally started going down - decomposing the bundle and using cosign's existing `VerifyBlobSignature()` - similar to the technique we used to add signing support. But it's a fair amount of work to decompose both the bundle and the trusted root into the format cosign is expecting, which would look very similar to code that's already in sigstore-go. Since I think we want to eventually get sigstore-go into cosign (and have more parts of cosign use sigstore-go in the future) I jumped ahead a bit in this pull request.

cmd/cosign/cli/verify/verify_bundle.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		identityPolicies = append(identityPolicies, verify.WithCertificateIdentity(certIdentity))
+	}
+
+	// Make some educated guesses about verification policy
+	verifierConfig := []verify.VerifierOption{}
+
+	if len(trustedroot.RekorLogs()) > 0 && !ignoreTlog {
+		verifierConfig = append(verifierConfig, verify.WithTransparencyLog(1), verify.WithObserverTimestamps(1))
```
haydentherapper, line 134: Should we use `WithIntegratedTimestamps` instead of `WithObserverTimestamps`?

cmd/cosign/cli/verify/verify_bundle.go
haydentherapper: I agree, I'm happy to see us cut over now. I've been looking over `VerifyBlobSignature` to see if there are any differences that we'd be missing, but I don't see anything concerning.

Online verification for Rekor is missing, but that's a good thing, I want to remove this. Bundles always have inclusion proofs and online lookups by default isn't a great design.

There is no support for verifying with certificate chains bundled in the verification material, but I know that's blocked on support in sigstore-go, and it's simple enough to work around with a trust root file.

haydentherapper (commented):
This is great, thanks for your work on this! Just one tiny comment.

cmd/cosign/cli/verify/verify_bundle.go
steiza: This is a good idea - in other contexts when I'm using sigstore-go I'll use `WithObserverTimestamps` to take advantage of a signed timestamp if it's available, but I think you're right that in cosign if you aren't using `--insecure-ignore-tlog`, you probably do want to specifically check the integrated timestamp.

haydentherapper (approved):
I'm good to merge as-is!
I'd like a little more testing ideally, could we add an e2e test under https://github.com/sigstore/cosign/blob/main/test/e2e_test.go as well?

steiza:
I would say land as-is. I'm already working on a conformance pull request (which [has an open question](https://github.com/sigstore/cosign/issues/3139#issuecomment-2253512195)), which we can roll additional e2e tests into as well.

PR #3808, opened 2024-08-04
move incremental builds per commit to GHCR instead of GCR

bobcallaway:
we can backfill the entries from gcr.io/projectsigstore/cosign/ci to GHCR as well, in case anyone is using these

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -67,5 +67,5 @@ jobs:
       - name: containers-cosign
         run: make sign-ci-containers
         env:
-          KO_PREFIX: gcr.io/projectsigstore/cosign/ci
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -67,5 +67,5 @@ jobs:
       - name: containers-cosign
         run: make sign-ci-containers
         env:
-          KO_PREFIX: gcr.io/projectsigstore/cosign/ci
+          KO_PREFIX: ghcr.io/sigstore/cosign/cosign/ci
```
haydentherapper, line 70: Should we write up a blog post/slack message for this change?

haydentherapper (approved):

hectorj2f (approved):

Issue #3832, opened 2024-08-14
Allow signing local image without registry access


bkabrda:
**Description**

Hi :wave:
I want to sign a local image that hasn't yet been uploaded to a registry (or the registry is not reachable right now) with `--upload=false --output-signature=signature.sig --output-certificate=certificate.crt`. Right now this fails with:

```
$ cosign sign -y --upload=false --output-signature=disconnected-fulcio.sig --output-certificate=disconnected-fulcio.crt foobarasd.com/myimage@sha256:2bbea7758536b170efcb168dc7cea3379908c2649af3e75ebac10161ddd513c2
Generating ephemeral keys...
Retrieving signed certificate...

<snip>

Successfully verified SCT...
Error: signing [foobarasd.com/myimage@sha256:2bbea7758536b170efcb168dc7cea3379908c2649af3e75ebac10161ddd513c2]: accessing image: Get "https://foobarasd.com/v2/": dial tcp: lookup foobarasd.com on 192.168.1.20:53: no such host
main.go:74: error during command execution: signing [foobarasd.com/myimage@sha256:2bbea7758536b170efcb168dc7cea3379908c2649af3e75ebac10161ddd513c2]: accessing image: Get "https://foobarasd.com/v2/": dial tcp: lookup foobarasd.com on 192.168.1.20:53: no such host
```

I think this should work, because to generate these artifacts locally we don't need to access the registry.

I have a simple change that I tested locally that I could submit as a PR if you folks think that this makes sense - please let me know. Thank you!

slimm609:
This would be very useful with `cosign save` to be able to sign and save a bundle, which then becomes portable to offline registries.  

right now the workflow is 

- push
- sign
- download (save)

which requires a temporary registry to get a offline bundle which is a lot of network activity and extra time.

bkabrda:
@slimm609 I didn't have time to work on this properly in the past 2 months, but I have some free time on my hands right now, so I'm going to revamp the PR that I submitted and try to make sure that your scenario is accounted for.

bkabrda:
@slimm609 after a lot of research I realized that your usecase would need to go through a completely different code path, so I didn't fix that in the PR that I submitted. However, your usecase actually makes a lot of sense I think, as it would actually be better to just save the bundle instead of producing a set of disconnected files (certificate, payload and signature). Once this issue is resolved, I would definitely like to work on that usecase as well (I'll open a new issue for it later), but I need to finish this up first to have at least something working.

[...]

PR #3841, opened 2024-08-20
Allow signing local image without registry access

bkabrda:
#### Summary

This PR attempts to address https://github.com/sigstore/cosign/issues/3832 - it will allow generating local signature for an image (when `--upload=false`) by using `--output-signature`/`--output-artifact`/`--output-payload` even when the remote registry is not accessible or the image hasn't yet been pushed there. Details are in the linked issue.

I only implemented this for the case the image is passed in by digest right now - it's easier to do and the code says we're going to eventually disable referencing images by tag, so I don't think it's strictly necessary there (but let me know if you think otherwise).

Verification of such signature requires the `--insecure-tlog-verify` (a.k.a.  `--private-infrastructure`) flag, which seems a little weird to me, but I think that's not something that I introduced by this PR. After some investigation, I think this is caused by the fact that the locally written files don't seem to represent a full bundle (like the one that can be generated with signing a blob or the one that gets pushed as an OCI artifact with `--upload=true`) and hence the locally created `fakeOCISignatures` object can't be used correctly in the `VerifyBundle` function in `pkg/cosign/verify.go`. I think maybe a better way to solve all of this would be to rather allow writing out full bundles for images, like we allow for `sign-blob`?

#### Release Note

* Enabled signing local images without access to registry

#### Documentation

I think the release note should be fine, but please correct me if I'm wrong.

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -180,11 +180,15 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 		}
 
 		if digest, ok := ref.(name.Digest); ok && !signOpts.Recursive {
-			se, err := ociremote.SignedEntity(ref, opts...)
-			if _, isEntityNotFoundErr := err.(*ociremote.EntityNotFoundError); isEntityNotFoundErr {
-				se = ociremote.SignedUnknown(digest)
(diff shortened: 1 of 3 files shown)


bkabrda:
Closing this PR, I'm doing some more research and I think I'll be able to find a better way to fix this.

PR #3844, opened 2024-08-20
Upgrade to TUF v2 client

cmurphy:
Use sigstore-go's TUF client to fetch the trusted_root.json from the TUF
mirror, if available. Where possible, use sigstore-go's verifiers which
natively accept the trusted root as its trusted material. Where there is
no trusted root available in TUF or sigstore-go doesn't support a use
case, fall back to the sigstore/sigstore TUF v1 client and the existing
verifiers in cosign.

TODO:
- [x] unit tests
- [x] e2e tests - waiting for https://github.com/sigstore/sigstore-go/pull/247 to be merged and percolated into the scaffolding action

Fixes https://github.com/sigstore/cosign/issues/3548

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -16,11 +16,14 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/attest"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/generate"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
+	"github.com/sigstore/cosign/v2/internal/ui"
+	"github.com/sigstore/cosign/v2/pkg/cosign"
 	"github.com/spf13/cobra"
 )
 
@@ -69,6 +72,7 @@ func Attest() *cobra.Command {
 			if err != nil {
 				return err
 			}
+
 			ko := options.KeyOpts{
 				KeyRef:                   o.Key,
 				PassFunc:                 generate.GetPass,
@@ -91,6 +95,13 @@ func Attest() *cobra.Command {
 				TSAServerName:            o.TSAServerName,
 				TSAServerURL:             o.TSAServerURL,
 				NewBundleFormat:          o.NewBundleFormat,
+			}
+			if o.Key == "" { // Get the trusted root if using fulcio for signing
+				trustedMaterial, err := cosign.TrustedRoot()
+				if err != nil {
+					ui.Warnf(context.Background(), "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
+				}
+				ko.TrustedMaterial = trustedMaterial
 			}
 			attestCommand := attest.AttestCommand{
 				KeyOpts:                 ko,
--- a/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
+++ b/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
@@ -17,12 +17,15 @@ package fulcioverifier
 
 import (
 	"context"
+	"crypto/x509"
 	"fmt"
 
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v2/internal/ui"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
+	"github.com/sigstore/sigstore-go/pkg/verify"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 )
 
@@ -32,12 +35,31 @@ func NewSigner(ctx context.Context, ko options.KeyOpts, signer signature.SignerV
 		return nil, err
 	}
 
-	// Grab the PublicKeys for the CTFE, either from tuf or env.
+	if ko.TrustedMaterial != nil && len(fs.SCT) == 0 {
+		// Detached SCTs cannot be verified with this function.
+		chain, err := cryptoutils.UnmarshalCertificatesFromPEM(fs.Chain)
+		if err != nil {
+			return nil, fmt.Errorf("unmarshalling cert chain from PEM for SCT verification: %w", err)
+		}
+		certs, err := cryptoutils.UnmarshalCertificatesFromPEM(fs.Cert)
+		if err != nil || len(certs) < 1 {
+			return nil, fmt.Errorf("unmarshalling cert from PEM for SCT verification: %w", err)
+		}
+		chain = append(certs, chain...)
+		chains := make([][]*x509.Certificate, 1)
+		chains[0] = chain
+		if err := verify.VerifySignedCertificateTimestamp(chains, 1, ko.TrustedMaterial); err != nil {
+			return nil, fmt.Errorf("verifying SCT using trusted root: %w", err)
+		}
+		ui.Infof(ctx, "Successfully verified SCT...")
+		return fs, nil
+	}
+
+	// There was no trusted_root.json or we need to verify a detached SCT, so grab the PublicKeys for the CTFE, either from tuf or env.
 	pubKeys, err := cosign.GetCTLogPubs(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("getting CTFE public keys: %w", err)
 	}
-
 	// verify the sct
 	if err := cosign.VerifySCT(ctx, fs.Cert, fs.Chain, fs.SCT, pubKeys); err != nil {
 		return nil, fmt.Errorf("verifying SCT: %w", err)
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -31,6 +31,7 @@ import (
 	"github.com/sigstore/cosign/v2/internal/ui"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
 	"github.com/sigstore/cosign/v2/pkg/cosign/cue"
+	"github.com/sigstore/cosign/v2/pkg/cosign/env"
 	"github.com/sigstore/cosign/v2/pkg/cosign/pivkey"
 	"github.com/sigstore/cosign/v2/pkg/cosign/pkcs11key"
 	"github.com/sigstore/cosign/v2/pkg/cosign/rego"
@@ -98,6 +99,21 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		return fmt.Errorf("constructing client options: %w", err)
 	}
 
+	trustedMaterial, err := cosign.TrustedRoot()
+	if err != nil {
+		ui.Warnf(ctx, "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
+	}
+
+	if options.NOf(c.CertChain, c.CARoots, c.CAIntermediates, c.TSACertChainPath) > 0 ||
+		env.Getenv(env.VariableSigstoreCTLogPublicKeyFile) != "" ||
+		env.Getenv(env.VariableSigstoreRootFile) != "" ||
+		env.Getenv(env.VariableSigstoreRekorPublicKey) != "" ||
+		env.Getenv(env.VariableSigstoreTSACertificateFile) != "" {
+		// trusted_root.json was found, but a cert chain was explicitly provided, or environment variables point to the key material,
+		// so don't overrule the user's intentions.
+		trustedMaterial = nil
+	}
+
 	co := &cosign.CheckOpts{
 		RegistryClientOpts:           ociremoteOpts,
 		CertGithubWorkflowTrigger:    c.CertGithubWorkflowTrigger,
@@ -112,23 +128,39 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		MaxWorkers:                   c.MaxWorkers,
 		UseSignedTimestamps:          c.TSACertChainPath != "" || c.UseSignedTimestamps,
 		NewBundleFormat:              c.NewBundleFormat,
+		TrustedMaterial:              trustedMaterial,
 	}
 	if c.CheckClaims {
 		co.ClaimVerifier = cosign.IntotoSubjectClaimVerifier
+	}
+
+	if c.TrustedRootPath != "" {
+		co.TrustedMaterial, err = root.NewTrustedRootFromPath(c.TrustedRootPath)
+		if err != nil {
+			return fmt.Errorf("loading trusted root: %w", err)
+		}
+	} else if options.NOf(c.CertChain, c.CARoots, c.CAIntermediates, c.TSACertChainPath) == 0 &&
+		env.Getenv(env.VariableSigstoreCTLogPublicKeyFile) == "" &&
+		env.Getenv(env.VariableSigstoreRootFile) == "" &&
+		env.Getenv(env.VariableSigstoreRekorPublicKey) == "" &&
+		env.Getenv(env.VariableSigstoreTSACertificateFile) == "" {
+		co.TrustedMaterial, err = cosign.TrustedRoot()
+		if err != nil {
+			ui.Warnf(ctx, "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
+		}
 	}
 
 	if c.NewBundleFormat {
 		if err = checkSigstoreBundleUnsupportedOptions(c); err != nil {
 			return err
 		}
-		co.TrustedMaterial, err = loadTrustedRoot(ctx, c.TrustedRootPath)
-		if err != nil {
-			return err
+		if co.TrustedMaterial == nil {
+			return fmt.Errorf("trusted root is required when using new bundle format")
 		}
 	}
 
 	// Ignore Signed Certificate Timestamp if the flag is set or a key is provided
-	if shouldVerifySCT(c.IgnoreSCT, c.KeyRef, c.Sk) && !c.NewBundleFormat {
+	if co.TrustedMaterial == nil && shouldVerifySCT(c.IgnoreSCT, c.KeyRef, c.Sk) && !c.NewBundleFormat {
 		co.CTLogPubKeys, err = cosign.GetCTLogPubs(ctx)
 		if err != nil {
 			return fmt.Errorf("getting ctlog public keys: %w", err)
@@ -136,7 +168,7 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 	}
 
 	// If we are using signed timestamps, we need to load the TSA certificates
-	if co.UseSignedTimestamps && !c.NewBundleFormat {
+	if co.UseSignedTimestamps && co.TrustedMaterial == nil && !c.NewBundleFormat {
 		tsaCertificates, err := cosign.GetTSACerts(ctx, c.TSACertChainPath, cosign.GetTufTargets)
 		if err != nil {
 			return fmt.Errorf("unable to load TSA certificates: %w", err)
@@ -154,15 +186,17 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 			}
 			co.RekorClient = rekorClient
 		}
-		// This performs an online fetch of the Rekor public keys, but this is needed
-		// for verifying tlog entries (both online and offline).
-		co.RekorPubKeys, err = cosign.GetRekorPubs(ctx)
-		if err != nil {
-			return fmt.Errorf("getting Rekor public keys: %w", err)
+		if co.TrustedMaterial == nil {
+			// This performs an online fetch of the Rekor public keys, but this is needed
+			// for verifying tlog entries (both online and offline).
+			co.RekorPubKeys, err = cosign.GetRekorPubs(ctx)
+			if err != nil {
+				return fmt.Errorf("getting Rekor public keys: %w", err)
+			}
 		}
 	}
 
-	if keylessVerification(c.KeyRef, c.Sk) {
+	if co.TrustedMaterial == nil && keylessVerification(c.KeyRef, c.Sk) {
 		if err := loadCertsKeylessVerification(c.CertChain, c.CARoots, c.CAIntermediates, co); err != nil {
 			return err
 		}
@@ -202,13 +236,15 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		}
 		if c.CertChain == "" {
 			// If no certChain is passed, the Fulcio root certificate will be used
-			co.RootCerts, err = fulcio.GetRoots()
-			if err != nil {
-				return fmt.Errorf("getting Fulcio roots: %w", err)
-			}
-			co.IntermediateCerts, err = fulcio.GetIntermediates()
-			if err != nil {
-				return fmt.Errorf("getting Fulcio intermediates: %w", err)
+			if co.TrustedMaterial == nil {
+				co.RootCerts, err = fulcio.GetRoots()
+				if err != nil {
+					return fmt.Errorf("getting Fulcio roots: %w", err)
+				}
+				co.IntermediateCerts, err = fulcio.GetIntermediates()
+				if err != nil {
+					return fmt.Errorf("getting Fulcio intermediates: %w", err)
+				}
 			}
 			co.SigVerifier, err = cosign.ValidateAndUnpackCert(cert, co)
 			if err != nil {
--- a/pkg/cosign/tuf.go
+++ b/pkg/cosign/tuf.go
@@ -1,0 +1,111 @@
+// Copyright 2025 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cosign
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/sigstore/cosign/v2/pkg/cosign/env"
+	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore-go/pkg/tuf"
+)
+
+func TrustedRoot() (root.TrustedMaterial, error) {
+	opts, err := setTUFOpts()
+	if err != nil {
+		return nil, fmt.Errorf("error setting TUF options: %w", err)
+	}
+	tr, err := root.NewLiveTrustedRoot(opts)
+	if err != nil {
+		return nil, fmt.Errorf("error getting live trusted root: %w", err)
+	}
+	return tr, nil
+}
+
+// setTUFOpts sets the TUF cache directory, the mirror URL, and the root.json in the TUF options.
+// The cache directory is provided by the user as an environment variable TUF_ROOT, or the default $HOME/.sigstore/root is used.
+// The mirror URL is provided by the user as an environment variable TUF_MIRROR. If not overridden by the user, the value set during `cosign initialize` in remote.json in the cache directory is used.
+// If the mirror happens to be the sigstore.dev production TUF CDN, the options are returned since it is safe to use all the default settings.
+// If the mirror is a custom mirror, we try to find a cached root.json. We must not use the default embedded root.json.
+// If the TUF options cannot be found through these steps, the caller should not try to use this TUF client to fetch the trusted root and should instead fall back to the legacy TUF client to fetch individual trusted keys.
+func setTUFOpts() (*tuf.Options, error) {
+	opts := tuf.DefaultOptions()
+	if tufCacheDir := env.Getenv(env.VariableTUFRootDir); tufCacheDir != "" { //nolint:forbidigo
+		opts.CachePath = tufCacheDir
+	}
+	err := setTUFMirror(opts)
+	if err != nil {
+		return nil, fmt.Errorf("error setting TUF mirror: %w", err)
+	}
+	if opts.RepositoryBaseURL == tuf.DefaultMirror {
+		// Using the default mirror, so just use the embedded root.json.
+		return opts, nil
+	}
+	err = setTUFRootJSON(opts)
+	if err != nil {
+		return nil, fmt.Errorf("error setting root: %w", err)
+	}
+	return opts, nil
+}
+
+func setTUFMirror(opts *tuf.Options) error {
+	if tufMirror := env.Getenv(env.VariableTUFMirror); tufMirror != "" { //nolint:forbidigo
+		opts.RepositoryBaseURL = tufMirror
+		return nil
+	}
+	// try using the mirror set by `cosign initialize`
+	cachedRemote := filepath.Join(opts.CachePath, "remote.json")
+	remoteBytes, err := os.ReadFile(cachedRemote)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil // `cosign initialize` wasn't run, so use the default
+	}
+	if err != nil {
+		return fmt.Errorf("error reading remote.json: %w", err)
+	}
+	remote := make(map[string]string)
+	err = json.Unmarshal(remoteBytes, &remote)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling remote.json: %w", err)
+	}
+	opts.RepositoryBaseURL = remote["mirror"]
+	return nil
+}
+
+func setTUFRootJSON(opts *tuf.Options) error {
+	// TUF root set by TUF_ROOT_JSON
+	if tufRootJSON := env.Getenv(env.VariableTUFRootJSON); tufRootJSON != "" { //nolint:forbidigo
+		rootJSONBytes, err := os.ReadFile(tufRootJSON)
+		if err != nil {
+			return fmt.Errorf("error reading root.json given by TUF_ROOT_JSON")
+		}
+		opts.Root = rootJSONBytes
+		return nil
+	}
+	// Look for cached root.json
+	cachedRootJSON := filepath.Join(opts.CachePath, tuf.URLToPath(opts.RepositoryBaseURL), "root.json")
+	if _, err := os.Stat(cachedRootJSON); !os.IsNotExist(err) {
+		rootJSONBytes, err := os.ReadFile(cachedRootJSON)
+		if err != nil {
+			return fmt.Errorf("error reading cached root.json")
+		}
+		opts.Root = rootJSONBytes
+		return nil
+	}
+	return fmt.Errorf("could not find cached root.json")
+}
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -29,6 +29,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"log"
 	"net/http"
 	"os"
 	"regexp"
@@ -38,7 +39,6 @@ import (
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/digitorus/timestamp"
 	"github.com/go-openapi/runtime"
-
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
@@ -66,6 +66,7 @@ import (
 	sgbundle "github.com/sigstore/sigstore-go/pkg/bundle"
 	"github.com/sigstore/sigstore-go/pkg/fulcio/certificate"
 	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore-go/pkg/tlog"
 	"github.com/sigstore/sigstore-go/pkg/verify"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
@@ -95,6 +96,9 @@ type CheckOpts struct {
 
 	// ClaimVerifier, if provided, verifies claims present in the oci.Signature.
 	ClaimVerifier func(sig oci.Signature, imageDigest v1.Hash, annotations map[string]interface{}) error
+
+	// TrustedMaterial contains trusted metadata for all Sigstore services. It is exclusive with RekorPubKeys, RootCerts, IntermediateCerts, CTLogPubKeys, and the TSA* cert fields.
+	TrustedMaterial root.TrustedMaterial
 
 	// RekorClient, if set, is used to make online tlog calls use to verify signatures and public keys.
 	RekorClient *client.Rekor
@@ -170,10 +174,6 @@ type CheckOpts struct {
 
 	// NewBundleFormat enables the new bundle format (Cosign Bundle Spec) and the new verifier.
 	NewBundleFormat bool
-
-	// TrustedMaterial is the trusted material to use for verification.
-	// Currently, this is only applicable when NewBundleFormat is true.
-	TrustedMaterial root.TrustedMaterial
 }
 
 type verifyTrustedMaterial struct {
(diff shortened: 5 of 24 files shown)


cmurphy:
There is still work to do here but I will be out for a couple of weeks so it might be worth getting some eyes on in the meantime.

jku:
> There is still work to do here but I will be out for a couple of weeks so it might be worth getting some eyes on in the meantime.

sorry, I missed this at the time: I will have a look today or monday

jku:
> Sigstore-go provides a way to check for a trusted root and automatically use it if available, but can also fetch individual targets as needed if the provided TUF mirror does not supply a trusted_root.json.

Can you specify what cosign does with sigstore-go? I can see there are some invidual target fetches in the code but does cosign now use trusted_root.json by default?

cmurphy:
> Can you specify what cosign does with sigstore-go?

Cosign uses sigstore-go piecemeal for various things, but the goal of this PR is to adopt sigstore-go's TUF client wrapper instead of the client wrapper provided by sigstore/sigstore.

> does cosign now use trusted_root.json by default?

It does not, that's part of the purpose of this PR and #3548. There is another PR in progress https://github.com/sigstore/cosign/pull/3854 that adds a --trusted-root flag that I will have to adjust this PR to conform with.

> sorry, I missed this at the time: I will have a look today or monday

I may now have to wait for that other PR to be fleshed out before I can make active progress on this, so don't feel rushed to review this. I can ping you again when it's in a more stable state.

pkg/cosign/fulcio.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+const (
+	// This is the root in the fulcio project.
+	fulcioTargetStr = `fulcio.crt.pem`
+	// This is the v1 migrated root.
+	fulcioV1TargetStr = `fulcio_v1.crt.pem`
+	// This is the untrusted v1 intermediate CA certificate, used or chain building.
+	fulcioV1IntermediateTargetStr = `fulcio_intermediate_v1.crt.pem`
+)
```
codysoyland, line 36: I wonder if these filenames should be configurable somewhere for private Sigstore TUF operators. The existing metadata format allows for additional targets to be added and discovered, but TUF v2 does not allow iterating over targets, so that strategy is unsupported, meaning this current diff does not necessarily support all private TUF deployments.

We could also provide a CLI utility to convert an old TUF v1 layout to a `trusted_root.json` and require private TUF maintainers to use it to generate a trusted root in order to support the next version of cosign. I kind of like this option as it fast tracks the adoption of the trusted root.

pkg/cosign/tuf.go
```diff
@@ -0,0 +1,91 @@
...
+func checkValidityPeriod(start, end time.Time) tufStatus {
+	now := time.Now()
+	if now.Before(start) {
+		return inactive
+	}
+	if now.After(end) {
+		return inactive
+	}
+	return active
+}
```
codysoyland, line 91: The validity period in the TrustedRoot should be compared to the timestamp provided by the timestamping service or transparency log. I understand this is not something that is known at the time that you are assembling the CheckOpts, and that makes this a hard problem, but I'm not sure we want to enforce this in this version of the code. I believe we still want to be able to verify a signature produced prior to a timestamping service's validity end date, even if the current time is after that date.

pkg/cosign/fulcio.go
cmurphy: > I wonder if these filenames should be configurable somewhere for private Sigstore TUF operators

These filenames are hardcoded in the [TUF v1 code](https://github.com/sigstore/sigstore/blob/745a80351bcaac1f814672fa14460f1f20ef0ff8/pkg/fulcioroots/fulcioroots.go#L38-L45) so recreating them here was an attempt to align with the old way of retrieving targets.

> The existing metadata format allows for additional targets to be added and discovered, but TUF v2 does not allow iterating over targets, so that strategy is unsupported, meaning this current diff does not necessarily support all private TUF deployments.

You're right, that was an oversight on my part. I had an earlier version of this that could support discovering targets from custom metadata that I will restore.

> We could also provide a CLI utility to convert an old TUF v1 layout to a trusted_root.json

We could do that as a last resort, but my hope was to maintain full backwards compatibility and ease users gently toward using trusted_root.json.

pkg/cosign/tuf.go
cmurphy: My goal here was to have something to substitute for the `Active`/`Expired` [status](https://github.com/sigstore/sigstore/blob/745a80351bcaac1f814672fa14460f1f20ef0ff8/pkg/tuf/status_type.go#L26-L27) from TUF v1 that does not seem to have an equivalent in TUF v2. Open to discarding this entirely or finding an alternative.

haydentherapper: For what it's worth, I don't think Cosign is effectively using Active/Expired now, as Cosign is only printing a message to stdout if you are verifying an entry with an "expired" key. This isn't really helpful to the user because there's nothing wrong with using an expired key, it's if the key is used to verify something outside of its validity window, which isn't currently implemented in Cosign.

If it'd be easier to just not deal with Active/Expired, I'd be supportive of that. We can either rely on sigstore-go down the line to do that check, or we can file an issue to implement that feature in Cosign's verification API when using the trusted root file. I'd lean towards the latter in line with https://github.com/sigstore/cosign/issues/3879.

cmurphy:
@steiza , @codysoyland , @haydentherapper and I met to discuss a path forward on this and we decided to shrink the scope of this PR by removing the fetch of trusted_root.json, to be addressed separately.

This PR dropped the fetch of trusted_root.json, but it now fails conformance tests because the SCT returned during signing from p-g-i is signed by the ctfe_2022.pub key, which the sigstore-go TUF client has no way to know about. This worked fine with p-g-i when trusted_root.json was considered, but shows that this approach will break any deployment that has rotated keys relying on custom metadata. I'm starting to think that using the new client must also be synchronized with using trusted_root.json and that the whole thing should be guarded by a flag that users can opt into. Thoughts?

steiza:
> but shows that this approach will break any deployment that has rotated keys relying on custom metadata

Oof, what a mess we've made for ourselves. Yeah, this is what makes the cosign modernization effort we're working on so tricky.

> I'm starting to think that using the new client must also be synchronized with using trusted_root.json and that the whole thing should be guarded by a flag that users can opt into.

This is what I'm thinking as well. #3854 could be updated to use `CheckOpts` as @codysoyland described in #3879. Then for those commands that support `--trusted-root` could also add something like `--tuf-url` + `--tuf-root` that would fetch the `trusted_root.json` for you, instead of having to supply it with `--trusted-root`.

cmurphy:
The latest change adds back the `root.NewLiveTrustedRoot` TUF fetch, now guarded by environment variables so the user can opt in. Using environment variables seems to be pretty common in cosign and it makes it so all the TUF logic is concentrated in the `pkg/cosign` package and we don't have to plumb through command line flags. It doesn't add a `TrustedMaterial` field to `CheckOpts` because it's not needed with this approach, the TUF functions work mostly the same as they did before. This could still work with #3854 where that PR would allow explicitly providing a trusted_root.json file instead of relying on either TUF client. I'd be happy to chat more about this on a call.

steiza:
I think we're still at odds a bit in terms of approach (but it's also possible I'm missing something!)

Ultimately, we have to decide what verification path we use in different scenarios, `sigstore-go` or `pkg/cosign/verify`.

Here's my understanding of the current state of things (which we can of course change):

| Scenario | Verification Path |
| --- | --- |
| Verifying with a new protocol buffer bundle (which requires a trusted root) | `sigstore-go` (landed in #3796) |
| Verifying disparate signed materials or an old bundle with a trusted root | `sigstore-go` (proposed in #3854) |
| Verifying anything with TUF v2 | `pkg/cosign/verify` (proposed here in #3844) |
| Verifying anything with TUF v2 | `sigstore-go` (suggested by @steiza - fetch the trusted root and then use proposed code in #3854) |

It is certainly possible to use `pkg/cosign/verify` here with TUF v2, but I think our current tests are missing the case where there are several sets of verification material and today `pkg/cosign/verify` picks one to use instead of checking all of them, including at least SCT verification, `VerifySet`, and TSA verification that I listed out on https://github.com/sigstore/cosign/pull/3854#issuecomment-2353573009.

We could make this pull request address those cases. But ultimately, we need #3844 and #3854 to agree on what the verification path should be when you're using TUF v2 - `sigstore-go` or `pkg/cosign/verify`.

steiza:
I was confused again this morning, and worried I lumped together too many use-cases, so I made a more detailed table:

| Signed materials | Verification materials | Verification materials provided by | Verification Path |
| --- | --- | --- | --- |
| Protocol buffer bundle | requires trusted root | trusted root file on disk | `sigstore-go` (landed in https://github.com/sigstore/cosign/pull/3796) |
| Protocol buffer bundle | requires trusted root | TUFv1 | N/A - we require a trusted root, which TUFv1 does not provide |
| Protocol buffer bundle | requires trusted root | TUFv2 | `sigstore-go` (suggested by @steiza - fetch the trusted root and then use code landed in https://github.com/sigstore/cosign/pull/3796) |
| Not protocol buffer bundle | trusted root | trusted root file on disk | `sigstore-go` (proposed in https://github.com/sigstore/cosign/pull/3854) |
| Not protocol buffer bundle | trusted root | TUFv1 | N/A - TUFv1 does not supply a trusted root, see "not trusted root" below |
| Not protocol buffer bundle | trusted root | TUFv2 | `pkg/cosign/verify` (proposed here in https://github.com/sigstore/cosign/pull/3844) **or** row below |
| Not protocol buffer bundle | trusted root | TUFv2 | `sigstore-go` (suggested by @steiza - fetch the trusted root and then use proposed code in https://github.com/sigstore/cosign/pull/3854) **or** row above |
| Not protocol buffer bundle | not trusted root (e.g. disparate material via flags or ENV vars) | TUFv1 | `pkg/cosign/verify` (already implemented) |
| Not protocol buffer bundle | not trusted root (e.g. disparate material via flags or ENV vars) | TUFv2 | This doesn't make sense - if you're using TUFv2 you're going to have a trusted root |

I think I ended up in the same place though.

In the case where we don't have a new protocol buffer bundle, and we have a trusted root that we fetched with TUFv2, we need to decide if the verification path is going to be:
- `pkg/cosign/verify` (in which case we need to audit and improve that verification path to handle the lists of possible verification material the trusted root provides) **or**
- `sigstore-go` where we've already implemented that functionality (which means modifying commands individually)

jku:
thanks for writing that up zach, I'm trying to catch up and this is useful

cmurphy:
@steiza thanks for the summary, I have a couple of minor clarifying questions:

> This doesn't make sense - if you're using TUFv2 you're going to have a trusted root

Do you mean ideally, or currently? Using the TUF v2 client doesn't guarantee you've created a trusted_root.json and added it to your TUF repository.

> pkg/cosign/verify (proposed here in #3844) or row below

Technically this PR does very little in `pkg/cosign/verify.go`. The scope of the TUF change also needs to include getting keys for the CT server which are needed during [signing](https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/sign/sign.go#L508).

jku:
> if new environment variables TUF_MIRROR, TUF_ROOT_JSON or
TUF_USE_TRUSTED_ROOT are set, use those to instantiate a TUF v2 client.

How do we expect these to get set? By end users?

steiza:
> Do you mean ideally, or currently? Using the TUF v2 client doesn't guarantee you've created a trusted_root.json and added it to your TUF repository.

Aha, I think I finally understand our different perspectives! I thought that when we use a TUF v2 client we were only going to fetch the `trusted_root.json` and not fetch individual files over TUF like `ctfe.pub` - but maybe that's that not the case? The way #3548 is worded made me think that we're doing the TUF v2 client upgrade and moving to using `trusted_root.json` at the same time.

If we are allowing fetching files like `ctfe.pub` over TUF v2, then I agree that we should use the contents of those files with the existing `pkg/cosign/verify` verification path (or signing, like you linked).

But if we're fetching `trusted_root.json` over TUF v2, that's when we need to decide if we're going to use `sigstore-go` or if we're going to make additional changes to `pkg/cosign/verify` in this pull request. The trusted root provides lists of possible verification material, which the `pkg/cosign/verify` was not originally written to handle.

For example, we verify the certificate with `intermediateCerts *x509.CertPool` that includes all the intermediate certs listed in the trusted root (great), but we [always pick the first chain to verify the SCT](https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L275-L282) (not great). I listed out two other similar cases on https://github.com/sigstore/cosign/pull/3854#issuecomment-2353573009, but I haven't done a close audit of `pkg/cosign/verify` either.

haydentherapper:
I've caught up on all the discussion, thanks Zach for the thorough chart!

https://github.com/sigstore/cosign/issues/3548 was not clear on what was in scope for the TUF v2 upgrade. When I wrote it, my goal was to move to TUF v2 independent of the trusted root changes. This was when the client was quite new, so I wanted to be able to test out TUF v2 without making any other changes, leveraging the client to fetch root material by name. As noted in the issue, we'd still need to support TUF v1 to fallback to the custom metadata and for private deployments that are using custom metadata.

There is less motivation for doing this change by itself as we have more confidence in the TUF client now, though I had hoped we could make the change by itself still because it will keep these PRs smaller. The original version of this PR - where you initialize metadata with the new TUF client and fallback to the old TUF client - was what I had in my head when I wrote the issue.

My concern with using env vars isn't necessarily that users have to set it - functionally, TUF v1 and v2 should be equivalent, so setting these are more for our testing during development than the end users. The concern is that we have to rely on the hardcoded target names without custom metadata. If we have to rotate the roots in the next N months, then the TUF v2 client won't pick up these changes unless we hardcode more filenames. This is why I had proposed to concurrently support TUF v2 and v1 clients - leverage TUF v2 for harcoded targets, TUF v1 for custom metadata, and merge.

I also recognize that this is functionally a no-op, that we get none of the benefits of the trusted root (validity windows) and still require provided TUF root to use custom metadata (though a separate topic, we will have to for quite awhile or else we'll break old Cosign clients).

At this point, I think we need to make a call on next steps. Are we going to interlink TUF v2 and the trusted root work, and say that a) TUF v1 is only for custom metadata and only support "legacy" verification with detached material/the old bundle format and b) if you provide either a trust root or verification bundle, the client will use TUF v2?

My only hesitance is around the sequencing of the refactoring. I want to make sure we keep PRs small so we can iterate faster. Having some intermediate PRs like this lets us test out well-scoped changes.

--------

Tangentially, addressing:

> pkg/cosign/verify (in which case we need to audit and improve that verification path to handle the lists of possible verification material the trusted root provides) or

I think it would be worthwhile to do this. This is the only way we can minimize the likelihood of introducing unexpected breaking changes. I'll respond on #3854 about this.

cmurphy:
I see now how trying to switch the internals of cosign's individual key material fetching functions isn't going to result in a seamless outcome. I'm afraid, however, that conditioning the which-TUF-client decision on what kind of bundle is in use is also going to miss several important pathways, namely signing (there is no bundle yet) and image verification (also no protobuf bundle). I've attempted to take a step back and enumerate everywhere we are obtaining and using trusted keys in some form:

## Key type: CT Log

| Action                                   | TUF fetch happens                                                                                                                           | Used                                                                                                                                                                                                                  |
|------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| creating a non-ephemeral-key signer      | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/sign/sign.go#L508                           | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/sign/sign.go#L515                                                                                                     |
| creating a fulcio signer                 | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go#L36 | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go#L42                                                                           |
| verifying an SCT in an image             | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify.go#L188                       | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/verify.go#L280 https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/verify.go#L296 |
| verifying an SCT in an image attestation | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_attestation.go#L128           | same as above                                                                                                                                                                                                         |
| verifying an SCT for a blob              | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_blob.go#L297                  | same as above                                                                                                                                                                                                         |
| verifying an SCT for a blob attestation  | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_blob_attestation.go#L195      | same as above                                                                                                                                                                                                         |

### Proposal

Where `cosign.GetCTLogPubs()` is called, use [`root.NewLiveTrustedRoot()`](https://github.com/sigstore/sigstore-go/blob/cf4aa434daa63ead3b449c1f495e98788f08c4ab/pkg/root/trusted_root.go#L397). If `CheckOpts` is available (verifier case), store the trusted root as a field on `CheckOpts`. Otherwise (signer case), use the trusted root directly or store on `KeyOpts`. If the returned trusted root, wherever stored, is non-nil, pass it to sigstore-go's [`verify.VerifySCT()`](https://github.com/sigstore/sigstore-go/blob/cf4aa434daa63ead3b449c1f495e98788f08c4ab/pkg/verify/sct.go#L32). Otherwise, use cosign's `cosign.VerifySCT()` or `cosign.VerifyEmbeddedSCT()`.

Caveat: cosign's version of `VerifySCT` has options for embedded and non-embedded SCTs in certificates, so will have to make some changes to make it compatible with sigstore-go's version ([embedded](https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/verify_sct.go#L109), [non-embedded](https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/verify_sct.go#L133)).

## Key type: Fulcio CA

| Action                                        | TUF fetch happens                                                                                                                                                                                                                                        | Used                                                                                                       |
|-----------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------|
| verifying certificate in an image             | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify.go#L224-L231 or https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify.go#L565-L572 | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/verify.go#L251 |
| verifying certificate in an image attestation | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_attestation.go#L196-L203                                                                                                                   | same as above                                                                                              |

The Fulcio CA is NOT retrieved from TUF for verifying the certificate for blobs or blob attestations - the Fulcio certificate must be passed in by a flag or extracted from the bundle.

### Proposal

Where `fulcio.GetRoots()` and `fulcio.GetIntermediates()` are called, use [`root.NewLiveTrustedRoot()`](https://github.com/sigstore/sigstore-go/blob/cf4aa434daa63ead3b449c1f495e98788f08c4ab/pkg/root/trusted_root.go#L397) and store the result on `CheckOpts`. If the trusted root is non-nil, pass it to sigstore-go's [`verify.VerifyLeafCertificate()`](https://github.com/sigstore/sigstore-go/blob/cf4aa434daa63ead3b449c1f495e98788f08c4ab/pkg/verify/certificate.go#L25). Otherwise, use cosign's `cosign.TrustedCert()`.

## Key type: Rekor pub

| Action                                   | TUF fetch happens                                                                                                                      | Used                                                                                                                                                                                                                                        |
|------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| verifying an SET in an image             | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify.go#L172                  | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/verify.go#L1087 and again https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/pkg/cosign/tlog.go#L492 (duplicate?) |
| verifying an SET in an image attestation | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_attestation.go#L154      | same as above                                                                                                                                                                                                                               |
| verifying an SET for a blob              | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_blob.go#L164             | same as above                                                                                                                                                                                                                               |
| verifying an SET for a blob attestation  | https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_blob_attestation.go#L182 | same as above                                                                                                                                                                                                                               |

### Proposal

Where `cosign.GetRekorPubs()` is called, use [`root.NewLiveTrustedRoot()`](https://github.com/sigstore/sigstore-go/blob/cf4aa434daa63ead3b449c1f495e98788f08c4ab/pkg/root/trusted_root.go#L397) and store the result on `CheckOpts`. If the trusted root is non-nil, pass it to sigstore-go's [`tlog.VerifySET()`](https://github.com/sigstore/sigstore-go/blob/cf4aa434daa63ead3b449c1f495e98788f08c4ab/pkg/tlog/entry.go#L266). Otherwise, use cosign's `cosign.VerifySET()`. Also, see if the calls to `VerifySET` can be de-duplicated.


---

@steiza 
> I thought that when we use a TUF v2 client we were only going to fetch the trusted_root.json and not fetch individual files over TUF like ctfe.pub

@haydentherapper 
> When I wrote it, my goal was to move to TUF v2 independent of the trusted root changes.
> ...
> I also recognize that this is functionally a no-op

I'm in agreement with @steiza now, I think separating TUF v2 from trusted root is a non-starter at this point. TUF v2 does not support custom metadata, so it's not just a no-op, it's really a big step backwards and makes users less secure. Really the *only* reason to use the TUF v2 client is to enable support for fetching the trusted root.

@jku 
> It feels wrong if users need to set an environment variable to make the software use the recommended trust root mechanism.

@haydentherapper 
> My concern with using env vars isn't necessarily that users have to set it - functionally, TUF v1 and v2 should be equivalent, so setting these are more for our testing during development than the end users

In this PR's current design (which we do not want to follow anymore), users would have had to set those env vars in order to opt into trusted root and TUF v2. In the proposal I've outlined above, I think we can make them optional, since we will always fall back to TUF v1 if we can't safely initialize a TUF v2 client or the TUF mirror does not have a trusted root.

cmurphy:
I've updated this PR to approximately implement the plan I laid out in [my comment](https://github.com/sigstore/cosign/pull/3844#issuecomment-2389974709). I've done some light manual testing but it will need significant e2e and unit testing. I wanted to make sure there is buy-in on this approach before setting to work on that. I think we should also add support for setting `--trusted-root` like @steiza started in https://github.com/sigstore/cosign/pull/3854 but I think that can be done in a separate PR.

steiza:
I think this approach is great! But I'll admit I don't fully understand the constraints of OCI verification, so we should probably get feedback from @codysoyland as well.

I think these changes around getting the trusted root could be used to improve the implementation in #3796 or (like you say) bring back some elements of #3854, but I agree that that can wait for a later pull request.

Nice work!

haydentherapper:
@cmurphy Fantastic summary of the work necessary to implement this! I'm supportive of all of the above.

> The Fulcio CA is NOT retrieved from TUF for verifying the certificate for blobs or blob attestations - the Fulcio certificate must be passed in by a flag or extracted from the bundle.

For verification, root certificates should be pulled through TUF through an [API in sigstore/sigstore](https://github.com/sigstore/sigstore/blob/main/pkg/fulcioroots/fulcioroots.go). We had started a refactor a long time ago to move TUF logic into this shared repo but never finished. Or is this referring to bring-your-own PKI where roots are provided via flag?

> I'm in agreement with @steiza now, I think separating TUF v2 from trusted root is a non-starter at this point. TUF v2 does not support custom metadata, so it's not just a no-op, it's really a big step backwards and makes users less secure. Really the only reason to use the TUF v2 client is to enable support for fetching the trusted root.

That makes sense, and if it's less work to not try and separate these two goals, let's not try to. Our goal is to adopt TUF v2 wholly, so no point in trying to have an intermediate no-op state.

cmurphy:
@haydentherapper 

> For verification, root certificates should be pulled through TUF through an API in sigstore/sigstore. We had started a refactor a long time ago to move TUF logic into this shared repo but never finished. Or is this referring to bring-your-own PKI where roots are provided via flag?

This was referring to the slightly different logic between `cosign verify-blob`/`cosign verify-blob-attestation` and `cosign verify`/`cosign verify-attestation` currently, where in the OCI case the fulcio roots are [fetched from TUF](https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify.go#L224-L231) if not provided via flag, while in the blob case [they aren't necessarily](https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_blob.go#L198-L201), though I think I missed that they [are fetched from TUF in the keyless case](https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/cmd/cosign/cli/verify/verify_blob.go#L280-L281) on my first pass.

cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
```diff
@@ -32,12 +34,26 @@ func NewSigner(ctx context.Context, ko options.KeyOpts, signer signature.SignerV
...
 	}
 
-	// Grab the PublicKeys for the CTFE, either from tuf or env.
+	if ko.TrustedMaterial != nil && len(fs.SCT) == 0 {
+		// We assume that if a trusted_root.json was found, the fulcio chain was included in it.
+		// fs.Chain will be ignored as root.VerifySignedCertificateTimestamp relies on the trusted root.
+		// Detached SCTs cannot be verified with this function.
+		certs, err := cryptoutils.UnmarshalCertificatesFromPEM(fs.Cert)
+		if err != nil || len(certs) < 1 {
+			return nil, fmt.Errorf("unmarshalling SCT from PEM: %w", err)
```
loosebazooka, line 43: Would "`unmarshalling certificate from PEM for SCT verification`" make more sense? or am I misreading this?

loosebazooka (commented):
I read over the history, and (perhaps I missed it) has this option been considered:

Is there a way to just build a trusted_root.json from the provided materials. We provide sensible defaults for config (like if provided on the command line -- they are just valid forever). Then only use the tursted_root code path for everything?

If that sounds dumb then I'll just finish up a regular code review here.

steiza:
> Is there a way to just build a trusted_root.json from the provided materials.

Yeah, so there are several things happening in parallel, if I understand your question correctly.

This pull request is "just" about upgrading cosign's TUF client from the existing version which fetches verification material as individual files to TUFv2 which fetches all the verification material in one `trusted_root.json` file. I put "just" in quotes, because that's actually quite a bit of work: we also have to adapt the various parts of the cosign verification path to make use of a `trusted_root.json` file where there might be multiple valid verification material options for a given time.

I think your question is "in the future, once cosign verification paths support making use of a `trusted_root.json` file, can we change the old verification paths that make use of individual files and instead assemble them into a `trusted_root.json` file and use the new verification path?"

My answer is that we could, although we might also delete those old verification paths and ask people to generate their own trusted root with something like #3876. But I think the work laid out in this pull request is a prerequisite, either way.

steiza:
@codysoyland and I talked about this pull request yesterday at the sigstore-go meeting. We think this is compatible with the work he's doing on OCI and is good to proceed!

codysoyland:
As @steiza mentioned, I've had a bit of time to look at the latest changes here, and I'm pleased with the recent changes! It will complement my work on [container image bundle verification](https://github.com/sigstore/cosign/pull/3889) quite well (we even both added `TrustedMaterial` to CheckOpts!). Apologies for not following this PR very closely recently! I'm close to getting that PR into review, perhaps next week. Looking forward to seeing you both at SigstoreCon in a few days!

cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go
cmurphy: Yes, will fix

cmurphy:
Addressed comments, added unit tests mainly for `initialize` and for the fulcio verifier which has the tricky detached SCT handling, added an e2e test that uses trusted_root.json, and fixed several issues uncovered through tests. Marking this ready for review because I think this is close to the final product, but I still need to debug the conformance and attestation tests which may introduce small changes, and I may not be able to get back to this for a bit.

cmurphy:
Unable to reproduce the conformance test issue because of a different python issue, also appearing here https://github.com/sigstore/cosign/actions/runs/12717693057/job/35454646340?pr=4006

jku:
> Unable to reproduce the conformance test issue because of a different python issue, also appearing here https://github.com/sigstore/cosign/actions/runs/12717693057/job/35454646340?pr=4006

I believe I have a fix: https://github.com/sigstore/sigstore-conformance/pull/179
 will try to remember to update here when it's released

haydentherapper:
Conformance fix is now merged.

jku:
> Conformance fix is now merged.

Yes, upgrading the action to v0.0.16 should work (`sigstore/sigstore-conformance@d658ea74a060aeabae78f8a379167f219dc38c38 # v0.0.16`)

haydentherapper:
Just need to rebase past https://github.com/sigstore/cosign/pull/4011

bkabrda:
Hi folks, thank you all for all your hard work on this, this is awesome.

I gave this PR a test with the private Sigstore instances that I'm working on and I ran into one problem: We're using RSA PSS keys to sign our TUF metadata and it seems that go-tuf v2 only got a support for these recently in https://github.com/theupdateframework/go-tuf/pull/625 - would it be possible to upgrade to a version of go-tuf that includes this PR? I'd be happy to help doing the updates once a new version of go-tuf is released, but not sure if/when that's going to happen.

jku:
> would it be possible to upgrade to a version of go-tuf that includes this PR?

Almost certainly: go-tuf just needs to make that release.

> one thing that would be really useful to help debugging this kind of issues is if there was a (debug) message saying that the trusted_root.json was in fact found, but failed to be parsed (and why).

it sounds like the failure is that go-tuf considers the repository invalid so it never even got to downloading `trusted_root.json` (trusted_root.json is the artifact that cosign is trying to download using TUF, it just happens to be similar in name as the TUF metadata file that contains the "invalid" key): I'm sure the cosign error message could be improved though

bkabrda:
> it sounds like the failure is that go-tuf considers the repository invalid so it never even got to downloading trusted_root.json

This does happen if the repository is considered invalid, but it also happens when `trusted_root.json` is invalid (in our case we had some issues in how we serialized `keyBytes` for some targets). Either way including a debugging message that shows the underlying error would cover both cases, so I believe that's the best way to go.

cmurphy:
This is passing tests now and should be ready for review. Apologies for the delay.

cmd/cosign/cli/attest.go
```diff
@@ -70,6 +73,12 @@ func Attest() *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			trustedMaterial, err := cosign.TrustedRoot()
+			if err != nil {
+				ui.Warnf(context.Background(), "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
+			}
```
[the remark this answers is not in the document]
steiza: ++, although I think this comment makes more sense on the `verify` commands where we have `--offline` as a strong single if we're doing air-gapped verification.

You **could** be doing `attest`/`sign` commands in a restricted network environment (maybe there's a HSM attached to the network) although I'm not sure we have a flag we can key off of (like `--offline` for `verify` commands). I don't know enough about `cosign` usage to know if we need to support signing in restricted network environments.

steiza (commented):
This is a monster pull request, although I did read it all the way through once (and then went back to several parts). Generally I think this is spot-on, but a few details I still have questions on.

cmd/cosign/cli/attest.go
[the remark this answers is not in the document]
cmurphy: I left the sign commands as-is because there is no flag to indicate where to get an offline trusted root. There should be, but I think that's outside of the scope of this already very large PR. I did rearrange the verify commands to make sure they don't try to fetch the TUF root online if a trusted root file is already provided.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -128,15 +145,15 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 	}
 
 	// Ignore Signed Certificate Timestamp if the flag is set or a key is provided
-	if shouldVerifySCT(c.IgnoreSCT, c.KeyRef, c.Sk) && !c.NewBundleFormat {
+	if co.TrustedMaterial == nil && shouldVerifySCT(c.IgnoreSCT, c.KeyRef, c.Sk) && !c.NewBundleFormat {
```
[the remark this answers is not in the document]
cmurphy: Fixed

steiza (commented):
I think once we fix the linter and e2e tests, this is good-to-go!

cmurphy:
https://github.com/sigstore/cosign/pull/4216 for the e2e test failures.

steiza (dismissed):
Nice work!

cmd/cosign/cli/attest.go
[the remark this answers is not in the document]
Hayden-IO: What happens now when you sign with a provided key rather than using an identity/fetching a certificate? Unless you're initializing a [signer using Fulcio](https://github.com/sigstore/cosign/blob/c03e2a8f883d54b0bfbadbac8497e6efcf6fb364/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go#L36) I'm fairly certain that there is no attempt to fetch TUF targets in the background. (If there is, that's arguably a bug in Cosign).

With just a key and signing a blob, signing should be offline (signing is not offline of course if writing to an OCI registry or if `--issue-certificate` is set to fetch a fulcio cert). It's also worth noting that Cosign may be used behind a firewall and even if signing requires a network connection (signing/attesting a container with a key), there may be a limited set of allowed domains for outbound traffic. This isn't hypothetical either, we've gotten bug reports/FRs about this in the past.

Is there a set of flag values we can use to determine if TUF targets are needed?

For Cosign v3, we do need to clarify what `--offline` does to actually mean "there are no outbound connections" and disallow certain flag combos.

cmd/cosign/cli/sign/sign.go
```diff
@@ -497,24 +499,33 @@ func signerFromKeyRef(ctx context.Context, certPath, certChainPath, keyRef strin
 	if _, err := cosign.TrustedCert(leafCert, rootPool, subPool); err != nil {
 		return nil, fmt.Errorf("unable to validate certificate chain: %w", err)
 	}
+	certSigner.Chain = certChainBytes
 	// Verify SCT if present in the leaf certificate.
 	contains, err := cosign.ContainsSCT(leafCert.Raw)
 	if err != nil {
 		return nil, err
 	}
-	if contains {
```
Hayden-IO: FWIW, this check isn't really meaningful and I'm tempted to say we just scrap it, to simplify this and limit when we need TUF metadata.

On signing with Fulcio, we [verify the SCT](https://github.com/sigstore/cosign/blob/c03e2a8f883d54b0bfbadbac8497e6efcf6fb364/cmd/cosign/cli/fulcio/fulcioverifier/fulcioverifier.go#L42) which is more of just an early check that verification will succeed when done by the consumer.

Here, this is when a certificate was fetched out of band, e.g private PKI, and will be attached to the container. It doesn't matter if the certificate contains an SCT or not, this is just provided by the signer. If it doesn't contain an SCT, then verifiers must later specify the `ignoreSCT` flag regardless.

Arguably, neither of these checks matter, as verification policy doesn't necessarily match what the signer might trust. I think this check here is even less meaningful since it's really just meant for private PKI which very likely doesn't use CT.

pkg/cosign/tuf.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	err = setTUFRootJSON(opts)
+	if err != nil {
+		return nil, fmt.Errorf("error setting root: %w", err)
+	}
+	return opts, nil
+}
+
+func setTUFMirror(opts *tuf.Options) error {
+	if tufMirror := env.Getenv(env.VariableTUFMirror); tufMirror != "" { //nolint:forbidigo
```
Hayden-IO, line 68: We currently have a flag for the mirror with `cosign initialize`, `--mirror`, and I believe also one for the root `--root`. Can we reuse these along with the env var?

pkg/cosign/verify.go
```diff
@@ -1220,6 +1257,38 @@ func VerifyBundle(sig oci.Signature, co *CheckOpts) (bool, error) {
...
+func (e *signedEntityForTimestamp) SignatureContent() (verify.SignatureContent, error) {
+	return e.sigContent, nil
+}
+
+func (s *sigContent) Signature() []byte {
+	return s.rawSig
+}
+
+func (s *sigContent) EnvelopeContent() verify.EnvelopeContent {
+	panic("not implemented")
```
Hayden-IO: Can we return nil? I'm hesitant to put any panics in here that might inadvertently get called.

pkg/cosign/tuf.go
cmurphy: It would be a little cumbersome to add two new flags for every sign and verify command and pipe those values through to here. I feel like environment variables are friendly to the user because they could just source an environment file and then all their cosign commands work for that TUF environment. How strongly do you feel about adding flags?

pkg/cosign/verify.go
cmurphy: If it gets called it's a programmer error, returning nil would hide that and make it harder to fix the bug. I could change the error message?

cmd/cosign/cli/sign/sign.go
cmurphy: >  it's really just meant for private PKI which very likely doesn't use CT.

"very likely" doesn't mean "never". I wouldn't feel good about burying a behavior change like this inside this giant PR. I can open a separate PR to remove this so discussion can be had on it independently and it can be referred back to if someone has questions about it later.

cmd/cosign/cli/attest.go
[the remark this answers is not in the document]
cmurphy: There's not an existing great set of flags for signing that can be used to declare that it needs to be airgapped. `--key` isn't adequate because it could refer to a KMS or K8s key and isn't necessarily offline. `--tlog-upload=false` and `--issue-certificate=false` also don't not fully encompass the scope of offline-ness.

I can try to move this closer to where it is actually needed.

cmd/cosign/cli/sign/sign.go
cmurphy: https://github.com/sigstore/cosign/pull/4225

Hayden-IO: Thanks for changing this!

pkg/cosign/tuf.go
Hayden-IO: Not strongly, we should note that this is something to resolve in v3 though.

cmd/cosign/cli/attest.go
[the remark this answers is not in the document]
Hayden-IO: If `--tlog-upload` and `--issue-certificate` are `false` and `--key` is used, then there should be no need for TUF metadata.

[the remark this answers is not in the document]
cmurphy: `--tlog-upload` actually isn't relevant because we don't verify the tlog checkpoint during signing (yet). `attest` and `attest-blob` don't use `--issue-certificate`, so I used `--key` for those and both `--key` and `--issue-certificate` for `sign` and `sign-blob`.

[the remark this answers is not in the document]
Hayden-IO: Ah, good call on not needing `tlog-upload`. This SGTM!

Hayden-IO:
Just need to resolve the conflict, this looks great!!

cmd/cosign/cli/attest.go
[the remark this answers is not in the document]
Hayden-IO: Was this added yet, I don't see these changes in the PR?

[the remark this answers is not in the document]
cmurphy: Oops, forgot to push them, they're there now.

haydentherapper (approved):
Amazing work on this!

PR #3876, opened 2024-09-11
Add trusted-root create helper command

### LINKED ISSUE #3700: Add support for providing a trust root file for private deployments

steiza:
#### Summary

Closes #3700

We want to help cosign users move from providing disparate verification material to a single file that contains the needed verification material.

Using a trusted root file makes it easier for users to rotate key material and specify what time period different keys were valid.

Once you create your trusted root with `cosign trusted-root create ...` you can then supply the output to verification commands that support it with `cosign verify* --trusted-root ...`

Here's an example, note that some flags can be specified more than once (for instance, if your certificate chain has rotated, or if you are supporting public good and private instance). You can specify as few or as much content as you'd like. You'll need to create or obtain certificate chains and public key files as needed:

```
$ go run cmd/cosign/main.go trusted-root create --certificate-chain fulcio-chain-public.pem --certificate-chain fulcio-chain-private.pem --timestamp-certificate-chain timestamp.pem --ctfe-key ctfe.pub --rekor-key rekor.pub
```

#### Release Note

* Added `cosign trusted-root create` command to assist users in creating trusted root files to use in cosign verification commands

#### Documentation

https://github.com/sigstore/docs/pull/327

### DIFF
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -23,6 +23,7 @@ type TrustedRootCreateOptions struct {
 	CAIntermediates  string
 	CARoots          string
 	CertChain        string
+	IgnoreSCT        bool
 	Out              string
 	RekorURL         string
 	TSACertChainPath string
@@ -52,6 +53,10 @@ func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.MarkFlagsMutuallyExclusive("ca-roots", "certificate-chain")
 	cmd.MarkFlagsMutuallyExclusive("ca-intermediates", "certificate-chain")
+
+	cmd.Flags().BoolVar(&o.IgnoreSCT, "ignore-sct", false,
+		"when set, do not include key for verifying certificate transparency "+
+			"log. Set this if you signed with a key instead of using Fulcio.")
 
 	cmd.Flags().StringVar(&o.Out, "out", "",
 		"path to output trusted root")
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -23,8 +23,8 @@ type TrustedRootCreateOptions struct {
 	CAIntermediates  string
 	CARoots          string
 	CertChain        string
-	IgnoreSCT        bool
-	IgnoreTlog       bool
+	CtfeKeyPath      string
+	RekorKeyPath     string
 	Out              string
 	TSACertChainPath string
 }
@@ -54,13 +54,12 @@ func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
 	cmd.MarkFlagsMutuallyExclusive("ca-roots", "certificate-chain")
 	cmd.MarkFlagsMutuallyExclusive("ca-intermediates", "certificate-chain")
 
-	cmd.Flags().BoolVar(&o.IgnoreSCT, "ignore-sct", false,
-		"when set, do not include key for verifying certificate transparency "+
-			"log. Set this if you signed with a key instead of using Fulcio.")
+	cmd.Flags().StringVar(&o.CtfeKeyPath, "ctfe-key", "",
+		"path to a PEM-encoded public key used by certificate authority for "+
+			"certificate transparency log.")
 
-	cmd.Flags().BoolVar(&o.IgnoreTlog, "ignore-tlog", false,
-		"when set, do not include key for verifying transparency. Set this if "+
-			"you did not sign with Rekor.")
+	cmd.Flags().StringVar(&o.RekorKeyPath, "rekor-key", "",
+		"path to a PEM-encoded public key used by transparency log like Rekor.")
 
 	cmd.Flags().StringVar(&o.Out, "out", "",
 		"path to output trusted root")
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -24,8 +24,8 @@ type TrustedRootCreateOptions struct {
 	CARoots          string
 	CertChain        string
 	IgnoreSCT        bool
+	IgnoreTlog       bool
 	Out              string
-	RekorURL         string
 	TSACertChainPath string
 }
 
@@ -58,11 +58,12 @@ func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
 		"when set, do not include key for verifying certificate transparency "+
 			"log. Set this if you signed with a key instead of using Fulcio.")
 
+	cmd.Flags().BoolVar(&o.IgnoreTlog, "ignore-tlog", false,
+		"when set, do not include key for verifying transparency. Set this if "+
+			"you did not sign with Rekor.")
+
 	cmd.Flags().StringVar(&o.Out, "out", "",
 		"path to output trusted root")
-
-	cmd.Flags().StringVar(&o.RekorURL, "rekor-url", "",
-		"address of rekor STL server")
 
 	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-certificate-chain", "",
 		"path to PEM-encoded certificate chain file for the RFC3161 timestamp authority. Must contain the root CA certificate. "+
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -1,0 +1,65 @@
+//
+// Copyright 2024 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package options
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type TrustedRootCreateOptions struct {
+	CAIntermediates  string
+	CARoots          string
+	CertChain        string
+	Out              string
+	RekorURL         string
+	TSACertChainPath string
+}
+
+var _ Interface = (*TrustedRootCreateOptions)(nil)
+
+func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.CAIntermediates, "ca-intermediates", "",
+		"path to a file of intermediate CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. "+
+			"The flag is optional and must be used together with --ca-roots, conflicts with "+
+			"--certificate-chain.")
+	_ = cmd.Flags().SetAnnotation("ca-intermediates", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.Flags().StringVar(&o.CARoots, "ca-roots", "",
+		"path to a bundle file of CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. Conflicts with --certificate-chain.")
+	_ = cmd.Flags().SetAnnotation("ca-roots", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.Flags().StringVar(&o.CertChain, "certificate-chain", "",
+		"path to a list of CA certificates in PEM format which will be needed "+
+			"when building the certificate chain for the signing certificate. "+
+			"Must start with the parent intermediate CA certificate of the "+
+			"signing certificate and end with the root certificate. Conflicts with --ca-roots and --ca-intermediates.")
+	_ = cmd.Flags().SetAnnotation("certificate-chain", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.MarkFlagsMutuallyExclusive("ca-roots", "certificate-chain")
+	cmd.MarkFlagsMutuallyExclusive("ca-intermediates", "certificate-chain")
+
+	cmd.Flags().StringVar(&o.Out, "out", "",
+		"path to output trusted root")
+
+	cmd.Flags().StringVar(&o.RekorURL, "rekor-url", "",
+		"address of rekor STL server")
+
+	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-certificate-chain", "",
+		"path to PEM-encoded certificate chain file for the RFC3161 timestamp authority. Must contain the root CA certificate. "+
+			"Optionally may contain intermediate CA certificates")
+}
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -36,7 +36,7 @@ func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
 		"path to a list of CA certificates in PEM format which will be needed "+
 			"when building the certificate chain for the signing certificate. "+
 			"Must start with the parent intermediate CA certificate of the "+
-			"signing certificate and end with the root certificate. Conflicts with --ca-roots and --ca-intermediates.")
+			"signing certificate and end with the root certificate.")
 	_ = cmd.Flags().SetAnnotation("certificate-chain", cobra.BashCompFilenameExt, []string{"cert"})
 
 	cmd.Flags().StringArrayVar(&o.CtfeKeyPath, "ctfe-key", nil,
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -20,51 +20,43 @@ import (
 )
 
 type TrustedRootCreateOptions struct {
-	CAIntermediates  string
-	CARoots          string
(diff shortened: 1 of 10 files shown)


cmd/cosign/cli/options/trustedroot.go
```diff
@@ -0,0 +1,65 @@
...
+	cmd.Flags().StringVar(&o.CertChain, "certificate-chain", "",
+		"path to a list of CA certificates in PEM format which will be needed "+
+			"when building the certificate chain for the signing certificate. "+
+			"Must start with the parent intermediate CA certificate of the "+
+			"signing certificate and end with the root certificate. Conflicts with --ca-roots and --ca-intermediates.")
+	_ = cmd.Flags().SetAnnotation("certificate-chain", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.MarkFlagsMutuallyExclusive("ca-roots", "certificate-chain")
+	cmd.MarkFlagsMutuallyExclusive("ca-intermediates", "certificate-chain")
+
```
cmurphy, line 55: I'm not seeing a way to provide keys for the CT log service, is that intentional?

cmd/cosign/cli/options/trustedroot.go
```diff
@@ -0,0 +1,65 @@
...
+	cmd.Flags().StringVar(&o.Out, "out", "",
+		"path to output trusted root")
+
+	cmd.Flags().StringVar(&o.RekorURL, "rekor-url", "",
```
cmurphy, line 59: What is the rationale for using the rekor URL here instead of an on-disk artifact? How do we know we can trust the key being served at this URL?

cmd/cosign/cli/options/trustedroot.go
```diff
@@ -0,0 +1,65 @@
...
+		"address of rekor STL server")
+
+	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-certificate-chain", "",
+		"path to PEM-encoded certificate chain file for the RFC3161 timestamp authority. Must contain the root CA certificate. "+
+			"Optionally may contain intermediate CA certificates")
+}
```
cmurphy, line 65: It's not uber-critical, but with the options here there's no way to provide a [uri](https://github.com/sigstore/protobuf-specs/blob/b288f87c935c02102866e2c8df2f3d66a812f446/protos/sigstore_trustroot.proto#L77) for either the CA or the TSA. Other tools like [trtool](https://github.com/kommendorkapten/trtool) allow the user to manually provide that information.

cmd/cosign/cli/trustedroot/trustedroot.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+
+	if c.Out != "" {
+		err = os.WriteFile(c.Out, trBytes, 0600)
+		if err != nil {
+			return err
+		}
+	} else {
+		ui.Infof(ctx, string(trBytes))
```
cmurphy, line 147: It looks like this writes to stderr rather than stdout, so it's not intuitive to capture it with a `>` redirect. Not super critical since you also have the `-out` option, just less intuitive to me.

cmurphy (commented):
Love this! Generating a trusted root has been a pain for me for the last few weeks.

cmd/cosign/cli/options/trustedroot.go
steiza: Sort of - I was trying to mirror the flags of commands like `cosign verify-blob ...` as much as possible, and I didn't see a flag for specifying the public key of the CT log. If there is one, and I missed it, I'm happy to add it!

cmd/cosign/cli/options/trustedroot.go
steiza: Another great question, with a similar answer - as much as possible I'm trying to mirror the flags and behavior of the `cosign verify* ...` commands.

The idea being when we go to deprecate these flags from `cosign verify* ...` commands, you will generate the trusted root you need by calling `cosign trusted-root create ...` with the deprecated flags you were passing to `cosign verify* ...`.

cmd/cosign/cli/options/trustedroot.go
steiza: Yeah, I'm leaning towards keeping this tool more minimal and mirroring the flags of existing `cosign verify* ...` commands, but people are of course welcome to edit the output to add additional context (including setting URIs or more sane valid timestamps!), or use other tooling like `trtool`.

cmd/cosign/cli/options/trustedroot.go
cmurphy: I don't think the flags for `verify*` are an ideal analog for what needs to be included in the trusted root. The `--rekor-url` flag used for the verify command is used in an online API call to verify the transparency log inclusion, but the trusted keys for the transparency log are fetched separately using the TUF configuration. We couldn't deprecate `--rekor-url`, or if we do it would be in favor of offline verification.

cmd/cosign/cli/options/trustedroot.go
steiza: This is a little tricky because of the diversity of private deployments of Sigstore. I added a `--ignore-sct` so that if the private deployment is using long-lived keys instead of a private Fulcio we don't try to add CT Log keys to their trusted root; otherwise we call `cosign.GetCTLogPubs()` which will make use of the environment variable or TUF target.

cmd/cosign/cli/options/trustedroot.go
steiza: Fair enough! I replace `--rekor-url` with `--ignore-tlog`, similar to `--ignore-sct`.

cmd/cosign/cli/options/trustedroot.go
steiza: Fair enough! I replaced `--rekor-url` with `--ignore-tlog`, with a similar implementation to `--ignore-sct`.

cmd/cosign/cli/options/trustedroot.go
```diff
@@ -0,0 +1,71 @@
...
+	_ = cmd.Flags().SetAnnotation("certificate-chain", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.MarkFlagsMutuallyExclusive("ca-roots", "certificate-chain")
+	cmd.MarkFlagsMutuallyExclusive("ca-intermediates", "certificate-chain")
+
+	cmd.Flags().BoolVar(&o.IgnoreSCT, "ignore-sct", false,
+		"when set, do not include key for verifying certificate transparency "+
+			"log. Set this if you signed with a key instead of using Fulcio.")
+
+	cmd.Flags().BoolVar(&o.IgnoreTlog, "ignore-tlog", false,
```
cmurphy, line 61: I'm still hesitant about the design of mirroring the flags of the `verify*` commands. This is a brand new command, it might as well have flag names that are intuitive in relation to its purpose. `trusted_root.json` is used for both signing and verification, so restricting the flags to just what is relevant to verification feels like an oversight. Using an explicit flag like `--certificate-chain` along with a negative flag like `--ignore-tlog` seems inconsistent.

It would make more sense to me if this command picked one direction and followed it for all key material: either every key is explicitly specified by a flag that points to an on-disk key (e.g. `cosign trusted-root --certificate-chain=fulcio-chain.pem --rekor-key=rekor.pub --ctfe-key=ctfe.pub`)
 or all keys including Fulcio certs are retrieved via the TUF client (and ignored if TUF can't find them).

steiza: > It would make more sense to me if this command picked one direction and followed it for all key material: either every key is explicitly specified by a flag that points to an on-disk key (e.g. cosign trusted-root --certificate-chain=fulcio-chain.pem --rekor-key=rekor.pub --ctfe-key=ctfe.pub)

This sounds good to me! I waffled a bit on files on disk vs. TUFv1 support, but I think supporting files on disk will be more accessible (it's easier for a TUFv1 user to download files and supply them via CLI than it is for someone using files on disk to set up a TUFv1 destination!)

cmurphy (approved):
lgtm!

cmd/cosign/cli/trustedroot/trustedroot.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	if c.CtfeKeyPath != "" {
+		ctLogPubKey, id, idBytes, err := getPubKey(c.CtfeKeyPath)
+		if err != nil {
+			return err
+		}
+
+		ctLogs[id] = &root.TransparencyLog{
```
haydentherapper, line 90: Do we need to set validity periods as well? Do we need to take in flags for start time, or default to now?

cmd/cosign/cli/options/trustedroot.go
```diff
@@ -0,0 +1,70 @@
...
+	CtfeKeyPath      string
+	RekorKeyPath     string
+	Out              string
+	TSACertChainPath string
+}
+
+var _ Interface = (*TrustedRootCreateOptions)(nil)
+
+func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.CAIntermediates, "ca-intermediates", "",
```
haydentherapper, line 35: How do cert pools work here? This seems incompatible with the trusted root, as you have to specify a chain.

I see below that you're flattening, assuming 1 intermediate per root and assuming these two lists are ordered. I think that's not always going to be accurate - i would assume a client would throw all their certs into a list without thinking about chain building.

I would recommend we drop these until we support pools in the trust root, and require that clients construct the chains themselves.

cmd/cosign/cli/options/trustedroot.go
```diff
@@ -0,0 +1,70 @@
...
+			"The flag is optional and must be used together with --ca-roots, conflicts with "+
+			"--certificate-chain.")
+	_ = cmd.Flags().SetAnnotation("ca-intermediates", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.Flags().StringVar(&o.CARoots, "ca-roots", "",
+		"path to a bundle file of CA certificates in PEM format which will be needed "+
+			"when building the certificate chains for the signing certificate. Conflicts with --certificate-chain.")
+	_ = cmd.Flags().SetAnnotation("ca-roots", cobra.BashCompFilenameExt, []string{"cert"})
+
+	cmd.Flags().StringVar(&o.CertChain, "certificate-chain", "",
```
haydentherapper, line 47: Should all of these options be plural, so that we can provide multiple chains, multiple logs, and multiple TSAs?

cmd/cosign/cli/trustedroot/trustedroot.go
steiza: Since we aren't setting `ValidityPeriodStart`, it will default to "0001-01-01T00:00:00Z".

Philisophically, I was thinking of this utility more as "enough of a trusted root to get you started" instead of a flexible tool to cover all use-cases. This will be a recurring theme in my responses! And of course once you have the skeleton of a trusted root file, the user can continue editing it and adding additional content.

cmd/cosign/cli/options/trustedroot.go
steiza: Initially we were trying to mirror other `cosign` flags, but we ended up moving away from that, so how about we specify a single root and a single intermediate file (that can have one or more intermediates)?

Again, this is in the spirit of "enough to get you started" and not a do-it-all utility.

cmd/cosign/cli/options/trustedroot.go
steiza: At this point, you can probably anticipate my response : D but I think we should stick to the bare minimum required to get folks off the ground, and let them iterate on top of the basic trusted root we give them.

steiza (commented):
If the bare-bones approach I describe makes sense, let me know and I can update things to match!

cmd/cosign/cli/trustedroot/trustedroot.go
haydentherapper: Agreed it shouldn't cover all cases, but if we're going to instruct users to go manually inspect it afterwards and update a set of values, providing flags to do that for them seems reasonable. Since validity period is a significant benefit of the trusted root, I'd prefer adding a flag for users to set the start time.

cmd/cosign/cli/options/trustedroot.go
haydentherapper: I'd lean towards just removing these flags entirely and only have --certificate-chain as an option. if you specify a root and its intermediates separately, you still have to know their order - might as well just use the chain flag at that point.

cmd/cosign/cli/options/trustedroot.go
haydentherapper: I feel less strongly on this one, but I think pluralizing is a small enough change that if we did that now, we'd cover future use cases (e.g. trusting both an internal deployment and public deployment)

cmd/cosign/cli/trustedroot/trustedroot.go
steiza: I added the ability to specify the validity start time for keys

cmd/cosign/cli/options/trustedroot.go
steiza: Only supporting chains sounds good to me! I updated this pull request to reflect that.

cmd/cosign/cli/options/trustedroot.go
steiza: This did end up being a pretty minor change! You can now use the flags multiple times to specify additional verification materials.

cmurphy (approved):

dmitris:
As discussed on the Sigstore Slack (#private-sigstore-users channel) - found a case causing a panic on nil pointer, raised a PR with a trivial fix (nil pointer check on block in the for loop condition): https://github.com/steiza/cosign/pull/2/files#diff-abb0ac148c9642335c919aaffc4700982c9d799bc98d528c4d7d921f2470f2b5R165
```go
for block, contents := pem.Decode(contents); block != nil; block, contents = pem.Decode(contents) {
```
it would be good to add a unit test covering the case of the nil block panic, just in case.  I can do it and send you another "PR-to-PR" - let me know if you'd like me to.

steiza:
Thank you @dmitris! I added your fix, along with tests, and added you as a co-author of the commit.

cmd/cosign/cli/trustedroot/trustedroot.go
```diff
@@ -0,0 +1,206 @@
...
+	"encoding/hex"
+	"encoding/pem"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+
+	"github.com/sigstore/cosign/v2/pkg/cosign"
```
dmitris, line 31: I think the sorted order is wrong - shouldn't this come  before `sigstore/sigstore-go`?
If I delete the blank line L30, the editor sorts the imports correctly.
```go
	"github.com/sigstore/cosign/v2/pkg/cosign"
	"github.com/sigstore/sigstore-go/pkg/root"
```

dmitris (changes requested):

cmd/cosign/cli/options/trustedroot.go
```diff
@@ -36,7 +36,7 @@ func (o *TrustedRootCreateOptions) AddFlags(cmd *cobra.Command) {
 		"path to a list of CA certificates in PEM format which will be needed "+
 			"when building the certificate chain for the signing certificate. "+
 			"Must start with the parent intermediate CA certificate of the "+
-			"signing certificate and end with the root certificate. Conflicts with --ca-roots and --ca-intermediates.")
```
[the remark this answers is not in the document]
steiza: Yes, so `cosign verify` supports both `--certificate-chain` and `--ca-roots`, but later we decided that `cosign trusted-root create` would only support `--certificate-chain`. There isn't a `--ca-roots` option for `cosign trusted-root create` anymore.

dmitris (approved):
👍

cmd/cosign/cli/options/trustedroot.go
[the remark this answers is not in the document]
dmitris: would `--certificate-chain` support PEM files with multiple chains, or only a single one?  If you have a PEM bundle with a dozen "top-level" self-signed CA root certificates, would you be able to use `cosign trusted-root create` to create a trusted root JSON file or not?  Would one had to "manually" split that bundle into multiple PEM files, one per certificate (or certificate chain)?  Can you use `--certificate-chain` multiple times?

As mentioned on Slack [link](https://sigstore.slack.com/archives/C049ALX6Q83/p1729699697044089?thread_ts=1729181857.607539&cid=C049ALX6Q83), if you currently try to pass a PEM bundle with multiple CA root certificates as `--certificate-chain`, the generated JSON file is incorrect: it puts all of them in the same chain, with one certificate being the root, and the rest - Intermediates, even though they are all self-signed roots.  The same issue is with parsing a timestamp certificate bundle with multiple certificates.  

If these are use cases that we do not intend to support, let's explicitly state it in the documentation.

haydentherapper (approved):
Great work @steiza!

PR #3877, opened 2024-09-11
Fix bug in attest-blob when using a timestamp authority with new bundles

steiza:
#### Summary

Fix bug in #3752

When adding bundle support to `attest-blob`, we sent the wrong data to the timestamp authority to sign.

You can test this by first attesting to a blob with a timestamp authority and a predicate with something like:

```
$ go run cmd/cosign/main.go attest-blob --bundle="attest-pgi.sigstore.json" --new-bundle-format=true --key=cosign.key --tlog-upload=false --timestamp-server-url="https://timestamp.githubapp.com/api/v1/timestamp" --tlog-upload=false --type=something --predicate="../sigstore-go/examples/sigstore-go-signing/intoto.txt" ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

Then the verification which previously failed (and now succeeds) looks like:

```
$ go run cmd/cosign/main.go verify-blob-attestation --trusted-root=trusted_root_with_timestamps.public.json --use-signed-timestamps=true --insecure-ignore-tlog=true --key=cosign.pub --bundle="attest-pgi.sigstore.json" --new-bundle-format=true --type=something ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```

#### Release Note

* Fixed bug that created bundles with incorrect signed timestamps when `attest-blob` was used with new bundles

#### Documentation

N/A

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -165,7 +165,21 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 	var rekorEntry *models.LogEntryAnon
 
 	if c.TSAServerURL != "" {
-		timestampBytes, err = tsa.GetTimestampedSignature(sig, client.NewTSAClient(c.TSAServerURL))
+		// sig is the entire JSON DSSE Envelope; we just want the signature for TSA
+		var envelope dsse.Envelope
+		err = json.Unmarshal(sig, &envelope)
+		if err != nil {
+			return err
+		}
+		if len(envelope.Signatures) == 0 {
+			return fmt.Errorf("envelope has no signatures")
+		}
+		envelopeSigBytes, err := base64.StdEncoding.DecodeString(envelope.Signatures[0].Sig)
+		if err != nil {
+			return err
+		}
+
+		timestampBytes, err = tsa.GetTimestampedSignature(envelopeSigBytes, client.NewTSAClient(c.TSAServerURL))
 		if err != nil {
 			return err
 		}
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -165,23 +165,35 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 	var rekorEntry *models.LogEntryAnon
 
 	if c.TSAServerURL != "" {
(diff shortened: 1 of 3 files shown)


haydentherapper (commented):
Can you also fix https://github.com/sigstore/cosign/blob/780780b11e0998512c034317fd7e98776153e59d/cmd/cosign/cli/attest/attest.go#L179, which I believe has the same issue?

haydentherapper:
Could you test that this doesn't break the existing verification flows for `verify-attestation` and `verify-blob-attestation` without using sigstore-go? I think we might have overloaded `sig` in a few places such that the verifier is expecting the DSSE envelope to have been signed over rather than the DSSE's sig, but I haven't dug in closely yet.

haydentherapper:
Grumble grumble, this will affect verification for non-bundle flows, see https://github.com/sigstore/cosign/blob/780780b11e0998512c034317fd7e98776153e59d/pkg/cosign/verify.go#L1140-L1154. We verify using either the signature or the signed payload.

codysoyland (approved):

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -165,9 +165,35 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
-		timestampBytes, err = tsa.GetTimestampedSignature(sig, client.NewTSAClient(c.TSAServerURL))
-		if err != nil {
-			return err
+		// We need to decide what signature to send to the timestamp authority.
+		//
+		// Historically, cosign sent `sig`, which is the entire JSON DSSE
+		// Envelope. However, when sigstore clients are verifying a bundle they
+		// will use the DSSE Sig field, so we choose what signature to send to
+		// the timestamp authority based on our output format.
+		if c.NewBundleFormat {
```
haydentherapper, line 174: Can we add a test under e2e_test as well? A copy of https://github.com/sigstore/cosign/blob/081dea1918e9536c1fe233aa2596301381967b3b/test/e2e_test.go#L771 that uses AttestBlobCommand?

haydentherapper (dismissed):
LGTM, thanks! Just a comment about a test.

cmd/cosign/cli/attest/attest_blob.go
steiza: Sure! This would have been a lot harder if I had not written https://github.com/sigstore/cosign/pull/3876 recently 😅

haydentherapper (approved):
Thanks!

PR #3889, opened 2024-09-25
Add support for new bundle specification for attesting/verifying OCI image attestations

codysoyland:
#### Summary

This PR adds support for the new [Cosign Bundle Specification](https://github.com/sigstore/cosign/blob/main/specs/BUNDLE_SPEC.md) in `cosign attest` and `cosign verify-attestation`.

Related: https://github.com/sigstore/cosign/issues/3139

To test, run the following (replacing `MY_IDENTITY`, `MY_ISSUER`, `MY_TRUSTED_ROOT` and `MY_IMAGE` as needed -- trusted root is optional). Note that the new OCI support requires passing `--new-bundle-format` into both commands.

```shell
go run ./cmd/cosign attest --predicate my-predicate.json --new-bundle-format MY_IMAGE
go run ./cmd/cosign verify-attestation --certificate-identity MY_IDENTITY --certificate-oidc-issuer MY_ISSUER --new-bundle-format --trusted-root=MY_TRUSTED_ROOT MY_IMAGE
```

Full example (using `crane`, but can instead use `docker tag`/`docker push`):

```shell
make cosign
docker run -d -p 5050:5000 ghcr.io/project-zot/zot-linux-arm64:latest
crane copy busybox:latest localhost:5050/busybox
echo '{"foo": "bar"}' > predicate.json
./cosign attest --predicate predicate.json --new-bundle-format localhost:5050/busybox:latest
# Dex workflow, assuming GitHub as provider
./cosign verify-attestation --certificate-identity=yourname@example.com --certificate-oidc-issuer=https://github.com/login/oauth --new-bundle-format localhost:5050/busybox:latest
```

To show that it uses the OCI 1.1 referrers API, you can use `oras`:

```shell
oras discover localhost:5050/busybox:latest
Discovered 1 artifact referencing latest
Digest: sha256:db142d433cdde11f10ae479dbf92f3b13d693fd1c91053da9979728cceb1dc68

Artifact Type                                   Digest
application/vnd.dev.sigstore.bundle.v0.3+json   sha256:3fb46d845fe437667a6b3ed45d2b11dea11c43f8fbe76dd642eb71de2a9b8b77
```


cmd/cosign/cli/attest/attest.go
```diff
@@ -220,14 +229,26 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
 	if err != nil {
 		return err
 	}
 
+	if c.KeyOpts.NewBundleFormat {
+		signerBytes, err := sv.Bytes(ctx)
+		if err != nil {
+			return err
+		}
+		bundleBytes, err := makeNewBundle(sv, rekorEntry, payload, signedPayload, signerBytes, timestampBytes)
```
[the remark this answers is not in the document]
codysoyland: This func exists from @steiza's previous work on adding bundle support, and I understand `NewBundle` in the name to refer to the "new format bundle" so I'm kind of okay with the current naming, but I get where you're coming from.

cmd/cosign/cli/attest/common.go
```diff
@@ -33,3 +37,15 @@ func predicateReader(predicatePath string) (io.ReadCloser, error) {
...
+func getEnvelopeSigBytes(envelopeBytes []byte) ([]byte, error) {
+	var envelope dsse.Envelope
+	err := json.Unmarshal(envelopeBytes, &envelope)
+	if err != nil {
+		return nil, err
+	}
+	if len(envelope.Signatures) == 0 {
+		return nil, fmt.Errorf("envelope has no signatures")
+	}
+	return base64.StdEncoding.DecodeString(envelope.Signatures[0].Sig)
```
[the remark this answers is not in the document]
codysoyland: Since cosign produced this signature, we know that it should have exactly one. There are several other places we assume there is only one signature in a DSSE, [including in sigstore-go](https://github.com/sigstore/sigstore-go/blob/main/pkg/bundle/signature_content.go#L100). I'm okay without checking in this context, but we might want to add some validation in the sigstore-go verifier (in the linked file) that will warn/err if there are multiple present when not expected.

cmd/cosign/cli/options/verify.go
```diff
@@ -222,7 +213,6 @@ type VerifyBlobAttestationOptions struct {
 	Key             string
 	SignaturePath   string
 	BundlePath      string
-	NewBundleFormat bool
```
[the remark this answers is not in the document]
codysoyland: Yes, good call, both of these (`NewBundleFormat` and `TrustedRootPath`) have moved up into `CertVerifyOptions` so aren't needed in the command-specific options.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -223,6 +237,19 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
 			co.SCT = sct
 		}
+	case c.TrustedRootPath != "":
+		if !c.NewBundleFormat {
+			return fmt.Errorf("unsupported: trusted root path currently only supported with --new-bundle-format")
+		}
+
+		// If a trusted root path is provided, we will use it to verify the bundle.
+		// Otherwise, the verifier will default to the public good instance.
+		if c.TrustedRootPath == "" {
```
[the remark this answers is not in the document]
codysoyland: Good call, I got that backwards!

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -327,3 +354,19 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 
 	return nil
 }
+
+func checkSigstoreBundleUnsupportedOptions(c *VerifyAttestationCommand) error {
+	if c.Cert != "" || c.CertRef != "" {
+		return fmt.Errorf("unsupported: certificate may not be provided using --cert when using --new-bundle-format (cert must be in bundle)")
```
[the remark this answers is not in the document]
codysoyland: I looked into this and both `Cert` and `CertRef` are mapped to `--certificate`, but only `CertRef` is used in the code. I removed the check for `Cert` and corrected the naming as well.

pkg/cosign/verify.go
```diff
@@ -156,13 +163,113 @@ type CheckOpts struct {
...
+	}
+
+	if !co.IgnoreTlog {
+		verifierOptions = append(verifierOptions, verify.WithTransparencyLog(1), verify.WithIntegratedTimestamps(1))
+	}
+	if co.UseSignedTimestamps {
+		verifierOptions = append(verifierOptions, verify.WithSignedTimestamps(1))
+	}
+	if co.IgnoreTlog && !co.UseSignedTimestamps {
+		verifierOptions = append(verifierOptions, verify.WithoutAnyObserverTimestampsUnsafe())
```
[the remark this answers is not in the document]
codysoyland: I like the idea, but it will only enter this conditional if the user has specified the flag `--insecure-ignore-tlog`, and I don't like to log from within library code (this is called from Policy Controller for example). We could add something to the CLI section of code that will warn if that option is provided, but I think that's kind of out of scope for this PR.

pkg/cosign/verify.go
```diff
@@ -1413,3 +1527,128 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+	var bundles = make([]*sgbundle.Bundle, 0, len(index.Manifests))
+	for _, result := range index.Manifests {
+		st, err := name.ParseReference(fmt.Sprintf("%s@%s", digest.Repository, result.Digest.String()))
+		if err != nil {
+			return nil, nil, err
+		}
+		bundle, err := ociremote.Bundle(st, co.RegistryClientOpts...)
+		if err != nil {
+			// There may be non-Sigstore referrers in the index, so we can ignore them.
+			// TODO: Should we surface any errors here (e.g. if the bundle is invalid)?
```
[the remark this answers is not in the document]
codysoyland: The above call `ociremote.Bundle` will error if the bundle does not parse correctly, and the containing func is only called by `verifyImageAttestationsSigstoreBundle`, which will actually do the bundle verification.

pkg/cosign/verify.go
```diff
@@ -1413,3 +1527,128 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+
+			attestations[index] = att
+			t.Done(nil)
+		}(bundle, i)
+
+		// wait till workers are available
+		t.Throttle()
+	}
+
+	for _, a := range attestations {
```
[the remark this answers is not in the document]
codysoyland: This logic was copied from `VerifyImageAttestation` (in this file), and I think it was done this way to avoid using a mutex to guard `checkedAttestations` since the above loop is done in a goroutine. It's probably simplest if kept in its current form.

pkg/cosign/verify.go
```diff
@@ -1413,3 +1527,128 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+	}
+
+	for _, verified := range bundlesVerified {
+		bundleVerified = bundleVerified || verified
+	}
+
+	if len(checkedAttestations) == 0 {
+		var combinedErrors []string
+		for _, err := range t.Errs() {
+			combinedErrors = append(combinedErrors, err.Error())
```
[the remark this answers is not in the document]
codysoyland: Similarly to your question above about `checkedAttestations`, I think this is done to avoid the shared memory problem with goroutines. I did notice however that the way the errors are assembled is kind of crude, so I'm cleaning it up. While doing so, I also noticed we're using the old `pkg/errors` error package, so I [submitted a PR](https://github.com/sigstore/cosign/pull/3956) to remove that.

test/e2e_test.go
```diff
@@ -361,7 +377,7 @@ func attestVerify(t *testing.T, predicateType, attestation, goodCue, badCue stri
 	}
 
 	// Now attest the image
-	ko := options.KeyOpts{KeyRef: privKeyPath, PassFunc: passFunc}
+	ko := options.KeyOpts{KeyRef: privKeyPath, PassFunc: passFunc, NewBundleFormat: newBundleFormat}
```
cmurphy, line 380: Maybe out of scope for this PR but a note for the future, this is setting `NewBundleFormat` without setting `TrustedRootPath`, which means the `TrustedMaterial` will default to using PGI. We're trying to avoid making external network calls to live services. The only reason that isn't a problem here is because this test is using a local key pair and is not uploading to Rekor, so it doesn't matter what verification material is used. But a useful test in the future might be to have this use ephemeral keys from Fulcio (localhost:5555) and upload the entry to Rekor (localhost:3000) so that the full verification path with a locally generated trust root could be tested.

pkg/cosign/verify.go
```diff
@@ -156,13 +163,113 @@ type CheckOpts struct {
 	// IgnoreTlog skip tlog verification
 	IgnoreTlog bool
 
+	// UseSignedTimestamps use signed timestamps if available
+	UseSignedTimestamps bool
```
cmurphy, line 167: Is this potentially something that could be split out into its own PR, so that each PR has a clear singular focus?

pkg/cosign/verify.go
```diff
@@ -156,13 +163,113 @@ type CheckOpts struct {
...
 	// Defaults to false.
 	ExperimentalOCI11 bool
+
+	NewBundleFormat bool
+
+	// TrustedMaterial is the trusted material to use for verification.
+	// Currently, this is only applicable when NewBundleFormat is true.
+	TrustedMaterial root.TrustedMaterial
+
+	// TODO: Add these to replace above fields?
```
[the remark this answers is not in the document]
codysoyland: I removed them for now.

pkg/cosign/verify.go
```diff
@@ -156,13 +163,113 @@ type CheckOpts struct {
...
+	root.TrustedMaterial
+	keyTrustedMaterial root.TrustedMaterial
+}
+
+func (v *verifyTrustedMaterial) PublicKeyVerifier(hint string) (root.TimeConstrainedVerifier, error) {
+	return v.keyTrustedMaterial.PublicKeyVerifier(hint)
+}
+
+// SigstoreGoOptions returns the verification options for verifying with sigstore-go.
+func (co *CheckOpts) SigstoreGoOptions() (trustedMaterial root.TrustedMaterial, verifierOptions []verify.VerifierOption, policyOptions []verify.PolicyOption, err error) {
```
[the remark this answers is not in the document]
codysoyland: I renamed it to `verificationOptions` and privatized it.

pkg/oci/remote/referrers.go
```diff
@@ -25,7 +25,9 @@ import (
 func Referrers(d name.Digest, artifactType string, opts ...Option) (*v1.IndexManifest, error) {
 	o := makeOptions(name.Repository{}, opts...)
 	rOpt := o.ROpt
-	rOpt = append(rOpt, remote.WithFilter("artifactType", artifactType))
+	if artifactType != "" {
```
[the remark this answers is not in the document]
codysoyland: The `artifactType` for bundles is something like `application/vnd.dev.sigstore.bundle.v0.3+json`. Since the version number is baked into the type, we can't use the filter feature here. Therefore, `getBundles` uses an empty string as the artifactType in the call to Referrers.

pkg/cosign/verify.go
codysoyland: That's a good idea for something we can split out... I think this is needed because the existing timestamp verification in cosign doesn't work with TUF/PGI, and is only enforced if you provide a timestamp cert chain to the verification subcommands.

pkg/cosign/verify_bundle.go
```diff
@@ -0,0 +1,35 @@
...
+package cosign
+
+import (
+	"context"
+
+	"github.com/sigstore/sigstore-go/pkg/verify"
+)
+
+// VerifyNewBundle verifies a SigstoreBundle with the given parameters
+func VerifyNewBundle(_ context.Context, co *CheckOpts, artifactPolicyOption verify.ArtifactPolicyOption, bundle verify.SignedEntity) (*verify.VerificationResult, error) {
```
codysoyland, line 25: Consider splitting this into separate PR and refactoring existing `verifyNewBundle`.

codysoyland: https://github.com/sigstore/cosign/pull/4013

pkg/cosign/verify.go
codysoyland: https://github.com/sigstore/cosign/pull/4006

codysoyland:
I rebased and squashed all commits into one as there have been a couple of items refactored into separate PRs (https://github.com/sigstore/cosign/pull/4006 and https://github.com/sigstore/cosign/pull/4013) and this ongoing PR was getting a bit messy.

I still need to finish #4013 and rebase this one again, at which point I will be ready for another review pass.

Thanks everyone who has been patiently reviewing and helping me iterate on this!

cmd/cosign/cli/options/verify.go
```diff
@@ -219,11 +218,9 @@ func (o *VerifyDockerfileOptions) AddFlags(cmd *cobra.Command) {
 
 // VerifyBlobAttestationOptions is the top level wrapper for the `verify-blob-attestation` command.
 type VerifyBlobAttestationOptions struct {
-	Key             string
-	SignaturePath   string
-	BundlePath      string
-	NewBundleFormat bool
-	TrustedRootPath string
```
codysoyland, line 226: Note these have moved to `CommonVerifyOptions`

pkg/cosign/verify.go
```diff
@@ -1413,3 +1527,123 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+	}
+
+	for _, a := range attestations {
+		if a != nil {
+			checkedAttestations = append(checkedAttestations, a)
+		}
+	}
+
+	for _, verified := range bundlesVerified {
+		atLeastOneBundleVerified = atLeastOneBundleVerified || verified
```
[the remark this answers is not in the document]
codysoyland: This logic was copied from https://github.com/sigstore/cosign/blob/737c83c71cd42ba8db843f84beadcd59b3aa6e0a/pkg/cosign/verify.go#L1054, but I renamed `bundleVerified` to `atLeastOneBundleVerified` so it's more clear. The effect is that if there are multiple bundles associated with an image, verification will succeed if at least one of them is verified.

I can write a comment about this if it helps!

codysoyland:
This has been rebased onto the merged refactors and is ready for review.

pkg/cosign/verify.go
```diff
@@ -585,6 +587,10 @@ func VerifyImageSignatures(ctx context.Context, signedImgRef name.Reference, co
 		}
 	}
 
+	if co.NewBundleFormat {
+		return nil, false, errors.New("bundle support for image signatures is not yet implemented")
```
steiza, line 591: Interesting! I didn't know this feature existed. For policy controller users they will definitely be using image attestations, but do we have an idea about how common are / what use cases people are using image signatures?

steiza (commented):
There's a lot here! Sorry this took me so long to review; just a few minor questions on the changes.

Do I understand the Codecov report correctly that 1% of these changes are covered via tests? Any thoughts on how to increase that?

pkg/cosign/verify.go
codysoyland: Good question! IIRC, a Rekor analysis showed pretty strong usage of signatures, but I expect attestations will outpace them. I'd like to fast follow with support for image signatures. Assuming we go with option three in https://github.com/sigstore/cosign/issues/3927, it will share much of the same code as image attestations.

codysoyland:
> Do I understand the Codecov report correctly that 1% of these changes are covered via tests? Any thoughts on how to increase that?

True, there are no remaining unit tests in this PR after the last rebase, but Codecov may not take into account the new e2e tests which should cover quite a bit. I'll try manually running coverage and see what I can reasonably cover with more tests.

dmitris:
@codysoyland - thanks for pushing this forward, eagerly awaiting this feature!  If I can help with testing or anything else, please let me know! 

Could you resolve the conflicts, please? Thanks.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -214,6 +228,17 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 			}
 			co.SCT = sct
 		}
+	case c.TrustedRootPath != "":
+		if !c.NewBundleFormat {
+			return fmt.Errorf("unsupported: trusted root path currently only supported with --new-bundle-format")
+		}
+
+		// If a trusted root path is provided, we will use it to verify the bundle.
+		// Otherwise, the verifier will default to the public good instance.
```
haydentherapper, line 237: Do we need to explicitly called `FetchTrustedRoot` here and in `verify.go`? This is based on looking at https://github.com/sigstore/cosign/blob/b4be5f7a4305d195ee5a587e98903dfced269333/cmd/cosign/cli/verify/verify_blob.go#L403, where the PGI TUF metadata is loaded in for verify_blob/verify_blob_attestation before calling the internal verification function.

pkg/cosign/verify.go
```diff
@@ -1506,3 +1515,123 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+	for _, result := range index.Manifests {
+		st, err := name.ParseReference(fmt.Sprintf("%s@%s", digest.Repository, result.Digest.String()))
+		if err != nil {
+			return nil, nil, err
+		}
+		bundle, err := ociremote.Bundle(st, co.RegistryClientOpts...)
+		if err != nil {
+			// There may be non-Sigstore referrers in the index, so we can ignore them.
+			// TODO: Should we surface any errors here (e.g. if the bundle is invalid)?
+			continue
```
haydentherapper, line 1550: Could we return an error at the end of this function if no bundles are found?

pkg/cosign/verify.go
```diff
@@ -1506,3 +1515,123 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+				dsse, ok := bundle.Content.(*protobundle.Bundle_DsseEnvelope)
+				if !ok {
+					return fmt.Errorf("bundle does not contain a DSSE envelope")
+				}
+				payload, err := json.Marshal(dsse.DsseEnvelope)
+				if err != nil {
+					return fmt.Errorf("marshaling DSSE envelope: %w", err)
+				}
+
+				// TODO: Add additional data to oci.Signature (Cert, Rekor Bundle, Timestamp, etc)
```
haydentherapper, line 1597: nit: can you clarify what the additional data is, just for future readers to know?

pkg/cosign/verify.go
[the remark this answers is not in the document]
haydentherapper: Adding onto this - Do we _want_ to support multiple bundles on a single container? We shouldn't be bound to Cosign's previous decisions and it might be worth revisiting this now.

There are valid use-cases for multiple bundles (rotating a signing key/identity for example), but I just want to take a second to make sure we really do want to support this.

pkg/oci/remote/signatures.go
```diff
@@ -49,6 +52,42 @@ func Signatures(ref name.Reference, opts ...Option) (oci.Signatures, error) {
 	}, nil
 }
 
+func Bundle(ref name.Reference, opts ...Option) (*sgbundle.Bundle, error) {
+	o := makeOptions(ref.Context(), opts...)
+	img, err := remoteImage(ref, o.ROpt...)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: We can check for a specific predicate type here using
```
haydentherapper, line 61: To clarify, do you mean checking for `application/vnd.dev.sigstore.bundle`? If so, I think hardcoding this is fine and we don't need this TODO.

pkg/oci/remote/signatures.go
```diff
@@ -49,6 +52,42 @@ func Signatures(ref name.Reference, opts ...Option) (oci.Signatures, error) {
...
+		return nil, err
+	}
+	if len(layers) != 1 {
+		return nil, errors.New("expected exactly one layer")
+	}
+	mediaType, err := layers[0].MediaType()
+	if err != nil {
+		return nil, err
+	}
+	if !strings.HasPrefix(string(mediaType), "application/vnd.dev.sigstore.bundle") {
```
haydentherapper, line 75: Should we be checking for specific bundle versions so that we don't attempt to unmarshal to an unsupported structure?

haydentherapper (commented):
This is really amazing work @codysoyland! 

Going to read up on OCI 1.1 and take another pass later, though we should ask the authors of the original OCI 1.1 code as well to review those bits.

pkg/cosign/verify.go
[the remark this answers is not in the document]
codysoyland: IMO, there are not many downsides to supporting multiple bundles and it's pretty much "free" with OCI, but it's good to challenge that assumption. I guess you could say a downside is that we need this logic here to decide if invalid attestations should cause a verification failure, but I would argue that this logic is natural if the [principle of monotonicity](https://github.com/in-toto/attestation/blob/main/spec/v1/README.md) is followed.

I would also highlight that each bundle may only contain one attestation, so if a user wishes to store both an SBOM and vulnerability scan attestation referencing the same image, multiple bundles must be allowed.

codysoyland:
> @codysoyland - thanks for pushing this forward, eagerly awaiting this feature! If I can help with testing or anything else, please let me know!

Thank you! I'd be interested in any manual test results you may have!

> Could you resolve the conflicts, please? Thanks.

I just rebased and resolved the conflict. 👍🏻

cmd/cosign/cli/verify/verify_attestation.go
codysoyland: Yes, good call! I just pushed a fix.

pkg/oci/remote/signatures.go
codysoyland: I'll remove the TODO as it's just an optimization suggestion -- the manifest pointing to the bundle is annotated with the in-toto predicate type, and we can filter it here so it doesn't have to download, parse, and verify every bundle.

pkg/oci/remote/signatures.go
codysoyland: The call to `b.UnmarshalJSON` below takes care of the version-specific validation. I also added a call to b.MinVersion to enforce bundle v0.3 or newer.

codysoyland:
I generated coverage with e2e testing. [Here is a link to the report.](https://gist.github.com/codysoyland/de83ab37e3c5422442e87e1443de8ff5) TL;DR: Most of the "happy paths" are covered, but many error cases are not covered, which is pretty typical for the e2e tests. I'll take another pass to see if there's any low hanging fruit for additional testing, but I don't think the coverage of this PR is much different from our existing coverage, however I would still like to improve it.

dmitris:
@codysoyland "stepped on" some of the files you're modifying with https://github.com/sigstore/cosign/pull/3889 - could you rebase and resolve the conflicts, please?

cmd/cosign/cli/options/verify.go
```diff
@@ -56,6 +58,13 @@ func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().IntVar(&o.MaxWorkers, "max-workers", cosign.DefaultMaxWorkers,
 		"the amount of maximum workers for parallel executions")
+
+	cmd.Flags().StringVar(&o.TrustedRootPath, "trusted-root", "",
+		"Path to a Sigstore TrustedRoot JSON file.")
```
dmitris, line 63: should it also say: ` Requires --new-bundle-format to be set.`?

codysoyland: Yes, good point! I pushed that change.

codysoyland:
> Going to read up on OCI 1.1 and take another pass later, though we should ask the authors of the original OCI 1.1 code as well to review those bits.

This is a great idea! The parts that I think need attention from a reviewer with more OCI experience would be the logic in:

- `pkg/oci/remote` (`WriteAttestationNewBundleFormat` and `Bundle`)
- `pkg/cosign` (`getBundles`)

Perhaps @hectorj2f or @vaikas might be able to take a look?

pkg/oci/remote/write.go
```diff
@@ -221,3 +223,126 @@ func (taggable taggableManifest) RawManifest() ([]byte, error) {
...
+		Layers: []v1.Descriptor{
+			{
+				MediaType: types.MediaType(bundleMediaType),
+				Digest:    blobDigest,
+				Size:      blobSize,
+			},
+		},
+		Subject: &v1.Descriptor{
+			MediaType: types.OCIManifestSchema1,
+			Digest:    desc.Digest,
```
lcarva, line 297: Is this right? `desc` is derived from `ref` -> `signTarget` -> `d` which is of type `name.Repository`.

The digest used here should be the exact digest of the image being described. `WriteAttestationNewBundleFormat` should probably take `name.Digest` instead of `name.Repository` to guarantee this.

pkg/oci/remote/write.go
```diff
@@ -221,3 +223,126 @@ func (taggable taggableManifest) RawManifest() ([]byte, error) {
...
+		},
+		Layers: []v1.Descriptor{
+			{
+				MediaType: types.MediaType(bundleMediaType),
+				Digest:    blobDigest,
+				Size:      blobSize,
+			},
+		},
+		Subject: &v1.Descriptor{
+			MediaType: types.OCIManifestSchema1,
```
lcarva, line 296: Could use `desc.MediaType` here.

pkg/oci/remote/write.go
codysoyland: Thanks for reviewing, and good catch here, I believe you're right! I'll fix.

codysoyland:
I found `github.com/google/go-containerregistry/pkg/registry` which was very easy to use to add some unit tests to the OCI interactions on top of what is already covered by e2e tests. I just pushed a test that verifies the response when there are no attestations. I'll try to add a few more test cases to improve coverage.

bkabrda:
Hi folks :wave:, I've been able to play around with this PR today a little bit and AFAICT it works really good, but I have some notes - see below.

What I tested:
* Attesting with both a specified private key and with the OIDC flow (I tried mixing both types of attestations for a single image)
* Verifying attestations, both with a specified public key and with cert identity+OIDC issuer details.
* I also went through the format of the created OCI artifacts and checked it against the specification.

Some notes:
* Verifying prints out a lot of various structs (I guess that's known and expected because the PR is still being worked on)
* Creating attestation using the OIDC flow prints the certificate twice (might also be known temporary debug output)
* The `dsseEnvelope.signatures` objects don't seem to have the `keyid` attribute. I don't think this is a problem, but still wanted to mention it in case somebody thinks otherwise.
* Verifying attestation with the cert identity+OIDC issuer segfaults when there is also an existing attestation created using a private key. I can look into this further tomorrow to see if I can find what the issue is, for now here's the stack trace:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xe1287b]

goroutine 112 [running]:
github.com/sigstore/cosign/v2/pkg/cosign.(*verifyTrustedMaterial).PublicKeyVerifier(0xc00006f728?, {0xc0009800c0?, 0xc000a2cd20?})
	/home/bkabrda/programming/sigstore/cosign/pkg/cosign/verify.go:185 +0x1b
github.com/sigstore/sigstore-go/pkg/bundle.(*PublicKey).CompareKey(0x2dc0f60?, {0x2dc0f60, 0xc0014af0e0}, {0x3864c20?, 0xc0014aef20?})
	/home/bkabrda/go/pkg/mod/github.com/sigstore/sigstore-go@v0.7.0/pkg/bundle/verification_content.go:64 +0x3c
github.com/sigstore/sigstore-go/pkg/verify.VerifyArtifactTransparencyLog({0x38687d0, 0xc00078f2f0}, {0x3864c20, 0xc0014aef20}, 0x1, 0x1)
	/home/bkabrda/go/pkg/mod/github.com/sigstore/sigstore-go@v0.7.0/pkg/verify/tlog.go:119 +0x667
github.com/sigstore/sigstore-go/pkg/verify.(*SignedEntityVerifier).VerifyTransparencyLogInclusion(0xc0005edbc0?, {0x38687d0?, 0xc00078f2f0?})
	/home/bkabrda/go/pkg/mod/github.com/sigstore/sigstore-go@v0.7.0/pkg/verify/signed_entity.go:683 +0x5d
github.com/sigstore/sigstore-go/pkg/verify.(*SignedEntityVerifier).Verify(0xc00037d5e0, {0x38687d0, 0xc00078f2f0}, {0xc0005edbc0?, {0xc0005a2258?, 0xc000505500?, 0xc00051d718?}})
	/home/bkabrda/go/pkg/mod/github.com/sigstore/sigstore-go@v0.7.0/pkg/verify/signed_entity.go:520 +0xd1
github.com/sigstore/cosign/v2/pkg/cosign.VerifyNewBundle({0x3?, 0x10000c000505500?}, 0x0?, 0xc0005edbc0, {0x38687d0, 0xc00078f2f0})
	/home/bkabrda/programming/sigstore/cosign/pkg/cosign/verify_bundle.go:34 +0x85
github.com/sigstore/cosign/v2/pkg/cosign.verifyImageAttestationsSigstoreBundle.func1.1({0x385f070?, 0xc00019eee0?}, 0xc00051d7a8?, 0x40b58b?, 0xc00006ffa0, {0xc000886100, 0x4, 0xc0001121e0?}, 0x3, 0xc00078f2f0)
	/home/bkabrda/programming/sigstore/cosign/pkg/cosign/verify.go:1589 +0x4d
github.com/sigstore/cosign/v2/pkg/cosign.verifyImageAttestationsSigstoreBundle.func1(0xc0004de000?, 0x3)
	/home/bkabrda/programming/sigstore/cosign/pkg/cosign/verify.go:1616 +0x6e
created by github.com/sigstore/cosign/v2/pkg/cosign.verifyImageAttestationsSigstoreBundle in goroutine 1
	/home/bkabrda/programming/sigstore/cosign/pkg/cosign/verify.go:1586 +0x26a
```

In terms of the code and how we'd share common functionality with https://github.com/sigstore/cosign/issues/3927, I think this looks really good; we might need to do a tiny bit of internal refactoring when working on it, but that shouldn't concern this PR.

bkabrda:
I think the traceback that I mentioned above is caused by the code added [here](https://github.com/sigstore/cosign/pull/4013/files#diff-8a85c8e688d61e16b8af8e09832ed2bef89c1163b0e9601a8363c782c387c006R231) - this condition only adds `keyTrustedMaterial` when we're trying to verify with an explicitly provided key. If we have a keyful attestation and are trying to verify in keyless mode, `keyTrustedMaterial` will always be nil; eventually the code gets [here](https://github.com/sigstore/cosign/pull/4013/files#diff-8a85c8e688d61e16b8af8e09832ed2bef89c1163b0e9601a8363c782c387c006R188) (through [this](https://github.com/sigstore/sigstore-go/blob/main/pkg/bundle/verification_content.go#L63) sigstore-go code) and fails because `keyTrustedMaterial` is not set.

One (hacky) solution could be providing an "empty verifier" for this case which would always fail verification (because that's what we want AFAICS), but we could also explicitly prune keyful attestations when doing keyless verification.

codysoyland:
@bkabrda I really appreciate this review, this is very thorough!

I think most of these items should be pretty low hanging fruit.

codysoyland:
> Verifying prints out a lot of various structs (I guess that's known and expected because the PR is still being worked on)

The output is a bit verbose but unless I'm mistaken, this PR doesn't change the existing output. It prints a success message followed by the DSSE. Here's an example:
```
Verification for registry.local:5010/busybox --
  - The code-signing certificate was verified using trusted certificate authority certificates
{"payload":"eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjAuMSIsInByZWRpY2F0ZVR5cGUiOiJodHRwczovL2Nvc2lnbi5zaWdzdG9yZS5kZXYvYXR0ZXN0YXRpb24vdjEiLCJzdWJqZWN0IjpbeyJuYW1lIjoicmVnaXN0cnkubG9jYWw6NTAxMC9idXN5Ym94IiwiZGlnZXN0Ijp7InNoYTI1NiI6IjQ5OGEwMDBmMzcwZDhjMzc5MjcxMThlZDgwYWZlOGFkYzM4ZDFlZGNiZmMwNzE2MjdkMTdiMjVjODhlZmNhYjAifX1dLCJwcmVkaWNhdGUiOnsiRGF0YSI6IntcImZvb1wiOiBcImJhclwifVxuIiwiVGltZXN0YW1wIjoiMjAyNS0wMy0xMlQxODoxNjo1OVoifX0=","payloadType":"application/vnd.in-toto+json","signatures":[{"sig":"MEYCIQCJ71NJXjPibT4gYm8KmGpSUiuTvo5huzQmC56gNCYXwAIhAMux5/ZHkGTkmvF7/4vhp+UwR55ON6BxSAXxPbJlAH0q"}]}
```

> Creating attestation using the OIDC flow prints the certificate twice (might also be known temporary debug output)

Thanks for bringing that up... this is logged [here](https://github.com/codysoyland/cosign/blob/5ee87498f1ded89bdf4f9c58958effa09a25f563/cmd/cosign/cli/sign/sign.go#L600-L612), because the parent method (`SignerVerifier.Bytes`) is called twice. I would opt for removing this logging completely, but I'm not sure if it makes sense to do that in this PR.

Generally I would like to refine cosign's output in a future PR, as it could use some love.

> The dsseEnvelope.signatures objects don't seem to have the keyid attribute. I don't think this is a problem, but still wanted to mention it in case somebody thinks otherwise.

Shouldn't be a problem as they `keyid` is an optional field according to [the DSSE spec](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md#parsing-rules). 👍🏻 

> Verifying attestation with the cert identity+OIDC issuer segfaults when there is also an existing attestation created using a private key.

Good catch... I fixed it in my latest commit by checking if `keyTrustedMaterial` is `nil` and returning an error if so. I was able to test a scenario like you described (multiple attestations signed with different key types) and it behaved as expected. 🎉

pkg/cosign/verify.go
```diff
@@ -1500,3 +1517,133 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+				if err != nil {
+					return fmt.Errorf("marshaling DSSE envelope: %w", err)
+				}
+
+				// We will return a slice of `[]oci.Signature` from this function for compatibility
+				// with the rest of the codebase. To do that, we wrap the verification output in a
+				// `oci.Signature` using static.NewAttestation(). This type may contain additional
+				// data such as the certificate chain, and rekor/tsa data, but for now we only use
+				// the payload (DSSE). TODO: Add additional data to returned `oci.Signature`. This
+				// can be done by passing a list of static.Option to NewAttestation (e.g. static.WithCertChain()).
```
haydentherapper, line 1610: To check, is this not an issue that the OCI struct is missing additional verification data because we already have a bundle?

haydentherapper (approved):
Fantastic work on this PR Cody!

I haven’t throughly reviewed the OCI bits since I’m not very familiar with container registries, but it’s been well tested and looks good.

pkg/cosign/verify.go
codysoyland: I don't *think* it's an issue right now, but I would like to go back and add that missing data in a later PR.

The result of this is that the return value of `VerifyImageAttestations` contains a `[]oci.Signature` that is a correct list but missing some embedded data, for example if you iterated through them and tried to fetch the cert chain for each one, it would be missing those chains, e.g.:

```go
for _, sig in checkedAttestations {
	fmt.Println(sig.Chain())
}
```

For the purposes of the cosign CLI, this doesn't seem to make a difference, but it may affect library callers. To complete this, we really ought to update sigstore-go to make it easier to extract the missing data ([the values needed to implement `oci.Signature`](https://github.com/sigstore/cosign/blob/647eca1376ea75b3af88cd65f8f35f9e37ecd671/pkg/oci/signatures.go#L34-L72)), or we can add all the boilerplate here to extract those values from the bundle.

PR #3904, opened 2024-10-11
Fix cosign copy --only for signatures

marcofranssen:
#### Summary

There is a bug in the `cosign copy` command for the deprecated flag, as well documentation is invalid for the CLI option.

See
https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/copy/copy.go#L192
requires to have value `sig` instead of `sign`.

Also aligned the option docs order to align with the order of the
example. https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/copy.go#L40

#### Release Note

Fixes the `cosign copy` command to have correct documentation and proper fallback implementation for the deprecated --sig-only commandline option.

#### Documentation

N.A.

### DIFF
--- a/cmd/cosign/cli/copy/copy.go
+++ b/cmd/cosign/cli/copy/copy.go
@@ -80,7 +80,10 @@ func CopyCmd(ctx context.Context, regOpts options.RegistryOptions, srcImg, dstIm
 	}
 
 	onlyFlagSet := false
-	tags := parseOnlyOpt(copyOnly, sigOnly)
+	tags, err := parseOnlyOpt(copyOnly, sigOnly)
+	if err != nil {
+		return err
+	}
 	if len(tags) > 0 {
 		onlyFlagSet = true
 	} else {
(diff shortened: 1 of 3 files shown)


cpanato (commented):
can we add tests?

marcofranssen:
I propose to do that in follow up pr. There have been historically 0 tests on this command.

haydentherapper:
We recognize that much of cmd/ is untested but we should use this opportunity to improve coverage. I'd recommend adding tests for the changed function, and refactoring if needed so you can have a more testable function.

marcofranssen:
Well I only have limited time, so don't want to blow up the scope of this PR. I'm just a happy user of cosign doing my contribution, trying to improve things that I can do. To refactor tests and such I need much more time which I don't have right now. 

So I would appreciate at least the improvements in this PR can be merged.

haydentherapper (approved):
Merging as this is straightforward enough, thanks!

Issue #1775, opened 2022-04-19
cosigned: allow configuring hash algorithm for signature verification


nsmith5:
**Description**

The cosign CLI allows setting `--signature-digest-algorithm` to one of `sha224|sha256|sha384|sha512` to accommodate [the many hash algorithms that might have been used with elliptical KMS keys](https://github.com/sigstore/sigstore/blob/eaea65e41486b908672b6b48eea37cf99ec7ca14/pkg/signature/kms/aws/client.go#L152-L163). The [current `ClusterImagePolicy`](https://github.com/sigstore/cosign/blob/623d50f9b77ee85886a166daac648455e65003ec/pkg/apis/cosigned/v1alpha1/clusterimagepolicy_types.go) API doesn't have a way to specify the hashing algorithm for verification so it defaults to sha256 and incorrectly rejects signatures with a different hash function.

hectorj2f:
@nsmith5 Good catch!

asraa:
FYI I think there's also some major complications around using non-sha256 when you're doing a keyless signature (uploading a rekor entry sort-of-requires that the sig was based off sha256, since we can't get the hash algorithm info that was used from a SignerVerifier right now)

https://github.com/sigstore/cosign/blob/80fe5a31a398c45395d09d23e237c676ddad709f/pkg/cosign/tlog.go#L187-L189

nsmith5:
Ah ok, my use-case doesn't use Rekor so I didn't hit that. Is that something we can / should change in Rekor ( allow other hash algorithms) or in cosign (restrict to SHA256 only)?

asraa:
> Ah ok, my use-case doesn't use Rekor so I didn't hit that. 

Gotcha! Hmmm I think ideally this is fixed through SignerVerifier exposing the correct SHA, then making cosign pull that info in when creating the rekor entry. Rekor would also need to actually use a customize-able SHA here (https://github.com/sigstore/rekor/blob/c39c0beb340273f0d729f1b853b49691b6445eae/pkg/pki/x509/x509.go#L75) when verifying.

I guess currently we should drop an error that we can't upload if someone uses `--signature-digest-algorith`

hectorj2f:
It sounds like a plan, then I assume we have to add this functionality to both rekor and cosigned.

tuminoid:
I think I'm hitting this one still today. Discussing it in [Slack](https://sigstore.slack.com/archives/C01PZKDL4DP/p1730288293793379).

PR #3917, opened 2024-10-30
WIP: add support for variable digests and paddings, ie. stop hardcoding SHA256 in verify.go

tuminoid:
Work-in-progress!

#### Summary

`cosign verify` has a switch called `--signature-digest-algorithm`, which apparently does nothing. You can have SHA256 signature, pass flag for `--signature-digest-algorithm sha512` and nothing complains. It also does not make SHA512 supported. Also, only PKCS#1 is supported. PSS is not. 

Namely, this PR stops hardcoding SHA256 and assuming PKCS#1 in verify.go, in minimal way. It makes verify work for SHA512/PSS for example. Let's say its a starting point. Feedback welcome.

It does not:
- [ ] fix the command line flag being ignored
- [ ] fix any other issue listed in #1775 

#### Release Note

Partially-fixes: #1775 

#### Documentation

TODO, though this is a bugfix, so not aiming to change anything, just make things work.

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/asn1"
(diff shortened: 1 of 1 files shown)


haydentherapper:
I believe the `signature-digest-algorithm` flag was specifically for KMS keys, as it's used in https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify.go#L198.

Note there was prior work on cryptographic agility: https://github.com/sigstore/cosign/pull/3497

I recognize the difference between these two approaches requires far more work for the earlier PR, but I don't want to shoehorn RSA-PSS support in in one specific verification workflow. I'd want to explore pulling in parts of that PR to support additional algorithms and digests for both signing and verification.

tuminoid:
Thanks @haydentherapper! We discussed this also in slack, and as expected, this was very simplistic approach. To actually get support for SHA512/PSS in Cosign/Sigstore, one would need:

- implement same support for Cosign SDKs in all supported languages
- Refactor signing/verifying library
- Implement support in Sigstore services (Fulcio/Rekor)
- Fix all clients

IMO this kind of effort must be thus coordinated and roadmapped properly, with approved designs, and hence outside of individual contributor scope.

I will close my PR.

PR #3938, opened 2024-11-19
Add e2e tests using a TUF mirror

cmurphy:
Add a test that, instead of overriding trusted keys with environment variables, actually initializes a TUF cache using a local TUF mirror. This will prepare us to safely make changes to the TUF client.

Metadata for the public good instance goes through a different code path, since those keys are embedded in the code. We can't test signing and verifying with the public good instance hermetically, so we rely on conformance tests to catch issues with that path.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,7 @@ require (
 	github.com/spf13/viper v1.19.0
 	github.com/spiffe/go-spiffe/v2 v2.3.0
 	github.com/stretchr/testify v1.9.0
+	github.com/theupdateframework/go-tuf/v2 v2.0.1
 	github.com/transparency-dev/merkle v0.0.2
 	github.com/withfig/autocomplete-tools/integrations/cobra v1.2.1
 	github.com/xanzy/go-gitlab v0.109.0
@@ -241,7 +242,6 @@ require (
 	github.com/tchap/go-patricia/v2 v2.3.1 // indirect
 	github.com/thales-e-security/pool v0.0.2 // indirect
 	github.com/theupdateframework/go-tuf v0.7.0 // indirect
-	github.com/theupdateframework/go-tuf/v2 v2.0.1 // indirect
 	github.com/titanous/rocacheck v0.0.0-20171023193734-afe73141d399 // indirect
 	github.com/tjfoc/gmsm v1.4.1 // indirect
 	github.com/urfave/negroni v1.0.0 // indirect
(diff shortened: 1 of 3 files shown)


haydentherapper (dismissed):
This looks great!

haydentherapper:
@cpanato any idea on why the goreleaser snapshot is failing?

jku (approved):
Looks good. Some notes:
* `ConsistentSnapshot = false` means the client will use some code paths that are different than prod (using consistent snapshot here would require  adding version prefixes to snapshot/targets filenames and hash prefixes to artifact filenames). I think this might be irrelevant in this case (considering you just want to verify the code calling the tuf client isn't broken)
* if  you want to simplify, you could use a single key to sign all roles (maybe not worth it if that part is copied from an example). Could get rid of the second loop in setupTUF().
* this PR does not include trusted_root.json yet but I guess you'd add that once you have code that would use it?

Seems good to me as is.

jku:
> `hook failed: shell: '/bin/bash -c if [ -n "$(git --no-pager diff --exit-code go.mod go.sum)" ]; then exit 1; fi': exit status 1: [no output]`

well that looks fun.
* `--exit-code` looks like a red herring as the test does not actually use the git exit code (because "-n")
* `$()` eats the git diff output so we can't see the what this failed on
* ...but there might also be a quoting failure here: notice how in the log there is no starting quote after `bash -c` even though there is in sources

fixing the test to use git exit code instead of comparing strings would at least allow us to see what happens and might avoid the potential quoting issue -- `git --no-pager diff --exit-code go.mod go.sum` looks like it should work as is: no `bash -c` needed?

haydentherapper (approved):

haydentherapper:
@cpanato we might need to try to free up more disk space, IIRC this is usually the culprit for this test failing

jku:
> @cpanato we might need to try to free up more disk space, IIRC this is usually the culprit for this test failing

this seems unlikely to be the reason in this case: it's not actually building anything yet at the point of failure.

jku:
> I can try out the refactor I mentioned: it should give some more leads.

I was not expecting it but turns out the patch does simply create a `go mod tidy` change.

There's two commits in https://github.com/jku/cosign/commits/tuf-e2e-tests/ -- @cmurphy feel free to take both or just do  a `go mod tidy` (I can submit the the hook refactor separately).

haydentherapper (approved):

cmurphy:
@jku wow thanks for catching that, I updated go.mod here if you want to submit your git-diff fix as its own PR

PR #3939, opened 2024-11-21
feat: sign will log entry Id

### LINKED ISSUE #3805: sign: Include rekor entry ID in output
## Summary

cosign's `sign*` commands currently output a signature's transparency log index number (as of v2.2.4). As an enhancement, it would be helpful if cosign also output the rekor entry ID.

## Details

rekor entries can be searched using an index number. However, the index number does not include the log tree (or "shard") ID. These identifiers are created by [rekor log sharding](https://docs.sigstore.dev/logging/sharding/).

The rekor entry ID consists of the shard ID and an entry-specific UUID (the latter appears to be the [hash of the Merkle tree leaf](https://github.com/sigstore/rekor/blob/v1.3.6/pkg/api/entries.go#L116-L122)). Knowing the entry ID allows end users to determine which shard the entry exists in.

It would be helpful to include it in cosign's output for informational and historical purposes.

## A quick note on terminology

For those unfamiliar, the use of the term `entry ID` appears to be relatively recent. The original value was referred to as a "UUID". The `rekor-cli` still refers to the entry ID as a UUID. The following GitHub issues and pull requests provide some context for this change.

- https://github.com/sigstore/rekor/issues/625
- https://github.com/sigstore/rekor/pull/623 
- https://github.com/sigstore/rekor/issues/353 
- https://github.com/sigstore/rekor/issues/487

goatwu1993:
#### Summary

Fix #3805

### DIFF
--- a/internal/pkg/cosign/rekor/signer.go
+++ b/internal/pkg/cosign/rekor/signer.go
@@ -23,11 +23,14 @@ import (
 	"io"
 	"os"
 
+	"github.com/go-openapi/runtime"
 	"github.com/sigstore/cosign/v2/internal/pkg/cosign"
+	"github.com/sigstore/cosign/v2/internal/ui"
 	cosignv1 "github.com/sigstore/cosign/v2/pkg/cosign"
 	cbundle "github.com/sigstore/cosign/v2/pkg/cosign/bundle"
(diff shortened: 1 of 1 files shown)


Hayden-IO:
Closing as we push out a major release. There's a lot of places we need to clean up UX.

PR #3943, opened 2024-11-25
goreleaser: Refactor git-diff hook

jku:
The git-diff hook implementation prevents us from seeing the failure details.

Simplify so that the git diff output ends up in stdout: otherwise functionality should remain same.

CC @cpanato as original author.


cpanato (approved):
lgtm

thanks

Issue #1880, opened 2022-05-14
proposal: remove pkg/errors dependency


zchee:
**Description**

The pkg/errors package has been deprecated.
I think remove this dependency will more maintainability.

zchee:
cc: @dlorenc

dlorenc:
cc @imjasonh

imjasonh:
+1 to this!

If you're volunteering to do the cleanup that's great, please cc me on any PRs you send. And if not that's okay too, just let me know and I'll take care of it.

imjasonh:
I added a [quick WIP depcheck_test](https://github.com/sigstore/cosign/compare/main...imjasonh:depcheck-pkg-errors) to help ensure we don't depend on this and uhh...we have a lot of transitive deps on it:

```
--- FAIL: TestNoDeps (0.71s)
    --- FAIL: TestNoDeps/github.com/sigstore/cosign/cmd/cosign (0.71s)
        depcheck.go:126: CheckNoDependency() = github.com/sigstore/cosign/cmd/cosign depends on banned dependency github.com/pkg/errors
            github.com/sigstore/cosign/cmd/cosign
            github.com/sigstore/cosign/cmd/cosign/cli  # Also: [github.com/sigstore/cosign/cmd/cosign/cli/attach github.com/sigstore/cosign/cmd/cosign/cli/attest github.com/sigstore/cosign/cmd/cosign/cli/copy github.com/sigstore/cosign/cmd/cosign/cli/download github.com/sigstore/cosign/cmd/cosign/cli/fulcio github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioverifier github.com/sigstore/cosign/cmd/cosign/cli/generate github.com/sigstore/cosign/cmd/cosign/cli/rekor github.com/sigstore/cosign/cmd/cosign/cli/sign github.com/sigstore/cosign/cmd/cosign/cli/triangulate github.com/sigstore/cosign/cmd/cosign/cli/upload github.com/sigstore/cosign/cmd/cosign/cli/verify github.com/sigstore/cosign/pkg/policy github.com/sigstore/cosign/pkg/sget]
            github.com/sigstore/cosign/cmd/cosign/cli/options
            github.com/awslabs/amazon-ecr-credential-helper/ecr-login
[...]

PR #1887, opened 2022-05-16
Remove dependency on deprecated github.com/pkg/errors

zchee:
#### Summary

Remove dependency on deprecated github.com/pkg/errors.

#### Ticket Link

Update: https://github.com/sigstore/cosign/issues/1880

#### Release Note

???

PR #3956, opened 2024-12-10
Remove dependency on github.com/pkg/errors

codysoyland:
Signed-off-by: Cody Soyland <codysoyland@github.com>

#### Summary

This removes the deprecated `github.com/pkg/errors` package, which has been replaced by the standard library. This work was previously done in https://github.com/sigstore/cosign/pull/1887, but several invocations have been gradually added since then.

Fixes https://github.com/sigstore/cosign/issues/1880 again :)

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcio.go
+++ b/cmd/cosign/cli/fulcio/fulcio.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/go-jose/go-jose/v3/jwt"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/sign/privacy"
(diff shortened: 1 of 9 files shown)


haydentherapper (approved):
Thank you!

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -209,7 +209,7 @@ func NewClient(fulcioURL string) (api.LegacyClient, error) {
 // or a path to an identity token via the --identity-token flag
 func idToken(s string) (string, error) {
 	// If this is a valid raw token or is empty, just return it
-	if _, err := jose.ParseSigned(s); err == nil || s == "" {
+	if _, err := jwt.ParseSigned(s); err == nil || s == "" {
```
codysoyland, line 212: I forgot to mention it in the PR description, but the reason I replaced `go.step.sm/crypto/jose` with `github.com/go-jose/go-jose/v3/jwt` is that the former is a wrapper around the latter (that we don't benefit from) and includes references to the `github.com/pkg/errors` package.

PR #3965, opened 2024-12-12
bump sigstore-conformance to v0.0.14 release

bobcallaway:
This updates to the latest release of sigstore-conformance


codysoyland (commented):
I'm a little -1 on this, as a change in conformance testing can cause builds to start failing (so we'd need to fix it right away or override branch protection). Dependabot PRs should already take care of the version bumps, and we can see the test failures before they land in main. Of course, releases only happen so often so it's nice to know about failures early...

Alternatively, we can have two conformance workflows, one with the latest pinned release and the second running against main, and only use branch protection on the pinned one so that branch protection is stable. What do you think of that?

bobcallaway:
> I'm a little -1 on this, as a change in conformance testing can cause builds to start failing (so we'd need to fix it right away or override branch protection). Dependabot PRs should already take care of the version bumps, and we can see the test failures before they land in main. Of course, releases only happen so often so it's nice to know about failures early...
> 
> Alternatively, we can have two conformance workflows, one with the latest pinned release and the second running against main, and only use branch protection on the pinned one so that branch protection is stable. What do you think of that?

aside: dependabot wasn't working here, because this was back pinned to v0.0.11 which was released in March...

bobcallaway:
> I'm a little -1 on this, as a change in conformance testing can cause builds to start failing (so we'd need to fix it right away or override branch protection). Dependabot PRs should already take care of the version bumps, and we can see the test failures before they land in main. Of course, releases only happen so often so it's nice to know about failures early...
> 
> Alternatively, we can have two conformance workflows, one with the latest pinned release and the second running against main, and only use branch protection on the pinned one so that branch protection is stable. What do you think of that?

i'll downscope this change to just get to the latest release. I still vote +1 on this because if we have a valid conformance test that isn't passing on a client, then the client should be updated ASAP in my opinion (and block other changes/releases).

PR #3979, opened 2024-12-17
Add nightly conformance test workflow

codysoyland:
#### Summary

This is an alternative / supplement to https://github.com/sigstore/cosign/pull/3978, as way of automatically running conformance tests against the current main branch of sigstore-conformance. This will run nightly at midnight UTC and report an issue on failure.

### DIFF
--- a/.github/workflows/conformance-nightly.yml
+++ b/.github/workflows/conformance-nightly.yml
@@ -50,7 +50,7 @@ jobs:
             const { owner, repo } = context.repo;
             const runId = context.runId;
             const issueTitle = 'Conformance Tests Failed';
-            const issueBody = `The nightly conformance tests have failed. Please check the logs for more details.\n\nWorkflow run: https://github.com/${owner}/${repo}/actions/runs/${runId}\n\ncc @sigstore/security-response-team`;
+            const issueBody = `The nightly conformance tests have failed. Please check the logs for more details.\n\nWorkflow run: https://github.com/${owner}/${repo}/actions/runs/${runId}\n\ncc @sigstore/security-response-team @sigstore/cosign-codeowners`;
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-nightly.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Create Issue on Failure
+        if: failure()
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { owner, repo } = context.repo;
+            const runId = context.runId;
+            const issueTitle = 'Conformance Tests Failed';
+            const issueBody = `The nightly conformance tests have failed. Please check the logs for more details.\n\nWorkflow run: https://github.com/${owner}/${repo}/actions/runs/${runId}\n\ncc @sigstore/security-response-team`;
```
haydentherapper, line 53: Should this go to cosign-codeowners rather than the SRC?

bobcallaway: well given that it might be a security issue i'd rather the SRC triage it first

codysoyland: We can tag whoever, so it probably makes sense to tag either codeowners or both SRC+codeowners.

By the way, the API has a way of [setting assignees](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#create-an-issue--parameters), but it doesn't seem to support groups, only individual users. That's why I'm using the issue body to ping the group.

bobcallaway: then maybe just tag both please

bobcallaway (approved):

haydentherapper (approved):

PR #4003, opened 2025-01-08
Add more complex TUF tests

cmurphy:
Add tests for invalid keys, custom metadata, and TSA verification.

These tests do not test the signing flow for when Fulcio issues a detached SCT, which can be affected by the TUF client, because the ephemral CA Fulcio backend cannot issue detached SCTs.


cpanato (approved):
cool

thanks!

haydentherapper (approved):
Great work on this!

jku (approved):
looks good to me, cases look reasonable.

cosign shouldn't really test tuf internals of course... but since the client does not yet use trusted_root.json and needs to do this meddling in the TUF internal, the test makes sense as well (and should translate to similar simpler tests in the future when we can just test the trusted_root.json parsing)

PR #4006, opened 2025-01-10
Add UseSignedTimestamps to CheckOpts, refactor TSA options

codysoyland:
#### Summary

This PR adds `UseSignedTimestamps` to `CheckOpts` and refactors TSA options to improve readability, consistency, and error handling.

This is partially split out of https://github.com/sigstore/cosign/pull/3889, where we will need `CheckOpts.UseSignedTimestamps` in order to set the options needed by sigstore-go's verifier.


haydentherapper (approved):
This looks great, thanks for the cleanup!

When we document this change in the changelog, we should mention that it is a breaking API change if you're calling `cosign.Verify` with timestamps, but a) we have no semver for the API, and b) I don't think this is a common usecase.

haydentherapper (approved):

Issue #3269, opened 2023-09-29
cosign sign `--allow-insecure-registry` does not behave as expected over http/https proxy


jkk-intel:
**Description**

Having `--allow-insecure-registry` should work for BOTH insecure and secure registries (as the name suggests "allow"). If `--allow-insecure-registry` flag is set, cosign will fail to resolve DNS for even secure registries (e.g. myregistry.azurecr.io)

e.g. 

```
$ export HTTP_PROXY="mycorp.proxy.com"
$ export HTTPS_PROXY="mycorp.proxy.com"
$ cosign sign --key=cosign.key "myregistry.azurecr.io/my-image:some-tag"
Pushing signature to: myregistry.azurecr.io/my-image:some-tag
# WORKS !
```

```
$ export HTTP_PROXY="mycorp.proxy.com"
$ export HTTPS_PROXY="mycorp.proxy.com"
$ cosign sign --key=cosign.key --allow-insecure-registry "myregistry.azurecr.io/my-image:some-tag"
Error: signing [myregistry.azurecr.io/my-image:some-tag@sha256]: accessing image: Get "https://myregistry.azurecr.io/v2/": dial tcp: lookup myregistry.azurecr.io: no such host
main.go:74: error during command execution: signing [myregistry.azurecr.io/my-image:some-tag@sha256]: accessing image: Get "https://myregistry.azurecr.io/v2/": dial tcp: lookup myregistry.azurecr.io: no such host
# FAILS
```

**Version**

```
GitVersion:    v2.0.2
GitCommit:     871448050b924a7946ebe47678f23aae09ef432d
GitTreeState:  clean
BuildDate:     2023-04-24T17:31:42Z
GoVersion:     go1.20.3
Compiler:      gc
Platform:      darwin/amd64
```

MinerYang:
Same issue here, 
```
COSIGN_LOG=debug cosign sign -y --allow-insecure-registry --key cosign.key --verbose=true <registry>/project-1732/alpine:latest
WARNING: Image reference <registry>/project-1732/alpine:latest uses a tag, not a digest, to identify the image to sign.
    This can lead you to sign a different image than the intended one. Please use a
    digest (example.com/ubuntu@sha256:abc123...) rather than tag
    (example.com/ubuntu:latest) for the input to cosign. The ability to refer to
    images by tag will be removed in a future release.

2024/11/22 09:27:45 --> GET https://<registry>/v2/
[...]

PR #4007, opened 2025-01-12
fix: set tls config while retaining other fields from default http transport

nianyush:
Signed-off-by: Nianyu Shen <xiaoyu9964@gmail.com>

#### Summary
original issue https://github.com/sigstore/cosign/issues/3269
Proxy from environment variable is not respected while setting insecure true.
At https://github.com/sigstore/cosign/blob/87c08b00fca8a96a2ef756cee3953372a5783a45/cmd/cosign/cli/options/registry.go#L154, its initializing an empty transport with only insecure set but those default behaviors from default transport is also gone. In this case, proxy value from envs is removed. The fix is to clone a default transport object and override the tlsconfig inside it.


haydentherapper (approved):
Thank you!

PR #4013, opened 2025-01-14
Refactor verifyNewBundle into library function

codysoyland:
#### Summary

This is a refactor of the bundle verification logic that replaces [verifyNewBundle](https://github.com/sigstore/cosign/blob/accc80a840524890790c5c368fee26e22ec60bda/cmd/cosign/cli/verify/verify_bundle.go#L65) with a re-written library function `cosign.VerifyNewBundle`, which accepts a simplified set of arguments, using [`CheckOpts`](https://github.com/sigstore/cosign/blob/accc80a840524890790c5c368fee26e22ec60bda/pkg/cosign/verify.go#L87) to supply the majority of the configuration. It has been moved out of the `cli` package into the `cosign` package and made public.

This is partially extracted from https://github.com/sigstore/cosign/pull/3889.


codysoyland:
@haydentherapper @cmurphy @steiza I would appreciate a look at this whenever you have a moment!

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -124,8 +114,44 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
 		Offline:                      c.Offline,
 		IgnoreTlog:                   c.IgnoreTlog,
 		UseSignedTimestamps:          c.TSACertChainPath != "" || c.UseSignedTimestamps,
+		NewBundleFormat:              c.KeyOpts.NewBundleFormat || checkNewBundle(c.BundlePath),
```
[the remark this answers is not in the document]
steiza: It's a good question, but we are also hoping to make `NewBundleFormat` be the default soon, when we start cutting cosign 3.x releases.

steiza (commented):
Nice work! I don't have any further comments other than the things already raised by @malancas.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -127,8 +107,85 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
+		}
+	case c.CertRef != "":
+		cert, err = loadCertFromFileOrURL(c.CertRef)
+		if err != nil {
+			return err
+		}
+	}
+
+	if co.NewBundleFormat {
+		if options.NOf(c.RFC3161TimestampPath, c.TSACertChainPath, c.RekorURL, c.CertChain, c.CARoots, c.CAIntermediates, c.CertRef, c.SigRef, c.SCTRef) > 1 {
```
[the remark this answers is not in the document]
codysoyland: Yes! Good callout, I believe you're correct it should be `>= 1`/ `> 0`. I'll modify.

cmd/cosign/cli/verify/verify_blob_attestation.go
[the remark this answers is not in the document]
codysoyland: Yeah, I agree we try to make `NewBundleFormat` the default in a future release.

As you mention, there is a lot of duplication between the commands, not just this CheckOpts struct... I'm hoping we can follow up and do some more de-duplication at some point. It's a little hard to refactor since the variables are stored in per-command structs. We can refactor more easily if we move a lot of config into a share structs such was what was done with `CertVerifyOptions`.

codysoyland:
I will take a look at the conformance failures Monday. Thanks all!

malancas (approved):
LGTM 🚀

haydentherapper:
Needs a rebase to fix the tests that didn't run

pkg/cosign/verify.go
```diff
@@ -165,6 +170,89 @@ type CheckOpts struct {
...
+
+	if co.SigVerifier != nil {
+		// We are verifying with a public key
+		policyOptions = append(policyOptions, verify.WithKey())
+		newExpiringKey := root.NewExpiringKey(co.SigVerifier, time.Time{}, time.Time{})
+		vTrustedMaterial.keyTrustedMaterial = root.NewTrustedPublicKeyMaterial(func(_ string) (root.TimeConstrainedVerifier, error) {
+			return newExpiringKey, nil
+		})
+	} else { //nolint:gocritic
+		// We are verifiying with a certificate
```
cmurphy, line 239:
```suggestion
		// We are verifying with a certificate
```

pkg/cosign/verify.go
```diff
@@ -165,6 +170,89 @@ type CheckOpts struct {
...
+	verifierOptions = make([]verify.VerifierOption, 0)
+
+	if co.SigVerifier != nil {
+		// We are verifying with a public key
+		policyOptions = append(policyOptions, verify.WithKey())
+		newExpiringKey := root.NewExpiringKey(co.SigVerifier, time.Time{}, time.Time{})
+		vTrustedMaterial.keyTrustedMaterial = root.NewTrustedPublicKeyMaterial(func(_ string) (root.TimeConstrainedVerifier, error) {
+			return newExpiringKey, nil
+		})
+	} else { //nolint:gocritic
```
cmurphy, line 238: This condition makes sense but it's a little different than what we had before, where `WithSignedCertificateTimestamps` is not conditioned on having a public key versus a cert https://github.com/sigstore/cosign/blob/7fc127b4e8226035ca7a7be9f423a653f09c3989/cmd/cosign/cli/verify/verify_bundle.go#L165-L167 - is this an intentional change in functionality or am I misreading the refactor?

pkg/cosign/verify.go
```diff
@@ -165,6 +170,89 @@ type CheckOpts struct {
...
+		// We are verifiying with a certificate
+		if !co.IgnoreSCT {
+			verifierOptions = append(verifierOptions, verify.WithSignedCertificateTimestamps(1))
+		}
+	}
+
+	if !co.IgnoreTlog {
+		verifierOptions = append(verifierOptions, verify.WithTransparencyLog(1), verify.WithIntegratedTimestamps(1))
```
cmurphy, line 246: Should we check if there are rekor keys in the trusted root before adding this policy, as it was before? https://github.com/sigstore/cosign/blob/7fc127b4e8226035ca7a7be9f423a653f09c3989/cmd/cosign/cli/verify/verify_bundle.go#L157 same for TSAs below

cmurphy (commented):
Generally lgtm, couple minor questions and a typo

pkg/cosign/verify.go
codysoyland: It feels dangerous to ignore the user input (IgnoreTlog, UseSignedTimestamps) if the trusted root doesn't contain rekor/tsa entries. I think the previous behavior was incorrect but I could be misunderstanding.

haydentherapper: Good question! I'd prefer what's here now, that verification should err out if your trusted root doesn't have the correct key material yet the user has specified that they want verification. Otherwise, if I can convince you to use my maliciously crafted trusted root that is missing key material, log/TSA verification would be unexpectedly skipped (it's a bit of a convoluted threat model, as you shouldn't be trusting some arbitrary trusted root anyways)

cmd/cosign/cli/verify/verify_bundle.go
```diff
@@ -36,169 +32,17 @@ import (
...
-
-	sev, err := verify.NewSignedEntityVerifier(trustedmaterial, verifierConfig...)
-	if err != nil {
-		return nil, err
-	}
-
-	return sev.Verify(bundle, verify.NewPolicy(artifactOpt, identityPolicies...))
-}
-
 func AssembleNewBundle(ctx context.Context, sigBytes, signedTimestamp []byte, envelope *dsse.Envelope, artifactRef string, cert *x509.Certificate, ignoreTlog bool, sigVerifier signature.Verifier, pkOpts []signature.PublicKeyOption, rekorClient *client.Rekor) (*sgbundle.Bundle, error) {
```
haydentherapper, line 46: For a later PR, we should move this to somewhere under pkg/ rather than be under the cli package. It looks like this is only used by the `cosign bundle` command.

cmd/cosign/cli/verify/verify_blob_attestation.go
```diff
@@ -92,19 +95,6 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
...
-	if c.KeyOpts.NewBundleFormat || checkNewBundle(c.BundlePath) {
-		if options.NOf(c.RFC3161TimestampPath, c.TSACertChainPath, c.RekorURL, c.CertChain, c.CARoots, c.CAIntermediates, c.CertRef, c.SCTRef) > 1 {
-			return fmt.Errorf("when using --new-bundle-format, please supply signed content with --bundle and verification content with --trusted-root")
-		}
-		_, err = verifyNewBundle(ctx, c.BundlePath, c.TrustedRootPath, c.KeyRef, c.Slot, c.CertVerifyOptions.CertOidcIssuer, c.CertVerifyOptions.CertOidcIssuerRegexp, c.CertVerifyOptions.CertIdentity, c.CertVerifyOptions.CertIdentityRegexp, c.CertGithubWorkflowTrigger, c.CertGithubWorkflowSHA, c.CertGithubWorkflowName, c.CertGithubWorkflowRepository, c.CertGithubWorkflowRef, artifactPath, c.Sk, c.IgnoreTlog, c.UseSignedTimestamps, c.IgnoreSCT)
-		if err == nil {
-			fmt.Fprintln(os.Stderr, "Verified OK")
-		}
-		return err
-	} else if c.TrustedRootPath != "" {
```
haydentherapper, line 104: Looks like this is missing from verify_blob_attestation?

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -348,3 +381,25 @@ func payloadBytes(blobRef string) ([]byte, error) {
...
+		return "", nil, err
+	}
+	return hexAlg, digestBytes, nil
+}
+
+func loadTrustedRoot(ctx context.Context, trustedRootPath string) (*root.TrustedRoot, error) {
+	if trustedRootPath != "" {
+		return root.NewTrustedRootFromPath(trustedRootPath)
+	}
+	ui.Infof(ctx, "no --trusted-root specified; fetching public good instance verification material via TUF")
```
haydentherapper, line 401: I'd guess the vast majority of use cases won't specify a trusted root, so I'd remove this.

haydentherapper (commented):
few small nits, looks great!

cmd/cosign/cli/verify/verify_blob_attestation.go
codysoyland: Thanks for catching that! I've pushed a fix.

haydentherapper (approved):

PR #4016, opened 2025-01-17
add `sha256:...` as alternative way to specify a blob

wolfv:
Hey, looking for some early feedback on this change to make validation of blobs without providing the actual blob easier.

#### Summary

With this change, one can validate blobs without the blob - just by `sha256:<hash>`.

#### Release Note

- Allow passing a hash instead of an actual blob / file by using `sha256:<hash>`.

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob_attestation.go
+++ b/cmd/cosign/cli/verify/verify_blob_attestation.go
@@ -28,6 +28,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
+	"regexp"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/options"
@@ -75,6 +77,9 @@ type VerifyBlobAttestationCommand struct {
 	SignaturePath       string // Path to the signature
 	UseSignedTimestamps bool
 }
+
+// Add validation helper
+var sha256RegExp = regexp.MustCompile("^[a-f0-9]{64}$")
 
 // Exec runs the verification command
 func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath string) (err error) {
@@ -127,29 +132,39 @@ func (c *VerifyBlobAttestationCommand) Exec(ctx context.Context, artifactPath st
 	}
 	var h v1.Hash
 	if c.CheckClaims {
-		// Get the actual digest of the blob
-		var payload internal.HashReader
-		f, err := os.Open(filepath.Clean(artifactPath))
-		if err != nil {
-			return err
-		}
-		defer f.Close()
-		fileInfo, err := f.Stat()
-		if err != nil {
-			return err
-		}
-		err = payloadsize.CheckSize(uint64(fileInfo.Size()))
-		if err != nil {
-			return err
-		}
+		var hexDigest string
+		if strings.HasPrefix(artifactPath, "sha256:") {
+			// log that we are using a sha256 hash
+			fmt.Fprintf(os.Stderr, "Using sha256 hash %s\n", artifactPath)
+			hexDigest = strings.TrimPrefix(artifactPath, "sha256:")
+			if !sha256RegExp.MatchString(hexDigest) {
+				return fmt.Errorf("invalid sha256 hash format: must be 64 hex characters")
+			}
+		} else {
+			// Get the actual digest of the blob
+			var payload internal.HashReader
+			f, err := os.Open(filepath.Clean(artifactPath))
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			fileInfo, err := f.Stat()
+			if err != nil {
+				return err
+			}
+			err = payloadsize.CheckSize(uint64(fileInfo.Size()))
+			if err != nil {
+				return err
+			}
 
-		payload = internal.NewHashReader(f, sha256.New())
-		if _, err := io.ReadAll(&payload); err != nil {
-			return err
+			payload = internal.NewHashReader(f, sha256.New())
+			if _, err := io.ReadAll(&payload); err != nil {
+				return err
+			}
+			hexDigest = hex.EncodeToString(payload.Sum(nil))
 		}
-		digest := payload.Sum(nil)
 		h = v1.Hash{
-			Hex:       hex.EncodeToString(digest),
+			Hex:       hexDigest,
 			Algorithm: "sha256",
 		}
 		co.ClaimVerifier = cosign.IntotoSubjectClaimVerifier


haydentherapper:
@woodruffw, @kommendorkapten and I had a discussion in https://github.com/sigstore/protobuf-specs/issues/444. There was some concern that providing only the hash of an attestation would lead to a lack of verification of the document itself.

wolfv:
Sure, but in our case the server application already checks the hash and we don't want to send the file around to the microservice to validate the signature :)

Also sigstore-python already supports this.

But also fine with me if it isn't accepted!

woodruffw:
Hey @wolfv! Good to see you 🙂 

Per https://github.com/sigstore/protobuf-specs/issues/444, I think there are two subtly different cases here: when doing a "bare" verification (`cosign verify-blob`) having a hash instead of an artifact is OK, since the signature is over a "bare" payload and the hash is equivalently binding to it.

However, in the attestation verification context, just a hash is insufficient (since it binds to the subject's digest, but not the subject's name.

So, I think `cosign` _could_ support `sha256:...` in the "bare" case, but forbid it for the attestation case. OTOH, I could see that being confusing for users -- I'm curious what @haydentherapper thinks.

(In `sigstore-python` we allow `sha256:...` in both cases, which was arguably a mistake 😅 -- I'm going to investigate removing or augmenting it in the next stable release. In particular, if we decide to keep it, I'll likely add some kind of required `--subject` flag to complement the digest.)

haydentherapper:
I agree with William. This PR is for the attestation case, but if this were for the "bare" case, that can be accepted.

With that said, I'm realizing that Cosign's attestation verification here is incomplete, missing subject name verification. I'd be open to a change that accepts a digest _and_ a subject name, with the accompanying changes for the latter.

codysoyland:
> However, in the attestation verification context, just a hash is insufficient (since it binds to the subject's digest, but not the subject's name.

Just to be clear, the current implementation doesn't check the file name -- it requires a file on disk and just computes its digest. This PR just adds the ability to specify a digest as an input instead of a file. We can discuss validating that the subject name matches, but that is orthogonal to this PR.

codysoyland:
@haydentherapper @woodruffw I'm curious how you'd respond to https://github.com/sigstore/cosign/issues/4019? It's a very related ask with regard to passing a subject digest as an input for attestation verification, but in the context of verifying git commits (not files on disk) using cosign. I think it's a reasonable ask and I'm not quite sure I understand the argument against providing digests as input for attestation verification.

wolfv:
Yeah after diving deeper into the whole sigstore / in-toto validation topic I understand that you would want to validate the in-toto statement subject.name against the file name of the artifact that we're verifying. 

I guess that makes a lot of sense, but this PR seemed to work without providing a subject or alternative filename so I do not think that's currently implemented (as @codysoyland mentions).

We'll do these kind of validations on our end either way but yeah, might be nice to (also) add a `--subject` flag and require that together with a sha hash, but I think that would be a different PR :)

codysoyland:
> In contrast, it's not clear what the user is intending when they pass a digest during attestation verification

Interesting, I hadn't considered that the user's intent might be unclear if we offer a flag to specify a digest. The `cosign verify-blob-attestation` command help text is fairly clear that the blob is checked against the attestation subjects, so I guess it feels like a natural extension to allow passing the digest as a means connecting the bundle to the artifact, but you're right we need to be careful about the documentation/UX here.

I'm actually more receptive to the notion of disallowing digest as inputs in "raw" message signature verification (as opposed to attestations)... In https://github.com/sigstore/protobuf-specs/issues/444, @kommendorkapten brought up three scenarios where providing a digest may be considered harmful, and two applied to message signatures: poor crypto library support for verifying signatures with only a message digest, and some algorithms (i.e. ed25519) incapable of verification with just a digest. The third argument was related to api security and could possible apply to DSSE, and I admit I don't understand the threat model aside from some potentially confusing UX like you pointed out.

> Alternatively, we could firm up the client spec to make it clear where the division of responsibilities between Sigstore clients and downstreams lies

Hard agree! Sigstore is mainly concerned with tying an identity to a claim/subject, and it seems sensible for the verifier to open the attestation and confirm the subject is present, but the breakdown in responsibilities is definitely unclear to the average user. This is sort of covered in [the client spec](https://github.com/sigstore/architecture-docs/blob/main/client-spec.md#45-signature-verification):

> Verifier SHOULD accept the raw artifact and compute the message digest to minimize any risk for confusion attacks.

kommendorkapten:
I think I can't add much content to the discussion here, most of pitfalls have been mentioned with passing in digests. And as @codysoyland mentions the client spec mentions that

> Verifier SHOULD accept the raw artifact and compute the message digest to minimize any risk for confusion attacks

So i think the spec is ok, it's more about getting the cli flags in order.

But another thing briefly mentioned here, and something I took up in https://github.com/sigstore/protobuf-specs/issues/444 is the effect this would have in the future for PQC. The current standardized algorithms follows Ed25519 in that they require the complete message. There are pre-hashed versions defined too, but will they be common/supported? My understanding now is that Ed25519ph is not very common/supported, and exposing an API that will break for the common scenarios is not that good (note that his will be a primary concerns for blobs, not in-toto/dsse as the dsse envelope _is_ the message being signed and verified).

codysoyland:
@wolfv I spoke with @kommendorkapten this morning, and we're in agreement that we could accept this PR with some caveats/changes. We want to make sure that the UI is clear and explicit about what is checked when you enter a digest. I would recommend:

- Add a new flag, something like `--expect-subject-digest=sha256:xxxxx` (this name is up for debate)
  - Add help text to this flag to make it obvious what it does
  - Do not inspect blob filename to see if it looks like a digest
  - This allows for arbitrary algorithms instead of just `sha256` (for @TomHennen's case, he would like to verify git commit shas)
- You will need to modify the [logic here](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify.go#L436-L442) to make the blob filename not required if the new flag is provided.

haydentherapper:
Agreed with @codysoyland on the suggested changes. You'll also need to update https://github.com/sigstore/cosign/blob/accc80a840524890790c5c368fee26e22ec60bda/pkg/cosign/verifiers.go#L79-L87 to remove the assumption the digest is `sha256`.

> Note that subject.name is [not a required field in statements](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md), so if the proposal is to check that value against something it won't work for a number of use cases.

TIL!

> the effect this would have in the future for PQC. The current standardized algorithms follows Ed25519 in that they require the complete message. There are pre-hashed versions defined too, but will they be common/supported? 

Let's discuss more outside of this PR, but the short answer is that for now, I'd say it's on the verifier to know if providing only a digest is possible for the given algorithm. This is meant to be an optimization, and the common use case is still to provide the blob.

PR #3967, opened 2024-12-12
Revert "Fix formatting of CODEOWNERS"

haydentherapper:
Reverts sigstore/cosign#3957

One of Sigstore's admins helps maintain a number of repos without explicitly being a codeowner. It seems like almost every repo across the Sigstore org has this same syntax error, which means effectively we haven't been using codeowners files for approvals. I have a few thoughts on ideas to fix this, but for now, I'm reverting this change so the admin can continue merging PRs.

A few ideas for fixes:

* Remove requiring codeowners approvals (each repo should be making that decision, and it's not ideal for the lack of notifications)
* Remove push restrictions (-1 on this, this mandates those with maintain can merge, but not those with write, which lets us implement review vs merge permissions)
* Allow admins to bypass branch protection rules (-1 on this, insider risk mitigation)

PR #4030, opened 2025-01-31
Fix CODEOWNERS syntax

codysoyland:
#### Summary

Each line should start with a file pattern. See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners


codysoyland:
> We actually had to revert this when this was last applied because it broke Carlos’s ability to approve PRs. #3967 has more info.

Gotcha, I thought I had fixed this, so this is good to know.

Issue #4019, opened 2025-01-21
Support signing 'raw' in-toto statements


TomHennen:
**Description**

I'd like the ability to sign an [in-toto Statement](https://github.com/in-toto/attestation/tree/main/spec#attestation-spec:~:text=Statement%3A%20Binds%20the%20attestation%20to%20a%20particular%20subject%20and%20unambiguously%20identifies%20the%20types%20of%20the%20predicate.) that I generate with other tooling with Sigstore and then later use Sigstore to verify the Signature on the attestation while leaving evaluation of the contents of the attestation to other custom tooling.

cosign currently supports signing arbitrary [in-toto _predicates_](https://github.com/in-toto/attestation/tree/main/spec#attestation-spec:~:text=Predicate%3A%20Contains%20arbitrary%20metadata%20about%20a%20subject%20artifact%2C%20with%20a%20type%2Dspecific%20schema.) via `cosign attest-blob --predicate <FILE> --type <TYPE> -hash <HASH>  ...` ([docs](https://github.com/sigstore/cosign/blob/main/doc/cosign_attest-blob.md)), but this does not work at the _Statement_ layer.

That works well if the user doesn't need any control over the `subject` field beyond specifying the `sha256` digest.  However there are some cases (like signing [the verification summary for the SLSA Source track](https://slsa.dev/spec/draft/source-requirements#summary-attestation)), where the ability to use custom digests in the subject and to set `annotations` on subjects is required.

FWIW this capability [has also been requested](https://github.com/sigstore/gitsign/issues/611) in gitsign but would be specific to git repos and would not support non-git usage.  (There's also been a question raised of if it would be helpful to have this in a tool that is currently supported or not CC @steiza).

A nice workflow might look something like:

Generate
```
# Outputs an unsigned vuln predicate like https://github.com/in-toto/attestation/blob/main/spec/predicates/vulns_02.md
$ vuln_scan_repo github.com/foo/bar --output vulns.intoto.json.unsigned
Scanned COMMIT abc123
# Signs the statement using Sigstore, storing the results in a bundle for later use
$ cosign attest-blob --statement vulns.intoto.json.unsigned --bundle vulns.intoto.json
```

Use
```
# Check the signature...
$ cosign verify-blob-attestation --statement-only --bundle vulns.intoto.json --output-file vulns.intoto.json.unsigned
# Process the results, failing if there are any critical vulns
$ vuln_scan_check --no-critical vulns.intoto.json.unsigned
[...]

PR #4032, opened 2025-02-03
Add support for attesting with complete statement

codysoyland:
#### Summary

Add support for signing attestations given a complete statement. This is an alternative to passing in a `predicate` and having cosign generate the statement.

This adds a new flag to `attest` and `attest-blob`, `--statement`, which accepts a complete in-toto statement. If provided, the predicate flag is ignored and the statement is signed.

Fixes https://github.com/sigstore/cosign/issues/4019


codysoyland:
We may go another direction with this, see https://github.com/sigstore/cosign/issues/4019#issuecomment-2632439701

steiza:
Thank you @codysoyland for getting the ball rolling on this! I've rolled your changes forward (and added verification support) in #4306.

PR #4047, opened 2025-02-05
Enable fetching signatures without remote get.

janisz:
**Summary**
When using cosign as a dependency and attempting to fetch signatures for a specific signed entity, the signed entity is always fetched from the registry.

In case the signed entity has been fetched previously, it'd be nice to introduce a function similar to the `FetchAttestations` that allows to fetch signatures from the signed entity directly, without reaching out to the registry to fetch the signed entity beforehand.

With the current changes in the PR, the error will now not include the reference anymore. However, the changes were also made to `FetchAttestations` in the same way.
If we want to include this information, we may wrap the returned error from `FetchAttestations`/`FetchSignatures` with the reference, so the content of the error will stay the same.

**Release Note**
NONE

**Refs:** 
- https://github.com/sigstore/cosign/pull/3007

### DIFF
--- a/pkg/cosign/fetch.go
+++ b/pkg/cosign/fetch.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"runtime"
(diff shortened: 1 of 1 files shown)


janisz:
@dmitris Could you take a look?

bkabrda (approved):
LGTM. The change seems to be consistent with what the attestation code is doing; there seems to be no change in behavior other than a minor formatting of the error output, but that shouldn't be an issue.

haydentherapper (approved):
Thanks!

Issue #4048, opened 2025-02-06
Cosign copy command requires "--only" since 2.4.2


MPritsch:
**Description**

Since Version 2.4.2 there is an error when using the "cosign copy" command:
```
Error: invalid value for --only: , only following values are supported, map[att:{} sbom:{} sig:{}]
```

To reproduce. Not that we first checked the functionality with version 2.4.1:

```
docker run --rm -it  --entrypoint sh cgr.dev/chainguard/wolfi-base                                                                                                                     
/ # apk update
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
 [https://packages.wolfi.dev/os]
OK: 124859 distinct packages available

/ # apk add cosign=2.4.1-r0
(1/1) Installing cosign (2.4.1-r0)
Executing busybox-1.37.0-r0.trigger
OK: 83 MiB in 16 packages

/ # cosign copy -f alpine ttl.sh/asdofjweofjwef
Copying index.docker.io/library/alpine@sha256:56fa17d2a7e7f168a043a2712e63aed1f8543aeafdcee47c58dcffe38ed51099 to ttl.sh/asdofjweofjwef:sha256-56fa17d2a7e7f168a043a2712e63aed1f8543aeafdcee47c58dcffe38ed51099...
...

/ # apk add cosign=2.4.2-r0
(1/1) Upgrading cosign (2.4.1-r0 -> 2.4.2-r0)
Executing busybox-1.37.0-r0.trigger
OK: 98 MiB in 16 packages

/ # cosign copy -f alpine ttl.sh/asdofjweofjwef
Error: invalid value for --only: , only following values are supported, map[att:{} sbom:{} sig:{}]
error during command execution: invalid value for --only: , only following values are supported, map[att:{} sbom:{} sig:{}]
```

**Version**

2.4.2

MPritsch:
I suspect this change is responsible: https://github.com/sigstore/cosign/pull/3904

bobcallaway:
#4049 should fix this; you can workaround this in the meantime by passing an explicit value for `--only` (e.g. `--only=sig`)

Muaytie666:
EIPS

bobcallaway:
> Can cosign please make a new release (presumably 2.4.3) with this fix?
> 
> This issue is very troublesome and would be fantastic to see addressed sooner rather than later.

There's a fix already at main and a clear workaround, can you clarify why neither works for you?

candrews:
[...]

PR #4049, opened 2025-02-06
fix parsing error in --only for cosign copy

bobcallaway:
If the `--only` option was omitted, the default tagSet became `[]string{""}` which is strictly not `[]string{}`... and parsing fails.

This switches the command line arg over to `StringSliceVar` and adds unit tests.

Fixes: #4048


haydentherapper (approved):

whou-v (commented):
Help

PR #4050, opened 2025-02-06
Add support for passing different SignerVerifier LoadOptions

ret2libc:
#### Summary
This is a preparation PR for https://github.com/sigstore/cosign/issues/3271 and it splits some changes required to completely add support for ed25519ph, with the idea of helping the reviewers. It was split from https://github.com/sigstore/cosign/pull/3479 .

#### Release Note
- Adds functions `SignerFromKeyOptsWithSVOpts`, `LoadPrivateKeyWithOpts`, `ValidateAndUnpackCertWithOpts`, `VerifyImageSignaturesWithOpts`, `VerifyLocalImageSignaturesWithOpts`, `VerifyBlobSignatureWithOpts`, `VerifyImageSignatureWithOpts`, `VerifyImageAttestationsWithOpts`, `VerifyLocalImageAttestationsWithOpts`, `VerifyBlobAttestationWithOpts`, `VerifyImageAttestationWithOpts`, `VerifierForKeyRefWithOpts`, `LoadPublicKeyRawWithOpts`, `SignerVerifierFromKeyRefWithOpts`, `PublicKeyFromKeyRefWithOpts`.

#### Documentation
With recent sigstore, we introduced `signature.LoadVerifierWithOpts` (and other versions for `Signer`, `SignerVerifier`). Adding support for ed25519ph would require passing some options around, however cosign is still using `signature.LoadVerifier`. This PR prepares the cosign codebase to accept `LoadOptions` and pass them around, without actually changing (yet) any behaviour.

Doing that required adding a lot of `*WithOpts` functions to avoid breaking other users who may rely on those exported functions. Please let me know if that's a good approach.

Once this is done, we can work on a smaller PR that actually adds supports for ed25519ph and pass the correct extra `LoadOptions` around.

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -18,7 +18,6 @@ package sign
 import (
 	"bytes"
 	"context"
-	"crypto"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
@@ -138,7 +137,7 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 	ctx, cancel := context.WithTimeout(context.Background(), ro.Timeout)
 	defer cancel()
 
-	sv, err := SignerFromKeyOpts(ctx, signOpts.Cert, signOpts.CertChain, ko)
+	sv, err := SignerFromKeyOptsWithSVOpts(ctx, signOpts.Cert, signOpts.CertChain, ko)
 	if err != nil {
 		return fmt.Errorf("getting signer: %w", err)
 	}
@@ -391,8 +390,8 @@ func signerFromSecurityKey(ctx context.Context, keySlot string) (*SignerVerifier
 	}, nil
 }
 
-func signerFromKeyRef(ctx context.Context, certPath, certChainPath, keyRef string, passFunc cosign.PassFunc) (*SignerVerifier, error) {
-	k, err := sigs.SignerVerifierFromKeyRef(ctx, keyRef, passFunc)
+func signerFromKeyRef(ctx context.Context, certPath, certChainPath, keyRef string, passFunc cosign.PassFunc, opts ...signature.LoadOption) (*SignerVerifier, error) {
+	k, err := sigs.SignerVerifierFromKeyRefWithOpts(ctx, keyRef, passFunc, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("reading key: %w", err)
 	}
@@ -521,12 +520,12 @@ func signerFromKeyRef(ctx context.Context, certPath, certChainPath, keyRef strin
 	return certSigner, nil
 }
 
-func signerFromNewKey() (*SignerVerifier, error) {
+func signerFromNewKey(opts ...signature.LoadOption) (*SignerVerifier, error) {
 	privKey, err := cosign.GeneratePrivateKey()
 	if err != nil {
 		return nil, fmt.Errorf("generating cert: %w", err)
 	}
-	sv, err := signature.LoadECDSASignerVerifier(privKey, crypto.SHA256)
+	sv, err := signature.LoadSignerVerifierWithOpts(privKey, opts...)
 	if err != nil {
 		return nil, err
 	}
@@ -559,7 +558,7 @@ func keylessSigner(ctx context.Context, ko options.KeyOpts, sv *SignerVerifier) 
 	}, nil
 }
 
-func SignerFromKeyOpts(ctx context.Context, certPath string, certChainPath string, ko options.KeyOpts) (*SignerVerifier, error) {
+func SignerFromKeyOptsWithSVOpts(ctx context.Context, certPath string, certChainPath string, ko options.KeyOpts, svOpts ...signature.LoadOption) (*SignerVerifier, error) {
 	var sv *SignerVerifier
 	var err error
 	genKey := false
@@ -567,11 +566,11 @@ func SignerFromKeyOpts(ctx context.Context, certPath string, certChainPath strin
 	case ko.Sk:
 		sv, err = signerFromSecurityKey(ctx, ko.Slot)
 	case ko.KeyRef != "":
-		sv, err = signerFromKeyRef(ctx, certPath, certChainPath, ko.KeyRef, ko.PassFunc)
+		sv, err = signerFromKeyRef(ctx, certPath, certChainPath, ko.KeyRef, ko.PassFunc, svOpts...)
 	default:
 		genKey = true
 		ui.Infof(ctx, "Generating ephemeral keys...")
-		sv, err = signerFromNewKey()
+		sv, err = signerFromNewKey(svOpts...)
 	}
 	if err != nil {
 		return nil, err
@@ -582,6 +581,10 @@ func SignerFromKeyOpts(ctx context.Context, certPath string, certChainPath strin
 	}
 
 	return sv, nil
+}
+
+func SignerFromKeyOpts(ctx context.Context, certPath string, certChainPath string, ko options.KeyOpts) (*SignerVerifier, error) {
+	return SignerFromKeyOptsWithSVOpts(ctx, certPath, certChainPath, ko)
 }
 
 type SignerVerifier struct {
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -65,7 +65,7 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 		return nil, err
 	}
 
-	sv, err := SignerFromKeyOpts(ctx, "", "", ko)
+	sv, err := SignerFromKeyOptsWithSVOpts(ctx, "", "", ko)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -132,7 +132,7 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	// each access.
 	ref = digest // nolint
 
-	sv, err := sign.SignerFromKeyOpts(ctx, c.CertPath, c.CertChainPath, c.KeyOpts)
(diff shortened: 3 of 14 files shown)


ret2libc:
cc @haydentherapper

ret2libc:
I'm going to add a few tests to this.

haydentherapper:
What do you think about limiting this to code under the `--new-bundle-format` flag, where [VerifyBundle](https://github.com/sigstore/cosign/blob/51dd1ce9665172c6b3c62f17b1a9a9be9266ef17/pkg/cosign/verify_bundle.go#L25) is called? Then most changes only need to occur in sigstore-go, we won't have to be concerned about unexpectedly breakages for current users, and we know the code will carry on to cosign 3.x.

ret2libc:
> What do you think about limiting this to code under the `--new-bundle-format` flag, where [VerifyBundle](https://github.com/sigstore/cosign/blob/51dd1ce9665172c6b3c62f17b1a9a9be9266ef17/pkg/cosign/verify_bundle.go#L25) is called? Then most changes only need to occur in sigstore-go, we won't have to be concerned about unexpectedly breakages for current users, and we know the code will carry on to cosign 3.x.

Sounds good, but we still want to modify the "sign"/"sign-blob" flow, don't we? At least to test the signing process through cosign and then the verification process in the new bundle format, correct?

haydentherapper (commented):
Reviewed this for the Fulcio PR. This LGTM and I realize my comment about `sign-blob` isn't immediately relevant as we need to make all these changes first.

https://github.com/sigstore/cosign/blob/8dd9e9b72406bcf501fe29e21b28b85d46f2a267/pkg/cosign/tlog.go#L240 needs to be updated as well.

haydentherapper:
If we do the Fulcio change to accept ecdsa-p384-sha-384 and ecdsa-p521-sha-512, I'd add an end-to-end test that verifies signing and verification with these keys. You can generate a key, import it into the Cosign key format, then sign with `IssueCertificate:true`.

ret2libc:
@haydentherapper could you approve the workflows please?

cmd/cosign/cli/sign/sign.go
```diff
@@ -536,18 +536,45 @@ func signerFromNewKey() (*SignerVerifier, error) {
 	}, nil
 }
 
+// adaptSignerVerifierToFulcio adapts, if necessary, the SignerVerifier to be
+// used to interact with Fulcio.
+//
+// This is needed in particular for ED25519 keys with the pre-hashed version of
+// the algorithm, which is not supported by Fulcio. This function creates a
+// ED25519 SignerVerifier based on that instead.
+func adaptSignerVerifierToFulcio(sv *SignerVerifier) (*SignerVerifier, error) {
```
haydentherapper, line 545: re: our offline chat about conditioning this on Rekor - it it technically supported to request a certificate for a key without writing to Rekor (using the `--issue-certificate` flag). I'm going to guess that both using this flag and using an ed25519 key affects almost no users.

If there is an easy way to condition using ed25519ph on whether Rekor will be used, and specifically for hashedrekord and not dsse (so using sign/sign-blob and not attest/attest-blob), let's do that. Otherwise, I think the impact of a breakage would be very minimal.

cmd/cosign/cli/fulcio/fulcio.go
```diff
@@ -113,6 +113,10 @@ type Signer struct {
 }
 
 func NewSigner(ctx context.Context, ko options.KeyOpts, signer signature.SignerVerifier) (*Signer, error) {
+	return NewSignerWithAdapter(ctx, ko, signer, signer)
+}
+
+func NewSignerWithAdapter(ctx context.Context, ko options.KeyOpts, signer signature.SignerVerifier, fulcioSigner signature.SignerVerifier) (*Signer, error) {
```
haydentherapper, line 119: can you add a func comment for why we have two signers now?

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -263,3 +270,31 @@ func extractCertificate(ctx context.Context, sv *SignerVerifier) ([]byte, error)
...
+
+func getHashFunction(sv *SignerVerifier) (crypto.Hash, error) {
+	pubKey, err := sv.PublicKey()
+	if err != nil {
+		return crypto.Hash(0), fmt.Errorf("error getting public key: %w", err)
+	}
+
+	// TODO: Ideally the SignerVerifier should have a method to get the hash function
+	// Cosign uses ED25519ph by default for ED25519 keys
+	algo, err := signature.GetDefaultAlgorithmDetails(pubKey, signatureoptions.WithED25519ph())
```
haydentherapper, line 282: this would be another place to condition the PH variant on if we're uploading to rekor

haydentherapper (commented):
Overall LGTM

cmd/cosign/cli/sign/sign.go
ret2libc: the sign-blob path already has a `shouldUpload` check. I'm planning on extending `KeyOpts` with 
```go
	// DefaultLoadOptions may be set to control the behaviour of
	// `LoadDefaultSigner/Verifier` family of functions. Some public/private key
	// types have ambiguities with regards to the signing algorithm to use (e.g.
	// RSA can be RSASSA-PSS or RSASSA-PKCS1v15). This is a way to control that.
	//
	// By default, Ed25519ph is used for ed25519 keys and RSA-PKCS1v15 is used
	// for RSA keys.
	DefaultLoadOptions *[]signature.LoadOption
```

then we can do:
```go
	if !shouldUpload {
		// To maintain backwards compatibility with older cosign versions,
		// we do not use ed25519ph for ed25519 keys when the signatures are not
		// uploaded to the Tlog.
		ko.DefaultLoadOptions = &[]signature.LoadOption{}
	}
```

now i just have to modify a few methods to pass down these `DefaultLoadOptions` into `cosign.LoadPrivateKey`. WDYT?

haydentherapper: SGTM!

cmd/cosign/cli/options/key.go
```diff
@@ -53,4 +56,13 @@ type KeyOpts struct {
...
 	InsecureSkipFulcioVerify bool
+
+	// DefaultLoadOptions may be set to control the behaviour of
+	// `LoadDefaultSigner/Verifier` family of functions. Some public/private key
+	// types have ambiguities with regards to the signing algorithm to use (e.g.
+	// RSA can be RSASSA-PSS or RSASSA-PKCS1v15). This is a way to control that.
+	//
+	// By default, Ed25519ph is used for ed25519 keys and RSA-PKCS1v15 is used
+	// for RSA keys.
+	DefaultLoadOptions *[]signature.LoadOption
```
haydentherapper, line 67: I think I’m missing something, but where is this option set?

haydentherapper (commented):
This looks great! LGTM

cmd/cosign/cli/options/key.go
ret2libc: In `sign_blob.go`:
```golang
	shouldUpload, err := ShouldUploadToTlog(ctx, ko, nil, tlogUpload)
	if err != nil {
		return nil, fmt.Errorf("upload to tlog: %w", err)
	}

	if !shouldUpload {
		// To maintain backwards compatibility with older cosign versions,
		// we do not use ed25519ph for ed25519 keys when the signatures are not
		// uploaded to the Tlog.
		ko.DefaultLoadOptions = &[]signature.LoadOption{}
	}

```

ret2libc:
@haydentherapper i've added a confirmation message in https://github.com/sigstore/cosign/pull/4050/commits/b4ea782297ecdf3e6128b599898ba065caf99816 related to your comment in https://github.com/sigstore/cosign/pull/3497#issuecomment-2814152197 . Does it make sense to you?

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -110,6 +112,36 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 	}
 	defer sv.Close()
 
+	hashFunction, err := getHashFunction(sv, ko.DefaultLoadOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	if hashFunction != crypto.SHA256 && !ko.NewBundleFormat && (shouldUpload || (!ko.Sk && ko.KeyRef == "")) {
+		ui.Infof(ctx, "Non SHA256 hash function is not supported for old bundle format. Use --new-bundle-format to use the new bundle format or use different signing key/algorithm.")
```
Hayden-IO, line 142: Nice, looks good!

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -110,6 +112,36 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
...
+		if !ko.SkipConfirmation {
+			if err := ui.ConfirmContinue(ctx); err != nil {
+				return nil, err
+			}
+		}
+		ui.Infof(ctx, "Continuing with non SHA256 hash function and old bundle format")
+	}
+
+	if payloadPath == "-" {
```
Hayden-IO: The e2e test failure on line 91 is due to moving payload initialization to later.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -86,6 +100,15 @@ func SignBlobCmd(ro *options.RootOptions, ko options.KeyOpts, payloadPath string
 			ClientSecret:     ko.OIDCClientSecret,
 			RedirectURL:      ko.OIDCRedirectURL,
 		})
+		if err != nil {
+			return nil, fmt.Errorf("retrieving ID token: %w", err)
+		}
+
+		payload, closePayload, err := getPayload(ctx, payloadPath, protoHashAlgoToHash(keypair.GetHashAlgorithm()))
```
Hayden-IO, line 107: This LGTM - In https://github.com/sigstore/cosign/pull/4368/files#diff-b3e868f876ca97e010ac2714a8fb988e76b89c44fbb226ee3f3b18d88b139500R70, when self-managed keys are provided, I'm only allowing the default algorithm mapping.

ret2libc: I've commented there, but we should use `defaultLoadOptions = cosign.GetDefaultLoadOptions()` to get the right options

PR #4062, opened 2025-02-17
sort Go imports

dmitris:
#### Summary
The current order of the Go imports in the file with large blocks of such imports is "all over the places" and hard to explain but clearly it is not a sorted one.
I propose to keep all the third-party (non-stdlib) imports in just one block to make it easier for the IDEs to properly sort those imports on addition or deletion.

#### Release Note
NONE

#### Documentation
n/a


cpanato (approved):
seems ok

i was planning to do similar

thanks

PR #4077, opened 2025-02-21
Fix replace with compliant image mediatype

### LINKED ISSUE #4076: Wrong mediatype on attestation replace
**Description**

Using replace option on `cosing attest`, an attestation with wrong mediatype is created, namely mediaType `application/vnd.docker.distribution.manifest.v2+json` instead of `application/vnd.oci.image.manifest.v1+json`. This results in tool like skopeo throwing an error:

```
FATA[0001] Error parsing manifest for image: unsupported docker v2s2 media type: "application/vnd.oci.image.manifest.v1+json"
```

**Version**

Latest and greatest

stgrace:
Fixes #4076 

#### Summary

Bug in replacing attestation.

#### Release Note

NONE

#### Documentation

No docs

### DIFF
--- a/pkg/oci/mutate/signatures.go
+++ b/pkg/oci/mutate/signatures.go
@@ -17,10 +17,10 @@ package mutate
 
 import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
-	"github.com/google/go-containerregistry/pkg/v1/empty"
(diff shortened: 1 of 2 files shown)


bobcallaway:
Could you please write a test that ensures we don’t regress this in the future?

stgrace:
Hi @bobcallaway 

Added a test for this specific issue. Let me know if you require any changes

PR #4091, opened 2025-02-28
fix: increase e2e_test docker compose timeout to 180s

ramonpetgrave64:
#### Summary

When running `./test/e2e_test.sh` on a personal machine, it can a lot longer than 60 seconds for the test containers to be ready. This PR increases the timeout to 3 minutes. In the CI, the containers seem to be ready at between 20 to 30 seconds, but increasing to 3 minutes I think should be fine.

Example invocation:

```
...
+ for repo in rekor fulcio
+ pushd rekor
...
+ '[' 12 -eq 18 ']'
...
+ '[' 3 == 3 ']'
...
+ for repo in rekor fulcio
+ pushd fulcio
...
+ '[' 7 -eq 18 ']'
...
+ '[' 3 == 3 ']'
...
+ echo 'running tests'
...
```

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -73,7 +73,7 @@ for repo in rekor fulcio; do
     count=0
     until [ $(${docker_compose} ps | grep -c "(healthy)") == 3 ];
     do
-        if [ $count -eq 6 ]; then
+        if [ $count -eq 18 ]; then
            echo "! timeout reached"
            exit 1
         else


haydentherapper:
@ramonpetgrave64 Did you investigate why it was taking awhile to start up? This seems like a regression if we need to bump the timeout so significantly.

PR #4097, opened 2025-03-03
enhancement: use jq to check for test container readiness

ramonpetgrave64:

#### Summary

Improves the way the e2e_test.sh checks for container readiness with jq.

Currently, there are exactly 3 healthcheck probes in each of fulcio and rekor's docker-compose.ymls. In the future, we should consider adding more. This PR is in preparation for that future.

The jq expression checks that all of the `.Health` properties are either 
 
- `"healthy"`, when a container does have a healthcheck defined
- `""`, when a container does not have a healthcheck defined

The script still waits until the containers are not `"unhealthy"`

## Testing process

Running e2e_test.sh locally

```log
...
++ docker compose ps --format '{{json .}}'
++ jq -s -e '.[] | .Health == "healthy" or .Health == ""'
WARN[0000] /usr/local/google/home/rpetgrave/rekor/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
+ false true true true true
...
++ docker compose ps --format '{{json .}}'
++ jq -s -e '.[] | .Health == "healthy" or .Health == ""'
WARN[0000] /usr/local/google/home/rpetgrave/rekor/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
+ true true true true true
...
+ echo 'running tests'
running tests
...
```

In the log, we see `+ false true true true true` when some containers are still not yet ready, and `+ true true true true true` when all of the containers are ready.

#### Release Note

Improve container readiness checks for e2e tests.

#### Documentation

None

### DIFF
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -71,7 +71,7 @@ for repo in rekor fulcio; do
     ${docker_compose} up -d
     echo -n "waiting up to 60 sec for system to start"
     count=0
-    until [ $(${docker_compose} ps | grep -c "(healthy)") == 3 ];
+    until $(docker compose ps --format '{{json .}}' | jq -s -e '.[] | .Health == "healthy" or .Health == ""');
     do
         if [ $count -eq 18 ]; then
            echo "! timeout reached"
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -71,7 +71,7 @@ for repo in rekor fulcio; do
     ${docker_compose} up -d
     echo -n "waiting up to 60 sec for system to start"
     count=0
-    until $(docker compose ps --format '{{json .}}' | jq -s -e '.[] | .Health == "healthy" or .Health == ""');
+    until $(${docker_compose} ps --format '{{json .}}' | jq -s -e '.[] | .Health == "healthy" or .Health == ""');
     do
         if [ $count -eq 18 ]; then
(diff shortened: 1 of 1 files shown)


test/e2e_test.sh
```diff
@@ -71,7 +71,7 @@ for repo in rekor fulcio; do
     ${docker_compose} up -d
     echo -n "waiting up to 60 sec for system to start"
     count=0
-    until [ $(${docker_compose} ps | grep -c "(healthy)") == 3 ];
+    until $(${docker_compose} ps --format '{{json .}}' | jq -s -e '.[] | .Health == "healthy" or .Health == ""');
```
haydentherapper, line 74: It looks like checking for an empty string doesn't only mean the healthcheck is undefined. I just tested this command locally with a failing deployment (docker compose up for Fulcio where the database and Trillian fails to start). The containers are stuck in a loop restarting, not showing as unhealthy, with `.Health == ""`.

haydentherapper (changes requested):

test/e2e_test.sh
ramonpetgrave64: Looking into this, it may be better to only check for the readiness, or health, of rekor-server and fulcio-server.

Some related work https://github.com/sigstore/rekor/pull/2397

PR #4117, opened 2025-03-17
feat: use pending scaffolding portable Makefile in e2e test script

ramonpetgrave64:

#### Summary

pending: https://github.com/sigstore/scaffolding/pull/1517

- Uses the new Makefile to startup fulcio and rekor containers. 
  - we could use the Action, but we still want to be able to do all this locally, so we use the Makefile directly.
- attaches the fakeoidc container to the fulcio network by the system's hostname, rather than ip.

## Testing

```
./test/e2e_test.sh
```

The github action worklfow "CI-Tests"'s job "e2e-tests" uses this script and should pass.


jku:
Can you describe the planned end state of this PR? Would this  use the action from scaffolding?

ramonpetgrave64:
redoing in https://github.com/sigstore/cosign/pull/4161

PR #4159, opened 2025-04-10
Added LoadFileOrURLUsingReader

### LINKED ISSUE #4112: verify-blob laods entire payload into memory
**Description**

Currently, when calling cosign’s verify-blob subcommand, it will actually load the entire message into memory first, before passing it to the `SignerVerifier.VerifySignature()` method. This means that verifying a 1GB file will use an additional 1GB of memory. This is likely undesirable, so this issue is requesting that it instead pass along an `io.Reader` of the file, so it can be streamed, instead of pre-loaded.

- https://github.com/sigstore/cosign/blob/647eca1376ea75b3af88cd65f8f35f9e37ecd671/cmd/cosign/cli/verify/verify_blob.go#L181

krisharyan117:
Closes #4112 
#### Release Note

Added a new function that returns io.Reader instead of the entire message called LoadFileOrURLUsingReader

### DIFF
--- a/cmd/cosign/cli/verify/verify_blob.go
+++ b/cmd/cosign/cli/verify/verify_blob.go
@@ -374,7 +374,18 @@ func payloadBytes(blobRef string) ([]byte, error) {
 	if blobRef == "-" {
 		blobBytes, err = io.ReadAll(os.Stdin)
 	} else {
-		blobBytes, err = blob.LoadFileOrURL(blobRef)
+		reader, err := blob.LoadFileOrURLUsingReader(blobRef)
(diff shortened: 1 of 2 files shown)


krisharyan117:
@ramonpetgrave64 please review this

krisharyan117:
Yes but I am facing  difficulty passing io.Reader in signature, err := static.NewSignature(blobBytes, sig, opts...). 

Can you give me a direction on how should I approach this?

Hayden-IO:
Closing for now, feel free to discuss solutions on the issue.

PR #4161, opened 2025-04-10
feat: use test containers action from scaffolding

ramonpetgrave64:
#### Summary

pending https://github.com/sigstore/scaffolding/pull/1544

 - using dev branch ramonpetgrave64/scaffolding/actions/test_containers@test-containers

Uses the new test_containers action to launch containers for testing purposes.

Some of this repo's code has been moved to scaffolding.

#### Testing

pre-submit workflows pass.

#### Followups

- https://github.com/sigstore/scaffolding/issues/1555


test/e2e_tsa_test.go
```diff
@@ -175,22 +177,29 @@ func generateMTLSKeys(t *testing.T, td string) (string, string, string, string,
...
-	viper.Set("tls-ca", timestampCACert)
-	viper.Set("tls-key", timestampServerKey)
-	viper.Set("tls-certificate", timestampServerCert)
-	tsaAPIServer := tsaserver.NewRestAPIServer("localhost", 3000, []string{"https"}, false, 10*time.Second, 10*time.Second)
-	tsaServer := httptest.NewServer(tsaAPIServer.GetHandler())
-	tsaClient, err := tsaclient.GetTimestampClient(tsaServer.URL)
+	close := func() {}
+	tsaURLPrefix := os.Getenv("TSA_URL")
+	if tsaURLPrefix == "" { // use tsa as a library instead.
+		slog.Info("Using TSA as a library")
```
ramonpetgrave64, line 187: working here https://github.com/sigstore/cosign/actions/runs/14392419854/job/40361993996?pr=4161#step:5:6890

test/helpers.go
```diff
@@ -503,12 +504,12 @@ func registryClientOpts(ctx context.Context) []remote.Option {
 // setLocalEnv sets SIGSTORE_CT_LOG_PUBLIC_KEY_FILE, SIGSTORE_ROOT_FILE, and SIGSTORE_REKOR_PUBLIC_KEY for the locally running sigstore deployment.
 func setLocalEnv(t *testing.T, dir string) error {
 	// fulcio repo is downloaded to the user's home directory by e2e_test.sh
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return fmt.Errorf("error getting home directory: %w", err)
+	home := os.Getenv(cloneDirEnvKey) //nolint:forbidigo
+	if home == "" {
+		t.Fatalf("clone directory env key not set: %s", cloneDirEnvKey)
 	}
```
jku, line 510: I think it would make a lot of sense to just embed the fulcio/config/ctfepubkey.pem test key here in the source file

.github/workflows/golangci-lint.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@1481404843c368bc19ca9406f87d6e0fc97bdcfd # v7.0.0
         with:
-          version: v2.0
+          version: v2.1.2
```
haydentherapper, line 44: Can be `v2.1` which should pull the latest patch release for a given minor version

.github/workflows/e2e-tests.yml
```diff
@@ -47,6 +48,11 @@ jobs:
           go-version-file: 'go.mod'
           check-latest: true
 
+      # docker/setup-docker doesn't worth with macos-13 and above runners.
+      - if: matrix.with-containers && runner.os != 'macOS'
```
haydentherapper, line 52: Do we need a container set up here? The e2e that run here do not currently use Kind since the [e2e tests don't run when `cross` is included](https://github.com/sigstore/cosign/blob/d6c40c940fb04cd62156a69b32f46df5a1012bfa/test/e2e_test.go#L16)

.github/workflows/tests.yaml
```diff
@@ -164,8 +164,15 @@ jobs:
           go install sigs.k8s.io/kind@v0.20.0
           kind create cluster
```
haydentherapper, line 166: Can we remove Kind?

test/helpers.go
haydentherapper: +1, if this is just to access the CT pub key, let's just bake it in, it'll simplify the test setup here and in e2e_test.go

test/e2e_tsa_test.go
```diff
@@ -175,22 +177,29 @@ func generateMTLSKeys(t *testing.T, td string) (string, string, string, string,
...
-	viper.Set("tls-port", 3000)
-	viper.Set("tls-ca", timestampCACert)
-	viper.Set("tls-key", timestampServerKey)
-	viper.Set("tls-certificate", timestampServerCert)
-	tsaAPIServer := tsaserver.NewRestAPIServer("localhost", 3000, []string{"https"}, false, 10*time.Second, 10*time.Second)
-	tsaServer := httptest.NewServer(tsaAPIServer.GetHandler())
-	tsaClient, err := tsaclient.GetTimestampClient(tsaServer.URL)
+	close := func() {}
+	tsaURLPrefix := os.Getenv("TSA_URL")
+	if tsaURLPrefix == "" { // use tsa as a library instead.
```
haydentherapper, line 186: Do we need this or can we simplify and assume the TSA is running?

test/e2e_test.sh
```diff
@@ -14,97 +14,18 @@
...
 
-set -ex
-
-docker_compose="docker compose"
-if ! ${docker_compose} version >/dev/null 2>&1; then
-    docker_compose="docker-compose"
-fi
-
-echo "setting up OIDC provider"
-pushd ./test/fakeoidc
```
haydentherapper, line 25: Also can remove the fakeoidc module

.github/workflows/e2e-tests.yml
ramonpetgrave64: I think this is the only tests that uses a TSA server.

.github/workflows/tests.yaml
ramonpetgrave64: There's a test that specifically uses the k8s api.

 - https://github.com/sigstore/cosign/blob/d6c40c940fb04cd62156a69b32f46df5a1012bfa/test/e2e_test.go#L2014

test/e2e_tsa_test.go
ramonpetgrave64: No, because currently the Action doesn't work on macOS, and these tests are intended to be cross-os

 - https://github.com/sigstore/cosign/pull/4161/files/1ff5abf60eccda2cca517ab65e56a9342ec0b497

test/helpers.go
ramonpetgrave64: embedded

test/e2e_test.sh
ramonpetgrave64: removed.

.github/workflows/e2e-tests.yml
haydentherapper: I'm not following why we need the action here, where is the TSA server needed?

test/e2e_tsa_test.go
haydentherapper: Can this be tracked as a future improvement then? Would be nice to clean this up.

ramonpetgrave64: tracking in https://github.com/sigstore/scaffolding/issues/1555

.github/workflows/tests.yaml
ramonpetgrave64: Added the comment. What do you mean by splitting into its own test?

.github/workflows/e2e-tests.yml
ramonpetgrave64: Needed in test/e2e_tsa_test.go

.github/workflows/golangci-lint.yml
ramonpetgrave64: changed.

.github/workflows/e2e-tests.yml
haydentherapper: Re-reviewing the change made in test/e2e_tsa_test.go, that's actually changing what it's testing, which is mTLS. Longer term, we can change this test to reuse the TSA spun up by the action, but that will require providing mTLS flags to the server. For now, we'll need to stick with spinning up the TSA via a library.

For e2e_test.go, there's no testing of mTLS so we can use the containerized TSA, but we'll need to refactor a lot of tests to use a common setup function, since each test that uses a TSA spins up a service (like in https://github.com/sigstore/cosign/blob/d6c40c940fb04cd62156a69b32f46df5a1012bfa/test/e2e_test.go#L567-L568). I'm OK with following up later to fix that.

.github/workflows/tests.yaml
haydentherapper: Sorry, I meant a comment in the action for why `kind` is needed.

Since `kind` is only needed for one test, it could be in its own e2e test so that it can run in parallel with other tests and improve overall runtime. We can do that later on, there's a few places we can clean up testing in cosign.

test/e2e_test.go
```diff
@@ -557,8 +557,14 @@ func downloadTSACerts(downloadDirectory string, tsaServer string) (string, strin
...
 func TestSignVerifyWithTUFMirror(t *testing.T) {
-	home, err := os.UserHomeDir() // fulcio repo was downloaded to $HOME in e2e_test.sh
-	must(err, t)
+	dir, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Copied from https://github.com/sigstore/fulcio/blob/273116884c6e247a98cb28a9acc3e5844c4f9b5c/config/ctfe/pubkey.pem
+	// This must be an abosulte path.
+	fulcioPubkey := filepath.Join(dir, "testdata", "fulcio", "config", "ctfe", "pubkey.pem")
```
haydentherapper, line 566: Can we use https://pkg.go.dev/embed, like:

```
//go:embed testdata/ctfe/pubkey.pem
var ctPubKey string
```

haydentherapper, line 566: Name should be `ct` rather than `fulcio`, since it's the CT log's public key.

ramonpetgrave64: No, because the the `tuf.update()` requires a path string, not a file object.

ramonpetgrave64: changed.

.github/workflows/e2e-tests.yml
ramonpetgrave64: Added to https://github.com/sigstore/scaffolding/issues/1555

haydentherapper: Bumping, the changes in e2e_tsa_test are changing what's being tested to not test mTLS.

haydentherapper (commented):
The issue with the containers starting up might be fixed now, I added healthchecks for Trillian and fixed a healthcheck for MySQL in Rekor.

ramonpetgrave64:
What's your question here? https://github.com/sigstore/cosign/pull/4161#discussion_r2089728739

haydentherapper:
> What's your question here? [#4161 (comment)](https://github.com/sigstore/cosign/pull/4161#discussion_r2089728739)

Sorry, missed this comment. `test/e2e_tsa_test.go` tests mTLS, which requires specific server flags to enable TLS connections. The TSA spun up from `setup-sigstore-env` doesn't set up TLS, so we can't just swap out initializing the TSA via the library for what's spun up in the testing script. To use the TSA from the testing script, we'd need a flag to set up the TSA with TLS (generally, this would be useful for Rekor and Fulcio as well which have TLS flags). Also, given the amount of time to spin up all services, I'd like to first have the testing script support selectively spinning up services so we aren't significantly increasing the runtime of `e2e_tsa_test.go`.

`test/e2e_test.go` doesn't require a TSA with TLS so we can use the TSA spun up from `setup-sigstore-env`, which just requires some refactoring.

test/e2e_test.sh
```diff
@@ -14,97 +14,18 @@
...
-      "IssuerURL": "$OIDC_URL",
-      "ClientID": "sigstore",
-      "Type": "email"
-    }
-  }
-}
-EOF
-popd
+# The service containers needed should first be launched by
+# scaffolding/actions/setup-sigstore-env/run-containers.sh
```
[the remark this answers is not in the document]
ramonpetgrave64: we won't have to because the Action uses `compose up --wait`, and the service healthchecks make sure that they are ready.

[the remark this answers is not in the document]
Hayden-IO: I was meaning when this is run in a local environment, to gracefully fail if I run `test/e2e_test.sh` without having first run the setup script.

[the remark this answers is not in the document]
ramonpetgrave64: added.

.github/workflows/tests.yaml
ramonpetgrave64: I think later. the "kind" step is always ~10 seconds.

ramonpetgrave64:
@haydentherapper Refactored the `e2e_test_tsa.go` a bit to run tow cases: one with the TSA container without mutual TLS, and one with the the TSA library and mutual TLS..

Issue #4098, opened 2025-03-04
cosign doesn't correctly respect individual TSA certificate chains


bkabrda:
**Description**

I have a private Sigstore instance, in which I attempted to rotate the TSA certificate chain. When trying to verify any blob/image after the rotation, I get the following error:

```
$ cosign --verbose verify-blob --certificate-identity=kilgore@kilgore.trout --bundle signed.bundle to-sign --rfc3161-timestamp=timestamp.txt --use-signed-timestamps
Error: unable to load TSA certificates: TSA certificate chain must contain exactly one leaf certificate
main.go:74: error during command execution: unable to load TSA certificates: TSA certificate chain must contain exactly one leaf certificate
```

This is because the `GetTSACerts` function in `pkg/cosign/tsa.go` only expects to parse out a single certificate chain, because of the assumption that these individual chains are provided as individual TUF targets.

However in order to be able to properly rotate the TSA certificate chain, one needs to provide the whole TSA certificate chain as a single TUF target (which is possible and I do it). However, when I attempt to introduce a new TSA chain as a new TUF target and keeping the old one as "Expired", I get the above error when verifying any artifact.

I think the `GetTSACerts` function should behave more like `GetRekorPubs`, which returns multiple keys and all of them are considered valid for verification.

I think this will actually work fine in the future with https://github.com/sigstore/cosign/pull/3844 because of how the [timestamp verification function from sigstore-go](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go#L87) works. I can submit a PR to fix this if you think it's still worth it before https://github.com/sigstore/cosign/pull/3844 gets merged - I think it's worth because of private Sigstore deployments that might be slower to upgrade.

**Version**

2.4.1, but applies to latest main branch as well AFAICS

haydentherapper:
A fix would be welcome, though if it’s a significant change, feel free to not and we will fix it indirectly with Sigstore-go.

haydentherapper:
We don't offer any gurantees around API breakages for Cosign so making a breaking change would be fine. I'm OK with either approach.

bkabrda:
As I mentioned in https://github.com/sigstore/cosign/pull/4174, I don't see a way to fix this without a regression; because #3844 was merged, I no longer need this so much, so I'm closing this issue - feel free to reopen if anyone wants to try to address it.

PR #4174, opened 2025-04-24
Respect multiple TSA certificate chains in trust root

bkabrda:
Note: right now the CI is failing, because this PR needs updating sigstore/sigstore to a version containing https://github.com/sigstore/sigstore/pull/2068 which wasn't released yet. I'm not sure if the best course is to update the dependency to the specific commit or if it's better to wait until a new version is released - please let me know.

#### Summary
Fixes https://github.com/sigstore/cosign/issues/4098 - cosign will now respect multiple TSA certificate chains provided through the trust root and will try to verify the signed timestamp with all of them (and fail if none of the TSA chains verify the signed timestamp).

#### Release Note

Support for multiple TSA certificate chains in the trust root was added.

#### Documentation

I don't believe this requires a documentation change.

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -156,9 +156,7 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		if err != nil {
 			return fmt.Errorf("unable to load TSA certificates: %w", err)
 		}
-		co.TSACertificate = tsaCertificates.LeafCert
-		co.TSARootCertificates = tsaCertificates.RootCert
-		co.TSAIntermediateCertificates = tsaCertificates.IntermediateCerts
+		co.TSACertificateChains = tsaCertificates
 	}
 
 	if !c.IgnoreTlog {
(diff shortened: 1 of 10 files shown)


haydentherapper:
You can use the sigstore/sigstore version at the commit.

haydentherapper:
nvm, I cut a release :) https://github.com/sigstore/sigstore/releases/tag/v1.9.4

bkabrda:
@haydentherapper thanks for the quick release of the sigstore library. I updated the PR to incorporate that, but I noticed that one e2e test is failing; it looks like it might be related to my change, so I'll try to take a look at it later today to see what's going wrong.

bkabrda:
Ok, I understand what's going wrong; in the ["nonstandard key names with valid usage"](https://github.com/sigstore/cosign/blob/982ce6d7ce7eda081d9f11d8201324dca3f7b193/test/e2e_test.go#L652), it is assumed that even if multiple targets with "Usage: TSA" are used, they are still just parts of one certificate chain, but my code change expects these to be full certificate chains and therefore it fails on parsing them as such.

Even though I dislike the behavior, I think technically my change is breaking it and so probably shouldn't be merged... WDYT @haydentherapper? I currently don't see a fully backwards compatible way of making this work with the "old-style" tuf targets and so probably the best way to go would be to wait for the trusted_root.json support to get merged.

bkabrda:
Because I don't see a way to fix this without a regression, and because this is effectively solved by https://github.com/sigstore/cosign/pull/3844 for my usecase, I'm closing this PR. If anyone sees a way to fix my PR and needs it, feel free to take the code and modify it to submit a new PR.

PR #4197, opened 2025-05-07
sort Go imports alphabetically

dmitris:
#### Summary
sort Go imports alphabetically - extension of https://github.com/sigstore/cosign/pull/4196 for a number of other files with the similar "import mess" 😄 

#### Release Note
NONE

#### Documentation
n/a

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -23,6 +23,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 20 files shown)


haydentherapper (approved):
Thanks!

cpanato:
I would prefer to group to std libs, then external lib, then sigstore libs then project pkgs


But if the alphabetical was the decision I am ok with that

haydentherapper:
Sorry, I'm ok with either approach! I'm not sure we've followed this for all Go repos

dmitris:
I think it is much easier to keep the third-party imports sorted alphabetically - then you can add/remove imports, and rely on the editor doing the sorting when you hit "Save".

I would not mind if the sigstore libs would be surrounded by the empty lines like a "block" - but that already makes the IDE-sorting more difficult since it doesn't work across the empty lines.  I feel that the easiest and most reliable thing is to have all the third-party libraries in one block (without blank lines), and just rely on the IDE sorting them alphabetically.

PR #4213, opened 2025-05-28
More linting coverage

scop:
#### Summary

More linting coverage, hopefully yielding fewer future bugs and cleaner code.

#### Release Note

NONE

I guess these are not that interesting for cosign users.

#### Documentation

NONE

PR #4216, opened 2025-05-30
Make e2e service health check more flexible

cmurphy:
In 62a6617 in rekor, health checks were added to the trillian services, which means now five of the five services will report "healthy" when running. The cosign e2e tests were checking that exactly 3 were healthy, so if they started too fast and 4 or more were healthy, the tests would fail. This change makes the check more flexible. It is not very sophisticated, as it's technically possible the wrong 3 services are healthy, but it should still be good enough to serve its purpose in CI.

### DIFF
--- a/test/e2e_test.sh
+++ b/test/e2e_test.sh
@@ -71,7 +71,7 @@ for repo in rekor fulcio; do
     ${docker_compose} up -d
     echo -n "waiting up to 60 sec for system to start"
     count=0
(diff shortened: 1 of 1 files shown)


haydentherapper:
I added a fix for this in https://github.com/sigstore/cosign/pull/4213/files#diff-c51c51ea1233078ae52e43453d38ef2751183093452770a9b74bc88828541aed, WDYT about this approach?

PR #4225, opened 2025-06-06
Don't verify SCT for a private PKI cert

cmurphy:
Remove the need to fetch TUF keys when signing with a private key and attaching a non-Fulcio certificate to the artifact bundle. Verifiers will still need to check whether the certificate contains an SCT and have a policy for verifying it.

Relates to https://github.com/sigstore/cosign/pull/3844/files#r2122409378


haydentherapper (approved):
Looks like this was never checked as part of sign-blob/attest(-blob) either.

PR #4242, opened 2025-06-12
Add Rekor v2 support for trusted-root create

cmurphy:
When deducing the Rekor log key ID, cosign universally assumes a Rekor v1 type checkpoint, which is not C2SP compliant. Rekor v2 is compliant for all different types of keys, which means the log ID must be calculated differently. This affects the `trusted-root create` tool which must generate the log ID from the public key. This change adds the ability for the trusted-root command to parse a ":" in the --rekor-key flag to indicate that the trusted material should be generated for a Rekor v2 log and that the origin string following the ":" should be used to calculate it. This is backwards compatible and will not affect Rekor v1 which needs no origin string.

This addresses the issue strictly for this command so that trusted_root files can be created for Rekor v2 servers. A later change will make more general changes to the TUF client to ensure the trusted material is generated properly for the server it relates to.

#### Release Note

Added support in `cosign trusted-root create` for generating a transparency log trusted root entry with a C2SP-conformant key ID, which is required for use with Rekor v2. Use `--rekor-key=</path/to/key>,<server-name>` to get key material for this kind of transparency log.

### DIFF
--- a/cmd/cosign/cli/trustedroot/trustedroot.go
+++ b/cmd/cosign/cli/trustedroot/trustedroot.go
@@ -23,9 +23,11 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/sigstore/cosign/v2/pkg/cosign"
+	"github.com/sigstore/rekor-tiles/pkg/note"
 	"github.com/sigstore/sigstore-go/pkg/root"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
@@ -91,9 +93,21 @@ func (c *CreateCmd) Exec(_ context.Context) error {
 	}
 
 	for i := 0; i < len(c.RekorKeyPath); i++ {
-		tlogPubKey, id, idBytes, err := getPubKey(c.RekorKeyPath[i])
+		keyParts := strings.SplitN(c.RekorKeyPath[i], ",", 2)
+		keyPath := keyParts[0]
+		tlogPubKey, id, idBytes, err := getPubKey(keyPath)
 		if err != nil {
 			return err
+		}
+		var origin string
+		if len(keyParts) > 1 {
+			origin = keyParts[1]
+		}
+		if origin != "" {
+			id, idBytes, err = getCheckpointID(origin, *tlogPubKey)
+			if err != nil {
+				return err
+			}
 		}
 
 		startTime := time.Unix(0, 0)
@@ -240,4 +254,12 @@ func getPubKey(path string) (*crypto.PublicKey, string, []byte, error) {
 	}
 
 	return &pubKey, keyID, idBytes, nil
+}
+
+func getCheckpointID(origin string, key crypto.PublicKey) (string, []byte, error) {
+	_, id, err := note.KeyHash(origin, key)
+	if err != nil {
+		return "", nil, err
+	}
+	return hex.EncodeToString(id), id, nil
 }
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/trustedroot/trustedroot.go
```diff
@@ -219,3 +233,11 @@ func getPubKey(path string) (*crypto.PublicKey, string, []byte, error) {
 
 	return &pubKey, keyID, idBytes, nil
 }
+
+func getCheckpointID(origin string, key crypto.PublicKey) (string, []byte, error) {
+	_, id, err := note.KeyHash(origin, key)
+	if err != nil {
+		return "", nil, err
+	}
+	return hex.EncodeToString(id), id, nil
```
Hayden-IO, line 264: This should be base64 encoded - https://github.com/sigstore/root-signing-staging/blob/b2f403d77cfc959af5a3359fdcb7b57a06fd5f6d/targets/trusted_root.json#L29

Hex encoding was for Rekor v1

haydentherapper (changes requested):

cmd/cosign/cli/trustedroot/trustedroot.go
cmurphy: The hex encoded string is not actually used in the trusted root, it's just a key in the map that is used for logging https://github.com/sigstore/sigstore-go/blob/c43b348797029a3a4483db857c16dd88d34a6cae/pkg/root/trusted_root_create.go#L37-L43

The bytes `id` is base64 encoded.

Hayden-IO: Ah yes, I missed this was returning the non-encoded bytes as well. Nvm!

haydentherapper (dismissed):

Hayden-IO:
Can you run `make docgen`?

Hayden-IO:
I'm not seeing anything different with file creation, I wonder if the windows test is a flake.

Hayden-IO:
Looks like it's not a flake, I'd add additional logging to determine where it's failing. Windows paths are fun.

cmd/cosign/cli/trustedroot/trustedroot_test.go
```diff
@@ -41,10 +42,16 @@ func TestCreateCmd(t *testing.T) {
...
+	rekorV1KeyPath := filepath.Join(td, "rekor.v1.pub")
+	makeKey(t, rekorV1KeyPath)
+	rekorV2KeyPath := filepath.Join(td, "rekor.v2.pub")
+	makeKey(t, rekorV2KeyPath)
+
 	outPath := filepath.Join(td, "trustedroot.json")
 
 	trustedrootCreate := CreateCmd{
 		CertChain:        []string{fulcioChainPath},
+		RekorKeyPath:     []string{rekorV1KeyPath, rekorV2KeyPath + ":rekor.server"},
```
jku: I don't know what the postfix is for here but I believe windows does not take kindly to ":" in filenames

jku: ah, the purpose was this `--rekor-key=</path/to/key>:<server-name> ` -- I think you'll have to invent another scheme if it needs to work in Windows (unless the postfix isn't part of the filename -- then I don't know what the issue is)

jku: yeah the code seems fine, it's not using the whole string as a filepath -- I'll hide this, it's not a useful thread

haydentherapper (dismissed):

haydentherapper (approved):

PR #4243, opened 2025-06-13
Add baseUrl and Uri to trusted-root create command

cmurphy:
Clients using Rekor v2 need the name of the log server in order to create a checkpoint verifier, so it is useful to include it in the trust root. This change adds that functionality for all key material.


cmd/cosign/cli/trustedroot/trustedroot.go
```diff
@@ -47,7 +51,11 @@ func (c *CreateCmd) Exec(_ context.Context) error {
 	rekorTransparencyLogs := make(map[string]*root.TransparencyLog)
```
Hayden-IO, line 52: Should we check that these URI arrays aren't empty?

haydentherapper (approved):

cmd/cosign/cli/trustedroot/trustedroot.go
cmurphy: We can't require them to be populated, that would be a breaking change. And baseURL isn't a required field in the protobuf spec or for Rekor v1.

steiza (approved):
Makes sense! Initially I had thought of this command as useful for making a trusted root for *verification* (where you don't need the URLs), but it could also be useful in signing where you need to contact the servers (and thus know their URLs).

PR #4264, opened 2025-06-25
avoid double-loading trustedroot from file

dmitris:
#### Summary
In [cmd/cosign/cli/verify/verify_attestation.go](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_attestation.go)
we call root.NewTrustedRootFromPath twice both times with the same condition if c.TrustedRootPath != "" :
* L137-138 https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_attestation.go#L137-L138
* L271, 278 https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_attestation.go#L271-L278

This change removes the second duplicate one - a check for non-nil `co.TrustedMaterial` added as a reminder to developers and to guard against potential future changes.

On `#cosign` Slack, @znewman01 [said](https://sigstore.slack.com/archives/C01PZKDL4DP/p1750864075461449?thread_ts=1750860467.217359&cid=C01PZKDL4DP):
> From a quick look, it seems like these checks could be combined

#### Release Note
NONE

#### Documentation
n/a

### DIFF
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -276,10 +276,6 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		// If a trusted root path is provided, we will use it to verify the bundle.
 		// Otherwise, the verifier will default to the public good instance.
 		// co.TrustedMaterial is already loaded from c.TrustedRootPath above,
-		// the check below is to serve as a reminder to developers and to future-proof.
-		if co.TrustedMaterial == nil {
-			panic("internal cosign error co.TrustedMaterial is nil with non-empty c.TrustedRootPath")
-		}
 	case c.CARoots != "":
 		// CA roots + possible intermediates are already loaded into co.RootCerts with the call to
 		// loadCertsKeylessVerification above.
--- a/cmd/cosign/cli/verify/verify_attestation.go
+++ b/cmd/cosign/cli/verify/verify_attestation.go
@@ -275,9 +275,10 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 
 		// If a trusted root path is provided, we will use it to verify the bundle.
 		// Otherwise, the verifier will default to the public good instance.
-		co.TrustedMaterial, err = root.NewTrustedRootFromPath(c.TrustedRootPath)
-		if err != nil {
-			return fmt.Errorf("creating trusted root from path: %w", err)
+		// co.TrustedMaterial is already loaded from c.TrustedRootPath above,
+		// the check below is to serve as a reminder to developers and to future-proof.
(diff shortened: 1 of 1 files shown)


Hayden-IO:
Can you sign the DCO?

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -275,9 +275,10 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
```
cmurphy, line 275: > unsupported: trusted root path currently only supported with --new-bundle-format

I don't think this is true anymore, the trusted material loaded from the trusted root path at the beginning of this file is now useful for verifying any bundle https://github.com/sigstore/cosign/blob/c36801ef34aa24eb01d064ab12da66a83b9c3ff4/pkg/cosign/verify.go#L1192-L1206

so I think this whole switch case could be deleted.

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -275,9 +275,10 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 
 		// If a trusted root path is provided, we will use it to verify the bundle.
 		// Otherwise, the verifier will default to the public good instance.
-		co.TrustedMaterial, err = root.NewTrustedRootFromPath(c.TrustedRootPath)
-		if err != nil {
-			return fmt.Errorf("creating trusted root from path: %w", err)
+		// co.TrustedMaterial is already loaded from c.TrustedRootPath above,
+		// the check below is to serve as a reminder to developers and to future-proof.
+		if co.TrustedMaterial == nil {
+			panic("internal cosign error co.TrustedMaterial is nil with non-empty c.TrustedRootPath")
```
cmurphy: If we don't delete the switch case then we could still just delete this, this check is done earlier in the file https://github.com/sigstore/cosign/blob/c36801ef34aa24eb01d064ab12da66a83b9c3ff4/cmd/cosign/cli/verify/verify_attestation.go#L153-L157

dmitris: I meant to leave it to guard against future changes, as mentioned in the comment
```go
// the check below is to serve as a reminder to developers and to future-proof.
```

dmitris: I removed the comment and the duplicate check in 9e0a2b9b.

Regarding removing the whole switch case - I would suggest to start with this "small change" and then as a follow-up someone (maybe you @cmurphy 😄 ) could remove the whole switch case.

cmurphy (approved):

haydentherapper (approved):

Hayden-IO:
Thank you!

PR #4280, opened 2025-07-10
Add signing-config create command

haydentherapper:
This introduces a new command `cosign signing-config create` to generate a SigningConfig message. This will first be used by our testing infrastructure to create a SigningConfig, and then by users adopting the bundle format in Cosign v2 and eventually as a requirement for v3 when signing with private deployments.


go.mod
```diff
@@ -34,9 +34,9 @@ require (
 	github.com/sigstore/fulcio v1.7.1
 	github.com/sigstore/protobuf-specs v0.4.3
 	github.com/sigstore/rekor v1.3.10
-	github.com/sigstore/rekor-tiles v0.1.6
+	github.com/sigstore/rekor-tiles v0.1.7-0.20250624231741-98cd4a77300f
 	github.com/sigstore/sigstore v1.9.5
-	github.com/sigstore/sigstore-go v1.0.0
+	github.com/sigstore/sigstore-go v1.0.1-0.20250710071215-2d48428552e7
```
Hayden-IO: I'm using the latest sigstore-go commit currently since I need SigningConfig v0.2 support.

cmd/cosign/cli/signingconfig.go
```diff
@@ -0,0 +1,73 @@
...
+func signingConfigCreate() *cobra.Command {
+	o := &options.SigningConfigCreateOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a Sigstore protobuf signing config",
+		Long: `Create a Sigstore protobuf signing config by supplying verification material for Fulcio, Rekor, OIDC, and TSA services.
+Each service is specified via a repeatable flag (--fulcio, --rekor, --oidc-provider, --tsa) that takes a comma-separated list of key-value pairs.`,
+		Example: `cosign signing-config create \
+    --fulcio="url=https://fulcio.sigstore.dev,api-version=1,start-time=2024-01-01T00:00:00Z,end-time=2025-01-01T00:00:00Z,operator=sigstore.dev" \
```
cmurphy, line 46: This API is pretty different from the `trusted-root create` API. If we were designing from scratch I see the merits of doing it this way, but since we have a preexisting pattern I think it would be more user-friendly to follow the same pattern.

Hayden-IO: This API allows users to provide multiple service deployments, such as multiple Rekor shards, whereas the `trusted-root create` API allows for only one deployment per service.

I've filed https://github.com/sigstore/cosign/issues/4282 to track copying over this API to `trusted-root create`.

Hayden-IO: https://github.com/sigstore/cosign/pull/4285 for `trusted-root create`. There's a little duplication, using the same `parseKVs` function, I'll update this once we merge one or the other.

cmurphy (approved):

pkg/cosign/tlog.go
```diff
@@ -478,7 +478,7 @@ func VerifyTLogEntryOffline(ctx context.Context, e *models.LogEntryAnon, rekorPu
 		if err != nil {
 			return fmt.Errorf("decoding log ID: %w", err)
 		}
-		entry, err := tlog.NewEntry(entryBytes, *e.IntegratedTime, *e.LogIndex, logID, e.Verification.SignedEntryTimestamp, e.Verification.InclusionProof)
+		entry, err := tlog.NewEntry(entryBytes, *e.IntegratedTime, *e.LogIndex, logID, e.Verification.SignedEntryTimestamp, e.Verification.InclusionProof) //nolint rekor v1 codepath
```
cmurphy: Nit: could use `//nolint:<name of the linter that's failing>` to disable just one linter and ensure this is checked by other linters.

Hayden-IO: Unfortunately for ignoring staticcheck lint failures, you have to update .golangci.yaml. Just updated.

matheuscscp:
Hi @haydentherapper @cmurphy,

I couldn't help but notice that this PR is adding 1.4k lines in `go.sum` with tons of versions of the same packages, see the screenshots below.

Is this really necessary? What import was introduced in this PR causing this huge explosion of imports? In CNCF Flux we integrate with cosign for verifying OCI artifacts in-cluster and for that we need to import `github.com/sigstore/cosign/v2`. When bumping from `v2.5.2` to `v2.5.3` we noticed this huge increase in the imported dependencies.

Screenshot from 2025-09-03 13-06-51

Screenshot from 2025-09-03 13-07-10

Screenshot from 2025-09-03 13-07-19

Screenshot from 2025-09-03 13-07-34

Hayden-IO:
This is coming from our import of github.com/sigstore/rekor-tiles through github.com/sigstore/sigstore-go. We can take a look at reducing the number of imported packages.

go.sum isn't a view of all dependencies from what I understand though, so we haven't significantly increased the number of dependencies.

Hayden-IO:
https://github.com/sigstore/sigstore-go/issues/519

PR #4285, opened 2025-07-14
Allow multiple services to be specified for trusted-root create

### LINKED ISSUE #4282: Update trusted-root create API to support multiple service deployments
**Description**

https://github.com/sigstore/cosign/pull/4280 introduced `cosign signing-config create` with an API that supports multiple service deployments. Rather than have a flag per-service-configuration, instead we use a comma-delimited string of service configuration, and can provide this string multiple times for each deployment.

Let's reuse this API for `trusted-root create` to improve usability of the command.

Hayden-IO:
This provides the same API as in the new `cosign signing-config create` command from #4280. To avoid this being a breaking change, the previous flags will still be supported, but will be removed in Cosign v3.

Fixes #4282

### DIFF
--- a/cmd/cosign/cli/options/trustedroot.go
+++ b/cmd/cosign/cli/options/trustedroot.go
@@ -20,6 +20,11 @@ import (
 )
 
 type TrustedRootCreateOptions struct {
+	Fulcio []string
+	CTFE   []string
+	TSA    []string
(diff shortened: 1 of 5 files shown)


cmurphy (commented):
This LGTM but I think @steiza should weigh in.

Issue #4284, opened 2025-07-14
Keyless verification fails without embedded SCT with --insecure-ignore-sct=true flag


namewasntfound:
**Description**

Keyless verification is failing even when the trusted_root.json in TUF includes the certificate chain of Fulcio. This issue arises specifically when the Fulcio certificates used for signing do not contain an embedded SCT and when passing the  --insecure-ignore-sct=true flag.

**Expected behavior:**

Verification should succeed if the Fulcio CA certificate chain is trusted and included in trusted_root.json, when passing  --insecure-ignore-sct=true flag and when the Fulcio certificate does not contain embedded SCT.

**Version**

2.5.2

Hayden-IO:
Can you provide the commands you used to sign and verify? 

FYI @cmurphy @steiza

namewasntfound:
Hi @haydentherapper,

For context, we’ve deployed a PoC instance using a private Fulcio, Rekor v1, and TSA. We're currently signing OCI images using only Fulcio and TSA (no transparency log). Here are the commands we're using:

**Signing:**

```
cosign sign <image> \
  --fulcio-url=<private_fulcio_url> \
  --insecure-skip-verify=true \
  --tlog-upload=false \
  --timestamp-server-url=<private_tsa_url>
```

**Verifying:**

```
cosign verify <image> \
  --insecure-ignore-tlog=true \
  --use-signed-timestamps=true \
  --insecure-ignore-sct=true \
  --certificate-identity=<cert_identity> \
  --certificate-oidc-issuer=<cert_oidc_issuer>
```

Hayden-IO:
Another datapoint: @cpanato let me know https://github.com/kubernetes-sigs/release-sdk/blob/main/test/integration/sign_test.go#L216 is failing, which also sets `IgnoreSCT` to `true`.

cmurphy:
@namewasntfound can you share the text of the error you encountered?

Was this working for you before and has stopped working recently?

namewasntfound:
Hi @cmurphy,

The last version we worked with successfully was cosign v2.5.0. At that time, our TUF repository contained individual targets for the trust material - specifically, the Rekor key, CTFE key and Fulcio certificates. Verification worked as expected with that setup and we didn't encounter any problems.

[...]

PR #4294, opened 2025-07-15
Fix cert verification logic for trusted-root/SCTs

cmurphy:
Previously, the logic for verifying the Fulcio leaf cert was gated on whether the verification material contained a trusted root and whether the cert had an embedded SCT. This was because the upgraded TUF client that fetched the trusted root and passed it to the sigstore-go verification routines was developed concurrently with updates to sigstore-go[1], and sigstore-go didn't initially support returning the cert chain that would allow verifying a detached SCT. When the change in sigstore-go landed, the API call was updated but the logic and the comment was not fixed. This meant that a trusted root file could not be used to verify leaf certs, unnecessarily. Since the trusted root was set, the TUF client didn't fetch the Fulcio certs independently, so the cert failed to verify. This change removes the restriction and adds a test that uses trusted root.

[1] https://github.com/sigstore/sigstore-go/pull/300

Fixes https://github.com/sigstore/cosign/issues/4284

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -343,21 +343,14 @@ func ValidateAndUnpackCertWithIntermediates(cert *x509.Certificate, co *CheckOpt
 
 	// Now verify the cert, then the signature.
 
-	shouldVerifyEmbeddedSCT := !co.IgnoreSCT
-	contains, err := ContainsSCT(cert.Raw)
-	if err != nil {
-		return nil, err
-	}
-	shouldVerifyEmbeddedSCT = shouldVerifyEmbeddedSCT && contains
-	// If trusted root is available and the SCT is embedded, use the verifiers from sigstore-go (preferred).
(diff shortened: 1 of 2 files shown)


cmurphy:
Linter failure addressed in https://github.com/sigstore/cosign/pull/4295

bobcallaway:
#4295 has merged, this probably just needs a rebase

pkg/cosign/verify.go
```diff
@@ -343,21 +343,18 @@ func ValidateAndUnpackCertWithIntermediates(cert *x509.Certificate, co *CheckOpt
 
 	// Now verify the cert, then the signature.
 
-	shouldVerifyEmbeddedSCT := !co.IgnoreSCT
 	contains, err := ContainsSCT(cert.Raw)
 	if err != nil {
 		return nil, err
 	}
```
bobcallaway: any benefit to moving this down lower right before it is used?

haydentherapper (approved):

PR #4304, opened 2025-07-23
feat(cli): add `--payload-media-type` to `sign` (fixes #4300)

### LINKED ISSUE #4300: `cosign sign --payload` hardcodes layer MIME to `application/vnd.dev.cosign.simplesigning.v1+json`
**Description**

When using `cosign sign --payload` you are free to provide any payload you want. This is officially documented in https://docs.sigstore.dev/cosign/signing/signing_with_containers/#sign-and-upload-a-generated-payload-in-another-format-from-another-tool

```
$ cosign sign --key cosign.key --payload=statement.json $IMAGE
```

(I'm using `cosign sign` to sign arbitrary in-toto statements while #4019 is being discussed)

At the moment, when doing this, even if you sign a `README.md` as per the example, the OCI layer MIME type is hardcoded to `application/vnd.dev.cosign.simplesigning.v1+json` which is not technically correct. This can be worked around by ignoring the mime type in implementations that have to read the layer, but it would be nice if the MIME was truthful.

Additionally, because the MIME is incorrect, `cosign verify` fails in cryptic ways since it tries to parse the payload as a SimpleSigning JSON object. Here's an example:

```
$ cosign verify --key=cosign.pub $IMAGE
Error: no matching signatures: invalid or missing digest in claim: 
error during command execution: no matching signatures: invalid or missing digest in claim:
```

Obviously guessing the mime type, especially if it's JSON is not possible, so my suggestion would be to provide a `--payload-type` flag to allow setting the type (and falling back to `application/vnd.dev.cosign.simplesigning.v1+json` if not provided, so nobody is broken by this, at least initially.

Cheers

**Version**

```
$ cosign version

GitVersion:    v2.5.0
GitCommit:     unknown
GitTreeState:  clean
BuildDate:     unknown
GoVersion:     go1.24.3
Compiler:      gc
Platform:      linux/amd64
```

arubegonsan:
#### Summary

This commit introduces a new `--payload-media-type` flag to the `cosign sign` command.

When signing with a custom payload using the `--payload` flag, users can now also specify the MIME type of that payload via `--payload-media-type`.

This media type is then stored in the `mediaType` field of the signature layer descriptor in the resulting OCI manifest. This allows consumers of the signature to correctly interpret the payload's content type.

If `--payload` is not used, this flag is ignored.

#### Release Note

* Fixes #4300: introduced the `--payload-media-type` flag in `cosign sign`.

#### Documentation

Needs documentation change.


cmd/cosign/cli/sign/sign.go
```diff
@@ -233,7 +235,11 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 	}
 
 	var s icos.Signer
-	s = ipayload.NewSigner(sv)
+	var staticOpts []static.StaticOption
```
cmurphy, line 238: It looks like the majority of the changed lines in this PR come from renaming this type. It's creating a lot of noise, and you'll find that it trips up the linter because it creates a "stutter". I don't think it's necessary to rename it because it's clear what it's for based on the package name. Could you undo this rename?

Issue #4221, opened 2025-06-05
Brainstorming: Cosign v3


haydentherapper:
**Description**

We've begun brainstorming what changes we'd like to make as part of the next major release of Cosign. We've been working on a number of related issues over the last year, tagged ["pre-theseus"](https://github.com/sigstore/cosign/issues?q=is%3Aissue%20label%3Apre-theseus), to modernize Cosign and bring it into conformance with newer Sigstore SDKs and clients that implement the Sigstore client specification. The next major release of Cosign, Cosign v3, will continue our modernization efforts.

We've copied in the notes from our most recent brainstorm, outlining the changes we'd like to make, to be formalized later on. Please chime in with thoughts or concerns!  If you're interested in discussing Cosign v3, come join our bi-weekly discussions on Thursdays at 1pm PST/4pm EST, on the Sigstore community calendar.

--------------

* Default to using the bundle protobuf specification  
    * Drop support for blob / attestations using old format  
    * Review tooling to migrate old bundles to new format \- [issue](https://github.com/sigstore/cosign/issues/3794). There is currently `cosign bundle create`.  
      * [https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify\_bundle.go\#L127-L174](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_bundle.go#L127-L174)   
    * Allow containers to use old format with special flag  
      * Container "bundle" doesn't have an upgrade CLI tool today  
        * This isn't just a format change, it also involves the referrer spec and what we're signing over  
        * [https://github.com/sigstore/cosign/issues/3927](https://github.com/sigstore/cosign/issues/3927) for containers with signatures as OCI artifacts  
  * Default to using trusted root verification material  
    * Drop support for providing verification material any other way  
  * Review existing subcommands and eliminate most of them  
    * Container manipulation commands should move to crane  
    * Things outside of blobs / attestations / container images should be removed  
  * Dropping support for online Rekor verification  
    * Verification path \-\> just sigstore-go?  
      * Yes please\! cosign v2.4.0 has been using sigstore-go for trusted root verifications without any users reporting issues 
  * Signing changes  
    * Use sigstore-go for signing\! This isn't top priority, but it adds features like support for signing config.  
    * For container signing we'll have to do some surgery around how cosign handles signing today  
    * Add `cosign signing-config create`   
  * cosign as an API?  
    * Eh, maybe in v4; use sigstore-go for now  
[...]

PR #4307, opened 2025-07-25
Default to using the new protobuf format

steiza:
#### Summary

This is a first step towards #4221, where we default `--new-bundle-format` to `true` instead of `false`.

We never did add protobuf support to `cosign sign` (see #3927 and #3139) - maybe we want to wait until that is done? Otherwise we could just have `--new-bundle-format` default to `true` when we add it to `cosign sign`.

#### Release Note

- Modified `attest`, `attest-blob`, `sign-blob`, `dockerfile-verify`, `manifest-verify`, `verify`, `verify-attestation`, `verify-blob`, `verify-blob-attestation` so that `--new-bundle-format` defaults to `true` instead of `false`.

#### Documentation

Ran `make docgen` as part of this PR.

### DIFF
--- a/cmd/cosign/cli/options/attest.go
+++ b/cmd/cosign/cli/options/attest.go
@@ -107,5 +107,5 @@ func (o *AttestOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.RecordCreationTimestamp, "record-creation-timestamp", false,
 		"set the createdAt timestamp in the attestation artifact to the time it was created; by default, cosign sets this to the zero value")
 
-	cmd.Flags().BoolVar(&o.NewBundleFormat, "new-bundle-format", false, "attach a Sigstore bundle using OCI referrers API")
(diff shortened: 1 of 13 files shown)


steiza:
The scaffolding doesn't support a `trusted_root.json` file yet (https://github.com/sigstore/scaffolding/issues/1001) so I guess this is blocked for now.

Hayden-IO:
@steiza Check out https://github.com/sigstore/scaffolding/tree/main/actions/setup-sigstore-env! My goal is to get all Sigstore projects switched over to this, and deprecate/remove most of the scaffolding repo infra since a) it's outdated and b) if you need k8s infra, just use helm-charts.

steiza:
https://github.com/sigstore/scaffolding/tree/main/actions/setup-sigstore-env does look promising, but I think we'd either need cosign to support signing config (which I know you're working on, but is probably a fairly large project) or have that Action output Fulcio / Rekor URLs until cosign supports signing configs.

... but thinking about this more, we have `cosign` installed, so maybe I can just run `cosign trusted-root create ...` to make the trusted root I need.

PR #4316, opened 2025-07-29
Have cosign sign support bundle format

steiza:
#### Summary

This is an implementation of #3927, but it also finishes the last secondary goal on #3139, which is just over 2 years old 🥳 

More explicitly, it allows you to use `cosign sign` to product Sigstore protobuf signatures that we want everyone to move to, as well as slight updates to `cosign verify` so it can verify those bundles.

These bundle signatures will show up in `cosign tree` if you include `--experimental-oci11=true`, but it will not work with `cosign download`, which does not look like it has been updated in awhile.

I think this is the last step before we enable `--new-bundle-format=true` everywhere (and maybe also `--experimental-oci11=true` in some places).

You can test it with something like:
```
$ go run cmd/cosign/main.go sign --key cosign.key --new-bundle-format=true --tlog-upload=false ghcr.io/steiza/alpine@sha256:a19367999603840546b8612572e338ec076c6d1f2fec61760a9e11410f546733
```

And then verify with something like:
```
$ go run cmd/cosign/main.go verify --key cosign.pub --new-bundle-format=true --insecure-ignore-tlog=true ghcr.io/steiza/alpine@sha256:a19367999603840546b8612572e338ec076c6d1f2fec61760a9e11410f546733
```

Here's an example of what that bundle looks like:
```
$ oras blob fetch --output - ghcr.io/steiza/alpine@sha256:fa5d401b87a89e0e1d847862c6552fe48981bcef75cb2b6cdf3c6656aaf64c9c | jq
{
  "verificationMaterial": {
    "publicKey": {
      "hint": "KphQpmcaLAPdFYX3K1YoGLXC6+jkCGRydpUd45IP1Xo="
    }
  },
  "dsseEnvelope": {
    "payload": "eyJ0eXBlIjoiaHR0cHM6Ly9pbi10b3RvLmlvL1N0YXRlbWVudC92MSIsInN1YmplY3QiOlt7ImRpZ2VzdCI6eyJzaGEyNTYiOiJhMTkzNjc5OTk2MDM4NDA1NDZiODYxMjU3MmUzMzhlYzA3NmM2ZDFmMmZlYzYxNzYwYTllMTE0MTBmNTQ2NzMzIn19XSwicHJlZGljYXRlX3R5cGUiOiJodHRwczovL3NpZ3N0b3JlLmRldi9jb3NpZ24vc2lnbi92MSJ9",
    "payloadType": "application/vnd.in-toto+json",
    "signatures": [
      {
        "sig": "MEUCIQC+txDVbNH9SDGvJHagR98s7ocfBkFAJTCCRQIU/S8YkwIgLquuB9sgLKsL2Z2dUAbrMDDK08R29gXIlW9PKW3gBJY="
      }
    ]
  }
}
```

And here's what the old signature looked like:
```
$ cosign download signature ghcr.io/steiza/alpine@sha256:a19367999603840546b8612572e338ec076c6d1f2fec61760a9e11410f546733 | jq
{
  "Base64Signature": "MEQCIGdpIWC5SbvxFdNX4qrjXzA/0unT7NnvZhqUDI3CykSrAiA+hAnjmCLmKU7NMRVNa1jrorP4c9/uOIJK8+F+R670gQ==",
  "Payload": "eyJjcml0aWNhbCI6eyJpZGVudGl0eSI6eyJkb2NrZXItcmVmZXJlbmNlIjoiZ2hjci5pby9zdGVpemEvYWxwaW5lIn0sImltYWdlIjp7ImRvY2tlci1tYW5pZmVzdC1kaWdlc3QiOiJzaGEyNTY6YTE5MzY3OTk5NjAzODQwNTQ2Yjg2MTI1NzJlMzM4ZWMwNzZjNmQxZjJmZWM2MTc2MGE5ZTExNDEwZjU0NjczMyJ9LCJ0eXBlIjoiY29zaWduIGNvbnRhaW5lciBpbWFnZSBzaWduYXR1cmUifSwib3B0aW9uYWwiOm51bGx9",
  "Cert": null,
  "Chain": null,
  "Bundle": null,
  "RFC3161Timestamp": null
}
```

#### Release Note

* Added `cosign sign --new-bundle-format=true` to store signature in Sigstore protobuf bundles

#### Documentation

Doc updates are in PR


cmd/cosign/cli/attest/attest.go
```diff
@@ -251,7 +252,12 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		if err != nil {
 			return err
 		}
-		bundleBytes, err := makeNewBundle(sv, rekorEntry, payload, signedPayload, signerBytes, timestampBytes)
+		var pubKey *crypto.PublicKey
+		pk, err := sv.PublicKey()
```
Hayden-IO, line 256: Do we need to handle the error here?

cmd/cosign/cli/sign/sign.go
```diff
@@ -216,6 +230,95 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 	return nil
 }
 
+func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpts, signOpts options.SignOptions, sv *SignerVerifier) error {
+	digestParts := strings.Split(digest.DigestStr(), ":")
```
Hayden-IO, line 234: Is there currently an expectation that the digest of the image is provided when signing a container? Could we compute this ourselves? _Should_ we compute this regardless to verify the provided digest is correct?

cmd/cosign/cli/sign/sign.go
```diff
@@ -216,6 +230,95 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
...
+	if len(digestParts) != 2 {
+		return fmt.Errorf("unable to parse digest %s", digest.DigestStr())
+	}
+
+	subject := intotov1.ResourceDescriptor{
+		Digest: map[string]string{digestParts[0]: digestParts[1]},
+	}
+	predicateType := "https://sigstore.dev/cosign/sign/v1"
```
Hayden-IO: Can we make this a constant somewhere, I assume this will have to be publicly used eventually?

cmd/cosign/cli/attest/attest.go
```diff
@@ -251,7 +252,12 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+		if err == nil {
+			pubKey = &pk
+		}
+		bundleBytes, err := cbundle.MakeNewBundle(pubKey, rekorEntry, payload, signedPayload, signerBytes, timestampBytes)
```
Hayden-IO, line 260: Why does MakeNewBundle take both a signer and the public key? Can this be simplified?

pkg/cosign/bundle/protobundle.go
```diff
@@ -63,3 +74,63 @@ func MakeProtobufBundle(hint string, rawCert []byte, rekorEntry *models.LogEntry
 
 	return bundle, nil
 }
+
+func MakeNewBundle(pubKey *crypto.PublicKey, rekorEntry *models.LogEntryAnon, payload, sig, signer, timestampBytes []byte) ([]byte, error) {
+	// Determine if signature is certificate or not
```
Hayden-IO:
```suggestion
	// Determine if the signer is a certificate or not
```

Hayden-IO: Though as I mentioned above, it'd be nice to simplify this method given there's two closely related inputs - pubkey and signer

test/e2e_test.go
```diff
@@ -834,6 +834,44 @@ func TestSignVerifyWithTUFMirror(t *testing.T) {
...
+		KeyRef:          pubKeyPath,
+		IgnoreTlog:      true,
+		NewBundleFormat: true,
+	}
+
+	args := []string{imgName}
+
+	must(cmd.Exec(ctx, args), t)
+}
+
```
Hayden-IO, line 1047: If possible, can we have an e2e test that demonstrates interactions with fulcio and rekor as well? There's an example above and in my signingconfig PR.

Hayden-IO:
What do we want to do about `experimental-oci11`? Can we add some documentation to its flag description that it's unrelated to `new-bundle-format`?

cmd/cosign/cli/attest/attest.go
steiza: This was on purpose - if we have an error `pubKey` will remain `nil` (because we have a certificate).

cmd/cosign/cli/sign/sign.go
steiza: Yeah, before we get to this point `cosign sign` parses the OCI reference and uses that to compute the digest. Here we're just splitting the algorithm from the digest (that `cosign sign` has already computed) in order to create the in-toto subject.

test/e2e_test.go
steiza: I was able to get Rekor working the e2e test, but not Fulcio 😞

cmd/cosign/cli/attest/attest.go
steiza: This is a holdover from how the private function was before - it first tried to get a certificate from `cryptoutils.UnmarshalCertificatesFromPEM(signer)` and if it was unable to then it did `pubKey, err := sv.PublicKey()`.

When I made this function public, it was creating an import loop with `sign.SignerVerifier`, so I opted to replace that option with `*crypto.PublicKey`. We could try to cram / overload `signerBytes`, but it's more work for the places we call `MakeBundle()` 🤷

steiza:
Here's a good command for testing compatibility with GitHub Artifact Attestations:

```
$ go run cmd/cosign/main.go verify --certificate-oidc-issuer "https://token.actions.githubusercontent.com" --certificate-identity-regexp "https://github.com/github/artifact-attestations-workflows/.github/workflows/ghcr-publish.yml@refs/heads/main" --new-bundle-format=true ghcr.io/github/artifact-attestations-workflows@sha256:06e6818324cb20e1ea6f22da7fb925323fb57251c98768c88b93872b50590f92
```

test/e2e_test.go
Hayden-IO: Have you tried `cosign sign --new-bundle-format` without providing a `--key`? Just want to make sure it is working with ephemeral keys/Fulcio as expected.

What was the error you got in testing? I can try to get something working later on in the SigningConfig PR.

cmd/cosign/cli/sign/sign.go
Hayden-IO: Perfect! Just wanted to check that wasn't just a user-provided digest.

cmd/cosign/cli/attest/attest.go
Hayden-IO: Eh, we can clean this up later with a larger refactor.

test/e2e_test.go
steiza: Yes, sorry, I should have phrased that better. I was unable to get Fulcio working **locally in the e2e test**. I was able to get Fulcio verification working generally (see https://github.com/sigstore/cosign/pull/4316#issuecomment-3161544529).

But I uploaded [what I was working on locally](https://github.com/sigstore/cosign/pull/4316/commits/80da7fbd4b712133d08f9d0a4b42a043b2fba257), and it seemed to work in the GitHub workflow!

haydentherapper (approved):

haydentherapper (approved):

PR #4318, opened 2025-07-31
Default to using the new protobuf format

steiza:
#### Summary

This is a first step towards #4221, where we default `--new-bundle-format` to `true` instead of `false`.

We never did add protobuf support to cosign sign (see #3927 and #3139) - maybe we want to wait until that is done? Otherwise we could just have `--new-bundle-format` default to `true` when we add it to cosign sign.

#### Release Note

- Modified `attest`, `attest-blob`, `sign-blob`, `dockerfile-verify`, `manifest-verify`, `verify`, `verify-attestation`, `verify-blob`, `verify-blob-attestation` so that `--new-bundle-format` defaults to `true` instead of `false`.

#### Documentation

Ran `make docgen` as part of this PR.

### DIFF
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -124,9 +124,10 @@ jobs:
       run: |
         ./cosign sign-blob README.md --fulcio-url ${FULCIO_URL} --rekor-url ${REKOR_URL} --bundle blob.sigstore.json --yes --identity-token ${OIDC_TOKEN}
 
+    # TODO - cosign sign doesn't support new bundle format yet
     - name: Verify with cosign
       run: |
-        ./cosign verify --rekor-url ${REKOR_URL} --allow-insecure-registry ${demoimage} --certificate-identity https://kubernetes.io/namespaces/default/serviceaccounts/default --certificate-oidc-issuer "https://kubernetes.default.svc.cluster.local"
+        ./cosign verify --new-bundle-format=false --rekor-url ${REKOR_URL} --allow-insecure-registry ${demoimage} --certificate-identity https://kubernetes.io/namespaces/default/serviceaccounts/default --certificate-oidc-issuer "https://kubernetes.default.svc.cluster.local"
 
     - name: Verify custom attestation with cosign, works
       run: |
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -122,7 +122,7 @@ jobs:
 
     - name: Sign a blob
       run: |
-        ./cosign sign-blob README.md --fulcio-url ${FULCIO_URL} --rekor-url ${REKOR_URL} --output-certificate cert.pem --output-signature sig --yes --identity-token ${OIDC_TOKEN}
+        ./cosign sign-blob README.md --fulcio-url ${FULCIO_URL} --rekor-url ${REKOR_URL} --bundle blob.sigstore.json --yes --identity-token ${OIDC_TOKEN}
 
     - name: Verify with cosign
       run: |
@@ -152,7 +152,7 @@ jobs:
 
     - name: Verify a blob
       run: |
-        ./cosign verify-blob README.md --rekor-url ${REKOR_URL} --certificate ./cert.pem --signature sig --certificate-identity https://kubernetes.io/namespaces/default/serviceaccounts/default --certificate-oidc-issuer "https://kubernetes.default.svc.cluster.local"
+        ./cosign verify-blob README.md --bundle blob.sigstore.json --certificate-identity https://kubernetes.io/namespaces/default/serviceaccounts/default --certificate-oidc-issuer "https://kubernetes.default.svc.cluster.local"
 
     - name: Collect diagnostics
       if: ${{ failure() }}
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -111,6 +111,15 @@ jobs:
         REPOSITORY=${PWD}/repository
         ./cosign initialize --root ${ROOT} --mirror file://${REPOSITORY}
 
+    # Get verification material from servers
+    # NOTE: never do this in production! This is a work-around until we upgrade sigstore/scaffolding
+    - run: curl -k ${REKOR_URL}/api/v2/trustBundle | jq -r ".chains[0].certificates[]" > fulcio.pem
+
+    - run: curl -k ${REKOR_URL}/api/v1/log/publicKey > rekor.pub
+
+    - name: Create trusted_root.json
+      run: ./cosign trusted-root create --certificate-chain fulcio.pem --rekor-key rekor.pub --out trusted_root.json
+
     - name: Sign demoimage with cosign
       run: |
         ./cosign sign --rekor-url ${REKOR_URL} --fulcio-url ${FULCIO_URL} --yes --allow-insecure-registry ${demoimage} --identity-token ${OIDC_TOKEN}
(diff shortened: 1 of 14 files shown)


Hayden-IO:
Overall LGTM. Let's hold off on this PR until we land https://github.com/sigstore/cosign/pull/4316 and https://github.com/sigstore/cosign/pull/4319, do a pass over the other open PRs, and review the Cosign v3 brainstorm to see if there's anything else we want to get in. Then let's cut a new release of Cosign, hopefully the last v2 release, and then start merging these breaking changes.

steiza:
So, in some senses this is not a breaking change, in that we've made it so that `verify-blob` and `verify-blob-attestation` will "fall back" to trying the old bundle format if they can't parse the provided bundle using the protobuf:

```
$ cat test.sigstore.json 
{"base64Signature":"MEUCIEjqaXVxJgbngDBIXX9DnZ9yuQBK2Y2dU2C1O74T0RvlAiEA9WI9il4+bHMVtzGl8lxrX3SxCfdg5IMUT90Zk5roOu8="}
```

Verifies if you explicitly say it's the old bundle format:
```
$ go run cmd/cosign/main.go verify-blob --bundle=test.sigstore.json --new-bundle-format=false --key=cosign.pub --use-signed-timestamps=false --insecure-ignore-tlog=true ../sigstore-go/examples/sigstore-go-signing/hello_world.txt 
...
Verified OK
```

... but also verifies if you omit the bundle format (defaults to true):
```
$ go run cmd/cosign/main.go verify-blob --bundle=test.sigstore.json --key=cosign.pub --use-signed-timestamps=false --insecure-ignore-tlog=true ../sigstore-go/examples/sigstore-go-signing/hello_world.txt 
...
Verified OK
```

I spent some time looking into providing this fallback behavior for OCI verify commands (`verify`, `verify-attestation`) as well. It's technical possible of course, but it would either add quite a few requests to your registry for each verify (essentially doubling the requests) or require a somewhat substantial refactor in order to query once and then pass the results along to `pkg/cosign/verify.go` functions - what do you think?

steiza:
There's some issues with how the trusted root file is working in `verify` - I'll have to take a look later

steiza:
I couldn't get the scaffolding working with the new bundle format - I think there was some issue with how the trusted root was constructed - but we can update it when we overhaul the scaffolding test.

steiza:
Okay! With #4346 I think the old scaffolding is working with the new protobuf bundle format. I also added code for `verify` and `verify-attestation` to detect if you are using the old signature format and to gracefully fallback. If tests pass, I think this is ready to review.

Hayden-IO:
The one downside with this is that while it's not a breaking change for verification of old metadata, it is a breaking change for verifying new metadata, in that you can't use an old client to verify new signatures. Are we OK with that?

steiza:
> The one downside with this is that while it's not a breaking change for verification of old metadata, it is a breaking change for verifying new metadata, in that you can't use an old client to verify new signatures. Are we OK with that?

I think so. I would hope that an organization would upgrade it's toolchain (for signing and verifying) to a new version at the same time.

If we don't think that's reasonable, the alternative (which I'm okay with) is cutting a last cosign 2.x release before this lands and have this be the start of cosign v3!

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -121,6 +121,18 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
 		UseSignedTimestamps:          c.TSACertChainPath != "" || c.UseSignedTimestamps,
 		NewBundleFormat:              c.NewBundleFormat,
 	}
+
+	// Check to see if we are using the new bundle format or not
+	if !c.LocalImage {
+		ref, err := name.ParseReference(images[0], c.NameOptions...)
+		if err == nil && c.NewBundleFormat {
+			newBundles, _, err := cosign.GetBundles(ctx, ref, co)
```
Hayden-IO, line 129: We'll make an extra network call with `GetBundles` - do you think that's fine? Or should we store the bundles it finds and pass those to the verification function? Not sure if that's straightforward though.

haydentherapper (commented):
Fantastic!

cmd/cosign/cli/verify/verify.go
```diff
@@ -141,6 +141,17 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		NewBundleFormat:              c.NewBundleFormat,
 	}
 
+	// Check to see if we are using the new bundle format or not
+	if !c.LocalImage {
```
Hayden-IO, line 145: Just to confirm, this should handle backwards compatibility for verification of either the new or old bundle format?

.github/workflows/kind-verify-attestation.yaml
```diff
@@ -127,16 +136,16 @@ jobs:
 
     - name: Sign a blob
       run: |
-        ./cosign sign-blob README.md --fulcio-url ${FULCIO_URL} --rekor-url ${REKOR_URL} --output-certificate cert.pem --output-signature sig --yes --identity-token ${OIDC_TOKEN}
+        ./cosign sign-blob README.md --fulcio-url ${FULCIO_URL} --rekor-url ${REKOR_URL} --bundle blob.sigstore.json --yes --identity-token ${OIDC_TOKEN}
 
     - name: Verify with cosign
       run: |
-        ./cosign verify --rekor-url ${REKOR_URL} --allow-insecure-registry ${demoimage} --certificate-identity https://kubernetes.io/namespaces/default/serviceaccounts/default --certificate-oidc-issuer "https://kubernetes.default.svc.cluster.local"
+        ./cosign verify --trusted-root=${trustedroot} --allow-insecure-registry ${demoimage} --certificate-identity https://kubernetes.io/namespaces/default/serviceaccounts/default --certificate-oidc-issuer "https://kubernetes.default.svc.cluster.local"
```
cmurphy, line 143: `--trusted-root` shouldn't be necessary here, it's already set for these local services in the `cosign initialize` step (both the air-gapped version and remote version)

cmd/cosign/cli/options/sign.go
```diff
@@ -142,8 +142,7 @@ func (o *SignOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&o.RecordCreationTimestamp, "record-creation-timestamp", false, "set the createdAt timestamp in the signature artifact to the time it was created; by default, cosign sets this to the zero value")
 
-	// TODO: have this default to true as a breaking change
-	cmd.Flags().BoolVar(&o.NewBundleFormat, "new-bundle-format", false, "expect the signature/attestation to be packaged in a Sigstore bundle")
+	cmd.Flags().BoolVar(&o.NewBundleFormat, "new-bundle-format", true, "expect the signature/attestation to be packaged in a Sigstore bundle")
```
cmurphy, line 145: Not your change, but why is the word "expect" here? The sign command is creating the bundle in a format it decides, not expecting anything of an existing bundle.

cmd/cosign/cli/verify/verify_attestation.go
steiza: Yeah, I [mention this above](https://github.com/sigstore/cosign/pull/4318#issuecomment-3165589864):

> I spent some time looking into providing this fallback behavior for OCI verify commands (`verify`, `verify-attestation`) as well. It's technical possible of course, but it would either add quite a few requests to your registry for each verify (essentially doubling the requests) or require a somewhat substantial refactor in order to query once and then pass the results along to `pkg/cosign/verify.go` functions - what do you think?

As you can see, I opted to implement increasing the number of requests 😆 While testing / debugging these changes though, I discovered there was already another place in `cmd/cosign/cli/verify/verify_attestation.go` where we were doing this, in addition to `pkg/cosign/verify.go` and `pkg/oci/remote/write.go`. In short, I don't think it's an issue.

cmd/cosign/cli/verify/verify.go
steiza: Yes, so this version of the code has `c.NewBundleFormat` default to `true`, so it tests to make sure you aren't using an old format bundle.

The only case that doesn't work here is if you explicitly specify `--new-bundle-format=false` while pointing it at a new bundle, but I'm not sure how you'd end up in that case.

.github/workflows/kind-verify-attestation.yaml
steiza: I think that now that https://github.com/sigstore/cosign/pull/4347 landed (which I wrote after this pull request) you're right!

I'm going to leave this as-is since this pull request has been going for a month and a half, but I support testing / removing `--trusted-root` from `kind-verify-attestation.yaml`.

cmd/cosign/cli/options/sign.go
steiza: Yeah, I think this is just copy/pasted from verify commands - we should change this! But I'm going to leave this as-is to not further increase the scope of this pull request.

PR #4319, opened 2025-08-01
Add support for SigningConfig for sign-blob/attest-blob, support Rekor v2

haydentherapper:
#### Summary

This is the first PR to leverage sigstore-go's signing API, which is gated on whether or not a signing config is provided. While we could continue to reuse Cosign's signing APIs by passing in the signing config URLs to KeyOpts, we would not have support for Rekor v2 signing, which has only been implemented in sigstore-go. When a SigningConfig is provided, it will also force usage of the new bundle format.

You can see how much simpler sign/attest-blob will be once we start refactoring the CLI code.

There are two follow ups for this PR: Adding support for SigningConfig for attest (and eventually sign) and adding support for self-managed keys, which just requires a few Keypair implementations that wrap SignerVerifier.

This PR is split into two commits. The first refactors the authentication layer so it can be reused without Cosign's fulcio.Signer (which we can eventually move out to a utilities library), and the second adds support for SigningConfig.

#### Release Note

* Added support for SigningConfig for sign-blob and attest-blob
* Added support for Rekor v2 signing and verification


Hayden-IO:
Verified working against our staging Rekor v2 instance:

```
curl -LO https://raw.githubusercontent.com/sigstore/root-signing-staging/refs/heads/main/metadata/root_history/1.root.json
cosign initialize --mirror="https://tuf-repo-cdn.sigstage.dev" --root=1.root.json

cosign sign-blob --use-signing-config --bundle sigstore.json README.md

cosign verify-blob --new-bundle-format --bundle sigstore.json --certificate-identity $EMAIL --certificate-oidc-issuer https://accounts.google.com --use-signed-timestamps README.md
```

And for attestations...

```
cosign attest-blob --use-signing-config --statement intoto.txt --bundle sigstore.att.json

cosign verify-blob-attestation --new-bundle-format --bundle sigstore.att.json --certificate-identity $EMAIL --certificate-oidc-issuer https://accounts.google.com --use-signed-timestamps --digestAlg sha256 --digest="b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9"
```

Hayden-IO:
Proof it's using Rekor v2:

```
cat sigstore.json | jq -r ".verificationMaterial.tlogEntries[0]".inclusionProof.checkpoint.envelope

log2025-alpha1.rekor.sigstage.dev
19056
bfZQI4+dbAj12+FU+t1/DwP+/M5lrQGODznnOG2OYsM=

— log2025-alpha1.rekor.sigstage.dev 8w1amS0/tukAkFuM+b+hDou8IeUyeyfAnUphXTQ/ntwKodJlMfaQGJ7blOD0OdUam7+NiWaHhW3+wQ96q2pOAZw1fgg=
```

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -165,6 +163,61 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		}
 	}
 
+	if c.SigningConfig != nil {
+		// TODO: Only ephemeral keys are currently supported
```
cmurphy: Discussed offline but will still note here, I think it's important to get this added quickly as it's a bigger use case in cosign than in sigstore-go.

pkg/cosign/bundle/sign.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+func SignData(content sign.Content, keypair sign.Keypair, idToken string, signingConfig *root.SigningConfig, trustedMaterial root.TrustedMaterial) ([]byte, error) {
+	var opts sign.BundleOptions
+
+	if trustedMaterial != nil {
+		opts.TrustedRoot = trustedMaterial
+	}
+
+	if idToken != "" {
```
cmurphy, line 37: Could fetching the idToken be moved here, instead of duplicating in sign_blob.go and attest_blob.go? This is already where the rest of the SelectService logic seems to be happening.

cmd/cosign/cli/attest/attest_blob.go
Hayden-IO: Created issues to track this, will get this implemented right after.

pkg/cosign/bundle/sign.go
Hayden-IO: Refactored identity token retrieval into its own method that's now called from sign_blob and attest_blob. When we move this out into a shared library and work towards v3, I'll want to refactor this a bit more - for example, providing an identity token via flag that's either the token itself or a path isn't the cleanest UX.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -165,6 +162,50 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		}
 	}
 
+	if c.SigningConfig != nil {
+		// TODO(#4327): Only ephemeral keys are currently supported
+		// Need to add support for self-managed keys (e.g. PKCS11, KMS, on disk)
+		// and determine if we want to store certificates for those as well.
```
steiza, line 168: In the meantime, should we check for key flags like `--key`, `--sk`, or `--slot` and return an error to the user? Otherwise I'm worried we're silently ignoring those options if you're using signing config.

Hayden-IO:
> What about OCI support for signing config / Rekor v2? When would it make sense to also add this functionality to attest and sign?

Tracking this in https://github.com/sigstore/cosign/issues/4324. My plan would be to merge this PR and your PR, then I'll circle back to add support for `sign` and `attest` when using the new bundle format.

cmd/cosign/cli/attest/attest_blob.go
Hayden-IO: Good idea, returning an error if any self-managed key is provided with a signing config.

steiza (approved):
This is great - nice work!

PR #4320, opened 2025-08-01
Verify subject with bundle only when checking claims

haydentherapper:
Previously, when --check-claims was set to false and a bundle in the new format was provided, we'd still try to check the in-toto subject digest and algorithm. These values weren't being set since they were conditioned on checking claims. Now, we skip digest verification if check-claims is false with a new bundle.


cmurphy (approved):

steiza (approved):
Nice catch!

PR #4336, opened 2025-08-11
feat: add oci artifact abstraction

arewm:
#### Summary

In order to have a more consistency when pushing artifacts to OCI registries with and without the referrer's API, I wanted to create an abstraction that would simplify maintenance.

Since there would be many changes required for this update, I wanted to start with submitting the abstraction for review. I can then follow up with making changes to the relevant read-only CLI commands and then finally to the write CLI commands as well.

Related: #4335

### DIFF
--- a/pkg/oci/artifacts/adaptive.go
+++ b/pkg/oci/artifacts/adaptive.go
@@ -1,0 +1,231 @@
+//
+// Copyright 2024 The Sigstore Authors.
+//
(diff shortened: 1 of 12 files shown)


arewm:
After talking with @steiza and @haydentherapper , it makes sense that we don't want to do this amount of work right now. I will close this issue and continue to iterate on the conversation.

Issue #4281, opened 2025-07-11
Update end-to-end tests to use Scaffolding containers


haydentherapper:
**Description**

Context: https://github.com/sigstore/scaffolding/tree/main/actions/setup-sigstore-env

The fakeoidc container was reimplemented in this setup as well. We can greatly simplify the [end-to-end testing bash script](https://github.com/sigstore/cosign/blob/main/test/e2e_test.sh) by leveraging the [run-containers script](https://github.com/sigstore/scaffolding/blob/main/actions/setup-sigstore-env/run-containers.sh).

Previous impl: https://github.com/sigstore/cosign/pull/4161

Hayden-IO:
Note: This is necessary for testing Rekor v2.

PR #4337, opened 2025-08-11
Don't require timestamps when verifying with a key

haydentherapper:
This supports signing and verification with Rekor v2 with a user-provided signing key. Timestamps will only be required for verifying Fulcio certificates.


Hayden-IO:
Unfortunately there's not a great way to test this. https://github.com/sigstore/cosign/issues/4281 covers adding support for Rekor v2 for end-to-end testing, at which point we can add a test for key upload without needing a TSA.

Edit: I've added a test that tests signing and verification without Rekor and without timestamps, which gets partial coverage. What's missing is a log that doesn't provide timestamps, which is just Rekor v2.

steiza (approved):
Makes sense to me!

PR #4351, opened 2025-08-18
Relax go directive to 1.24.x

Hayden-IO:
One of the dependencies has hardcoded a specific version of go, which forces all clients to use that version of Go in their own modules. This is unnecessarily restrictive, rather libraries should specify the minimum Go version necessary to build, and consumers should use the latest patch release when building to pick up bug fixes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,14 @@
 module github.com/sigstore/cosign/v2
 
-go 1.24.3
+go 1.24.0
+
+toolchain go1.24.6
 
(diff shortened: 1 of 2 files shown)


Hayden-IO:
Once https://github.com/buildkite/shellwords/pull/3 is in and buildkite/agent is updated as well, we can bump the buildkite/agent dep back up.

cpanato (approved):

PR #4357, opened 2025-08-22
refactor: extract function to write referrer attestations

arewm:
The only way that we can generate attestations is when we also produce protobuf bundles. While we are working to use bundles by default from the CLI, we should enable the cosign api to be used as a dependency to also assist in migrating to the referrer's api from the tag-based strategy.

Co-Authored-By: Gemini
Signed-off-by: arewm <arewm@users.noreply.github.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

#### Release Note
Added functions `WriteReferrer` and `WriteAttestationsReferrer` to the API enabling the ability to write
generic attestations with the referrer's API.

#### Documentation
No changes are made to the CLI interface.

### DIFF
--- a/pkg/oci/remote/write.go
+++ b/pkg/oci/remote/write.go
@@ -224,7 +224,9 @@ func (taggable taggableManifest) MediaType() (types.MediaType, error) {
 	return taggable.mediaType, nil
 }
 
-func WriteAttestationNewBundleFormat(d name.Digest, bundleBytes []byte, predicateType string, opts ...Option) error {
+// WriteReferrer writes a referrer manifest for a given subject digest.
+// It uploads the provided layers and creates a manifest that refers to the subject.
+func WriteReferrer(d name.Digest, artifactType string, layers []v1.Layer, annotations map[string]string, opts ...Option) error {
 	o := makeOptions(d.Repository, opts...)
 
 	signTarget := d.String()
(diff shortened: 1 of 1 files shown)


pkg/oci/remote/write.go
```diff
@@ -316,6 +311,50 @@ func WriteAttestationNewBundleFormat(d name.Digest, bundleBytes []byte, predicat
...
+	annotations := map[string]string{
+		"org.opencontainers.image.created": time.Now().UTC().Format(time.RFC3339),
+	}
+
+	// We have to pick an artifactType for the referrer manifest. The attestation
+	// layers themselves are DSSE envelopes, which wrap in-toto statements.
+	// For discovery, the artifactType should describe the semantic content (the
+	// in-toto statement) rather than the wrapper format (the DSSE envelope).
+	// Using the in-toto media type is the most appropriate and conventional choice,
+	// as policy engines and other tools will query for attestations using this type.
```
steiza, line 354: I had to really think through this (mostly because as mentioned I haven't spent much time with `cosign`'s OCI functions), but I think this is the right answer.

Retrieving an attestation that doesn't use the new protobuf bundles:
```
$ go run cmd/cosign/main.go download attestation ghcr.io/steiza/actions-oidc-proxy:v1 | jq
{
  "payloadType": "application/vnd.in-toto+json",
  "payload": "eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjAuMSIsInByZWRpY2F0ZVR5cGUiOiJodHRwczovL2Nvc2lnbi5zaWdzdG9yZS5kZXYvYXR0ZXN0YXRpb24vdjEiLCJzdWJqZWN0IjpbeyJuYW1lIjoiZ2hjci5pby9zdGVpemEvYWN0aW9ucy1vaWRjLXByb3h5IiwiZGlnZXN0Ijp7InNoYTI1NiI6ImEwNGMyOTk2YzRjNDJhYTUxNzQ5MmJjMGQ4ZTA5Y2EwMDJhNTk5ZDEyZjVlODJiZDUzNzNlZDYzMTNhZjZhNzkifX1dLCJwcmVkaWNhdGUiOnsiRGF0YSI6IntcInBhc3NlZFwiOiB0cnVlfVxuIiwiVGltZXN0YW1wIjoiMjAyNS0wOC0xNVQxODoyMjoyMVoifX0=",
  "signatures": [
    {
      "keyid": "",
      "sig": "MEYCIQD+TJnd3/GuORNfKMMpbviudNGyy+G9qUw0nOtQABEEnAIhAKiTsFRVWdPRSVN/xD3LxGkb/5xPz4f8voN6e16jdePa"
    }
  ]
}
```

`application/vnd.in-toto+json` makes sense!

steiza (commented):
I know this is still a draft, but I just wanted to say that I support this approach!

arewm:
@steiza , I am happy for this change to be non-draft if you want to actually review it. I just put it in draft so that I could reference it in my comment on the issue. 

I didn't add any test cases for this, but this section of code was also not very well tested. I can add some if you would like.

steiza:
> I didn't add any test cases for this, but this section of code was also not very well tested. I can add some if you would like.

I'm in favor of testing generally, but to unit test this it seems like we'd need to mock quite a bit (at least `remote.Head`, `remote.WriteLayer`, and `remote.Put`).

Another option we have for testing is to implement something in https://github.com/sigstore/cosign/blob/main/test/e2e_test.go. Do you want to take a look at that file and see if it makes sense to extend an existing test or add a new one to cover these changes? Don't feel like you have to add something - if it doesn't make sense that's fine too - but I think it's worth at least checking. cc @haydentherapper has more cosign testing experience than I do.

arewm:
I think that the primary intention of `e2e_test.go` is to test the CLI. We do test the new bundle format in that test, so that will be used to ensure that this refactoring is backwards compatible. Since we are not adding CLI functionality to generate referring attestations with the legacy format, it doesn't seem appropriate to include tests there.

The additional tests that would be added would be to ensure that the `artifactType` is what we actually what we tried to set it to.

Hayden-IO:
At this point, `e2e_test.go` is a bit of everything. If you needed an OCI index to avoid mocking out calls, I believe we've set one up in the e2e tests already. Some additional tests would be great though I recognize testing in Cosign can sometimes be 90% of the time spent on a feature, so if it's becoming too much of a pain, just let us know.

arewm:
I had Claude help to write some test, but I didn't exercise it any more from the e2e tests as there is no additional CLI interface added. I looks like it went ahead and wrote tests for more than what I added though to cover some of the gap.

steiza (approved):
Looks good to me - nice work!

Issue #4324, opened 2025-08-04
Support SigningConfig for Sign and Attest


haydentherapper:
**Description**

Follow up from #4319. Need to decide on if we use sigstore-go's signing API, passing in the to-be-signed container or provenance, or continue to use Cosign's sign API. If we go with the latter, we'll need to extract URLs from the signing config and plumb them through to the sign API, and pull in Rekor v2's client from sigstore-go for Rekor v2 support. I'm not yet sure which of these solutions is more straightforward - the first would be preferred, since that's in line with the eventual refactoring and simplification of Cosign's codebase.

Hayden-IO:
We need to include the changes from https://github.com/sigstore/cosign/pull/4367 (adding `--trusted-root`) and https://github.com/sigstore/cosign/pull/4368.

PR #4367, opened 2025-08-26
Update conformance test binary for signing config

haydentherapper:
The signing config will now be provided on the sign path to test Rekor v2, along with the trusted root for verifying bundles on the sign path.

This also adds support for providing a trusted root with sign-blob/attest-blob. Currently, you can either provide just the CT log key or we'll fetch the trusted root from the initialized TUF repo. Since we are providing the trusted root for staging during signing now, this PR also lets the user provide the trusted root they'll use for verifying during signing.

This also updates the signing config flag name to drop `-path` for uniformity. This is fine to do since we haven't cut a release of Cosign with this feature yet.

Fixes https://github.com/sigstore/cosign/issues/4356

### DIFF
--- a/cmd/cosign/cli/attest_blob.go
+++ b/cmd/cosign/cli/attest_blob.go
@@ -90,11 +90,18 @@ func AttestBlob() *cobra.Command {
 				NewBundleFormat:          o.NewBundleFormat,
 			}
 			if o.Key == "" && env.Getenv(env.VariableSigstoreCTLogPublicKeyFile) == "" { // Get the trusted root if using fulcio for signing
-				trustedMaterial, err := cosign.TrustedRoot()
-				if err != nil {
-					ui.Warnf(context.Background(), "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
+				if o.TrustedRootPath != "" {
+					ko.TrustedMaterial, err = root.NewTrustedRootFromPath(o.TrustedRootPath)
+					if err != nil {
+						return fmt.Errorf("loading trusted root: %w", err)
+					}
+				} else {
+					trustedMaterial, err := cosign.TrustedRoot()
+					if err != nil {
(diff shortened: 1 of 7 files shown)


cpanato (dismissed):

Hayden-IO:
Note that this is used both by Cosign (for verifying an SCT during signing) and sigstore-go when called from Cosign (when a signing config is provided).

jku (approved):
looks good to me.

I suppose this path does not get exercised yet in the PR tests since the conformance test uses release version of conformance... Manually running the conformance-nightly workflow on this branch might work though?

cmd/cosign/cli/attest_blob.go
```diff
@@ -90,11 +90,18 @@ func AttestBlob() *cobra.Command {
...
+					ko.TrustedMaterial, err = root.NewTrustedRootFromPath(o.TrustedRootPath)
+					if err != nil {
+						return fmt.Errorf("loading trusted root: %w", err)
+					}
+				} else {
+					trustedMaterial, err := cosign.TrustedRoot()
+					if err != nil {
+						ui.Warnf(context.Background(), "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
+					}
+					ko.TrustedMaterial = trustedMaterial
```
steiza, line 103: It's a little weird to me that we always verify after signing, but I guess this is how it was before (if you aren't using a managed key for signing).

steiza (approved):
Nice!

I was going to complain that we aren't making the corresponding changes to `attest` and `sign`, but I guess that will come along with https://github.com/sigstore/cosign/issues/4324? Do we want to note there that we should include `--trusted-root` when we do that work?

cmd/cosign/cli/attest_blob.go
Hayden-IO: It's not a bad idea to make sure that what we're returning to the caller is verifiable.

Also if we enforce this now, this is something we can walk back later on if we find it to be unnecessary. I could imagine one use case is where the trust root is distributed separately from the signing config (for some reason...), effectively having disjoint signers and verifiers.

PR #4368, opened 2025-08-26
Support self-managed keys when signing with sigstore-go

### LINKED ISSUE #4327: Support self-managed keys with sigstore-go signing
**Description**

Follow up from https://github.com/sigstore/cosign/pull/4319. The sigstore-go signing API takes a `Keypair` rather than a `SignerVerifier`. We'll need to create `Keypair` implementations for PIV and KMS keys, which can use the existing `SignerVerifier` implementations for each.

haydentherapper:
This creates a wrapper around the Keypair interface when a SignerVerifier is provided for signing with KMS or any other provided keys. This also retains support for --issue-certificate to request a certificate for a managed key.

Fixes #4327


Hayden-IO:
Verified this works with Rekor v2:

```
curl -LO https://raw.githubusercontent.com/sigstore/root-signing-staging/refs/heads/main/metadata/root_history/1.root.json
cosign initialize --root 1.root.json --mirror="https://tuf-repo-cdn.sigstage.dev" 

cosign generate-key-pair

cosign sign-blob --new-bundle-format --bundle artifact.sigstore.json --yes --use-signing-config --key cosign.key README.md
cosign verify-blob --new-bundle-format --bundle artifact.sigstore.json --key cosign.pub README.md

cosign sign-blob --new-bundle-format --bundle artifact.sigstore.json --yes --use-signing-config --key cosign.key --issue-certificate README.md
cosign verify-blob --new-bundle-format --bundle artifact.sigstore.json --certificate-identity $email --certificate-oidc-issuer https://accounts.google.com --use-signed-timestamps README.md

cosign attest-blob --use-signing-config --statement intoto.txt --bundle sigstore.att.json --key cosign.key
cosign verify-blob-attestation --new-bundle-format --bundle sigstore.att.json --digestAlg sha256 --digest="b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9" --key cosign.pub

cosign attest-blob --use-signing-config --statement intoto.txt --bundle sigstore.att.json --key cosign.key --issue-certificate
cosign verify-blob-attestation --new-bundle-format --bundle sigstore.att.json --digestAlg sha256 --digest="b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9" --certificate-identity $email --certificate-oidc-issuer https://accounts.google.com --use-signed-timestamps
```

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -157,30 +158,50 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
-		// and determine if we want to store certificates for those as well.
+		var keypair sign.Keypair
+		var ephemeralKeypair bool
+		var idToken string
+		var err error
+
+		// Set to false so sigstore-go fetches the Fulcio certificate,
+		// otherwise SignerFromKeyOpts would through the old signing path
+		issueCertForKey := c.IssueCertificateForExistingKey
+		c.IssueCertificateForExistingKey = false
```
steiza: I don't understand this code block. What flag to I pass to `cosign attest-blob` to specify this behavior? Where does this get used by sigstore-go?

Hayden-IO: Because we have too much duplication everywhere, I had missed adding `--issue-certificate` to `attest-blob` and `attest` - I just added this.

For context, `--issue-certificate` was added as part of https://blog.sigstore.dev/adopting-sigstore-incrementally-1b56a69b8c15/, to request a Fulcio certificate even if a key is provided. 

This flag isn't used by sigstore-go, it's used by `SignerFromKeyOpts`, which does a few things - it load a key from KMS/PIV/disk, it generates a key, and if `IssueCertificateForExistingKey` is set or an ephemeral key was generated, it requests a certificate from Fulcio. This is the root cause of the issue, the method does too much.

sigstore-go will be handling certificate requests, so I need to set `IssueCertificateForExistingKey` to `false` to prevent `SignerFromKeyOpts` from trying to generate a cert.

steiza: Ahhhh, okay, now I see. Still, mutating the configuration like this makes me a bit nervous (I've been bitten before in the `cosign` code base doing similar things!)

The "problem" is that `SignerFromKeyOpts` does two things:
- Constructs the `SignerVerifier`
- Calls `keylessSigner` in some cases

There aren't that many places we call `SignerFromKeyOpts`. Instead of mutating the config like this, would it make sense to have `keylessSigner` become a public method and call it when needed from the `cmd/cosign/cli` files? I'm not 100% sure if that's better or not, but I think it would allow us to avoid mutating the configuration like this.

Hayden-IO: Yea, I'm happy to try this out.
FYI @ret2libc, this would also solve the e2e test failure you were running into.

Hayden-IO: I refactored this and then searched for method usage across GitHub and found it's being used in some libraries. So one of two options:
* Call this out in the CHANGELOG - we provide no guarantees for API breakages, and because the method signature has changed, it'll force clients to make an update 
* Check if `SigningConfig` is set in `SignerFromKeyOpts` - This is kinda gross

internal/key/svkeypair.go
```diff
@@ -0,0 +1,122 @@
...
+	case *rsa.PublicKey:
+		keyAlg = "RSA"
+	case ed25519.PublicKey:
+		keyAlg = "ED25519"
+	default:
+		return nil, errors.New("unsupported key type")
+	}
+
+	// TODO: Ideally SignerVerifier would return its configured hash.
+	algo, err := signature.GetDefaultAlgorithmDetails(pubKey)
```
ret2libc: we should use `defaultLoadOptions = cosign.GetDefaultLoadOptions()` and pass `defaultLoadOptions` to this method, to ensure that ed25519ph is used instead of ed25519.

steiza (approved):
This looks good to me!

cmd/cosign/cli/attest/attest_blob.go
steiza: Yeah, I think a CHANGELOG entry is just fine.

internal/key/svkeypair.go
Hayden-IO: Thanks, added!

Hayden-IO:
Re-verified all commands work with after the rebase. Signing with a non-ecdsa-p256 key doesn't work with Rekor v2, but that's a [bug](https://github.com/sigstore/sigstore-go/blob/b9e47831ad9f0ada7a859ab92a224e5399221acd/pkg/sign/transparency.go#L128) in sigstore-go I'm fixing now, no need to block this PR.

steiza (approved):

PR #4386, opened 2025-09-04
Bump sigstore-go, support alternative hash algorithms with keys

haydentherapper:
sigstore-go now handles non-ECDSA-P-256 signatures with Rekor v2. To support verification, we also need a way to provide alternative hash algorithms to the default SHA-256. cosign verify already had a flag for this, so I added the flag to all verify commands. In the future, when we are only processing bundles, we can lookup the default hash algorithm given the key.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,9 @@ module github.com/sigstore/cosign/v2
 
 go 1.24.6
 
+// TODO: Remove
+replace github.com/sigstore/sigstore-go => github.com/haydentherapper/sigstore-go v0.0.0-20250904203112-9f828c3bdc96
+
 require (
 	cuelang.org/go v0.14.1
 	github.com/ThalesIgnite/crypto11 v1.2.5
--- a/cmd/cosign/cli/options/verify.go
+++ b/cmd/cosign/cli/options/verify.go
@@ -137,6 +137,7 @@ type VerifyAttestationOptions struct {
 	CertVerify          CertVerifyOptions
 	Registry            RegistryOptions
 	Predicate           PredicateRemoteOptions
+	SignatureDigest     SignatureDigestOptions
 	Policies            []string
 	LocalImage          bool
 }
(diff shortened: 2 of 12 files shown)


Hayden-IO:
Tested with:

```
openssl ecparam -genkey -name secp521r1 -noout -out ecdsa-p521-private.pem
cosign import-key-pair --key ecdsa-p521-private.pem
cosign sign-blob --new-bundle-format --bundle artifact.sigstore.json --yes --use-signing-config --key cosign.key README.md
cosign verify-blob --new-bundle-format --bundle artifact.sigstore.json --key cosign.pub --signature-digest-algorithm sha512 README.md
```

go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/sigstore/cosign/v2
 
 go 1.24.6
 
+// TODO: Remove
+replace github.com/sigstore/sigstore-go => github.com/haydentherapper/sigstore-go v0.0.0-20250904203112-9f828c3bdc96
```
Hayden-IO: Note: Will replace once https://github.com/sigstore/sigstore-go/pull/520 is in

steiza:
So in https://github.com/sigstore/sigstore-go/pull/520#discussion_r2326048942 we decided to not support ed25519ph with Fulcio for now, but with https://github.com/sigstore/cosign/pull/4050 we implemented `adaptSignerVerifierToFulcio` in `cmd/cosign/cli/sign/sign.go`.

As part of these changes, should we partially roll back the `adaptSignerVerifierToFulcio` changes that were made in #4050? cc @ret2libc

Hayden-IO:
> So in [sigstore/sigstore-go#520 (comment)](https://github.com/sigstore/sigstore-go/pull/520#discussion_r2326048942) we decided to not support ed25519ph with Fulcio for now, but with #4050 we implemented `adaptSignerVerifierToFulcio` in `cmd/cosign/cli/sign/sign.go`.
> 
> As part of these changes, should we partially roll back the `adaptSignerVerifierToFulcio` changes that were made in #4050? cc @ret2libc

I'm fine with leaving this in, and documenting this is a change between Cosign v2 and v3 when we switch over signing to sigstore-go.

steiza:
The test coverage shows up as high, but I don't think we have an automated test that covers verifying with a SHA512 digest? We should have at least one test that covers that case (if we don't already!)

> I'm fine with leaving this in, and documenting this is a change between Cosign v2 and v3 when we switch over signing to sigstore-go.

I'm not wild about releasing a feature we plan to immediately deprecate 🙃 but I'm also not sure how easy it is to untangle these things.

Hayden-IO:
Oh sorry, I was thinking we had merged the adapter in an earlier PR and it was already in a released version. Lemme take a pass at this tomorrow.

PR #4402, opened 2025-09-11
Add a terminal spinner while signing with sigstore-go

Hayden-IO:
Recording signatures to Rekor v2 can take up to 10 seconds. We want to avoid someone killing the process while waiting for a response from Rekor, otherwise the user will have to re-sign the artifact.

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -206,7 +206,7 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 			Data:        payload,
 			PayloadType: "application/vnd.in-toto+json",
 		}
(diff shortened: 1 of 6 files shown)


Hayden-IO:
https://github.com/user-attachments/assets/92f4e7b5-d796-48f6-8148-b02b11cfe602

bobcallaway:
perhaps too nuanced, but is there any benefit to having a visualization break down into something like:
- if calling fulcio, print a dot
- print dot when signature is created
- then use a spinner to denote an upload?

Hayden-IO:
I'd like a more fine-grained UX, but we couldn't call sigstore-go's `sign.Bundle` then, we'd have to copy in that function's code into Cosign and wrap each outbound call. Happy to explore this later on as part of additional UX cleanup.

PR #4404, opened 2025-09-12
Add CHANGELOG for v2.6.0

Hayden-IO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 v2.6.0 introduces a number of new features, including:
 
-* Signing an in-toto statement rather than Cosign constructing one from a predicate (#4306)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,59 @@
+# v2.6.0
+
+v2.6.0 introduces a number of new features, including:
+
+* Signing an in-toto statement rather than Cosign constructing one from a predicate (#4306)
```
steiza: Minor: verifying by providing a digest (instead of a file on disk) is also quite useful and might be notable here, for when the subject of your in-toto statement isn't a file on disk (like a git commit hash for SLSA source provenance).

But we also illustrate this in the example below, so I don't feel strongly that it must be called out here.

Hayden-IO: Updated to include this!

PR #4414, opened 2025-09-16
Fix signing and verification with ed25519 keys with bundles and Rekor

Hayden-IO:
With the recent changes we made to use sigstore-go rather than Cosign for signing and verification, ed25519 managed key support broke, because we were incorrectly specifying ed25519ph for dsse Rekor entries and not specifying ed25519ph for hashedrekord entries. This PR correctly sets load options for when signing and verifying a blob (using the prehash variant) and when signing/verifying attestations (using the pure variant). This also fixes a bug where the SignerVerifier Keypair didn't handle crypto.Hash(0) for ed25519, which specifies no hash when signing.

This has been tested with sign/verify, sign-blob/verify-blob, attest/verify-attestation, and attest-blob/verify-blob-attestation.

### DIFF
--- a/internal/key/svkeypair_test.go
+++ b/internal/key/svkeypair_test.go
@@ -34,6 +34,7 @@ import (
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
 )
 
 // mockSignerVerifier is a mock implementation of signature.SignerVerifier for testing.
@@ -61,7 +62,7 @@ func (m *mockSignerVerifier) VerifySignature(_, _ io.Reader, _ ...signature.Veri
 	return errors.New("not implemented")
 }
 
-func TestNewKMSKeypair(t *testing.T) {
+func TestNewSignerVerifierKeypair(t *testing.T) {
 	ecdsaPriv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
 		t.Fatalf("failed to generate ecdsa key: %v", err)
@@ -78,6 +79,7 @@ func TestNewKMSKeypair(t *testing.T) {
 	testCases := []struct {
 		name      string
 		sv        signature.SignerVerifier
+		prehash   bool
 		expectErr bool
 		errMsg    string
 	}{
@@ -103,6 +105,14 @@ func TestNewKMSKeypair(t *testing.T) {
 			expectErr: false,
 		},
 		{
+			name: "ED25519ph key",
+			sv: &mockSignerVerifier{
+				pubKey: ed25519Priv.Public(),
+			},
+			prehash:   true,
+			expectErr: false,
+		},
+		{
 			name: "Unsupported key type",
 			sv: &mockSignerVerifier{
 				pubKey: "not a key",
@@ -122,7 +132,11 @@ func TestNewKMSKeypair(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			kp, err := NewSignerVerifierKeypair(tc.sv, nil)
+			var loadOpts []signature.LoadOption
+			if tc.prehash {
+				loadOpts = []signature.LoadOption{options.WithED25519ph()}
+			}
+			kp, err := NewSignerVerifierKeypair(tc.sv, &loadOpts)
 			if tc.expectErr {
 				if err == nil {
 					t.Errorf("expected an error, but got none")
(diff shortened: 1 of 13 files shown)


Hayden-IO:
```
head -c 128 < /dev/urandom > artifact
cosign sign-blob --signing-config signing_config_v0.2.json --trusted-root trusted_root.json --bundle artifact.sigstore.json --yes --key cosign.key artifact
cosign verify-blob --new-bundle-format --trusted-root trusted_root.json --bundle artifact.sigstore.json --key cosign.pub artifact

cosign attest-blob --signing-config signing_config_v0.2.json --trusted-root trusted_root.json --statement intoto.txt --bundle intoto.sigstore.json --yes --key cosign.key
cosign verify-blob-attestation --bundle="intoto.sigstore.json" --key=cosign.pub --trusted-root trusted_root.json --digest="b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde8" --digestAlg="sha256"

cosign sign --signing-config signing_config_v0.2.json --trusted-root trusted_root.json --yes --key cosign.key localhost:5050/busybox:latest
cosign verify --new-bundle-format --key cosign.pub  --trusted-root trusted_root.json localhost:5050/busybox:latest

cosign attest --predicate predicate.json --signing-config signing_config_v0.2.json --trusted-root trusted_root.json --yes --key cosign.key localhost:5050/busybox:latest
cosign verify-attestation --new-bundle-format --key cosign.pub  --trusted-root trusted_root.json localhost:5050/busybox:latest
```

ret2libc (approved):
LGTM!

steiza:
I'm so confused 😅 so help me check my understanding.

Fulcio does not support ed25519ph, but Rekor does. And if someone is using a key they manage (instead of Fulcio), we want to support ed25519ph and use that with Rekor as well.

In #4386 we removed some ed25519ph support, but did so in a way that broke "pure" ed25519 for self-managed keys. So we still aren't supporting ed25519ph with Fulcio, but we are fixing ed25519 for self-managed keys, and you can also use a self-managed ed25519ph key with Rekor. Did I get that right?

internal/key/svkeypair_test.go
```diff
@@ -61,7 +62,7 @@ func (m *mockSignerVerifier) VerifySignature(_, _ io.Reader, _ ...signature.Veri
 	return errors.New("not implemented")
 }
 
-func TestNewKMSKeypair(t *testing.T) {
+func TestNewSignerVerifierKeypairKeypair(t *testing.T) {
```
cmurphy: Is the double `Keypair` a typo?

internal/key/svkeypair_test.go
```diff
@@ -102,6 +104,14 @@ func TestNewKMSKeypair(t *testing.T) {
 			},
 			expectErr: false,
 		},
+		{
+			name: "ED25519ph key",
```
cmurphy, line 108: This test passes for me even without all the other changes, I'm not sure it's proving that a bug was fixed

Hayden-IO:
Thanks y'all for the comments. I was a little preemptive putting this up, I need to fix e2e tests (they're failing because when the tlog isn't used, we use ed25519 - the prehash variant is only for hashedrekord).

@steiza, correct. This is to support self-managed ed25519 keys. A summary of what's supported and what's not:

Supported:
* ed25519 key, signing an artifact
* ed25519 key for an artifact + Rekor (ed25519ph + `hashedrekord`) <-- This is the only time the prehash variant is used
* ed25519 key signing an attestation
* ed25519 key for an attestation + Rekor (ed25519 + `dsse`)
* For completeness, ECDSA/RSA + Fulcio (or without Fulcio) + Rekor (or without Rekor) for artifacts and DSSEs

Maybe supported:
* ed25519 key signing an attestation + Fulcio + Rekor (ed25519 + `dsse`) - This should be possible but we might need to update how the SignerVerifier is loaded. I'll test this out.

Not supported:
* ed25519 key + Fulcio + Rekor (ed25519ph + fulcio + hashedrekord, since Fulcio doesn't know how to verify ed25519ph signatures)

internal/key/svkeypair_test.go
Hayden-IO: Ah, good catch. This is because I was passing a nil pointer as the load options, which caused the ed25519 test key to load the prehash signer variant. Without the changes, passing a pointer to an empty array causes the ed25519 test to fail as expected.

Hayden-IO:
@aaronlew02 For your recent PR, did you test if ed25519 keys now work? I'm hoping I can just close this out.

aaronlew02:
My PR (#4880) also addressed (albeit slightly differently) handling a non-prehashed algorithm in `SignData` to prevent a panic, so your fix and its test can likely be removed here. However, my PR only enabled signing with Ed25519 keys in `sign-blob`, so it looks like the rest of your changes are still necessary to enable the use of Ed25519 keys elsewhere.

PR #4415, opened 2025-09-17
Support multiple --sign-container-identity

### LINKED ISSUE #4140: Issue with --output-certificate in cosign sign-blob

**My question is:**

emilyzheng:
#### Summary

With this change, cosign sign can be run only once when an image has multiple pull references.

Closes #4140

#### Release Note

`cosign sign` supports more than one --sign-container-identity.


cmd/cosign/cli/options/sign.go
```diff
@@ -137,7 +137,7 @@ func (o *SignOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.IssueCertificate, "issue-certificate", false,
 		"issue a code signing certificate from Fulcio, even if a key is provided")
 
-	cmd.Flags().StringVar(&o.SignContainerIdentity, "sign-container-identity", "",
+	cmd.Flags().StringArrayVar(&o.SignContainerIdentities, "sign-container-identity", nil,
```
bobcallaway: I'd use `StringSliceVar` here so you can specify the values as:

`--sign-container-identity=docker.io/name1:tag1,docker.io/name2:tag2`

emilyzheng: Thanks, updated.

emilyzheng:
Added an e2e test.

Hayden-IO:
I wanted to mention that the codepath updated here is no longer going to be the default with Cosign v3. Have you explored implementing this when the new-bundle-format flag is used?

emilyzheng:
> I wanted to mention that the codepath updated here is no longer going to be the default with Cosign v3. Have you explored implementing this when the new-bundle-format flag is used?

Signing with `--new-bundle-format=true` doesn't seem to care about what `--sign-container-identity` specifies (SignContainerIdentity is used in signDigest, but not used in signDigestBundle). 

It makes sense because `cosign verify --new-bundle-format=true` always return the inputted image as docker-reference. It's unnecessary to sign the image with multiple identities to get expected docker-reference:
```
### Different tags
$ cosign verify --new-bundle-format=true --key sigstore.pub quay.io/emilyzheng/ubi9/php-82:1 | jq -r .[].critical.identity.'"docker-reference"'
quay.io/emilyzheng/ubi9/php-82:1

$ cosign verify --new-bundle-format=true --key sigstore.pub quay.io/emilyzheng/ubi9/php-82:2 | jq -r .[].critical.identity.'"docker-reference"'
quay.io/emilyzheng/ubi9/php-82:2

### Registry proxy
$ cosign verify --new-bundle-format=true --key sigstore.pub registry:4000/emilyzheng/ubi9/php-82:1 | jq -r .[].critical.identity.'"docker-reference"'
registry:4000/emilyzheng/ubi9/php-82:1
```

cmurphy (dismissed):

emilyzheng:
Does this need further reviews? Can it be merged?

cmd/cosign/cli/sign/sign.go
```diff
@@ -246,14 +256,21 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
...
-	if err != nil {
-		return err
-	}
+	ociSigs := make([]oci.Signature, len(payloads))
+	b64sigs := make([]string, len(payloads))
 
-	b64sig, err := ociSig.Base64Signature()
-	if err != nil {
-		return err
+	for i, payload := range payloads {
```
bobcallaway: can we use a different variable name than `payload`? we're shadowing the input arg.

Issue #4372, opened 2025-08-28
Output when verifying a container with the new bundle format is missing content


haydentherapper:
**Description**

**Version**

The output from `cosign verify --new-bundle-format --certificate-identity=$email --certificate-oidc-issuer=https://accounts.google.com --use-signed-timestamps localhost:5050/busybox:latest`:

```
[{"critical":{"identity":{"docker-reference":""},"image":{"docker-manifest-digest":""},"type":""},"optional":null}]
```

Note the empty strings.

crispysipper:
I experienced similar behavior with something like this:

```
cosign sign "${image}@${DIGEST}" \
  --yes \
  --new-bundle-format=true \
  --identity-token $SIGSTORE_ID_TOKEN \
  --fulcio-url <redacted> \
  --rekor-url <redacted> \
  --output-signature sig.json
```

the sign itself passes, nothing is actually written to `--output-signature` when `--new-bundle-format=true`

perhaps this is by design, I am just now getting familiarized with the new-bundle-format.

Hayden-IO:
That's by design, though we'll need to add checks that one or the other is set.

cc @steiza something for v3

steiza:
This week is a little crazy, but I took some time this morning to at least debug what was happening and fill in part of the information; see #4416. If this is urgent, someone could take that direction and finish it; otherwise it'll probably be a week or two before I can look at this again.

steiza:
This should now be fixed!

PR #4416, opened 2025-09-18
Partially populate the output of cosign verify when working with new bundles

steiza:
#### Summary

For #4372. Note that currently this only fills the the `critical` fields of the `docker-reference`, `docker-manifest-digest`, and `type`, and not any of the `optional` fields (which could be populated from sigstore-go verification results).

#### Release Note

* Updated `cosign verify` output when working with new bundles to include information about the verified container image

#### Documentation

N/A

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -54,7 +54,6 @@ import (
 	ociremote "github.com/sigstore/cosign/v2/pkg/oci/remote"
 	"github.com/sigstore/cosign/v2/pkg/oci/static"
 	"github.com/sigstore/cosign/v2/pkg/types"
-	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	"github.com/sigstore/rekor/pkg/generated/client"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	rekor_types "github.com/sigstore/rekor/pkg/types"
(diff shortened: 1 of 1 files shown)


steiza:
As it stands, this pull request will return something like:

```
[{"critical":{"identity":{"docker-reference":"ghcr.io/steiza/nginx:latest"},"image":{"docker-manifest-digest":"sha256:2d5b42edd09430be531f850d10c43bfb50087b8afa81eb8d83df4ee02b2d3a4b"},"type":"https://sigstore.dev/cosign/sign/v1"},"optional":null}]
```

steiza:
I should have known that #4316 was a little too easy! The content returned by `VerifyImageAttestations()` is very tailored to `verify-attestations`, and so when I reused that function for `verify` it didn't understand the JSON encoded information it was receiving.

I'm not a huge fan of this approach, but the alternative is pretty major surgery to `pkg/cosign/verify.go`, either making copy-pasting quite a bit of code to create a parallel codepath for `verify`, or refactoring the functions `VerifyImageAttestations()` calls to return much more information and then filter that down.

I think transforming the information we need in the cmd works for now, and when we simplify code paths to just use sigstore-go for verification it will be much easier to plumb the information we need back to the cmd, instead of transforming the information we receive today.

haydentherapper (approved):
Nice!

PR #4428, opened 2025-09-26
Fetch service URLs from the TUF PGI signing config by default

haydentherapper:
This will also use sigstore-go's signing API by default.

### DIFF
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -125,23 +125,29 @@ jobs:
         trustedroot=$(find ./repository/targets -name "*.trusted_root.json")
         echo "trustedroot=$trustedroot" >> $GITHUB_ENV
 
-    - name: Set SigningConfig
+    - name: Create SigningConfig
       run: |
-        signingconfig=$(find ./repository/targets -name "*.signing_config.v0.2.json")
-        echo "signingconfig=$signingconfig" >> $GITHUB_ENV
+        ./cosign signing-config create \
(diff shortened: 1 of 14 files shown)


Hayden-IO:
Test failing because we prefer --use-signing-config (using the public instance) over --signing-config (what's provided), will update.

.github/workflows/kind-verify-attestation.yaml
```diff
@@ -120,23 +120,34 @@ jobs:
...
+            --rekor-config="ANY" \
+            --tsa="url=${TSA_URL}/api/v1/timestamp,api-version=1,start-time=2024-01-01T00:00:00Z,operator=sigstore.dev" \
+            --tsa-config="EXACT:1" \
+            --out signingconfig.json
+        echo "signingconfig=signingconfig.json" >> $GITHUB_ENV
+
     - name: Sign demoimage with cosign
       run: |
-        ./cosign sign --rekor-url ${REKOR_URL} --fulcio-url ${FULCIO_URL} --yes --allow-insecure-registry ${demoimage} --identity-token ${OIDC_TOKEN}
+        ./cosign sign --signing-config=${signingconfig} --trusted-root=${trustedroot} --yes --allow-insecure-registry ${demoimage} --identity-token ${OIDC_TOKEN}
```
steiza, line 141: Minor: like @cmurphy was saying in https://github.com/sigstore/cosign/pull/4318/files#r2349940915, since we call `cosign initialize ...` above, we shouldn't need to set the `--trusted-root` here.

🤷 our tests are definitely due for an overhaul as we change the default behavior.

Hayden-IO: Agreed, let's clean this up as part of a test overhaul.

PR #4440, opened 2025-10-02
Create changelog for v3.0.0

haydentherapper:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+# v3.0.0
+
+Announcing the next major release of Cosign!
+
+Cosign v3 is a minor change from Cosign v2.6.x, with all of the new capabilities of recent
+releases **on by default**, but will still allow you to disable them if you need the older functionality.
+These new features include support for the standardized bundle format (`--new-bundle-fomat`), providing roots
+of trust for verification and service URLs for signing via one file (`--trusted-root`, `--signing-config`),
+and container signatures stored as an OCI Image 1.1 referring artifact.
+
+Learn more on our [v3 announcement blog post](https://blog.sigstore.dev/cosign-3-0-available/)! See
+the changelogs for [v2.6.0](#v260), [v2.5.0](#v250), and [v2.4.0](#v240) for more information on recent
+changes.
+
+If you have any feedback, please reach out on Slack or file an issue on GitHub.
+
+## Changes
+
+* Default to using the new protobuf format (#4318)
+* Fetch service URLs from the TUF PGI signing config by default (#4428)
+* Bump module version to v3 for Cosign v3.0 (#4427)
+
 # v2.6.1
 
 ## Bug Fixes


Hayden-IO:
Will merge after https://github.com/sigstore/cosign/pull/4428. Also depends on https://github.com/sigstore/sigstore-blog/pull/89 being merged.

steiza (approved):

vszakats:
Homebrew just bumped to cosign 3.0.0. 3.0.0 doesn't yet appear on the sigstore/cosign [release page](https://github.com/sigstore/cosign/releases).
The ChangeLog link added in this PR leads to a 404: https://blog.sigstore.dev/cosign-3-0-available/

This wouldn't be an issue, but 3.0.0 broke our CI with this error:
```
Error: must provide --bundle with --signing-config or --use-signing-config
error during command execution: must provide --bundle with --signing-config or --use-signing-config
```
https://github.com/curl/curl-for-win/actions/runs/18323148696/job/52181010177#step:3:10903
The command is: `cosign sign-blob -y --key=<key> --output-signature=<file>.cosign <file>`

Is there a pointer on what changed and what needs to be updated downstream?

steiza:
@vszakats sorry about this!

The shortest answer is that cosign v3 does indeed have breaking changes, but you can still go back to the old behavior with:

```
cosign sign-blob --new-bundle-format=false ...
```

However, we would strongly encourage looking at moving to the new bundle format soon (maybe when builds aren't breaking). I think you probably want to do something like this:

```
cosign sign-blob -y --key=cosign.key --bundle=all-linux-8.17.0-DEV_2449a5aa7497075494ad8316a6fa755467f1ef8b.sigstore.json all-linux-8.17.0-DEV_2449a5aa7497075494ad8316a6fa755467f1ef8b.zip
```

You can see the blog post draft describing the changes at https://github.com/sigstore/sigstore-blog/pull/89/files?short_path=fb897a4#diff-fb897a4aa79b17c0f08e8417fe87da2427f05735b20bc7a65ab415af420325d7.

vszakats:
@steiza Thanks for the examples and pointer, and your quick response!

If I understand correctly `.sigstore.json` is replacing `.cosign` as the "sidecar" file we are publishing and people can use to verify the blob. Also allowing offline verification.

The linked blog draft lead me terrified when reading `--new-bundle-format` was added by 2.6.0 (released ~3 weeks ago). But then saw the 2.5.0 release notes also listing it. If so, it would save a lot of headaches as 2.5.0 is the version present on the latest Debian, for example.

vszakats:
Just confirmed, this command-line does work with both 2.5.0 and 3.0.0:
```sh
cosign sign-blob -y --key=<key> --new-bundle-format --bundle=<file>.sigstore.json <file>
```
Ref: https://github.com/curl/curl-for-win/actions/runs/18325729261/job/52189603760

edit: with the command above 2.5.0 dumps the `.cosign` payload to stdout. This doesn't happen with 3.0.0. When adding `--output-signature=<file>.cosign` to the command-line, 2.5.0 creates the `.cosign` file as expected. In 3.0.0 this option is silently ignored and there is no `.cosign` file generated.

PR #4448, opened 2025-10-08
choose different signature filename for KMS-signed release signatures

bobcallaway:

### DIFF
--- a/README.md
+++ b/README.md
@@ -763,4 +763,4 @@ process](https://github.com/sigstore/.github/blob/main/SECURITY.md)
 
 ## Bundle files in GitHub Release Assets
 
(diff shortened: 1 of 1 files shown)


bobcallaway:
> Can we also link to the format of the bundle, since if a user wants to verify without Cosign they’ll need to know how to interpret that bundle?

do you have a link?

Hayden-IO:
https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto. Actually, can we link to https://docs.sigstore.dev/cosign/system_config/installation/#verifying-cosign-with-artifact-key, and then update that?

PR #4449, opened 2025-10-08
Deduplicate code in sign/attest* and verify* commands

cmurphy:
The sign/sign-blob/attest/attest-blob and verify/verify-blob/verify-attestation/verify-blob-attestation commands contained a lot of code that was either copy and pasted to each command, or, more insidiously, functionally the same but written slightly differently between commands. This PR is an effort to reduce the existing duplication, make common updates simpler, prevent the commands from further diverging, and overall reduce LOC. The result is hopefully that the remainder of the code left in each command is mostly specific to that command itself.

This PR is best reviewed commit-by-commit, but can be merged by squash-and-merge.

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -16,13 +16,9 @@
 package verify
 
 import (
-	"bytes"
 	"context"
 	"crypto"
-	"crypto/x509"
-	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -30,18 +26,14 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/in-toto/in-toto-golang/in_toto"
-	"github.com/sigstore/cosign/v3/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/sign"
 	cosignError "github.com/sigstore/cosign/v3/cmd/cosign/errors"
-	"github.com/sigstore/cosign/v3/internal/ui"
-	"github.com/sigstore/cosign/v3/pkg/blob"
 	"github.com/sigstore/cosign/v3/pkg/cosign"
 	"github.com/sigstore/cosign/v3/pkg/oci"
 	"github.com/sigstore/cosign/v3/pkg/oci/static"
 	sigs "github.com/sigstore/cosign/v3/pkg/signature"
 	"github.com/sigstore/protobuf-specs/gen/pb-go/dsse"
-	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature/payload"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/signcommon/common.go
```diff
@@ -389,3 +393,45 @@ func signerFromNewKey() (*SignerVerifier, error) {
...
+		return nil, nil, fmt.Errorf("rfc3161 timestamp is nil")
+	}
+	if ko.NewBundleFormat {
+		return timestampBytes, rfc3161Timestamp, nil
+	}
+	ts, err := json.Marshal(rfc3161Timestamp)
+	if err != nil {
+		return nil, nil, fmt.Errorf("marshalling timestamp: %w", err)
+	}
+	if err := os.WriteFile(ko.RFC3161TimestampPath, ts, 0600); err != nil {
```
Hayden-IO, line 439: When calling `cosign sign` or `cosign attest`, this file doesn't need to be written since it's included as an annotation on the image.

It looks like we support fetching a timestamp only with the new bundle format for `cosign sign` (I think this might have been a regression, I recall supporting it before the bundle format was added), but we support it with the old and new bundle format for `cosign attest`. Can we gate whether we write this file on if we're signing a container vs a blob?

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -169,52 +164,18 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		return nil
```
Hayden-IO: Do you think the above should be de-duped as well? There's commonality between attest, sign and attest-blob with the construction of the attestation, but the latter is written to a file rather than OCI. Not sure if this deduplication is worth it to add complexity to `WriteNewBundleWithSigningConfig`, which I guess would have to conditionally write to OCI or a file.

cmd/cosign/cli/verify/verify.go
```diff
@@ -151,21 +149,9 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-		co.TrustedMaterial, err = cosign.TrustedRoot()
-		if err != nil {
-			ui.Warnf(ctx, "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
-		}
+	err = SetTrustedMaterial(ctx, c.TrustedRootPath, c.CertChain, c.CARoots, c.CAIntermediates, c.TSACertChainPath, co)
+	if err != nil {
+		return fmt.Errorf("setting trusted material: %w", err)
 	}
 
 	if err = CheckSigstoreBundleUnsupportedOptions(*c, co); err != nil {
```
Hayden-IO, line 149: To confirm, we call `SetTrustedMaterial` before `CheckSigstoreBundleUnsupportedOptions` because we need to know if a trusted root was fetched when using the bundle format? Just wondering if we can exit out earlier, but seems like no.

cmd/cosign/cli/attest/attest_blob.go
cmurphy: Yes I'll update these

Hayden-IO:
This is fantastic! Thank you for splitting this change into the many commits, made this so straightforward to review!

cmd/cosign/cli/verify/verify.go
cmurphy: Right, this is also consistent with its behavior before.

cmd/cosign/cli/signcommon/common.go
cmurphy: > When calling cosign sign or cosign attest, this file doesn't need to be written since it's included as an annotation on the image.

I see where I went wrong, in `cosign sign` this function is only called by `signDigestBundle` when `NewBundleFormat` is set, so this path is skipped so no file is written. It's not the case for `cosign attest` which is my mistake.

> It looks like we support fetching a timestamp only with the new bundle format for cosign sign (I think this might have been a regression, I recall supporting it before the bundle format was added),

Fetching a timestamp without the new bundle format should still work, it's part of this I think: https://github.com/sigstore/cosign/blob/dccda706ef56cd6f8a40a8fad8992ff46fb218de/cmd/cosign/cli/sign/sign.go#L426-L437

> Can we gate whether we write this file on if we're signing a container vs a blob?

Yes will fix

PR #4456, opened 2025-10-10
Cache signing config when calling initialize

Hayden-IO:
When calling cosign initialize, the client will cache the trusted root file if available. This PR adds support for caching the signing config as well. The public-good instance's TUF repo includes this file. Private deployments likely don't use this file, so like with the trusted root, Cosign will print a warning rather than fail initialization.


cmd/cosign/cli/initialize/init_test.go
cmurphy: This should have signing_config.v0.2.json too

Hayden-IO: Whoops, thanks, added.

PR #4457, opened 2025-10-10
Deprecate offline flag

### LINKED ISSUE #4454: Offline verification makes network calls
**Description**

We have some CI jobs that perform lots of verifications using the `cosign` tool in offline mode. Our CI is in GitHub actions, which means that it's not airgapped, but we still want to test this code path.

What surprised us is that `cosign verify-blob --offline` can:
1. Perform network calls to get some TUF certificates.
2. Fail the operation, if the above calls fail.

We have been running this CI jobs for months now, and this issue has manifested once or twice. Here's a command that failed recently:

```
cosign verify-blob --offline --key /Users/runner/work/dangerzone/dangerzone/share/freedomofpress-dangerzone.pub --bundle var/folders/vk/nx37ffx50hv5djclhltc26vw0000gn/T/tmp5c90_qkr /var/folders/vk/nx37ffx50hv5djclhltc26vw0000gn/T/tmp75kgy45p
```

and its error:

> Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: error getting live trusted root: failed to create TUF client failed to load metadata: tuf refresh failed: Get "https://tuf-repo-cdn.sigstore.dev/13.root.json": dial tcp: lookup tuf-repo-cdn.sigstore.dev: no such host
> Error: getting Rekor public keys: updating local metadata and targets: error updating to TUF remote mirror: tuf: failed to download 13.root.json: Get "https://tuf-repo-cdn.sigstore.dev/13.root.json": dial tcp: lookup tuf-repo-cdn.sigstore.dev: no such host
> error during command execution: getting Rekor public keys: updating local metadata and targets: error updating to TUF remote mirror: tuf: failed to download 13.root.json: Get "https://tuf-repo-cdn.sigstore.dev/13.root.json": dial tcp: lookup tuf-repo-cdn.sigstore.dev: no such host

At the same time, we have verified that this command works in an airgapped mode, so it seems to me that there's a code path in `cosign` that does not ignore network errors, if it operates in offline mode, whereas other paths do

My questions are:
1. Is the `--offline` mode a truly offline mode, that makes absolutely no network requests, or is it a mode where `cosign` relies on the signed timestamps, but may update its certificate chain from a CDN?
2. Should `cosign` fail in a case like this?

**Version**

```
cosign version

GitVersion:    v2.5.2
GitCommit:     af5a988bb15a03919ccaac7a2ddcad7a9d006f38
GitTreeState:  clean
BuildDate:     2025-06-17T20:03:14Z
GoVersion:     go1.24.4
Compiler:      gc
Platform:      linux/amd64
```

Note that the platform is not that important, since we've seen it fail both on macOS and Linux. Also, we purposefully use an older `cosign` version, to check how it behaves after a while (e.g. here, it triggered a certificate refresh).

Hayden-IO:
The offline flag is misleading and is a no-op with the new Cosign v3 defaults. The flag's purpose was to prevent a client from falling back to verifying an artifact's inclusion in Rekor when a proof failed to verify. Most users thought offline verification forced the client to not make any network requests - a very reasonable assumption, but with TUF, network requests are a part of verification if the local TUF metadata has expired.

I've updated the README as well, though we need to make a far more comprehensive pass over the documentation since it's out of date given our new trusted-root/bundle flags.

Fixes #4454

### DIFF
--- a/cmd/cosign/cli/options/verify.go
+++ b/cmd/cosign/cli/options/verify.go
@@ -37,7 +37,8 @@ type CommonVerifyOptions struct {
 
 func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.Offline, "offline", false,
-		"only allow offline verification")
+		"only verify an artifact's inclusion in a transparency log using a provided proof, rather than querying the log. May still include network requests to retrieve service keys from a TUF repository")
+	_ = cmd.Flags().MarkDeprecated("offline", "To verify in an airgapped environment, provide a --bundle with the signature and verification material, and a --trusted-root file with the service keys and certificates")
 
 	cmd.Flags().StringVar(&o.TSACertChainPath, "timestamp-certificate-chain", "",
 		"path to PEM-encoded certificate chain file for the RFC3161 timestamp authority. Must contain the root CA certificate. "+
--- a/README.md
+++ b/README.md
@@ -141,6 +141,14 @@ The following checks were performed on these signatures:
 
 ### Verify a container in an air-gapped environment
 
+**Note:** This section is out of date.
+
+**Note:** Most verification workflows require periodically requesting service keys from a TUF repository.
+For airgapped verification of signatures using the public-good instance, you will need to retrieve the
+[trusted root](https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json) file from the production
+TUF repository. The contents of this file will change without notification. By not using TUF, you will need
+to build your own mechanism to keep your airgapped copy of this file up-to-date.
+
 Cosign can do completely offline verification by verifying a [bundle](./specs/SIGNATURE_SPEC.md#properties) which is typically distributed as an annotation on the image manifest.
 As long as this annotation is present, then offline verification can be done.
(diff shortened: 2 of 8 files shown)


cmd/cosign/cli/options/verify.go
```diff
@@ -37,7 +37,8 @@ type CommonVerifyOptions struct {
 
 func (o *CommonVerifyOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.Offline, "offline", false,
-		"only allow offline verification")
+		"[deprecated] only verify an artifact's inclusion in a transparency log using a provided proof, rather than querying the log. May still include network requests to retrieve service keys from a TUF repository")
```
cmurphy: FYI it looks like using `MarkDeprecated` hides the flag from the `--help` output so I'm not sure it matters what's in this help text.

If it does appear somewhere, it might be better to change `[deprecated]` to `DEPRECATED` to match the deprecation format of the `--attachment` flag.

Hayden-IO: Good point, I've removed `[deprecated]` since the flag help text is no longer used. Looking around, we also have `[DEPRECATED]` for `--sig-only` - I'll look at standardizing as we deprecate more flags.

PR #4458, opened 2025-10-10
Deprecate tlog-upload flag

### LINKED ISSUE #4503: cosign does not seem to honor --tlog-upload=false
**Description**

Cosign does not seem to honor --tlog-upload=false. Below is the command I executed

```
cosign sign --tlog-upload=false --key awskms:///arn:aws:kms:us-west-2:0123456789123:key/mrk-dummy --verbose=true ttl.sh/<redacted>@sha256:<redacted>
```

I have also set the below environment variable based on some references I saw online ;- 
```
rams3sh@rams3sh-MacBook-Pro test % env | grep -i cosign
COSIGN_PRIVATE_INFRASTRUCTURE=true
```

I get the below snippet in the debug logs :- 
```
2025/10/28 09:37:51 --> PATCH https://ttl.sh/v2/<redacted>?_state=<redacted>
2025/10/28 09:37:51 PATCH /v2/<redacted>/blobs/uploads/<redacted>?_state=<redacted> HTTP/1.1
Host: ttl.sh
User-Agent: cosign/v3.0.2 (darwin; arm64) go-containerregistry/v0.20.6
Content-Length: 4361
Content-Type: application/octet-stream
Accept-Encoding: gzip

{"mediaType":"application/vnd.dev.sigstore.bundle.v0.3+json", "verificationMaterial":{"publicKey":{"hint":"<redacted>"}, "tlogEntries":[{"logIndex":"647594071", "logId":{"keyId":"<redacted>"}, "kindVersion":{"kind":"dsse", "version":"0.0.1"}, "integratedTime":"1761624467", "inclusionPromise": .......
```
I checked the transparency logs in the below link :- 
https://search.sigstore.dev/?logIndex=647594071

And it seems like the entry has been created.

I am not sure if I am doing anything wrong in the above command execution. 

**Version**

```

GitVersion:    v3.0.2
GitCommit:     84449696f0658a5ef5f2abba87fdd3f8b17ca1be
GitTreeState:  "clean"
BuildDate:     2025-10-10T18:17:56Z
GoVersion:     go1.25.2
Compiler:      gc
Platform:      darwin/arm64
```

Hayden-IO:
Clients that don't want to use a transparency log should provide a signing config without tlog service instances rather than use this flag. This provides a detailed error message for what clients should do. 

Fixes #4503

### DIFF
--- a/cmd/cosign/cli/signcommon/common.go
+++ b/cmd/cosign/cli/signcommon/common.go
@@ -655,7 +655,7 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
 		return fmt.Errorf("cannot specify service URLs and use signing config")
 	}
 	if (useSigningConfig || signingConfigPath != "") && !tlogUpload {
-		return fmt.Errorf("--tlog-upload=false is not supported with --signing-config or --use-signing-config. Provide a signing config without a transparency log service")
+		return fmt.Errorf("--tlog-upload=false is not supported with --signing-config or --use-signing-config. Provide a signing config with --signing-config without a transparency log service, which can be created with `cosign signing-config create` or `curl https://raw.githubusercontent.com/sigstore/root-signing/refs/heads/main/targets/signing_config.v0.2.json | jq 'del(.rekorTlogUrls)'` for the public instance")
 	}
 	// Signing config requires a bundle as output for verification materials since sigstore-go is used
 	if (useSigningConfig || signingConfigPath != "") && !newBundleFormat && bundlePath == "" {
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -852,7 +852,7 @@ func TestSignVerifyWithTUFMirror(t *testing.T) {
 	}
 }
 
-func prepareSigningConfig(t *testing.T, fulcioURL, rekorURL, oidcURL, tsaURL string) string {
+func prepareSigningConfig(t *testing.T, fulcioURL, rekorURL, oidcURL, tsaURL string) string { //nolint: unparam
 	startTime := "2024-01-01T00:00:00Z"
 	fulcioSpec := fmt.Sprintf("url=%s,api-version=1,operator=fulcio-op,start-time=%s", fulcioURL, startTime)
 	rekorSpec := fmt.Sprintf("url=%s,api-version=1,operator=rekor-op,start-time=%s", rekorURL, startTime)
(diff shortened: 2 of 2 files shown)


Hayden-IO:
Hold to merge until https://github.com/sigstore/cosign/pull/4449 is in, then I'll rebase.

Hayden-IO:
Rebased and refactored common logic into a shared method.

Hayden-IO:
Going to update to override the signing config with a printed warning. I don't love doing this, but there's clearly a lot of usage still of tlog-upload and we didn't deprecate this in v2.6

Hayden-IO:
@cmurphy @steiza This is ready to go.

I initially had implemented this to err out like how we err out if any service URLs are provided along with a signing config. However, I think there's far more users specifying this flag while using the public instance than users who are providing URLs for private deployments. I chose to override the tlog service URLs in the signing config instead and log a deprecation message instead.

cmurphy:
> Clients that don't want to use a transparency log should provide a signing config without tlog service instances rather than use this flag

This actually seems kind of unfriendly to me. It feels like a lot of overhead to have to edit the signing config that cosign downloaded for you, which will get overwritten if you edited it in-place. The use case is often that the user is just trying out a cosign command and doesn't want to add a few more kilobytes of junk to the public log, so having to make them jump through this hoop is going to be confusing.

Hayden-IO:
> This actually seems kind of unfriendly to me. It feels like a lot of overhead to have to edit the signing config that cosign downloaded for you, which will get overwritten if you edited it in-place. The use case is often that the user is just trying out a cosign command and doesn't want to add a few more kilobytes of junk to the public log, so having to make them jump through this hoop is going to be confusing.

Transparency is a key part of Sigstore, and so putting a significant speed bump in place is reasonable. I don't think users are using this flag to avoid adding junk. We see this flag mentioned in blog posts online. It could be because of privacy concerns (users can use a key rather than certificate then), or because of private infrastructure, or I think because of Cosign v1 (when all the blog posts were written) defaulting to no Rekor so the posts mentioned the flag for parity. This is also a concern for verification - either users are defaulting to using the "insecure ignore tlog" flag, which should be strongly discouraged, or there's not a lot of verification so this hasn't come up frequently.

We need a well-lit path for private infrastructure. What's great about using the signing config and trusted root is that we can remove these flags and instead signers can explicitly configure what infra they use and verifiers can configure what's trusted, and we don't have to build a different path.

Regardless, what's implemented here is just a deprecation warning for now.

Hayden-IO:
@steiza Following up from the Go meeting (finally, sorry!), I reverted the changes to not mutate the signing config and instead err out. The error message is now more detailed and provides a one-liner to grab the latest signing config and drop tlogs.

crispysipper:
a 13 day deprecation warning wasn't much of a buffer, guys

Many of us use templating to support reusable CI/CD jobs and this is a breaking change now that it defaults to erroring out

Hayden-IO:
Hey, sorry for the breakage, but this was resolving an issue as the flag's value was being ignored. This flag should have been marked as deprecated when we released v3.0.0.

PR #4462, opened 2025-10-13
Remove --out flag from `cosign initialize`

apyrgio:
#### Summary

Remove any mention of the `--out` flag from the `cosign initialize` command, since it's no longer used.

#### Release Note

NONE


apyrgio:
Docgen failed because `make docgen` adds a stray newline at the end of the file 😕 . My editor linters removed it, so I'm adding it back.

Issue #4470, opened 2025-10-14
Missing or no-op Cosign v3 equivalent commands


apyrgio:
While experimenting a bit with Cosign v3, I realized that there are some Cosign v2 commands that no longer have an equivalent, if an image was signed with the new Sigstore bundle format:

* `cosign [download|attach] signature`: It will fail since it expects to read/write a manifest with tag `sha256-<digest>.sig`
* `cosign save`: It will download the image layers, but no the Sigstore bundle

Is there a plan to make these commands work with Cosign v3 or v4 in the future?

Hayden-IO:
Also flagging that `cosign tree` has not been updated.

b3n3d17:
Another [issue-4507](https://github.com/sigstore/cosign/issues/4507) also mentions that the `cosign delete` command may have the same problem

steiza:
`cosign save` and `delete` have not been updated, but now that https://github.com/sigstore/cosign/pull/4477 has landed I think everything else mentioned here is covered.

Hayden-IO:
Also https://github.com/sigstore/cosign/issues/4553, `cosign triangulate`

Munken:
Also #4564 `cosign copy`

steiza:
I believe we have now implemented support for protobuf bundles in all the commands listed, or otherwise have added deprecation warnings with `oras` / `cosign` commands you can use to get a similar result.

PR #4477, opened 2025-10-20
Add support to download and attach for protobuf bundles

steiza:
#### Summary

Continuing to make progress on #4470.

This adds protobuf bundle support for `cosign download attestation` and `cosign download signature` (which will work if you do `cosign sign` with the new protobuf format), as well as adding support for `cosign attach attestation` and `cosign attach signature`.

This does not add any support (yet) for `cosign save`.

#### Release Note

* Added protobuf bundle support to subcommands `download attestation`, `download signature`, `attach attestation`, and `attach signature` (where you should use the new `--bundle` flag).

#### Documentation

N/A

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -3290,6 +3290,7 @@ func TestSignDownloadAttachNewBundle(t *testing.T) {
 	ko := options.KeyOpts{KeyRef: privKeyPath, PassFunc: passFunc}
 	so := options.SignOptions{
 		NewBundleFormat: true,
+		Upload:          true,
 	}
 
 	must(sign.SignCmd(ro, ko, so, []string{imgName}), t)


cmurphy (commented):
> Support for attach signature would be somewhat strange, as it did not even support the old cosign bundle format (just raw signatures and certificates). If you do cosign sign and then cosign download signature, you should attach it with cosign attach attestation 

I'm finding this pretty confusing, and I think users will have a hard time understanding it. Could we make it so that `attach signature` works for bundles?

Alternatively - I think the problem stems from the naming, where downloading/attaching bundles just isn't very well described by the words "attestation" or "signature" any more. What about adding a new subcommand `attach bundle` instead of trying to coerce `attach signature` and `attach attestation` to do this?

steiza:
> I'm finding this pretty confusing, and I think users will have a hard time understanding it.

This was my first time using these commands and looking at how they were implemented, so I agree 😅. These commands are old (relative to the age of Cosign), implemented in 2021 and not really updated since 2023, so it's not surprising that Sigstore usage has evolved while these commands have not.

> Could we make it so that `attach signature` works for bundles?

I think this is the most promising path forward. Looking over the flags again, `--payload` is closest to a bundle. We could re-use `--payload` and check if it's a new protobuf bundles, and / or add a `--bundle` alias for `--payload` to make it clearer what's going on.

> Alternatively - I think the problem stems from the naming, where downloading/attaching bundles just isn't very well described by the words "attestation" or "signature" any more.

I agree generally, but I think we should wait a bit before tackling this problem. I want to make sure Cosign v3 is working for folks first, helping them transition to protobuf bundles, trusted roots, and signing configs, before we start changing the command set.

But this is very much in scope for Cosign v4, and it's not just `cosign attach signature` / `cosign attach attestation`! With protobuf bundles, `cosign sign` and `cosign attest` are also quite similar, and (as you point out) it's not clear we need separate `sign` / `attest` commands generally. But this is a larger change, and I think we should hold off for a bit to when we're ready to iterate on the set of commands as a whole.

cmd/cosign/cli/attach/attach.go
```diff
@@ -37,16 +38,58 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, signed
 	}
 
 	for _, payload := range signedPayloads {
-		if err := attachAttestation(ctx, ociremoteOpts, payload, imageRef, regOpts.NameOptions()); err != nil {
+		fmt.Fprintf(os.Stderr, "Using payload from: %s", payload)
```
cmurphy, line 41: Tiny nit - why `fmt.Fprintf` here instead of `ui.Infof`?

steiza: I just moved this line out from inside `attachAttestation()` 🤷 - we definitely have a mix of the two throughout Cosign.

I'm going to leave this as-is, but this would be good to clean up at some point.

PR #4489, opened 2025-10-22
Pass insecure registry flags through to referrers

### LINKED ISSUE #4466: Image sign with flag  --allow-http-registry=true fails
**Description**

Using to sign image with http local registry fails with flags `--allow-http-registry=true` or/and ` --allow-insecure-registry=true`. With version 2.6-2.6.1 signing works.

Versions: 3.0-3.0.2

cmurphy:
Fixes #4466

- [x] Tests

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
@@ -85,6 +86,9 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	ociremoteOpts, err := c.RegistryOptions.ClientOpts(ctx)
 	if err != nil {
 		return err
+	}
+	if c.RegistryOptions.AllowHTTPRegistry || c.RegistryOptions.AllowInsecure {
+		ociremoteOpts = append(ociremoteOpts, ociremote.WithNameOptions(name.Insecure))
 	}
 	digest, err := ociremote.ResolveDigest(ref, ociremoteOpts...)
 	if err != nil {
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -186,6 +186,7 @@ jobs:
       run: go test -tags=e2e,registry -v ./test/...
       env:
         COSIGN_TEST_REPO: insecure-registry.notlocal:5001
+        TUF_ROOT_JSON: ${{ github.workspace }}/root.json
 
     - name: Setup local insecure OCI 1.1 registry
       run: |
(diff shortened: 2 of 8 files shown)


steiza:
I'm not sure if this is working right, or if I'm doing something wrong while testing this.

I'm working off the instructions on https://github.com/tstromberg/sigstore-the-local-way?tab=readme-ov-file#11-starting-a-local-registry.

I think `cosign` does something special with `localhost`, because even on `main` this works:
```
$ go run cmd/cosign/main.go verify --key cosign.pub --allow-insecure-registry=true --allow-http-registry=true localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a:latest 

Verification for localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a:latest --
...
```

But if I set up a DNS alias in `/etc/hosts`:

```
$ cat /etc/hosts | grep 'asdf.example.com'
127.0.0.1 asdf.example.com
$ ping asdf.example.com
PING asdf.example.com (127.0.0.1): 56 data bytes
```

... then verification fails, both on `main` and with these changes:
```
$ go run cmd/cosign/main.go verify --key cosign.pub --allow-insecure-registry=true --allow-http-registry=true asdf.example.com:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a:latest 
Error: no signatures found
```

Again, I might be doing something wrong here!

cmurphy:
> I think cosign does something special with localhost, because even on main this works:

Yep, cosign treats localhost and *.local as special because of https://github.com/google/go-containerregistry/pull/125

>  then verification fails, both on main and with these changes:

I think `no signatures found` could be a bundle/no bundle problem, was that image signed with an old version of cosign? Does it work when you add `--new-bundle-format=false`?

steiza:
> I think no signatures found could be a bundle/no bundle problem, was that image signed with an old version of cosign? Does it work when you add --new-bundle-format=false?

I don't think this is it, because I'm running off an up-to-date `main` and your branch. The only thing I'm changing between the verification commands is if the container starts with `localhost` or `asdf.example.com` (which resolves to `127.0.0.1`).

But just to be sure, I checked again and got the same behavior.

cmurphy:
@steiza you were right, and I fixed the tests to catch that issue.

cmd/cosign/cli/attest/attest.go
```diff
@@ -86,6 +87,9 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	if err != nil {
 		return err
 	}
+	if c.RegistryOptions.AllowHTTPRegistry || c.RegistryOptions.AllowInsecure {
+		ociremoteOpts = append(ociremoteOpts, ociremote.WithNameOptions(name.Insecure))
```
steiza, line 91: Minor: so our `RegistryOptions.AllowHTTPRegistry` maps to `name.Insecure` which both mean "it's okay if the registry uses plain HTTP without TLS.

But my interpretation of `RegistryOptions.AllowInsecure` is that the registry is using TLS, but with a self-signed certificate (or at least a certificate that doesn't verify with the machine's certificate store).

🤷 Are there people actually using that configuration? Or is everyone just using an HTTP registry? I guess we'll find out! I didn't see a "use TLS but allow certificates that don't verify with the machine's certificate store" on https://pkg.go.dev/github.com/google/go-containerregistry@v0.20.6/pkg/name.

cmurphy: Yeah, it didn't seem like go-containerregistry makes a distinction between them.

Issue #3599, opened 2024-03-14
Attestations require uploading entire payload to rekor


jonjohnsonjr:
**Description**

Recently ran into an issue where we have an absurdly large attestation (130MB) that is rejected by rekor due to its size.

Arguably, this is a bug in rekor, but I think it's really a bug in how cosign attestations work. Both the dsse and intoto record types require uploading the entire attestation payload to rekor. If they're small enough, rekor will store the intoto attestation. AFAICT, rekor doesn't store the dsse payload at all, but it _does_ require you to upload the entire thing.

Cosign should support a form of attestation that doesn't require the entire payload be uploaded to rekor.

This is part of why I initially proposed an OCI descriptor as the payload format, since embedding `data` is an optional thing. I think sigstore ended up using dsse so that the payload type could be signed, but I can't remember the details.

Anyway, it would be great if someone could tell me I'm just holding this wrong or something, but I'd love an easy way to attest a large thing. Should I just attest an OCI descriptor that points to the large thing? Is there an existing pattern that people use for this?

It would take a lot of load off of rekor if it could just verify a hash instead of the entire payload, so maybe what we need is a form of `cosign attest` that stores and verifies a hashedrekord? Does that even make sense? If someone can hold all of this in their head, I'd love your opinions on what the right path forward would be.

bobcallaway:
> Arguably, this is a bug in rekor, but I think it's really a bug in how cosign attestations work. Both the dsse and intoto record types require uploading the entire attestation payload to rekor. If they're small enough, rekor will store the intoto attestation. AFAICT, rekor doesn't store the dsse payload at all, but it _does_ require you to upload the entire thing.

Originally we had envisioned Rekor storing the full attestation (and implemented the `intoto` type to do just that), but after some experience and hindsight realized that it was probably better to not position Rekor to need full search capabilities (since folks want to be able to look up all log entries /attestations for arbitrary subjects). We added the `dsse` type to clean up some code issues, as well as start enforcing the pattern that Rekor wouldn't store the attestation when an entry was made with that type.

One way you could think of a DSSE envelope is that it has "embedded" signatures (similar to an RPM or an APK file), since the payload and signature are part of the same JSON document. 

[...]

PR #4490, opened 2025-10-22
feat(ISV-6377): let cosign use HashedRekord for TLOG verification of attestations

BorekZnovustvoritel:
Assisted-by: claude-4.5-sonnet (Cursor)
Signed-off-by:  Marek Szymutko <mszymutk@redhat.com>

Addresses the verification part of #3599 

#### Summary

Without this feature, the hashedRekords cannot be used as TLOG entries for attestations because the `cosign verify-attestation` would fail. The problem is described in #3599 but the brief summary is:

- `cosign attest` uses DSSE or in-toto payloads, sends the whole payload to the Rekor server
- In case of SBOMs, the attestations can be too large to be processed by Rekor, disallowing the usage of Rekor for attesting SBOMs
- This change only addresses the verification part. With this change, the command `cosign verify-attestation` now also supports HashedRekord TLOG entries, which makes sure that a digest is sufficient payload to be uploaded to Rekor.
- Additional changes are required for the full functionality, `cosign attest` doesn't support uploading the HashedRekord entries. A workaround must be used as of now, the workaround is described in [this write-up](https://github.com/BorekZnovustvoritel/Cosign-Rekor-Demo) (with the only difference that when using the version of Cosign built from this PR, the verification succeeds)

#### Release Note

* New feature: Support for HashedRekords TLOG entries was added to `cosign verify-attestation` 

#### Documentation

No user-facing API was changed in this PR, I believe comments in code are a sufficient change.

### DIFF
--- a/pkg/cosign/tlog.go
+++ b/pkg/cosign/tlog.go
@@ -23,6 +23,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"hash"
@@ -31,7 +32,8 @@ import (
 	"strings"
 
 	"github.com/go-openapi/strfmt"
-	"github.com/go-openapi/swag"
+	"github.com/go-openapi/swag/conv"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/cosign/v3/internal/ui"
 	"github.com/sigstore/cosign/v3/pkg/cosign/bundle"
 	"github.com/sigstore/cosign/v3/pkg/cosign/env"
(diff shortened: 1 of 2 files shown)


arewm:
@steiza, fya as I mentioned this to you on a call a couple weeks back.

Hayden-IO:
Hey y'all, sorry to leave this PR hanging. I will get back to it in a week or so. The approach seems reasonable, but I need to dedicate a bit of time to think through this and if there's any unintended consequences.

BorekZnovustvoritel:
(I'm just heartbeating so this doesn't become stale or it doesn't auto-close or something) :smile:

BorekZnovustvoritel:
Closing because the underlying issue has been resolved by another PR.

PR #4495, opened 2025-10-24
Support stdout output for attest-blob bundles

### LINKED ISSUE #4494: Support writing bundles to stdout
**Description**

I was messing around with using cosign to write multiple bundles to a [jsonl](http://jsonlines.org/). I'm not entirely sure if that's a good pattern or useful thing yet, but a nice way to support it would be if cosign allowed writing bundles to stdout, like:

```
cosign attest-blob --predicate ./predicate.json --bundle -  ./my-blob.bin
``` 

Then, I could redirect that to append to a jsonl file or pipe it to other things for other unknown use cases.

Perhaps, `--bundle` should *default* to `-` to assume stdout for attesting. (However, it probably shouldn't default to `-` for *stdin* for verifying, b/c people would get confused when they omitted it and the program just hung waiting for stdin input.)

ralphbean:
#### Summary

This enables `attest-blob --bundle=-` to write bundles to stdout with a trailing newline, allowing users to create JSONL files containing multiple attestations by redirecting and appending output.

This change adds support for the convention of using "-" to represent stdout. When the bundle path is "-", the bundle is written to stdout instead of a file, and the signature output is suppressed to avoid conflicts.

Changes:
- Add stdout detection in attest/attest_blob.go and signcommon/common.go
- Suppress signature output when bundle goes to stdout
- Add comprehensive test coverage in attest_blob_test.go
- Update flag description and add JSONL example to documentation

Example usage, appending two predicates about the same blob to the same jsonl file.
```
cosign attest-blob --key key.key --predicate pred1.json \ --type slsaprovenance --bundle=- blob.txt >> attestations.jsonl
cosign attest-blob --key key.key --predicate pred2.json \ --type slsaprovenance --bundle=- blob.txt >> attestations.jsonl
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closes #4494

#### Release Note

Added support for writing blob attestations to stdout.

#### Documentation

I don't think this needs any special docs update, but I'm open to it if you think there's somewhere it should be noted.

### DIFF
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -186,10 +186,15 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 			}
 		}
 
-		if err := os.WriteFile(c.BundlePath, contents, 0600); err != nil {
-			return fmt.Errorf("create bundle file: %w", err)
+		// If BundlePath is "-", write to stdout with trailing newline
+		if c.BundlePath == "-" {
+			fmt.Fprintln(os.Stdout, string(contents))
+		} else {
+			if err := os.WriteFile(c.BundlePath, contents, 0600); err != nil {
+				return fmt.Errorf("create bundle file: %w", err)
+			}
+			fmt.Fprintln(os.Stderr, "Bundle wrote in the file ", c.BundlePath)
 		}
-		fmt.Fprintln(os.Stderr, "Bundle wrote in the file ", c.BundlePath)
 	}
 
 	if c.OutputSignature != "" {
@@ -197,7 +202,8 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 			return fmt.Errorf("create signature file: %w", err)
 		}
 		fmt.Fprintf(os.Stderr, "Signature written in %s\n", c.OutputSignature)
-	} else {
+	} else if c.BundlePath != "-" {
+		// Only output signature to stdout if bundle is not going to stdout
 		fmt.Fprintln(os.Stdout, string(bundleComponents.SignedPayload))
 	}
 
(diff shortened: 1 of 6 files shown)


steiza (commented):
I like the idea of `cosign` supporting a way to construct a `bundles.sigstore.jsonl` file, but I'm hesitant to assume that the only `stdout` from subcommands like `attest-blob` will be the bundle itself.

Would it make sense to have these subcommands **append** to the `--bundle` path, instead of just calling `os.WriteFile()`, which truncates the file? Then, adapting the example you provided you'd do something like:

```
cosign attest-blob --key key.key --predicate pred1.json --type slsaprovenance --bundle=attestations.sigstore.jsonl blob.txt 
cosign attest-blob --key key.key --predicate pred2.json --type slsaprovenance --bundle=attestations.sigstore.jsonl blob.txt
```

I'm curious as to what other Cosign maintainers think. Also, if we made this change, we'd probably want to update `attest-blob` and `sign-blob` to behave in the same way.

ralphbean:
> I'm hesitant to assume that the only stdout from subcommands like attest-blob will be the bundle itself.

Yeah, there's an old issue lurking about this at  https://github.com/sigstore/cosign/issues/2510

ralphbean:
> Would it make sense to have these subcommands append to the --bundle path, instead of just calling os.WriteFile(), which truncates the file? 

As a user, this sounds nice. Consider, is it a breaking change? A user might theoretically be depending on that truncation, but in practice I can't imagine it mattering. At most, somebody's local flow might change if they were depending on writing an attestation over and over again as a test.

Is there a type-ambiguity? If you call attest-blob once with --bundle, you get a json file where .json is an appropriate suffix. If you call it twice, you get a json lines file where .jsonl is an appropriate suffix.

Perhaps an explicit `--append` option, defaulting to false would be the right choice (although, it does inflate the number of options and I believe there's an effort underway or done with v3 to reduce the number of options).

Issue #4474, opened 2025-10-18
`--upload=false` doesn't work in cosign v3


caarlos0:
**Description**

In v2.x, `cosign sign --upload=false` would skip pushing anything to the registry.
Judging by goreleaser tests, it now tries to push it anyway.

See: 

- https://github.com/goreleaser/goreleaser/blob/8c0e69642db37a8e49ce5353b38416ff03ee9468/internal/pipe/sign/sign_docker_test.go#L57
- https://github.com/goreleaser/goreleaser/actions/runs/18616945525/job/53082755040#step:23:109

It is failing with UNAUTHORIZED errors because tests are not logged in into the registry on purpose. Point is, it shouldn't have tried to upload...

**Version**

v3.0.2

caarlos0:
thank you! any plans for a release?

Hayden-IO:
We want to wrap up a few more of the open issues, then we'll cut a release.

PR #4499, opened 2025-10-25
Add --output-bundle and fix --upload for new bundle

cmurphy:
Without this change, --upload=false was not respected with the new bundle format. It also would not have made sense because there was no way to output the bundle locally. This change adds a flag --output-bundle so that the bundle can be created on disk without attaching it to the image, and also passes through the Upload parameter to bypass uploading it if desired.

Fixes https://github.com/sigstore/cosign/issues/4474

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -117,8 +117,17 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		return err
 	}
 
+	bundleOpts := signcommon.CommonBundleOpts{
+		Payload:       payload,
+		Digest:        digest,
+		PredicateType: types.CosignSignPredicateType,
+		BundlePath:    c.BundlePath,
+		Upload:        !c.NoUpload,
+		OCIRemoteOpts: ociremoteOpts,
+	}
+
 	if c.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, payload, digest, types.CosignSignPredicateType, c.BundlePath, !c.NoUpload, c.SigningConfig, c.TrustedMaterial, ociremoteOpts...)
+		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/cosign/cli/options/sign.go
```diff
@@ -97,6 +98,10 @@ func (o *SignOptions) AddFlags(cmd *cobra.Command) {
 		"write the certificate to FILE")
 	_ = cmd.MarkFlagFilename("output-certificate", certificateExts...)
 
+	cmd.Flags().StringVar(&o.OutputBundle, "output-bundle", "",
+		"write the bundle to FILE")
```
steiza: Minor: in `sign-blob` and `attest-blob` we just call this `--bundle`. Arguably `--output-bundle` is more descriptive, but `--bundle` would be more consistent. I slightly lean towards consistency here, but open to other opinions.

Separately, taking inspiration from `sign-blob` maybe the description could be something like:

```
write everything required to verify the image to a FILE
```

cmd/cosign/cli/attest/attest.go
```diff
@@ -118,7 +118,7 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	}
 
 	if c.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, payload, digest, types.CosignSignPredicateType, "", c.SigningConfig, c.TrustedMaterial, ociremoteOpts...)
+		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, payload, digest, types.CosignSignPredicateType, "", !c.NoUpload, c.SigningConfig, c.TrustedMaterial, ociremoteOpts...)
```
steiza: Minor: this is starting to be a **lot** of arguments to the `signcommon.Write*` functions! I'm fairly new to Go, but I think the idiomatic thing would be at some point to move these into a `signcommon.WriteBundleWithSigningConfigOptions` or some such. I'm not convinced this has to change in this pull request though, because this is an internal method that's easy to refactor later if needed.

steiza (commented):
Just a few minor comments, otherwise LGTM. Holding off on approving until you have a chance to respond (and don't forget to run `make docgen`!)

cmd/cosign/cli/attest/attest.go
cmurphy: I added a second commit to create a separate struct for these functions.

haydentherapper (commented):
Looks great! Should we also add the `--bundle` flag to `cosign attest` for uniformity?

cmurphy:
@haydentherapper done

Issue #4504, opened 2025-10-28
Annotations are not being displayed while verification


rabajaj0509:
**Description**

In release v3.0.2, the -a, --annotations option is no longer shown in the optional field of the hash during verification. This behavior was present in v2.6.1.

Command used to sign:
```
cosign sign --tlog-upload --use-signing-config=false --key "$KEY_FILE_PATH" -a metadata_hash="$METADATA_HASH" "$IMAGE_WITH_DIGEST"
```

Commmand used for verification:
```
./cosign-darwin-arm64 verify --insecure-ignore-tlog --key cosign.pub <registry>/<image>:<tag>
```
Output:
```
"optional":null}]
```

rabajaj0509:
Expected Output:
```
,"optional":{"metadata_hash":"ea6d50dcc0c5f54022544d3ba5c9043298674d86e1aa34db4d962bf4c13a1aaf"}}]
```

PR #4508, opened 2025-10-31
Fix signing/verifying annotations for new bundle

cmurphy:
Without this change, when the new bundle format is used, annotations were not being added to the payload, nor were they being checked during verification, but were still being reported as verified. This also had a side effect of the annotations not appearing in the verification output. This change fixes the issue by passing through the annotations to the statement builder during signing, and using a different in-toto `Statement` type to parse annotations from the statement during verification, as well as actually calling the claims verifier.

Fixes https://github.com/sigstore/cosign/issues/4504
The missing annotations in the display output was the least of the problems - the annotations were not being added or verified at all.

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -47,6 +47,7 @@ import (
 	"github.com/sigstore/cosign/v3/pkg/types"
 	sigPayload "github.com/sigstore/sigstore/pkg/signature/payload"
 	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	// Loads OIDC providers
 	_ "github.com/sigstore/cosign/v3/pkg/providers/all"
@@ -114,7 +115,7 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 				return fmt.Errorf("accessing image: %w", err)
 			}
 			if signOpts.NewBundleFormat {
-				err = signDigestBundle(ctx, digest, ko, signOpts)
+				err = signDigestBundle(ctx, digest, ko, signOpts, annotations)
 			} else {
 				err = signDigest(ctx, digest, staticPayload, ko, signOpts, annotations, se)
 			}
@@ -137,7 +138,7 @@ func SignCmd(ro *options.RootOptions, ko options.KeyOpts, signOpts options.SignO
 			}
 			digest := ref.Context().Digest(d.String())
 			if signOpts.NewBundleFormat {
-				err = signDigestBundle(ctx, digest, ko, signOpts)
+				err = signDigestBundle(ctx, digest, ko, signOpts, annotations)
 			} else {
 				err = signDigest(ctx, digest, staticPayload, ko, signOpts, annotations, se)
 			}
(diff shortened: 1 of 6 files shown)


Hayden-IO:
Caveat that I haven't looked at annotations in a long time - I thought that annotations were persisted in OCI, but it looks like no, they're just provided by users as additional data to sign over, and so they also must be provided during verification. This approach using in-toto differs because we're persisting the annotations as well, and arguably this is better because we're storing everything that was signed over.

Is this going to be a surprise to users? Does this risk inflating the attestation size? I'm not sure how most users are using annotations. The docs suggest to use it for signing over tags, so my guess is this is fine.

My only concern is that I haven't seen in-toto statements with annotations before (didn't realize this was supported by the spec!), so this might be an issue for cross-client verification that doesn't fully implement the in-toto spec.

cmurphy:
> I thought that annotations were persisted in OCI, but it looks like no, they're just provided by users as additional data to sign over,

What makes you say this?

Annotations were indeed persisted in OCI for non-protobuf-format signatures before this change:

```
$ go run cmd/cosign/main.go sign --new-bundle-format=false --use-signing-config=false -y --key cosign.key -a foo=bar localhost:5000/wolfi-base:latest
...
$ crane blob localhost:5000/wolfi-base@$(crane manifest localhost:5000/wolfi-base:sha256-ae238a181d95804645919b2671d50ae77477efbfb299544491346e2911125aaf.sig | jq -r .layers[0].digest) | jq .optional
```

That is also why they appeared in the verification output even if you weren't verifying annotations:

```
$ go run cmd/cosign/main.go verify --key cosign.pub localhost:5000/wolfi-base:latest | jq .[0].optional

Verification for localhost:5000/wolfi-base:latest --
{
  "Bundle": {
    "SignedEntryTimestamp": "MEUCICGmzQx/ryF/tIMNMnrgN4qxGoCIdwvjsSRXkntNIQeZAiEAgpd+9xnKN+7UBcv2FPn5Rw1t0uYm1A/rO3ih0guQReQ=",
    "Payload": {
      "body": "eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJraW5kIjoiaGFzaGVkcmVrb3JkIiwic3BlYyI6eyJkYXRhIjp7Imhhc2giOnsiYWxnb3JpdGhtIjoic2hhMjU2IiwidmFsdWUiOiJlYmU3Y2JiY2RiOGYzNjgzMDYzYjE1Yzc0ODhmMTRlZmRjMGRkYzU2MTVjMTRlMzhmNWIzY2NmMWQyMTRlODQyIn19LCJzaWduYXR1cmUiOnsiY29udGVudCI6Ik1FUUNJRGpudFh6QSttc0d6RTY0TFc3NmZSemRVd1lIb3Z3TlVvZ01URU5xQzh0akFpQlU3THNrRVM5VGxxWHFRNnh1ekJWTVd1OExaaFRLQ1dpUDh3OXM2RGcrVnc9PSIsInB1YmxpY0tleSI6eyJjb250ZW50IjoiTFMwdExTMUNSVWRKVGlCUVZVSk1TVU1nUzBWWkxTMHRMUzBLVFVacmQwVjNXVWhMYjFwSmVtb3dRMEZSV1VsTGIxcEplbW93UkVGUlkwUlJaMEZGTkc0emJHYzNjV0pOYWtkSmNrMHpNR3Q2V21sV1JFRlFlblZ0Y3dwelIwTmxVbFpOVlRST1NGZENSVXh6VFVSbVVIVkpZMlV6YkdaWFkxTndlV3BTSzB0NlNtTXlaRzVsU2tkT1EwMVFZMDVQT1d4WVpHOVJQVDBLTFMwdExTMUZUa1FnVUZWQ1RFbERJRXRGV1MwdExTMHRDZz09In19fX0=",
      "integratedTime": 1762198597,
      "logIndex": 663838986,
      "logID": "c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d"
    }
  },
  "foo": "bar"
}
```

Hayden-IO:
You're totally right, I missed [this code](https://github.com/sigstore/cosign/blob/8e7e0571a0ccad90a883da3676562eb6e1fe5ab4/cmd/cosign/cli/sign/sign.go#L349) where the OCI signatures (which include the user provided annotations) is uploaded.

haydentherapper (approved):
This looks good!

Issue #4506, opened 2025-10-30
Cosign cli doesn't pass in a cancellable context


frezbo:
**Description**

When signing multiple images, cosign cli cannot be `Ctrl + C`'ed since the `SignCmd` creates it's own context here https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign.go#L66

It's better to create a `signal.NotifyContext` in https://github.com/sigstore/cosign/blob/main/cmd/cosign/main.go#L38 here and pass it around, so all downstream context can be cancelled on signal interrups

cmurphy:
@frezbo it seems like you have a solid grasp on what needs to be fixed here, are you interested in making the PR?

frezbo:
I can take a look when I've some free time

PR #4514, opened 2025-11-09
fix: Use signal context

frezbo:
fix: Use signal context for `sign` cli package.

Fixes: #4506

Use proper context passing for `sign` cli package.

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -16,7 +16,6 @@
 package cli
 
 import (
-	"context"
 	"fmt"
 	"os"
 
@@ -94,7 +93,7 @@ race conditions or (worse) malicious tampering.
 
 		Args:             cobra.MinimumNArgs(1),
 		PersistentPreRun: options.BindViper,
-		RunE: func(_ *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			switch o.Attachment {
 			case "sbom":
 				fmt.Fprintln(os.Stderr, options.SBOMAttachmentDeprecation)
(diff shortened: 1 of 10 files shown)


cmd/cosign/main.go
```diff
@@ -35,7 +36,9 @@ import (
 
 func main() {
 	// Fix up flags to POSIX standard flags.
-	ctx := context.Background()
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
```
frezbo: this is the main fix

bobcallaway: this should also look for `SIGTERM` too, since we are distributing cosign in a container form factor

frezbo: fixed 👍

frezbo:
> I see you've made a lot of style changes and touched 62 files with this commit. Could you consider condensing this PR into a [single purpose commit](https://lxsameer.com/essays/git-etiquette#org4593396)? This would make it easier to review because there is less surface area to cover, and it keeps the [git log history](https://itnext.io/best-coding-practices-for-git-e9335c95f21d) clean which is really important for debugging with git blame or git bisect. The file mode formats, variable initializations and whitespace changes are a bit subjective and are not relevant to the bug you're fixing. If you feel like these changes are important, could you open a separate PR and also ensure there is a linter rule to make sure these style changes are objectively enforced going forward?

it's just doing standard gofmt formatting which seems like a standard practice to follow. I'll just change this PR just to fix passing signal context and only update tests that touch it, I'll leave up to the maintainers on fixing further lint issues

frezbo:
I've kept the changes to minimal and only updated tests that needs updating

frezbo:
Seems i missed some tests, updated them

PR #4525, opened 2025-11-10
Fix OCI with local certs

### LINKED ISSUE #4513: Signing is not working with self-signed certificate
**Question**
I want to sign an image in AWS ECR using cosign. I have generated a self signed certificate using `openssl` that provided me a private key and a .crt file, but to sign using local keys I need to import the private key to cosign format.

`cosign import-key-pair --key private-key.pem`

When I try to sign and attach a certificate, seems like it is omitting the certificate and signing only using the key, because the cert was generated with unconverted key and it does not match while signing.

`cosign sign --key <my-imported.key> --certificate <self-signed.crt> <Image>`

Is there a way to achieve signing using my own cert(self-signed)? I don't see any doc mentioning it is not possible.

FYI: I don't want to go with keyless signing and use Fulcio CA for issuing the certificate.

cmurphy:
This is now two commits - one fixing the old bundle, one fixing the new one when signing config is used.

# Fix cert attachment for new bundle with signing config

Without this change, if an image is signed with `--certificate` and also
using signing config and the new bundle format (the defaults), the logic
routes to sigstore-go signing which does not natively understand signing
with a cert that is not issued by Fulcio. This means that only the
public key is uploaded to the bundle, so the image cannot be verified
using the certificate. This change ensures the certificate is passed
through to the bundle if it is provided with `sign`, `attest`, and
`attest-blob` . The `sign-blob` command does not support
`--certificate`.

# Fix OCI verification with local cert - old bundle

`SigVerifier` is set on `CheckOpts` no matter whether the verifying
material is a public key or a certificate. The `keyBytes` function used
for comparing the verifier object was incorrectly assuming that
`SigVerifier` being non-nil meant that a local certificate was not
present and was returning the certificate's public key instead of the
certificate. This change fixes the function to check for a certificate
before returning the bytes from the public key.

It is also possible to provide a public key as an argument when a
certificate is in the signature bundle. It does not make sense to try to
provide both, but this has been allowed and asserted in
verify_blob_test.go. In this case, the comparer should ensure that the
certificate and public key are related to each other. The alternative is
to disallow providing a key as a command argument when a certificate is
present in the signature, but this could be considered a breaking
change.

This change only applies to verifying images using the old signature
format. For the new bundle format, the certificate verification goes
through a different path.

Fixes #4513

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -77,7 +77,7 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
 	}
 
 	if ko.SigningConfig != nil {
-		keypair, idToken, err := signcommon.GetKeypairAndToken(ctx, ko, "", "")
+		keypair, _, idToken, err := signcommon.GetKeypairAndToken(ctx, ko, "", "")
 		if err != nil {
 			return nil, fmt.Errorf("getting keypair and token: %w", err)
 		}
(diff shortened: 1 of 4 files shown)


Hayden-IO:
@cmurphy Do you know if your first commit fixes this issue, https://github.com/sigstore/cosign/issues/2632? I forgot I had looked into fixing this (https://github.com/sigstore/cosign/pull/2633) quite awhile ago.

cmurphy:
@haydentherapper yes this addresses the second part of that issue. You still need to provide the identity and issuer.

PR #4530, opened 2025-11-18
Protobuf bundle support for subcommands `save` and `load`

steiza:
#### Summary

Continuing to make progress on #4470.

To test:

- Run [local registries](https://github.com/tstromberg/sigstore-the-local-way?tab=readme-ov-file#11-starting-a-local-registry) on port `1338` (default) and port `1337` (to test loading)

- Build and sign an image:

```
sigstore/sigstore-go/examples/sigstore-go-verification$ ko build .
sigstore/cosign$ go run cmd/cosign/main.go sign --key cosign.key --use-signing-config=false --tlog-upload=false localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a:latest
```

- Save it from registry:

```
$ go run cmd/cosign/main.go save --dir save localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a
```

- Load it into mirror registry:

```
$ go run cmd/cosign/main.go load --dir save localhost:1337/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a
```

- List out files that are saved:

```
$ ls save/blobs/sha256/*
```

- Use `oras` to confirm they are in new registry:

```
$ oras manifest fetch localhost:1337/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a@sha256:...
```

#### Release Note

- Added protobuf bundle support to `cosign save` and `cosign load`

#### Documentation

N/A

### DIFF
--- a/test/e2e_insecure_registry_test.go
+++ b/test/e2e_insecure_registry_test.go
@@ -94,7 +94,7 @@ func TestInsecureRegistry(t *testing.T) {
 		}
 	}
 	must(sign.SignCmd(t.Context(), ro, ko, so, []string{imgName}), t)
-	mustErr(verify(pubKey, imgName, true, nil, "", false), t)
+	mustErr(verify(pubKey, imgName, true, nil, "", false, false), t)
 	cmd := cliverify.VerifyCommand{
 		KeyRef:      pubKey,
 		CheckClaims: true,
--- a/test/e2e_kms_test.go
+++ b/test/e2e_kms_test.go
@@ -59,7 +59,7 @@ func TestSecretsKMS(t *testing.T) {
 	privKey := kms
 
 	// Verify should fail at first
-	mustErr(verify(pubKey, imgName, true, nil, "", false), t)
+	mustErr(verify(pubKey, imgName, true, nil, "", false, false), t)
 
 	rekorURL := os.Getenv(rekorURLVar)
 
@@ -76,10 +76,10 @@ func TestSecretsKMS(t *testing.T) {
 		TlogUpload: true,
 	}
 	must(sign.SignCmd(t.Context(), ro, ko, so, []string{imgName}), t)
-	must(verify(pubKey, imgName, true, nil, "", false), t)
+	must(verify(pubKey, imgName, true, nil, "", false, false), t)
 
 	// Sign and verify with annotations
-	mustErr(verify(pubKey, imgName, true, map[string]any{"foo": "bar"}, "", false), t)
+	mustErr(verify(pubKey, imgName, true, map[string]any{"foo": "bar"}, "", false, false), t)
 	soAnno := options.SignOptions{
 		Upload:     true,
 		TlogUpload: true,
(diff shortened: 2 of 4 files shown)


pkg/oci/remote/write.go
```diff
@@ -92,6 +92,25 @@ func WriteSignedImageIndexImages(ref name.Reference, sii oci.SignedImageIndex, o
...
+	if err != nil {
+		return err
+	}
+	if bundles != nil { // will be nil if there are no associated bundles
+		if digest, ok := ref.(name.Digest); ok {
+			se, err := SignedEntity(ref, opts...)
+			if err != nil {
+				return err
+			}
+			err = WriteSignaturesExperimentalOCI(digest, se, opts...)
```
Hayden-IO, line 107: Definitely not at all an expert in this part of the codebase, so I'm trying to wrap my head around how `WriteSignaturesExperimentalOCI` works here. Is this now needed because we were writing referring artifacts to OCI using a different API?

steiza: Yeah, I also had to do a lot of digging in this code base to understand what was going on.

But that's exactly right - we don't want to "just" write the protobuf bundle as an OCI blob, we also need to write an OCI manifest that (1) points at the protobuf bundle blob *and* (2) has a subject referring to the signed image - which is how the registry knows that this is a "referring artifact".

Hayden-IO: That makes sense, thanks. So how were things working previously? Was it that we were looking up the referring artifact directly rather than going through the OCI manifest?

Hayden-IO: I guess to add to that, is there a change we need to make to the verification code?

steiza: Yeah, so previously signatures were pushed to a tag made from the digest of the image manifest.

So if your container was `example.com/myapp@sha256:abcd`, then you would ask for the signature at `example.com/myapp:sha256-abcd.sig` (note that this is a tag, even though it looks like a digest).

The verification code shouldn't need to change. New protobuf bundle signatures should always use the OCI referrers API. But I still need to update tests 😅

Hayden-IO: Sorry, by previously I meant with bundles and OCI. If we were only publishing the referring artifact and not the manifest that points to the referring artifact, how was verification working? How did the OCI library know how to look up the referring artifact without a manifest?

steiza:
I don't think this should be part of v3.0.3 anymore 😅 

Saving is pretty straightforward. For loading you basically need to implement something that behaves like the OCI referring API (look at the files for manifests that are the subject of the container image). Such a thing is possible, but I'm really not familiar with `sigstore/cosign/pkg/oci` and `google/go-containerregistry`.

Hayden-IO:
I'm wondering if not writing the manifest is a bug that needs to be fixed before a new release?

PR #4539, opened 2025-11-21
Protobuf bundle support for subcommand `clean`

steiza:
#### Summary

I think this really for-real completes #4470.

Testing is pretty easy! Run `go run cmd/cosign/main.go clean <some image you signed before>`.

#### Release Note

- Added protobuf bundle support to `cosign clean`

#### Documentation

N/A

### DIFF
--- a/cmd/cosign/cli/clean.go
+++ b/cmd/cosign/cli/clean.go
@@ -139,16 +139,22 @@ func CleanCmd(ctx context.Context, regOpts options.RegistryOptions, cleanType op
 	switch cleanType {
 	case options.CleanTypeSignature:
 		cleanTags = []name.Reference{sigRef}
+		if len(referrerRefs) > 0 {
+			ui.Warnf(ctx, "image has referrers, consider using --referrer")
+		}
 	case options.CleanTypeSbom:
 		cleanTags = []name.Reference{sbomRef}
 	case options.CleanTypeAttestation:
 		cleanTags = []name.Reference{attRef}
+		if len(referrerRefs) > 0 {
+			ui.Warnf(ctx, "image has referrers, consider using --referrer")
+		}
 	case options.CleanTypeReferrer:
 		cleanTags = referrerRefs
 	case options.CleanTypeAll:
 		cleanTags = append([]name.Reference{sigRef, attRef, sbomRef}, referrerRefs...)
 	default:
-		panic("invalid CleanType value")
+		return errors.New("invalid CleanType value")
 	}
 
 	for _, t := range cleanTags {


cmd/cosign/cli/clean.go
```diff
@@ -62,32 +62,76 @@ func CleanCmd(ctx context.Context, regOpts options.RegistryOptions, cleanType op
...
+				if err != nil {
+					return err
+				}
+				referrerRefs = append(referrerRefs, layerDigest)
+			}
+		}
+	}
+
+	var cleanTags []name.Reference
 	switch cleanType {
```
cmurphy, line 139: It seems like `--type` no longer has any meaning when it comes to referrers - when I sign with `sign` but then clean with `--type=attestation`, it just removes everything. That could come as a surprise to someone who's expecting those objects to be different.

cmd/cosign/cli/clean.go
```diff
@@ -62,32 +62,76 @@ func CleanCmd(ctx context.Context, regOpts options.RegistryOptions, cleanType op
...
+	referrerRefs := []name.Reference{}
+	digest, ok := ref.(name.Digest)
+	if !ok {
+		var err error
+		digest, err = ociremote.ResolveDigest(ref, ociRemoteOpts)
+		if err != nil {
+			return fmt.Errorf("resolving digest: %w", err)
+		}
+	}
+	idx, err := remote.Referrers(digest, remoteOpts...)
```
cmurphy, line 92: Is there any chance a manifest in the referrer index could not have been produced by cosign/sigstore and cause a surprise if it was deleted?

Does this also need to delete the blobs referenced in the manifests or is it enough to just delete the manifests?

cmd/cosign/cli/clean.go
```diff
@@ -103,13 +147,16 @@ func CleanCmd(ctx context.Context, regOpts options.RegistryOptions, cleanType op
 			case errors.As(err, &te) && te.StatusCode == http.StatusBadRequest:
 				// Docker registry >=v2.3 requires does not allow deleting the OCI object name directly, must use the digest instead.
 				// See https://github.com/distribution/distribution/blob/main/docs/content/spec/api.md#deleting-an-image
-				if err := deleteByDigest(t, remoteOpts...); err != nil {
-					if errors.As(err, &te) && te.StatusCode == http.StatusNotFound { //nolint: revive
+				tTag, ok := t.(name.Tag)
+				if ok {
```
cmurphy, line 172: What does this mean? It only tries to delete by digest if the reference is a tag?

steiza: We should only be down here if the `remote.Delete()` above failed, which will remove references if they are a tag or a digest. Despite its name, `deleteByDigest()` is really expecting a `name.Tag`, and so now we're checking to make sure what we're supplying it is a tag (which we probably should have been doing before...)

cmd/cosign/cli/clean.go
steiza: Yeah, unfortunately with the new protobuf bundle (where we're always trying to use the OCI1.1 referrer specification) both [attest](https://github.com/sigstore/cosign/blob/4e3042cebab7fefd0a64f803979b6cfb00859ddd/cmd/cosign/cli/attest/attest.go#L127) and [sign](https://github.com/sigstore/cosign/blob/4e3042cebab7fefd0a64f803979b6cfb00859ddd/cmd/cosign/cli/sign/sign.go#L192) are using `CosignSignPredicateType`, so I don't think there's an easy way to tell them apart.

I think eventually we might consider merging `sign` and `attest`, but of course that isn't where we are today.

Alternatively, we could only clean up referring items on `CleanTypeAll`? That could also be confusing. 🤷 Definitely open to feedback here.

cmd/cosign/cli/clean.go
steiza: It is certainly possible! Although maybe unlikely.

We queue up deleting the manifest by adding it to `referrerRefs` on line 106 below, and we queue up the layers in the manifest for deletion on line 120 below.

It's certainly possible to be more clever here - we could only delete layers if they have media type `application/vnd.dev.sigstore.bundle.v0.3+json` and only delete the manifest if all the layers are removed (otherwise rewrite the manifest to exclude the deleted layers)... but I think this command is really meant more for development so you can sign something, remove the signature, sign it again, and quickly iterate through those steps.

cmd/cosign/cli/clean.go
cmurphy: What if there was a new `--type=referrer` just for this? Then we could print a warning if referrers are found but a signature or attestation type is specified to let the user know they need to use specify the new type. I think at some point we just need to accept that referrers are different and can't be cajoled into behaving like classic `.sig`s or `.att`s.

cmd/cosign/cli/clean.go
cmurphy: I think it wouldn't hurt to check for the media type before deleting.

cmd/cosign/cli/clean.go
```diff
@@ -62,32 +63,90 @@ func CleanCmd(ctx context.Context, regOpts options.RegistryOptions, cleanType op
...
+						referrerRefs = append(referrerRefs, layerDigest)
+					}
+				}
+			}
+		}
+	}
+
+	var cleanTags []name.Reference
 	switch cleanType {
 	case options.CleanTypeSignature:
```
cmurphy, line 140: It might be nice to log a warning here if `len(referrerRefs) > 0` that this type isn't doing what the user wants it to do and they need to set `--type=referrer`.

cmurphy (commented):
One small suggestion but otherwise lgtm.

PR #4592, opened 2025-12-19
Support default services for trusted-root and signing-config creation

Hayden-IO:
When a client doesn't want to use a default service like Rekor while still using the public instance, a client would need to retrieve the signing configuration or trusted root, remove the tlog sections, and provide that to the sign/verify command. This first step can now be skipped by using `--with-default-services` and `--no-<service>` to omit it from the configuration file. This will provide a better user experience while maintaining the unified configuration interface between clients.

Ref: https://github.com/docker/actions-toolkit/issues/901


Hayden-IO:
Verified working with `cosign trusted-root create --with-default-services` and `cosign signing-config create --with-default-services --no-default-rekor`.

PR #4612, opened 2026-01-03
Optimize cosign tree performance by caching digest resolution

bobcallaway:
## Summary

Optimizes the `cosign tree` command by eliminating redundant digest resolution calls, resulting in a notable performance improvement due to omitting unnecessary network calls.

## Changes

- Extract the digest once from the `SignedEntity` that's already fetched
- Pass the digest reference instead of tag reference to `AttestationTag()`, `SignatureTag()`, `SBOMTag()`, and `Referrers()`
- Remove the explicit `ResolveDigest()` call in OCI 1.1 mode that was duplicating work

## Performance Impact

Tested with 10 iterations on `gcr.io/projectsigstore/cosign:v2.2.0`:
- **Before**: 1,946ms average
- **After**: 1,249ms average
- **Improvement**: 697ms faster (35%)

## Why This Works

The original code fetched the manifest multiple times:
1. Once in `SignedEntity(ref)` to get the signed entity
2. Again in `AttestationTag(ref)` to resolve tag → digest
3. Again in `SignatureTag(ref)` to resolve tag → digest
4. Again in `SBOMTag(ref)` to resolve tag → digest
5. Again in `ResolveDigest(ref)` for OCI 1.1 mode

Since `SignedEntity` already fetches and processes the manifest, we can extract the digest once and reuse it for all subsequent operations, avoiding 4+ redundant network calls.

### DIFF
--- a/cmd/cosign/cli/tree.go
+++ b/cmd/cosign/cli/tree.go
@@ -71,8 +71,16 @@ func TreeCmd(ctx context.Context, regOpts options.RegistryOptions, regExpOpts op
 		return err
 	}
 
+	// Get the digest from the signed entity to avoid redundant manifest fetches
+	// in subsequent tag resolution calls
+	digest, err := simg.Digest()
+	if err != nil {
+		return err
+	}
+	digestRef := ref.Context().Digest(digest.String())
+
 	// Handle the legacy mode first, always
-	attRef, err := ociremote.AttestationTag(ref, remoteOpts...)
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):
LGTM - IIUC, https://github.com/sigstore/cosign/blob/0faa2add684e298363ac2b12089148f9c033e01b/pkg/oci/remote/remote.go#L156-L161 is the chunk of code this change will hit instead of the chunk below it.

steiza (approved):
LGTM!

Issue #4550, opened 2025-11-27
Cosign 3.0.2 continues trying to reach out to TUF CDN even with local key, Nexus-only access


oarifbulut:
Cosign v3.0.2
Signing --> Local key pair (cosign.key, cosign.pub)
Registry -->  Sonatype Nexus
Build Agent --> Only has access to Nexus, no Internet egress
Goal --> Sign + verify an image through Nexus only, without contacting TUF, Rekor, CTLog, Fulcio

Hello team,
I would like to clarify the correct approach for a fully private signing and verification workflow using Cosign v3.0.2, where only a Nexus Registry is reachable and no network access to Sigstore TUF services is allowed.

here how I could sign the image 

cosign sign --key cosign.key --tlog-upload=false --use-signing-config=false IMAGE

how I couldn't verify it 

COSIGN_EXPERIMENTAL=0 cosign verify --key cosign.pub --insecure-ignore-tlog=true --insecure-ignore-sct=true --use-signing-config=false

OR 

cosign verify --key cosign.pub --offline=true --new-bundle-format=false --trusted-root .../trusted_root.json --local-image /path/to/dir

oarifbulut:
Is there anybody who has this info? +++

oarifbulut:
@haydentherapper Hello, do you have any idea on this? This would be highly apprecited if I could get any answer. I got stucked in something and need to overcome.  :(

Hayden-IO:
@oarifbulut Could you confirm when you're seeing the outbound network requests, on signing and/or verification? We shouldn't be fetching TUF metadata on signing when you provide a key and specify `--use-signing-config=false` ([code](https://github.com/sigstore/cosign/blob/dc8ad5e8252e0114400fe0950525168782dfc1c6/cmd/cosign/cli/signcommon/common.go#L650-L676)). 

If it is on verification, I think I've found the issue in [SetTrustedMaterial](https://github.com/sigstore/cosign/blob/dc8ad5e8252e0114400fe0950525168782dfc1c6/cmd/cosign/cli/verify/common.go#L174). We check if you provide any service trust roots, and if not, fetch trust root material from TUF. In your case, you only need a key to verify, so you provide no service roots and therefore we make a call to TUF.

oarifbulut:
[...]

PR #4613, opened 2026-01-04
Don't require a trusted root to verify offline with a key

Hayden-IO:
When signing offline using a key and opting out of Rekor, contents from
the TUF repo are not fetched because they aren't needed. Similarly with
verification, TUF shouldn't be required when no service roots are
needed, which is true when a) a public key is provided, b) Rekor is
skipped via flag, and c) timestamp verification is skipped via a flag.

Note that this affects the old bundle format verification path,
which allows for services to be skipped via flags. When a new bundle
format is provided, users will not be required to include a trusted root
if the above 3 conditions are satisfied.

Fixes https://github.com/sigstore/cosign/issues/4550

### DIFF
--- a/cmd/cosign/cli/verify/common.go
+++ b/cmd/cosign/cli/verify/common.go
@@ -43,7 +43,7 @@ import (
 )
 
 // CheckSigstoreBundleUnsupportedOptions checks for incompatible settings on any Verify* command struct when NewBundleFormat is used.
-func CheckSigstoreBundleUnsupportedOptions(cmd any, co *cosign.CheckOpts) error {
+func CheckSigstoreBundleUnsupportedOptions(cmd any, verifyOfflineWithKey bool, co *cosign.CheckOpts) error {
 	if !co.NewBundleFormat {
 		return nil
 	}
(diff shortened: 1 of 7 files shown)


Hayden-IO:
Verified by going offline and running:

Old bundle:

```
./cosign sign-blob --use-signing-config=false --tlog-upload=false --key=cosign.key --bundle=out.bundle --new-bundle-format=false README.md

./cosign verify-blob --bundle=out.bundle --private-infrastructure --key=cosign.pub --new-bundle-format=false README.md

# Throws an error as expected, trying to reach the TUF repo
./cosign verify-blob --bundle=out.bundle --key=cosign.pub --new-bundle-format=false README.md
```

New bundle:

```
./cosign sign-blob --use-signing-config=false --tlog-upload=false --key=cosign.key --bundle=out.bundle README.md

./cosign verify-blob --bundle=out.bundle --private-infrastructure --key=cosign.pub README.md

# Throws an error as expected trying to reach the TUF repo
./cosign verify-blob --bundle=out.bundle --key=cosign.pub README.md
```

You can also sign offline using new Cosign v3 flags:

```
./cosign signing-config create --out sc.json
./cosign trusted-root create --out trusted-root.json

./cosign sign-blob --signing-config=sc.json --trusted-root=trusted-root.json --key=cosign.key --bundle=out.bundle README.md
```

Hayden-IO:
Grumble, well I might actually have to handle the new bundle path, otherwise --trusted--root becomes a required flag when only a key is used. Even though it's unnecessary to fetch TUF metadata, most users probably don't care.

At some point, we should make `--offline` actually mean offline.

Hayden-IO:
Thanks @steiza, I've added a test that sets the TUF environment variables to invalid values, so we know that any TUF repo requests don't occur.

steiza (approved):
LGTM - thanks again!

Issue #4594, opened 2025-12-21
cosign does not generate signed timestamp anymore


augjoh:
**Description**

After upgrading from 3.0.2 -> 3.0.3 cosign does not generae the timestamp file anymore:

```
cosign sign-blob acme-client.SHA256 --bundle acme-client.bundle --use-signing-config=false --timestamp-server-url "${TSA_URL}" --rfc3161-timestamp acme-client.json
[...]
Generating ephemeral keys...
Retrieving signed certificate...
Successfully verified SCT...
Using payload from: acme-client.SHA256
	The sigstore service, hosted by sigstore a Series of LF Projects, LLC, is provided pursuant to the Hosted Project Tools Terms of Use, available at https://lfprojects.org/policies/hosted-project-tools-terms-of-use/.
	Note that if your submission includes personal data associated with this signed artifact, it will be part of an immutable record.
	This may include the email address associated with the account with which you authenticate your contractual Agreement.
	This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later, and is subject to the Immutable Record notice at https://lfprojects.org/policies/hosted-project-tools-immutable-records/.
By typing 'y', you attest that (1) you are not submitting the personal data of any other person; and (2) you understand and agree to the statement and the Agreement terms at the URLs listed above.
Timestamp fetched with time:  2025-12-20 20:21:11 +0000 UTC
tlog entry created with index: 774228167
Wrote bundle to file acme-client.bundle
[...]
```

Version 3.0.2 creates `acme-client.json` and writes additionally:

```
[...]
RFC3161 timestamp written to file acme-client.json
[...]
```

**Version**

 3.0.3

Issue #4534, opened 2025-11-19
Require bundle output or registry upload


aaronlew02:
**Description**

For signing and attestation commands, at least one form of verification material output should be required if signatures are not being uploaded to a registry. 

* For `sign` and `attest`, users requesting a new-format bundle are able to request a bundle output or registry upload but are not required to request either. 
* For `sign-blob` and `attest-blob`, users requesting a new-format bundle are able but not required to request a bundle output.

Without having uploaded any signatures to a registry, users will be unable to verify a signing event without these verification materials.

**Note:** As the old bundle format is a legacy concept, this requirement does not apply to users requesting an old-format bundle. However, the existing options in that case are listed:

* For `sign-blob` and `attest-blob`, users requesting an old-format bundle are able to request a bundle output, though a signature output is provided in the terminal by default regardless.
* For `attest`, users requesting an old-format bundle are able to request a bundle output or registry upload but are not required to request either. 
* For `sign`, users requesting an old-format bundle have a third option — a detached output of the signature, payload, and/or certificate — but are not required to request these, either.

PR #4371, opened 2025-08-27
Sc sign attest

haydentherapper:

PR #4618, opened 2026-01-05
Sign exclusively via sigstore-go

### LINKED ISSUE #4570: Restructuring signing code for shared implementation
**Description**

Currently, for `sign`, `sign-blob`, `attest`, and `attest-blob`, we fork the code on whether or not a user provides a signing config, e.g. [in sign-blob](https://github.com/sigstore/cosign/blob/901b44d65952108d545bb150b99def655e9d3e1b/cmd/cosign/cli/sign/sign_blob.go#L87-L104). When the user provides a signing config, we route signing to the sigstore-go library. Otherwise, we use the Cosign signing code. This choice introduced some unexpected issues with Cosign v3, with missing functionality.

This proposes restructuring the signing restructuring code to not have two divergent signing paths. Roughly, the signing code should be structured like so:

1. Setup: Retrieve service URLs, trust root, initialize any configuration
2. Sign: Sign a container/attestation/blob
3. Return: Construct required outputs, e.g. bundle or detached verification material

For (1), we can retrieve service URLs either from the environment or flags (like with Cosign v2.5-) or from a signing config.

For (2), this will require more design, but our goal should be to use sigstore-go to sign content in all commands. We may need to build an abstraction layer that transforms a container image into a byte stream to pass to sigstore-go (right now, Cosign's sign library does the opposite, taking in an OCI representation of a container, blob or attestation). I'd suggest we start refactoring `sign-blob` and `attest-blob` first.

For (3), again we should use the flags to determine what output the user requests. If we support both bundles and detached verification material concurrently, we can easily mark the latter deprecated without maintaining separate codepaths.

We have a common library for sign-related functionality already - https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/signcommon/common.go - which we can continue to expand.

This work will enable future refactoring to have flags shared between sign commands, move towards converging `sign` and `attest` implementations, deprecating v2 flags, and ultimately having a single signing implementation from sigstore-go while providing signing containers as a well-formed library.

aaronlew02:
Closes #4570

#### Summary

This change refactors the signing path in all signing and attestation commands such that all signing events occur via sigstore-go, not just those in which a signing config is used. 

#### Release Note

In cases where a signing config is not used, a signing config is constructed using TUF- or flag-provided URLs. When a legacy bundle is requested, a legacy bundle is constructed from the new-format bundle provided by sigstore-go.

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -19,13 +19,16 @@ import (
 	"context"
 	_ "crypto/sha256" // for `crypto.SHA256`
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/signcommon"
+	"github.com/sigstore/cosign/v3/internal/ui"
 	"github.com/sigstore/cosign/v3/pkg/cosign/attestation"
 	cbundle "github.com/sigstore/cosign/v3/pkg/cosign/bundle"
 	cremote "github.com/sigstore/cosign/v3/pkg/cosign/remote"
@@ -33,6 +36,8 @@ import (
 	ociremote "github.com/sigstore/cosign/v3/pkg/oci/remote"
 	"github.com/sigstore/cosign/v3/pkg/oci/static"
 	"github.com/sigstore/cosign/v3/pkg/types"
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	"github.com/sigstore/sigstore/pkg/signature"
 )
 
 // nolint
@@ -123,64 +128,107 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	bundleOpts := signcommon.CommonBundleOpts{
 		Payload:       payload,
 		Digest:        digest,
-		PredicateType: types.CosignSignPredicateType,
+		PredicateType: predicateURI,
 		BundlePath:    c.BundlePath,
 		Upload:        !c.NoUpload,
 		OCIRemoteOpts: ociremoteOpts,
 	}
 
-	if c.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
+	if c.SigningConfig == nil {
+		c.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(c.KeyOpts, c.TlogUpload)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
 	}
 
-	bundleComponents, closeSV, err := signcommon.GetBundleComponents(ctx, c.CertPath, c.CertChainPath, c.KeyOpts, c.NoUpload, c.TlogUpload, payload, digest, c.RekorEntryType)
+	bundleBytes, pubKey, pubKeyPem, hashAlgProto, err := signcommon.NewAttestationBundle(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
 	if err != nil {
-		return fmt.Errorf("getting bundle components: %w", err)
+		return fmt.Errorf("creating bundle: %w", err)
 	}
-	defer closeSV()
 
-	sv := bundleComponents.SV
+	if c.NewBundleFormat {
+		// TODO(#4534): Require bundle output or registry upload
+		if c.BundlePath != "" {
+			if err := os.WriteFile(c.BundlePath, bundleBytes, 0600); err != nil {
+				return fmt.Errorf("create bundle file: %w", err)
+			}
+			ui.Infof(ctx, "Wrote bundle to file %s", c.BundlePath)
+		}
 
-	if c.NoUpload && c.BundlePath == "" {
-		fmt.Println(string(bundleComponents.SignedPayload))
+		if !c.NoUpload {
+			if err := ociremote.WriteAttestationNewBundleFormat(digest, bundleBytes, bundleOpts.PredicateType, ociremoteOpts...); err != nil {
+				return fmt.Errorf("writing bundle: %w", err)
+			}
+		}
 		return nil
 	}
 
-	opts := []static.Option{static.WithLayerMediaType(types.DssePayloadType)}
-	if sv.Cert != nil {
-		opts = append(opts, static.WithCertChain(sv.Cert, sv.Chain))
+	var pb protobundle.Bundle
+	if err := protojson.Unmarshal(bundleBytes, &pb); err != nil {
+		return fmt.Errorf("unmarshalling bundle: %w", err)
 	}
 
-	if bundleComponents.RFC3161Timestamp != nil {
-		opts = append(opts, static.WithRFC3161Timestamp(bundleComponents.RFC3161Timestamp))
-	}
-
-	predicateType, err := options.ParsePredicateType(c.PredicateType)
+	bundleComponents, err := signcommon.ExtractComponentsFromProtoBundle(&pb)
 	if err != nil {
-		return err
+		return fmt.Errorf("extracting components from bundle: %w", err)
 	}
 
-	bundleOpts.PredicateType = predicateType
+	legacyBundleBytes, err := signcommon.NewLegacyBundleFromProtoBundleComponents(bundleComponents, pubKeyPem)
+	if err != nil {
+		return fmt.Errorf("creating legacy bundle: %w", err)
+	}
+
+	if c.BundlePath != "" {
+		if err := os.WriteFile(c.BundlePath, legacyBundleBytes, 0600); err != nil {
+			return fmt.Errorf("create bundle file: %w", err)
+		}
+		ui.Infof(ctx, "Wrote bundle to file %s", c.BundlePath)
+	}
+
+	if c.NoUpload {
+		return nil
+	}
+
+	certPem, chainPem := signcommon.EncodeCertificatesToPEM(bundleComponents.Certificates)
+
+	opts := []static.Option{
+		static.WithLayerMediaType(types.DssePayloadType),
+		static.WithAnnotations(map[string]string{
+			"predicateType": predicateURI,
+		}),
+	}
+	if certPem != nil {
+		opts = append(opts, static.WithCertChain(certPem, chainPem))
+	}
+
+	if len(bundleComponents.RFC3161Timestamps) > 0 {
+		opts = append(opts, static.WithRFC3161Timestamp(cbundle.TimestampToRFC3161Timestamp(bundleComponents.RFC3161Timestamps[0].GetSignedTimestamp())))
+	}
 
 	predicateTypeAnnotation := map[string]string{
-		"predicateType": predicateType,
+		"predicateType": predicateURI,
 	}
 	// Add predicateType as manifest annotation
 	opts = append(opts, static.WithAnnotations(predicateTypeAnnotation))
 
-	if bundleComponents.RekorEntry != nil {
-		opts = append(opts, static.WithBundle(cbundle.EntryToBundle(bundleComponents.RekorEntry)))
+	if len(bundleComponents.RekorEntries) > 0 {
+		opts = append(opts, static.WithBundle(signcommon.RekorBundleFromProtoTlogEntry(bundleComponents.RekorEntries[0])))
 	}
 
-	if c.KeyOpts.NewBundleFormat {
-		return signcommon.WriteBundle(ctx, sv, bundleComponents.RekorEntry, bundleOpts, bundleComponents.SignedPayload, bundleComponents.SignerBytes, bundleComponents.TimestampBytes)
+	ociSig, err := static.NewAttestation(bundleComponents.Signature, opts...)
+	if err != nil {
+		return fmt.Errorf("creating attestation: %w", err)
 	}
 
 	// We don't actually need to access the remote entity to attach things to it
 	// so we use a placeholder here.
 	se := ociremote.SignedUnknown(digest, ociremoteOpts...)
 
-	dd := cremote.NewDupeDetector(sv)
+	ddVerifier, err := signature.LoadVerifier(pubKey, signcommon.ProtoHashAlgoToHash(hashAlgProto))
+	if err != nil {
+		return fmt.Errorf("loading verifier: %w", err)
+	}
+	dd := cremote.NewDupeDetector(ddVerifier)
 	signOpts := []mutate.SignOption{
 		mutate.WithDupeDetector(dd),
 		mutate.WithRecordCreationTimestamp(c.RecordCreationTimestamp),
@@ -191,15 +239,10 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		signOpts = append(signOpts, mutate.WithReplaceOp(ro))
 	}
 
-	sig, err := static.NewAttestation(bundleComponents.SignedPayload, opts...)
-	if err != nil {
-		return err
-	}
-
 	// Attach the attestation to the entity.
-	newSE, err := mutate.AttachAttestationToEntity(se, sig, signOpts...)
+	newSE, err := mutate.AttachAttestationToEntity(se, ociSig, signOpts...)
 	if err != nil {
-		return err
+		return fmt.Errorf("attaching attestation: %w", err)
 	}
 
 	// Publish the attestations associated with this entity
--- a/cmd/cosign/cli/attest/attest_blob.go
+++ b/cmd/cosign/cli/attest/attest_blob.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"context"
 	"crypto"
-	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -32,11 +31,12 @@ import (
 	intotov1 "github.com/in-toto/attestation/go/v1"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/signcommon"
-	"github.com/sigstore/cosign/v3/pkg/cosign"
+	"github.com/sigstore/cosign/v3/internal/ui"
 	"github.com/sigstore/cosign/v3/pkg/cosign/attestation"
 	cbundle "github.com/sigstore/cosign/v3/pkg/cosign/bundle"
-	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	"github.com/sigstore/sigstore/pkg/signature"
+	"google.golang.org/protobuf/encoding/protojson"
 )
 
 // nolint
@@ -146,85 +146,92 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
 		BundlePath: c.BundlePath,
 	}
 
-	if c.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
+	if c.SigningConfig == nil {
+		var err error
+		c.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(c.KeyOpts, c.TlogUpload)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
 	}
 
-	bundleComponents, closeSV, err := signcommon.GetBundleComponents(ctx, c.CertPath, c.CertChainPath, c.KeyOpts, false, c.TlogUpload, payload, nil, c.RekorEntryType)
+	bundleBytes, _, pubKeyPem, _, err := signcommon.NewAttestationBundle(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
 	if err != nil {
-		return fmt.Errorf("getting bundle components: %w", err)
+		return fmt.Errorf("creating bundle: %w", err)
 	}
-	defer closeSV()
 
-	sv := bundleComponents.SV
+	if c.NewBundleFormat {
+		// TODO(#4534): Require bundle output or registry upload
+		if c.BundlePath != "" {
+			if err := os.WriteFile(c.BundlePath, bundleBytes, 0600); err != nil {
+				return fmt.Errorf("create bundle file: %w", err)
+			}
+			ui.Infof(ctx, "Wrote bundle to file %s", c.BundlePath)
+		}
+		return nil
+	}
 
-	signedPayload := cosign.LocalSignedPayload{}
+	var pb protobundle.Bundle
+	if err := protojson.Unmarshal(bundleBytes, &pb); err != nil {
+		return fmt.Errorf("unmarshalling bundle: %w", err)
+	}
 
-	if bundleComponents.RekorEntry != nil {
-		signedPayload.Bundle = cbundle.EntryToBundle(bundleComponents.RekorEntry)
+	bundleComponents, err := signcommon.ExtractComponentsFromProtoBundle(&pb)
+	if err != nil {
+		return err
 	}
 
 	if c.BundlePath != "" {
-		var contents []byte
-		if c.NewBundleFormat {
-			pubKey, err := sv.PublicKey()
-			if err != nil {
-				return err
-			}
-
-			contents, err = cbundle.MakeNewBundle(pubKey, bundleComponents.RekorEntry, payload, bundleComponents.SignedPayload, bundleComponents.SignerBytes, bundleComponents.TimestampBytes)
-			if err != nil {
-				return err
-			}
-		} else {
-			signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(bundleComponents.SignedPayload)
-			signedPayload.Cert = base64.StdEncoding.EncodeToString(bundleComponents.SignerBytes)
-
-			contents, err = json.Marshal(signedPayload)
-			if err != nil {
-				return err
-			}
+		contents, err := signcommon.NewLegacyBundleFromProtoBundleComponents(bundleComponents, pubKeyPem)
+		if err != nil {
+			return fmt.Errorf("creating legacy bundle: %w", err)
 		}
 
 		if err := os.WriteFile(c.BundlePath, contents, 0600); err != nil {
 			return fmt.Errorf("create bundle file: %w", err)
 		}
-		fmt.Fprintln(os.Stderr, "Bundle wrote in the file ", c.BundlePath)
+		ui.Infof(ctx, "Wrote bundle to file %s", c.BundlePath)
 	}
 
 	if c.OutputSignature != "" {
-		if err := os.WriteFile(c.OutputSignature, bundleComponents.SignedPayload, 0600); err != nil {
+		if err := os.WriteFile(c.OutputSignature, bundleComponents.Signature, 0600); err != nil {
 			return fmt.Errorf("create signature file: %w", err)
 		}
 		fmt.Fprintf(os.Stderr, "Signature written in %s\n", c.OutputSignature)
 	} else {
-		fmt.Fprintln(os.Stdout, string(bundleComponents.SignedPayload))
+		fmt.Fprintln(os.Stdout, string(bundleComponents.Signature))
 	}
 
 	if c.OutputAttestation != "" {
 		if err := os.WriteFile(c.OutputAttestation, payload, 0600); err != nil {
-			return fmt.Errorf("create signature file: %w", err)
+			return fmt.Errorf("create attestation file: %w", err)
 		}
 		fmt.Fprintf(os.Stderr, "Attestation written in %s\n", c.OutputAttestation)
 	}
 
 	if c.OutputCertificate != "" {
-		cert, err := cryptoutils.UnmarshalCertificatesFromPEM(bundleComponents.SignerBytes)
-		// signer is a certificate
-		if err != nil {
-			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Was a certificate used? ", err)
-			return nil
-
+		if len(bundleComponents.Certificates) == 0 {
+			return fmt.Errorf("no certificate found in bundle")
 		}
-		if len(cert) != 1 {
-			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Expected a single certificate")
-			return nil
-		}
-		bts := bundleComponents.SignerBytes
-		if err := os.WriteFile(c.OutputCertificate, bts, 0600); err != nil {
+		certPem, _ := signcommon.EncodeCertificatesToPEM(bundleComponents.Certificates)
+		if err := os.WriteFile(c.OutputCertificate, certPem, 0600); err != nil {
 			return fmt.Errorf("create certificate file: %w", err)
 		}
 		fmt.Fprintln(os.Stderr, "Certificate written to file ", c.OutputCertificate)
+	}
+
+	if c.RFC3161TimestampPath != "" {
+		if len(bundleComponents.RFC3161Timestamps) == 0 {
+			return fmt.Errorf("no RFC3161 timestamp found in bundle")
+		}
+		legacyTimestamp := cbundle.TimestampToRFC3161Timestamp(bundleComponents.RFC3161Timestamps[0].GetSignedTimestamp())
+		ts, err := json.Marshal(legacyTimestamp)
+		if err != nil {
+			return fmt.Errorf("marshalling timestamp: %w", err)
+		}
+		if err := os.WriteFile(c.RFC3161TimestampPath, ts, 0600); err != nil {
+			return fmt.Errorf("create timestamp file: %w", err)
+		}
+		fmt.Fprintln(os.Stderr, "Timestamp wrote in the file ", c.RFC3161TimestampPath)
 	}
 
 	return nil
--- a/cmd/cosign/cli/options/key.go
+++ b/cmd/cosign/cli/options/key.go
@@ -54,6 +54,7 @@ type KeyOpts struct {
 	// for valid values.
 	FulcioAuthFlow string
 
+	// Deprecated: SCT verification is no longer performed during signing/attestation.
 	// Modeled after InsecureSkipVerify in tls.Config, this disables
 	// verifying the SCT.
 	InsecureSkipFulcioVerify bool
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -21,30 +21,31 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	intotov1 "github.com/in-toto/attestation/go/v1"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
-	"github.com/sigstore/cosign/v3/cmd/cosign/cli/rekor"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/signcommon"
-	icos "github.com/sigstore/cosign/v3/internal/pkg/cosign"
-	ifulcio "github.com/sigstore/cosign/v3/internal/pkg/cosign/fulcio"
-	ipayload "github.com/sigstore/cosign/v3/internal/pkg/cosign/payload"
-	irekor "github.com/sigstore/cosign/v3/internal/pkg/cosign/rekor"
-	"github.com/sigstore/cosign/v3/internal/pkg/cosign/tsa"
 	"github.com/sigstore/cosign/v3/internal/pkg/cosign/tsa/client"
 	"github.com/sigstore/cosign/v3/internal/ui"
 	"github.com/sigstore/cosign/v3/pkg/cosign"
+	cbundle "github.com/sigstore/cosign/v3/pkg/cosign/bundle"
 	cremote "github.com/sigstore/cosign/v3/pkg/cosign/remote"
 	"github.com/sigstore/cosign/v3/pkg/oci"
 	"github.com/sigstore/cosign/v3/pkg/oci/mutate"
 	ociremote "github.com/sigstore/cosign/v3/pkg/oci/remote"
+	"github.com/sigstore/cosign/v3/pkg/oci/static"
 	"github.com/sigstore/cosign/v3/pkg/oci/walk"
 	"github.com/sigstore/cosign/v3/pkg/types"
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	"github.com/sigstore/sigstore-go/pkg/sign"
+	"github.com/sigstore/sigstore/pkg/signature"
 	sigPayload "github.com/sigstore/sigstore/pkg/signature/payload"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/structpb"
@@ -196,17 +197,37 @@ func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpt
 		OCIRemoteOpts: ociremoteOpts,
 	}
 
-	if ko.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, ko, signOpts.Cert, signOpts.CertChain, bundleOpts, ko.SigningConfig, ko.TrustedMaterial)
+	if ko.SigningConfig == nil {
+		shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, ko, digest, signOpts.TlogUpload)
+		if err != nil {
+			return fmt.Errorf("should upload to tlog: %w", err)
+		}
+		ko.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(ko, shouldUpload)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
 	}
 
-	bundleComponents, closeSV, err := signcommon.GetBundleComponents(ctx, signOpts.Cert, signOpts.CertChain, ko, false, signOpts.TlogUpload, payload, digest, "dsse")
+	bundleBytes, _, _, _, err := signcommon.NewAttestationBundle(ctx, ko, signOpts.Cert, signOpts.CertChain, bundleOpts, ko.SigningConfig, ko.TrustedMaterial)
 	if err != nil {
-		return fmt.Errorf("getting bundle components: %w", err)
+		return err
 	}
-	defer closeSV()
 
-	return signcommon.WriteBundle(ctx, bundleComponents.SV, bundleComponents.RekorEntry, bundleOpts, bundleComponents.SignedPayload, bundleComponents.SignerBytes, bundleComponents.TimestampBytes)
+	if signOpts.BundlePath != "" {
+		if err := os.WriteFile(signOpts.BundlePath, bundleBytes, 0600); err != nil {
+			return fmt.Errorf("create bundle file: %w", err)
+		}
+		ui.Infof(ctx, "Wrote bundle to file %s", signOpts.BundlePath)
+	}
+
+	if signOpts.Upload {
+		ui.Infof(ctx, "Pushing signature to: %s", digest.Repository)
+		if err := ociremote.WriteAttestationNewBundleFormat(digest, bundleBytes, bundleOpts.PredicateType, bundleOpts.OCIRemoteOpts...); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko options.KeyOpts, signOpts options.SignOptions,
@@ -234,61 +255,89 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 		payloads = append(payloads, payload)
 	}
 
-	sv, closeSV, err := signcommon.GetSignerVerifier(ctx, signOpts.Cert, signOpts.CertChain, ko)
-	if err != nil {
-		return fmt.Errorf("getting signer: %w", err)
-	}
-	defer closeSV()
-
-	dd := cremote.NewDupeDetector(sv)
-
-	var s icos.Signer
-	s = ipayload.NewSigner(sv)
-	if sv.Cert != nil {
-		s = ifulcio.NewSigner(s, sv.Cert, sv.Chain)
-	}
-
-	if ko.TSAServerURL != "" {
-		if ko.TSAClientCACert == "" && ko.TSAClientCert == "" { // no mTLS params or custom CA
-			s = tsa.NewSigner(s, client.NewTSAClient(ko.TSAServerURL))
-		} else {
-			s = tsa.NewSigner(s, client.NewTSAClientMTLS(ko.TSAServerURL,
-				ko.TSAClientCACert,
-				ko.TSAClientCert,
-				ko.TSAClientKey,
-				ko.TSAServerName,
-			))
-		}
-	}
-	shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, ko, digest, signOpts.TlogUpload)
-	if err != nil {
-		return fmt.Errorf("should upload to tlog: %w", err)
-	}
-	if shouldUpload {
-		rClient, err := rekor.NewClient(ko.RekorURL)
+	if ko.SigningConfig == nil {
+		shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, ko, digest, signOpts.TlogUpload)
 		if err != nil {
-			return err
+			return fmt.Errorf("should upload to tlog: %w", err)
 		}
-		s = irekor.NewSigner(s, rClient)
+		ko.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(ko, shouldUpload)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
 	}
+
+	keypair, certBytes, idToken, err := signcommon.GetKeypairAndToken(ctx, ko, signOpts.Cert, signOpts.CertChain)
+	if err != nil {
+		return fmt.Errorf("getting keypair and token: %w", err)
+	}
+	if closer, ok := keypair.(interface{ Close() }); ok {
+		defer closer.Close()
+	}
+
+	var tsaClientTransport http.RoundTripper
+	if ko.TSAClientCACert != "" || (ko.TSAClientCert != "" && ko.TSAClientKey != "") {
+		tsaClientTransport, err = client.GetHTTPTransport(ko.TSAClientCACert, ko.TSAClientCert, ko.TSAClientKey, ko.TSAServerName, 30*time.Second)
+		if err != nil {
+			return fmt.Errorf("getting TSA client transport: %w", err)
+		}
+	}
+	cbundleOpts := cbundle.SignOptions{TSAClientTransport: tsaClientTransport}
 
 	ociSigs := make([]oci.Signature, len(payloads))
 	b64sigs := make([]string, len(payloads))
 
-	for i, payload := range payloads {
-		ociSig, _, err := s.Sign(ctx, bytes.NewReader(payload))
-		if err != nil {
-			return err
-		}
-		ociSigs[i] = ociSig
+	var leafCertPem []byte
 
-		b64sig, err := ociSig.Base64Signature()
-		if err != nil {
-			return err
+	for i, payload := range payloads {
+		content := &sign.PlainData{
+			Data: payload,
 		}
+
+		bundleBytes, err := cbundle.SignData(ctx, content, keypair, idToken, certBytes, ko.SigningConfig, ko.TrustedMaterial, cbundleOpts)
+		if err != nil {
+			return fmt.Errorf("signing bundle: %w", err)
+		}
+
+		var pb protobundle.Bundle
+		if err := protojson.Unmarshal(bundleBytes, &pb); err != nil {
+			return fmt.Errorf("unmarshalling bundle: %w", err)
+		}
+
+		bundleComponents, err := signcommon.ExtractComponentsFromProtoBundle(&pb)
+		if err != nil {
+			return fmt.Errorf("extracting components from bundle: %w", err)
+		}
+
+		certPem, chainPem := signcommon.EncodeCertificatesToPEM(bundleComponents.Certificates)
+		if i == 0 {
+			leafCertPem = certPem
+		}
+
+		b64sig := base64.StdEncoding.EncodeToString(bundleComponents.Signature)
 		b64sigs[i] = b64sig
+
+		var opts []static.Option
+		if certPem != nil {
+			opts = append(opts, static.WithCertChain(certPem, chainPem))
+		}
+
+		if len(bundleComponents.RFC3161Timestamps) > 0 {
+			opts = append(opts, static.WithRFC3161Timestamp(cbundle.TimestampToRFC3161Timestamp(bundleComponents.RFC3161Timestamps[0].GetSignedTimestamp())))
+		}
+
+		if len(bundleComponents.RekorEntries) > 0 {
+			opts = append(opts, static.WithBundle(signcommon.RekorBundleFromProtoTlogEntry(bundleComponents.RekorEntries[0])))
+		}
+
+		ociSig, err := static.NewSignature(payload, b64sig, opts...)
+		if err != nil {
+			return fmt.Errorf("creating signature: %w", err)
+		}
+
+		ociSigs[i] = ociSig
 	}
 
+	// TODO(#4534): Require bundle output or registry upload
 	outputSignature := signOpts.OutputSignature
 	if outputSignature != "" {
 		// Add digest to suffix to differentiate each image during recursive signing
@@ -311,12 +360,18 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 	}
 
 	if signOpts.OutputCertificate != "" {
-		rekorBytes, err := sv.Bytes(ctx)
-		if err != nil {
-			return fmt.Errorf("create certificate file: %w", err)
+		var outBytes []byte
+		if len(leafCertPem) > 0 {
+			outBytes = leafCertPem
+		} else {
+			pubPem, err := keypair.GetPublicKeyPem()
+			if err != nil {
+				return fmt.Errorf("getting public key pem: %w", err)
+			}
+			outBytes = []byte(pubPem)
 		}
 
-		if err := os.WriteFile(signOpts.OutputCertificate, rekorBytes, 0600); err != nil {
+		if err := os.WriteFile(signOpts.OutputCertificate, outBytes, 0600); err != nil {
 			return fmt.Errorf("create certificate file: %w", err)
 		}
 		// TODO: maybe accept a --b64 flag as well?
@@ -346,6 +401,13 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 	if !signOpts.Upload {
 		return nil
 	}
+
+	hashAlgo := signcommon.ProtoHashAlgoToHash(keypair.GetHashAlgorithm())
+	ddVerifier, err := signature.LoadVerifier(keypair.GetPublicKey(), hashAlgo)
+	if err != nil {
+		return fmt.Errorf("loading verifier: %w", err)
+	}
+	dd := cremote.NewDupeDetector(ddVerifier)
 
 	// Attach the signature to the entity.
 	var newSE oci.SignedEntity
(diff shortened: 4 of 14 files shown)


cmd/cosign/cli/signcommon/common.go
```diff
@@ -660,19 +660,22 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
 		(useSigningConfig || signingConfigPath != "") {
-		if trustedRootPath != "" {
+		switch {
+		case trustedRootPath != "":
 			ko.TrustedMaterial, err = root.NewTrustedRootFromPath(trustedRootPath)
 			if err != nil {
 				return fmt.Errorf("loading trusted root: %w", err)
 			}
-		} else {
+		case useSigningConfig || signingConfigPath != "":
```
cmurphy: This seems like a big change in behavior, since previously it would fetch a trusted root from TUF no matter what. Can you explain why it needs to be tied to the signing config?

pkg/cosign/bundle/sign.go
```diff
@@ -79,11 +79,13 @@ func SignData(ctx context.Context, content sign.Content, keypair sign.Keypair, i
 		keyTrustedMaterial := root.NewTrustedPublicKeyMaterial(func(_ string) (root.TimeConstrainedVerifier, error) {
 			return key, nil
 		})
-		trustedMaterial := &verifyTrustedMaterial{
-			TrustedMaterial:    bundleOpts.TrustedRoot,
-			keyTrustedMaterial: keyTrustedMaterial,
+		if bundleOpts.TrustedRoot != nil && keyTrustedMaterial != nil {
```
cmurphy: Can you explain what you were running into that made this necessary? sigstore-go is always going to need the trusted root set in order to work, and `keyTrustedMaterial` is explicitly set on line 79.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -76,52 +62,33 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
...
-
-		var tsaClientTransport http.RoundTripper
-		if ko.TSAClientCACert != "" || (ko.TSAClientCert != "" && ko.TSAClientKey != "") {
-			tsaClientTransport, err = client.GetHTTPTransport(ko.TSAClientCACert, ko.TSAClientCert, ko.TSAClientKey, ko.TSAServerName, 30*time.Second)
-			if err != nil {
-				return nil, fmt.Errorf("getting TSA client transport: %w", err)
-			}
+	if ko.SigningConfig == nil {
+		if ko.RekorVersion == 0 {
+			ko.RekorVersion = 1
```
cmurphy: I think this is unnecessary since you have a default in the flag.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -132,88 +99,68 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
...
 	if err != nil {
-		return nil, fmt.Errorf("signing blob: %w", err)
+		return nil, fmt.Errorf("reading payload: %w", err)
+	}
+	content := &sign.PlainData{
+		Data: data,
+	}
+	certIdToken := idToken
+	if len(certBytes) > 0 {
+		certIdToken = ""
```
cmurphy: Could you add a comment here

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -132,88 +99,68 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
...
-			certBytes, err := extractCertificate(ctx, sv)
-			if err != nil {
-				return nil, err
-			}
-			signedPayload.Cert = base64.StdEncoding.EncodeToString(certBytes)
-
-			contents, err = json.Marshal(signedPayload)
-			if err != nil {
-				return nil, err
+			if rfc3161Timestamp != nil && ko.RFC3161TimestampPath != "" {
```
cmurphy: This is underneath `ko.BundlePath != ""`, but I think the timestamp path is mainly for the case of disparate materials. Would it make more sense to move this outside this conditional and put it near the `if output*` conditionals?

cmd/cosign/cli/signcommon/common.go
```diff
@@ -660,19 +660,22 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
 			ko.TrustedMaterial, err = cosign.TrustedRoot()
 			if err != nil {
 				ui.Warnf(ctx, "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
 			}
+		default:
+			ko.TrustedMaterial = nil
```
cmurphy: I don't think we'd ever want to unset the trusted material if it was set before? Can you leave a comment for why this is needed? What are you expecting sigstore-go to do if this is nil?

cmurphy (commented):
This looks really good! I left a few questions.

Even though this PR is pretty big, I think it might be good to do all of sign, attest, and attest-blob in this PR too so there's not so much risk of divergence as time goes on. Open to other opinions on that.

cmd/cosign/cli/sign/sign_blob.go
aaronlew02: e2e tests that don't set `ko.RekorVersion` will enter this logic with a default value of 0. WDYT about modifying the e2e tests to set this value where necessary?

cmd/cosign/cli/signcommon/common.go
aaronlew02: The e2e tests pass after undoing the changes to this function, so I likely overestimated the need for these changes and will keep them undone.

My initial reasoning was that since _all_ signing will be handled by sigstore-go, there are cases where the `trustedMaterial` input to `SignData` will need to be `nil` so as to prevent sigstore-go from verifying on the signing path despite user intention. However, currently, there's an imperative for the user to provide a trusted root with custom service URLs and no signing config (for example, for SCT verification), meaning that these defaults should remain.

cmd/cosign/cli/signcommon/common.go
aaronlew02: See [here](https://github.com/sigstore/cosign/pull/4618#discussion_r2734310071).

pkg/cosign/bundle/sign.go
aaronlew02: `SignData` has a `ko.TrustedMaterial` input, and when an e2e test calls `SignBlobCmd` without setting `ko.TrustedMaterial` (as it does not expect `SignData` to be called), `bundleOpts.TrustedRoot` is not set in `SignData`. A nil `TrustedMaterial` field in `trustedMaterial` leads to a nil pointer error, as sigstore-go attempts to access non-existent services from the `TrustedMaterial`.

cmd/cosign/cli/sign/sign_blob.go
cmurphy: Yeah don't be afraid to modify the e2e tests to mimic setting the flags.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -132,88 +111,60 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
...
+	data, err := io.ReadAll(&payload)
 	if err != nil {
-		return nil, fmt.Errorf("signing blob: %w", err)
+		return nil, fmt.Errorf("reading payload: %w", err)
+	}
+	content := &sign.PlainData{
+		Data: data,
+	}
+	// If a certificate is provided, clear the idToken so that SignData uses
+	// that certificate instead of fetching a new one from Fulcio.
```
cmurphy: I'm curious why this came up and why this wasn't a problem before this change? What do you think of changing SignData to handle this by just reversing the switch cases?

pkg/cosign/bundle/sign.go
cmurphy: If we're in this part of the switch statement, there is public key material that needs to be used to verify the bundle as part of `sign.Bundle`. If `bundleOpts.TrustedRoot` isn't set, [this whole section](https://github.com/sigstore/sigstore-go/blob/18f941817a298f6443f70e16c82f306a52928977/pkg/sign/signer.go#L160-L185) gets skipped. Could you change this to 1) remove the check for `keyTrustedMaterial` being nil since it is set just above, and 2) ensure `bundleOpts.TrustedRoot` is set using the `keyTrustedMaterial` with or without the trusted root?

cmd/cosign/cli/sign/sign_blob.go
aaronlew02: This is no longer an issue, so I've undone the change. This was likely being broken by one of the other changes that I've undone.

pkg/cosign/bundle/sign.go
aaronlew02: > If bundleOpts.TrustedRoot isn't set, [this whole section](https://github.com/sigstore/sigstore-go/blob/18f941817a298f6443f70e16c82f306a52928977/pkg/sign/signer.go#L160-L185) gets skipped.

Keyful signing without a signing config, as shown by [the `keyOpts` defined in `TestSignBlobBundle`](https://github.com/sigstore/cosign/blob/71291bec3bc6f822b2fd57317cde937214cf1539/test/e2e_test.go#L3172) (which would fail here), does not require a trusted root, and so my understanding is that we would want to skip this section in that circumstance. 

> remove the check for keyTrustedMaterial being nil since it is set just above

Done.

> ensure bundleOpts.TrustedRoot is set using the keyTrustedMaterial with or without the trusted root

With a nil trusted root, the above circumstance will result in a nil pointer error, and with an empty trusted root, verification will fail.

pkg/cosign/bundle/sign.go
```diff
@@ -79,11 +79,13 @@ func SignData(ctx context.Context, content sign.Content, keypair sign.Keypair, i
 		keyTrustedMaterial := root.NewTrustedPublicKeyMaterial(func(_ string) (root.TimeConstrainedVerifier, error) {
 			return key, nil
 		})
-		trustedMaterial := &verifyTrustedMaterial{
-			TrustedMaterial:    bundleOpts.TrustedRoot,
-			keyTrustedMaterial: keyTrustedMaterial,
+		if bundleOpts.TrustedRoot != nil {
```
cmurphy, line 83: Providing a trusted root to `sign.Bundle` needs to not be optional, because it will skip this verification step as part signing the bundle https://github.com/sigstore/sigstore-go/blob/3f2ee9eda9b2891647354114ff9cba1167bbcb52/pkg/sign/signer.go#L160-L185. The trusted root will need to be created from other sources the way you're doing for the signing config. This might also be a good opportunity to add an e2e test to make sure that not providing the trusted root doesn't accidentally cause a bundle to pass this step.

cmd/cosign/cli/options/key.go
```diff
@@ -27,6 +27,7 @@ type KeyOpts struct {
 	KeyRef               string
 	FulcioURL            string
 	RekorURL             string
+	RekorVersion         uint32
```
steiza: I'm a little nervous about adding this option - what should we do if people specify this **and** SigningConfig?

We could either make those options mutually exclusive (which is a pain) or always default to Rekor v1 when you aren't using SigningConfig and keep moving folks over to use SigningConfig sooner than later. I'm curious what @Hayden-IO thinks here as well.

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -233,60 +175,180 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
...
+	}
 	return sig, nil
 }
 
-// Extract an encoded certificate from the SignerVerifier. Returns (nil, nil) if verifier is not a certificate.
-func extractCertificate(ctx context.Context, sv *signcommon.SignerVerifier) ([]byte, error) {
-	signer, err := sv.Bytes(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("error getting signer: %w", err)
+func newSigningConfigFromKeyOpts(ko options.KeyOpts, shouldUpload bool) (*root.SigningConfig, error) {
```
steiza: This functions (and `extractElementsFromProtoBundle()` and `newLegacyBundleFromProtoBundleElements()`) are awesome!

It doesn't have to part of this pull request, but probably we are going to want to move them to `cmd/cosign/cli/signcommon/common.go` sooner than later.

steiza (commented):
Awesome work! The only thing I'm not quite sure how about is how we should handle `RekorVersion` as a user input argument.

pkg/cosign/bundle/sign.go
aaronlew02: > The trusted root will need to be created from other sources the way you're doing for the signing config.

That sounds like a good solution, but I believe creating a trusted root in many circumstances is not possible as is creating a signing config. For an artificial signing config, only service URLs are necessary, which means that one can always be created. For a trusted root, verification-related information (like certificates) is necessary, which the user is not currently required to provide and doesn't always seem possible to determine.

Is the verification step to which you linked always necessary for Cosign? For example, currently, [keyful signing without a signing config](https://github.com/sigstore/cosign/blob/661243c65e03f815d16f3be57b1e65c2683d4794/test/e2e_test.go#L3141) skips any verification on signing.

cmd/cosign/cli/sign/sign_blob.go
aaronlew02: That's a great idea. If this PR comes to include other commands, this will likely be moved to `common.go` here as well.

pkg/cosign/bundle/sign.go
cmurphy: I see you're right, verification material isn't required in cosign for signing unless there is a CT log involved, so we can't start requiring it now.

cmd/cosign/cli/sign/sign_blob.go
cmurphy: I recommend expanding this PR to the other sign/attest commands so that they are all in sync and there doesn't need to be a refactor to move things into signcommon.

cmd/cosign/cli/options/key.go
aaronlew02: This is a good question, and I'm also interested in what @Hayden-IO thinks here.

Hayden-IO: Yep, I agree with Zach. What was the need for this? The reason we went with creating a configuration for signing was to avoid manually specifying API version, as that’s not a great UX.

aaronlew02:
# Testing

The following commands can be used to test the signing commands (`sign`, `sign-blob`, `attest`, and `attest-blob`) across various use cases. Signing commands must be run on this PR's branch and verification commands should be run on the `main` branch. 

## 1. Bundle Output
These tests validate the behavior of the `--new-bundle-format` flag and the resulting artifacts (bundles and/or detached verification material).

### `cosign sign`
* **New Bundle (OCI Upload):**
  ```bash
  cosign sign --new-bundle-format=true [IMAGE_URI]
  cosign verify --new-bundle-format=true --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **Legacy Format (OCI Upload, Raw Local Files & Bundle):**
  ```bash
  cosign sign --new-bundle-format=false --output-signature sig.b64 --output-certificate cert.pem --output-payload payload.json --bundle sign.bundle --use-signing-config=false [IMAGE_URI]
  cosign verify --new-bundle-format=false --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  cosign verify --new-bundle-format=false --signature sig.b64 --certificate cert.pem --payload payload.json --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  cosign verify --new-bundle-format=false --signature sig.b64 --certificate cert.pem --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  # (Note: cosign verify does not support checking a local --bundle for remote OCI artifacts)
  ```
**Note:** The last two commands will fail at transparency log lookup because the provided certificate is not being attached to the signature. This is expected behavior and will be fixed by [#4737](https://github.com/sigstore/cosign/pull/4737).

### `cosign sign-blob`
* **New Bundle (Local File):**
  ```bash
  cosign sign-blob --new-bundle-format=true --bundle blob.bundle [BLOB]
  cosign verify-blob --new-bundle-format=true --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle blob.bundle [BLOB]
  ```
* **Legacy Format (Raw Local Files & Bundle):**
  ```bash
  cosign sign-blob --new-bundle-format=false --output-signature sig.b64 --output-certificate cert.pem --bundle blob.bundle [BLOB]
  cosign verify-blob --new-bundle-format=false --signature sig.b64 --certificate cert.pem --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB]
  cosign verify-blob --new-bundle-format=false --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle blob.bundle [BLOB]
  ```

### `cosign attest`
* **New Bundle (OCI Upload via Referrers):**
  ```bash
  cosign attest --new-bundle-format=true --predicate [PREDICATE] [IMAGE_URI]
  cosign verify-attestation --new-bundle-format=true --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **Legacy Format (OCI Upload via `.att` Tag & Local Bundle):**
  ```bash
  cosign attest --new-bundle-format=false --predicate [PREDICATE] --bundle attest.bundle --use-signing-config=false [IMAGE_URI]
  cosign verify-attestation --new-bundle-format=false --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  # (Note: cosign verify-attestation does not support checking a local --bundle for remote OCI artifacts)
  ```

### `cosign attest-blob`
* **New Bundle (Local File):**
  ```bash
  cosign attest-blob --new-bundle-format=true --predicate [PREDICATE] --bundle attest.bundle [BLOB]
  cosign verify-blob-attestation --new-bundle-format=true --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle attest.bundle [BLOB]
  ```
* **Legacy Format (Raw Local Files & Bundle):**
  ```bash
  cosign attest-blob --new-bundle-format=false --predicate [PREDICATE] --output-signature sig.b64 --output-certificate cert.pem --output-attestation att.json --bundle attest.bundle [BLOB]
  cosign verify-blob-attestation --new-bundle-format=false --signature sig.b64 --certificate cert.pem --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB]
  cosign verify-blob-attestation --new-bundle-format=false --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle attest.bundle [BLOB]
  ```

---

## 2. Transparency Log Upload
These tests validate the `--tlog-upload` boolean flag using explicit key-based signing.

### `cosign sign`
* **Upload True (Default):**
  ```bash
  cosign sign --key [KEY] --tlog-upload=true [IMAGE_URI]
  cosign verify --key [KEY] [IMAGE_URI]
  ```
* **Upload False:**
  ```bash
  cosign sign --key [KEY] --tlog-upload=false --use-signing-config=false [IMAGE_URI]
  cosign verify --key [KEY] --insecure-ignore-tlog=true [IMAGE_URI]
  ```

### `cosign sign-blob`
* **Upload True:**
  ```bash
  cosign sign-blob --key [KEY] --tlog-upload=true --bundle [BUNDLE] [BLOB]
  cosign verify-blob --key [KEY] --bundle [BUNDLE] [BLOB]
  ```
* **Upload False:**
  ```bash
  cosign sign-blob --key [KEY] --tlog-upload=false --use-signing-config=false --bundle [BUNDLE] [BLOB]
  cosign verify-blob --key [KEY] --insecure-ignore-tlog=true --bundle [BUNDLE] [BLOB]
  ```

### `cosign attest`
* **Upload True:**
  ```bash
  cosign attest --key [KEY] --predicate [PREDICATE] --tlog-upload=true [IMAGE_URI]
  cosign verify-attestation --key [KEY] [IMAGE_URI]
  ```
* **Upload False:**
  ```bash
  cosign attest --key [KEY] --predicate [PREDICATE] --tlog-upload=false --use-signing-config=false [IMAGE_URI]
  cosign verify-attestation --key [KEY] --insecure-ignore-tlog=true [IMAGE_URI]
  ```

### `cosign attest-blob`
* **Upload True:**
  ```bash
  cosign attest-blob --key [KEY] --predicate [PREDICATE] --tlog-upload=true --bundle [BUNDLE] [BLOB]
  cosign verify-blob-attestation --key [KEY] --bundle [BUNDLE] [BLOB]
  ```
* **Upload False:**
  ```bash
  cosign attest-blob --key [KEY] --predicate [PREDICATE] --tlog-upload=false --use-signing-config=false --bundle [BUNDLE] [BLOB]
  cosign verify-blob-attestation --key [KEY] --insecure-ignore-tlog=true --bundle [BUNDLE] [BLOB]
  ```

---

## 3. URLs vs. Signing Config
These tests validate that infrastructure endpoints provided via explicit flags, TUF, and a custom signing config behave correctly.

### `cosign sign`
* **Explicit URLs:**
  ```bash
  cosign sign --rekor-url [URL] --fulcio-url [URL] --use-signing-config=false [IMAGE_URI]
  cosign verify --rekor-url [URL] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **TUF Signing Config:**
  ```bash
  cosign sign --use-signing-config=true [IMAGE_URI]
  cosign verify --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **Local Signing Config File:**
  ```bash
  cosign sign --signing-config [CONFIG_FILE] [IMAGE_URI]
  cosign verify --trusted-root [TRUSTED_ROOT_FILE] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```

### `cosign sign-blob`
* **Explicit URLs:**
  ```bash
  cosign sign-blob --rekor-url [URL] --fulcio-url [URL] --use-signing-config=false --bundle blob.bundle [BLOB]
  cosign verify-blob --rekor-url [URL] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle blob.bundle [BLOB]
  ```
* **TUF Signing Config:**
  ```bash
  cosign sign-blob --use-signing-config=true --bundle blob.bundle [BLOB]
  cosign verify-blob --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle blob.bundle [BLOB]
  ```
* **Local Signing Config File:**
  ```bash
  cosign sign-blob --signing-config [CONFIG_FILE] --bundle blob.bundle [BLOB]
  cosign verify-blob --trusted-root [TRUSTED_ROOT_FILE] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle blob.bundle [BLOB]
  ```

### `cosign attest`
* **Explicit URLs:**
  ```bash
  cosign attest --predicate [PREDICATE] --rekor-url [URL] --fulcio-url [URL] --use-signing-config=false [IMAGE_URI]
  cosign verify-attestation --rekor-url [URL] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **TUF Signing Config:**
  ```bash
  cosign attest --predicate [PREDICATE] --use-signing-config=true [IMAGE_URI]
  cosign verify-attestation --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **Local Signing Config File:**
  ```bash
  cosign attest --predicate [PREDICATE] --signing-config [CONFIG_FILE] [IMAGE_URI]
  cosign verify-attestation --trusted-root [TRUSTED_ROOT_FILE] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```

### `cosign attest-blob`
* **Explicit URLs:**
  ```bash
  cosign attest-blob --predicate [PREDICATE] --rekor-url [URL] --fulcio-url [URL] --use-signing-config=false --bundle attest.bundle [BLOB]
  cosign verify-blob-attestation --rekor-url [URL] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle attest.bundle [BLOB]
  ```
* **TUF Signing Config:**
  ```bash
  cosign attest-blob --predicate [PREDICATE] --use-signing-config=true --bundle attest.bundle [BLOB]
  cosign verify-blob-attestation --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle attest.bundle [BLOB]
  ```
* **Local Signing Config File:**
  ```bash
  cosign attest-blob --predicate [PREDICATE] --signing-config [CONFIG_FILE] --bundle attest.bundle [BLOB]
  cosign verify-blob-attestation --trusted-root [TRUSTED_ROOT_FILE] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] --bundle attest.bundle [BLOB]
  ```

---

## 4. OCI Upload
These tests validate the ability to toggle pushing signatures and attestations directly to an OCI registry.

### `cosign sign`
* **Upload True (Default):**
  ```bash
  cosign sign --upload=true [IMAGE_URI]
  cosign verify --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **Upload False (Local Only):**
  ```bash
  cosign sign --upload=false --bundle local.bundle [IMAGE_URI]
  # Verification of local bundles for remote images is not supported directly via 'cosign verify'
  ```

### `cosign attest`
* **Upload True (Default):**
  ```bash
  cosign attest --predicate [PREDICATE] --no-upload=false [IMAGE_URI]
  cosign verify-attestation --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
  ```
* **Upload False (Local Only):**
  ```bash
  cosign attest --predicate [PREDICATE] --no-upload=true --bundle local.bundle [IMAGE_URI]
  # Verification of local bundles for remote images is not supported directly via 'cosign verify-attestation'
  ```

cmd/cosign/cli/options/key.go
aaronlew02: The `--rekor-version` flag was intended to allow specification of the version of a Rekor service provided by URL via the `--rekor-url` flag, so that the version could be specified in the resulting signing config.

I've removed the `--rekor-version` flag in favor of a default to v1 when a Rekor URL is provided via flag.

test/e2e_test.go
```diff
@@ -657,19 +657,8 @@ func TestSignVerifyWithTUFMirror(t *testing.T) {
 	tests := []struct {
 		name          string
 		targets       []targetInfo
-		wantSignErr   bool
 		wantVerifyErr bool
 	}{
-		{
-			name: "invalid CT key name with no usage",
```
cmurphy, line 664: Can you say more about why this test was removed?

In a scenario where 1) Fulcio is used to sign, and 2) TUF is not providing either a trusted root or legacy-style target names and usage identifiers, how is the sign command verifying the SCT? Does sigstore-go just not do that? Is that a change in behavior we're willing to accept?

aaronlew02: > Can you say more about why this test was removed?

This test now fails, as it expects an error on signing which no longer occurs.

> Does sigstore-go just not do that?

Correct. In fact, sigstore-go-based signing in Cosign has always avoided SCT verification on signing, as [`Bundle`](https://github.com/sigstore/sigstore-go/blob/d149fde00a0c53d2bab889cf447aef7f62a1cc8c/pkg/sign/signer.go#L55), which supports verification on signing when a trusted root is present, does **_not_** call `verify.WithSignedCertificateTimestamps`. The services in `BundleOpts` are populated according to the signing config, which does not contain CT logs.

> Is that a change in behavior we're willing to accept?

The disparity introduced here (i.e., the removal of SCT verification on signing) only applies to users who use a signing config. Given that this PR intends to minimize disparity, I'm open to continue mandating SCT verification if possible, but that creates a question of whether we want to depart from previous behavior by enabling this for all users.

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,142 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
 		OCIRemoteOpts: ociremoteOpts,
 	}
 
-	if c.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
+	if c.SigningConfig == nil {
+		c.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(c.KeyOpts, c.TlogUpload)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
```
steiza, line 141: I love this! This is exactly what we want - assembling a `SigningConfig` instead of having diverging codepaths depending on if you're using `SigningConfig` or not.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -146,87 +148,102 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
-		bts := bundleComponents.SignerBytes
-		if err := os.WriteFile(c.OutputCertificate, bts, 0600); err != nil {
+		certPem := pem.EncodeToMemory(pemBlock)
+		if err := os.WriteFile(c.OutputCertificate, certPem, 0600); err != nil {
 			return fmt.Errorf("create certificate file: %w", err)
 		}
 		fmt.Fprintln(os.Stderr, "Certificate written to file ", c.OutputCertificate)
 	}
 
+	if c.RFC3161TimestampPath != "" {
```
steiza, line 222: I guess we previously weren't outputting a signed timestamp if a user was doing `cosign attest-blob --new-bundle-format=false --rfc3161-timestamp-bundle`? Oops!

steiza (commented):
First of all, this is amazing work. This is exactly the way we want these changes to be made, and this will make it very easy to dramatically simplify the codebase as we deprecate the old output formats.

That said, I think I found a bug (albeit a very minor one):

I'm using an empty signing config:
```
$ cat empty_signing_config.json 
{"mediaType":"application/vnd.dev.sigstore.signingconfig.v0.2+json", "rekorTlogConfig":{}, "tsaConfig":{}}
```

I'm trying to sign with just a key, no PGI, and I get the immutable records notice (and I don't think that I should):
```
$ go run cmd/cosign/main.go sign-blob --signing-config empty_signing_config.json --key cosign.key --bundle blob-key.sigstore.json ../sigstore-go/examples/hello_world.txt

	The sigstore service, hosted by sigstore a Series of LF Projects, LLC, ...
```

For what it's worth, the resulting bundle looks correct and doesn't include a Rekor entry:
```
$ cat blob-key.sigstore.json | jq 
{
  "verificationMaterial": {
    "publicKey": {
      "hint": "KphQpmcaLAPdFYX3K1YoGLXC6+jkCGRydpUd45IP1Xo="
    }
  },
  "messageSignature": {
    "messageDigest": {
      "algorithm": "SHA2_256",
      "digest": "qUiQTy8PR5uPgZdpSzAYSw0u0cHNKh7A+4XSmaGSpEc="
    },
    "signature": "MEYCIQCnvxCdQtvWcAhzyVmZbbDF9du/MVTxufE3kBXfTMO87gIhAPIjoOWEra/nCRJ/FLIHW6/XBKNbLxt6hzQtfofBlcbB"
  }
}
```

cmd/cosign/cli/attest/attest_blob.go
Hayden-IO: It looks like this is handling https://github.com/sigstore/cosign/issues/4594, cool!

aaronlew02:
> I'm using an empty signing config ... I'm trying to sign with just a key, no PGI, and I get the immutable records notice (and I don't think that I should)

@steiza Thanks for pointing that out! That shouldn't be happening when a signing config is used. I've placed the call to `ShouldUploadToTlog` in `sign` and `sign-blob` under the `if ko.SigningConfig == nil` block to address this.

cmd/cosign/cli/attest/attest_blob.go
aaronlew02: @Hayden-IO Yes, but with a caveat: the use case in #4594 is for the new bundle format, and this PR only enables the RFC 3161 timestamp output for the old bundle format. This is in keeping with the other output flags, whose deprecation in the new-bundle use case is [documented](https://github.com/sigstore/cosign/blob/36f40082f3c507e131cb9d926b75b36606160483/cmd/cosign/cli/signcommon/common.go#L692). 

I have added a warning in `LoadTrustedMaterialAndSigningConfig` that `"--rfc3161-timestamp-path is deprecated when using --new-bundle-format and will be ignored"`.

**Note:** `--rfc3161-timestamp-path` is already deprecated for `sign-blob`, `attest-blob`, and `verify`, with a message stating, `"please use --bundle to provide the output bundle location, which will include the signed timestamp"`. However, RFC 3161 timestamps are not inserted into old-format bundles, because `LocalSignedPayload`, which forms the old-format bundle, does not contain an `RFC3161Timestamp`, unlike `SignedPayload`.

steiza (dismissed):
This LGTM!

There's a merge conflict, and a few PRs that are open that have overlapping functionality. @Hayden-IO I think we should prioritize landing this when the merge conflicts are resolved.

steiza (dismissed):

steiza (dismissed):

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	bundleOpts := signcommon.CommonBundleOpts{
 		Payload:       payload,
 		Digest:        digest,
-		PredicateType: types.CosignSignPredicateType,
```
cmurphy, line 126: @Hayden-IO it looks like you introduced the hardcoded PredicateType for the signing config path for attestations in https://github.com/sigstore/cosign/pull/4371, do you remember why or whether it was intentional? This looks like a good cleanup but just want to make sure it's not breaking something that had a good reason for it.

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+		}
+		if certPem != nil {
+			opts = append(opts, static.WithCertChain(certPem, chainPem))
+		}
+
+		if bundleComponents.RFC3161Timestamp != nil {
+			opts = append(opts, static.WithRFC3161Timestamp(cbundle.TimestampToRFC3161Timestamp(bundleComponents.RFC3161Timestamp.GetSignedTimestamp())))
+		}
+
+		predicateType, err := options.ParsePredicateType(c.PredicateType)
```
cmurphy: I think you have this already as `predicateURI` on line 79 (looks like the old version unnecessarily repeated this)

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+	if c.BundlePath != "" {
+		if err := os.WriteFile(c.BundlePath, legacyBundleBytes, 0600); err != nil {
+			return fmt.Errorf("create bundle file: %w", err)
+		}
+		ui.Infof(ctx, "Wrote bundle to file %s", c.BundlePath)
 	}
 
-	// Publish the attestations associated with this entity
-	return ociremote.WriteAttestations(digest.Repository, newSE, ociremoteOpts...)
+	if !c.NoUpload {
```
cmurphy: nit - could reverse this condition and return early here, it's a little more Go-idiomatic

cmd/cosign/cli/sign/sign.go
```diff
@@ -234,59 +261,103 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
...
 
-		b64sig, err := ociSig.Base64Signature()
+		bundleComponents, err := signcommon.ExtractComponentsFromProtoBundle(&pb)
 		if err != nil {
-			return err
+			return fmt.Errorf("extracting components from bundle: %w", err)
+		}
+
+		var certPem, chainPem []byte
+		for j, c := range bundleComponents.Certificates {
```
cmurphy: I'm curious why the need to iterate over certificates and separate the chain from the leaf only came up for sign.go and not the other commands?

test/e2e_test.go
cmurphy: Since we weren't verifying the SCT before when signing config was used, I'm open to keeping that logic the same for both paths with this change. I'm curious if @Hayden-IO has thoughts here.

cmurphy (commented):
This is looking great! I still have a couple of questions and minor suggestions but I'm fine with not holding this up and moving this forward as-is. The only other suggestion I might make, probably for a followup PR, is that there's a small amount of duplicated code in the clients that might be worthwhile to move to signcommon.

cmd/cosign/cli/signcommon/common.go
```diff
@@ -711,3 +588,162 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
 
 	return nil
 }
+
+func ExtractComponentsFromProtoBundle(bundle *protobundle.Bundle) (*BundleComponents, error) {
```
Hayden-IO, line 596: Can you add a comment for any exported functions?

cmd/cosign/cli/signcommon/common.go
```diff
@@ -711,3 +588,162 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
+
+	return bc, nil
+}
+
+func NewLegacyBundleFromProtoBundleComponents(bc *BundleComponents, keypair sign.Keypair) ([]byte, error) {
+	signedPayload := cosign.LocalSignedPayload{
+		Base64Signature: base64.StdEncoding.EncodeToString(bc.Signature),
+	}
+
+	if len(bc.Certificates) > 0 {
```
Hayden-IO, line 685: Can you use marshaling functions from https://github.com/sigstore/sigstore/blob/main/pkg/cryptoutils/certificate.go?

cmd/cosign/cli/signcommon/common.go
```diff
@@ -711,3 +588,162 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
+		cert := bc.Certificates[0]
+		pemBlock := &pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: cert.GetRawBytes(),
+		}
+		certPem := pem.EncodeToMemory(pemBlock)
+		signedPayload.Cert = base64.StdEncoding.EncodeToString(certPem)
+	} else if keypair != nil {
+		pubKeyPem, err := keypair.GetPublicKeyPem()
```
Hayden-IO: Why are we decoding and re-encoding a public key?

cmd/cosign/cli/signcommon/common.go
```diff
@@ -711,3 +588,162 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
+				LogIndex:       bc.RekorEntry.GetLogIndex(),
+				LogID:          hex.EncodeToString(bc.RekorEntry.GetLogId().GetKeyId()),
+			},
+		}
+	}
+
+	return json.Marshal(signedPayload)
+}
+
+func ProtoHashAlgoToHash(hashFunc pb_go_v1.HashAlgorithm) crypto.Hash {
```
Hayden-IO: It looks like this is only being used in sign_blob, can it stay there as an unexported function?

cmd/cosign/cli/signcommon/common.go
```diff
@@ -711,3 +588,162 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
+			ValidityPeriodStart: time.Now(),
+		})
+	}
+
+	var rekorServices []root.Service
+	var rekorConfig root.ServiceConfiguration
+	if ko.RekorURL != "" && shouldUpload {
+		rekorServices = append(rekorServices, root.Service{
+			URL:                 ko.RekorURL,
+			MajorAPIVersion:     1,
```
Hayden-IO, line 725: Let's make sure we document this in the function comment that this intentionally only works with Rekor v1.

cmd/cosign/cli/attest/attest.go
Hayden-IO: No, I really don't know why I did! I've been staring at this old PR for a bit and can't recall any reason why I hardcoded this. I'm guessing it was a bad copy-paste from https://github.com/sigstore/cosign/blob/a09afa97480a0a4a20ad6314600598b7bddc8c0c/cmd/cosign/cli/sign/sign.go#L193.

This was reported as an issue on slack a couple of weeks ago, this is a good bug fix. Arguably this could be moved out into its own PR but this is fine here.

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
+	if c.SigningConfig == nil {
+		c.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(c.KeyOpts, c.TlogUpload)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
 	}
 
-	bundleComponents, closeSV, err := signcommon.GetBundleComponents(ctx, c.CertPath, c.CertChainPath, c.KeyOpts, c.NoUpload, c.TlogUpload, payload, digest, c.RekorEntryType)
+	bundleBytes, keypair, err := signcommon.NewAttestationBundle(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
```
Hayden-IO: What is `keypair` used for? It looks like it's only used for the public key, not for signing. Could we return the public key instead? This would mean we could call `close` on the keypair in `NewAttestationBundle`.

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+		}
+		ui.Infof(ctx, "Wrote bundle to file %s", c.BundlePath)
 	}
 
-	// Publish the attestations associated with this entity
-	return ociremote.WriteAttestations(digest.Repository, newSE, ociremoteOpts...)
+	if !c.NoUpload {
+		var certPem, chainPem []byte
+		for i, cert := range bundleComponents.Certificates {
+			p := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.GetRawBytes()})
```
Hayden-IO: Can we use `cryptoutils.MarshalCertificateToPEM`?

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+			if i == 0 {
+				certPem = p
+			} else {
+				chainPem = append(chainPem, p...)
```
Hayden-IO: Let's double check that each PEM-encoded certificate in the chain is newline-delimited.

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+			return err
+		}
+		predicateTypeAnnotation := map[string]string{
+			"predicateType": predicateType,
+		}
+		// Add predicateType as manifest annotation
+		opts = append(opts, static.WithAnnotations(predicateTypeAnnotation))
+
+		if bundleComponents.RekorEntry != nil {
+			rb := &cbundle.RekorBundle{
```
Hayden-IO: We repeat this block of code twice, so could we have a function to convert between the new and old Rekor bundle?

cmd/cosign/cli/attest/attest.go
```diff
@@ -124,85 +132,145 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
+		ociSig, err := static.NewAttestation(bundleComponents.Signature, opts...)
+		if err != nil {
+			return fmt.Errorf("creating attestation: %w", err)
+		}
+
+		// We don't actually need to access the remote entity to attach things to it
+		// so we use a placeholder here.
+		se := ociremote.SignedUnknown(digest, ociremoteOpts...)
+
+		ddVerifier, err := signature.LoadVerifier(keypair.GetPublicKey(), crypto.SHA256)
```
Hayden-IO: This is an issue - we can't hardcode `crypto.SHA256` as the hash algorithm can vary. Either you could call https://github.com/sigstore/cosign/blob/a09afa97480a0a4a20ad6314600598b7bddc8c0c/internal/key/svkeypair.go#L85 or use the original signerverifier object from GetKeypairAndToken.

We've done a lot of work around crypto agility, so let's make sure that this change isn't going to affect that. Can you generate a ECDSA-P384 key, which will default to a sha-384 hashing algorithm. and verify that signing and verification still work?

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -146,87 +148,106 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
+	}
 
-			contents, err = cbundle.MakeNewBundle(pubKey, bundleComponents.RekorEntry, payload, bundleComponents.SignedPayload, bundleComponents.SignerBytes, bundleComponents.TimestampBytes)
-			if err != nil {
-				return err
-			}
-		} else {
-			signedPayload.Base64Signature = base64.StdEncoding.EncodeToString(bundleComponents.SignedPayload)
-			signedPayload.Cert = base64.StdEncoding.EncodeToString(bundleComponents.SignerBytes)
+	var cert *pb_go_v1.X509Certificate
```
Hayden-IO: Can this be initialized in the scope of the `c.OutputCertificate` block rather than outside? Or am I missing where `cert` is used?

cmd/cosign/cli/attest/attest_blob.go
Hayden-IO: Thanks for mentioning this difference, I took care of 4594 by adding the warning in a separate PR.

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -146,87 +148,106 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
-			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Was a certificate used? ", err)
-			return nil
-
+		if cert == nil {
+			return fmt.Errorf("no certificate found in bundle")
 		}
-		if len(cert) != 1 {
-			fmt.Fprintln(os.Stderr, "Could not output signer certificate. Expected a single certificate")
-			return nil
+		pemBlock := &pem.Block{
```
Hayden-IO: same here, let's use cryptoutils.

cmd/cosign/cli/sign/sign.go
```diff
@@ -196,17 +200,40 @@ func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpt
 		OCIRemoteOpts: ociremoteOpts,
 	}
 
-	if ko.SigningConfig != nil {
-		return signcommon.WriteNewBundleWithSigningConfig(ctx, ko, signOpts.Cert, signOpts.CertChain, bundleOpts, ko.SigningConfig, ko.TrustedMaterial)
+	if ko.SigningConfig == nil {
+		shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, ko, digest, signOpts.TlogUpload)
```
Hayden-IO, line 201: Let's not solve this in this PR, but we should be consistent with --tlog-upload support for sign/sign-blob and attest/attest-blob. Also, I noticed --upload is only for sign, not attest. Can you file an issue to track and follow up after this PR is merged?

cmd/cosign/cli/sign/sign.go
```diff
@@ -196,17 +200,40 @@ func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpt
...
+	}
+
+	if signOpts.BundlePath != "" {
+		if err := os.WriteFile(signOpts.BundlePath, bundleBytes, 0600); err != nil {
+			return fmt.Errorf("create bundle file: %w", err)
+		}
+		ui.Infof(ctx, "Wrote bundle to file %s", signOpts.BundlePath)
+	}
+
+	if signOpts.Upload {
```
Hayden-IO, line 223: As another follow up, we should mandate that at least upload or bundle-path is set, otherwise you're throwing away the signature. Can you either update the existing issue about mandating --bundle-path or create a new one and link that here in a comment?

cmd/cosign/cli/sign/sign_blob.go
Hayden-IO, line 1: Check out the comments on attest.go that are relevant here, I won't duplicate all the comments.

cmd/cosign/cli/signcommon/common.go
```diff
@@ -711,3 +588,162 @@ func LoadTrustedMaterialAndSigningConfig(ctx context.Context, ko *options.KeyOpt
...
+	}
+
+	if vm := bundle.GetVerificationMaterial(); vm != nil {
+		if chain := vm.GetX509CertificateChain(); chain != nil && len(chain.GetCertificates()) > 0 {
+			bc.Certificates = chain.GetCertificates()
+		} else if cert := vm.GetCertificate(); cert != nil {
+			bc.Certificates = []*pb_go_v1.X509Certificate{cert}
+		}
+		if tlogEntries := vm.GetTlogEntries(); len(tlogEntries) > 0 {
+			bc.RekorEntry = tlogEntries[0]
```
Hayden-IO: Can RekorEntry and RFC3161Timestamp be repeated to handle multiple entries or timestamp signatures?

If the legacy bundle construction can only handle one of each, then it's in that function we should extract only one from the components.

test/e2e_test.go
Hayden-IO: Is this removed because we can't guarantee that a trusted root is available?

I think we discussed this early on, and one idea was populating an incomplete trusted root with just a CT log public key. I believe the issue was sigstore-go couldn't handle this.

I'm OK with making both paths the same, but we should think through any cleanups we should do after this lands. For example, the "insecure skip ct verification" flag can be removed/deprecated as it's now ignored.

Hayden-IO: Related, how we are handling verification during signing? We can't mandate a trusted root be available as that's a breaking change.

Hayden-IO (changes requested):
Exciting to see this coming together!

test/e2e_test.go
cmurphy: I think the only thing we have been verifying during signing was the SCT, and only in the non-sigstore-go path, so with this change there is no more verification during signing.

aaronlew02: @Hayden-IO

> Is this removed because we can't guarantee that a trusted root is available?

This is removed because signing via sigstore-go will no longer verify SCTs on signing, regardless of the content of the trusted root.

> Related, how we are handling verification during signing?

Verification during signing remains the same for users who use a signing config (i.e., it does not occur).  SCT verification (the only signing-side verification handled by Cosign) is removed for users who do not use a signing config.

> We can't mandate a trusted root be available as that's a breaking change.

Is that necessary? Users who provide services not available in the initialized TUF repo are (and would continue to be) expected to provide a trusted root containing those services.

cmd/cosign/cli/attest/attest.go
aaronlew02: We call `EncodeToMemory`, which calls `Encode`, which [appends a newline to the footer of each certificate](https://github.com/golang/go/blob/96d6d38872a56eba26cd2e3dbd65bab138f9c599/src/encoding/pem/pem.go#L316).

test/e2e_test.go
Hayden-IO: > Verification during signing remains the same for users who use a signing config (i.e., it does not occur)

We currently do verify on signing - see [this snippet](https://github.com/sigstore/cosign/blob/a09afa97480a0a4a20ad6314600598b7bddc8c0c/pkg/cosign/bundle/sign.go#L40-L42) and initialization [here](https://github.com/sigstore/cosign/blob/a09afa97480a0a4a20ad6314600598b7bddc8c0c/cmd/cosign/cli/signcommon/common.go#L663-L679).

If you don't have a trusted root file in your TUF repo, you'll see a warning like `Could not fetch trusted_root.json from the TUF repository`, however those individual targets are only used in the older signing path, not this new one with sigstore-go.

For simplicity's sake, I'd rather not try to replicate just verifying the SCT. In Cosign v4 we'll likely only support a trusted root file anyways. The action item here would be to update the wording of the [warning](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/signcommon/common.go#L676C20-L676C77) to instead say something like "skipping verification after signing" instead of mentioning individual targets.

@cmurphy, since you did the work for upgrading the TUF client in Cosign, does this sound OK/correct?

aaronlew02: > We currently do verify on signing

Sorry, I misspoke. That's correct. I must have been thinking about Cosign-native verification (i.e., SCT verification), specifically.

cmd/cosign/cli/sign/sign.go
aaronlew02: > we should be consistent with --tlog-upload support for sign/sign-blob and attest/attest-blob

`--tlog-upload` is used in all of those commands to determine whether Rekor is included in the constructed signing config. Is the inconsistency that only for `sign` and `sign-blob` is the transparency log upload ultimately decided by user confirmation?

> Also, I noticed --upload is only for sign, not attest

Correct, though `attest` has `--no-upload`, which is simply the reverse condition.

cmd/cosign/cli/signcommon/common.go
aaronlew02: The intention was to normalize the encoded PEM output. However, `GetPublicKeyPem` calls `MarshalPublicKeyToPEM`, which should handle that. I've removed the decoding block, since it's redundant.

test/e2e_test.go
cmurphy: > @cmurphy, since you did the work for upgrading the TUF client in Cosign, does this sound OK/correct?

@Hayden-IO yes this is right, we only get to the point of needing the individual CT pub key target and using it to verify the SCT in the non-signing-config paths and when there is also no trusted root. In checking this I also noticed that there is a `--insecure-skip-verify` specifically for opting into not verifying the Fulcio SCT, so  as part of this I guess we should also deprecate that flag and update the wording to explain that it doesn't do anything anymore.

Hayden-IO: +1 to deprecating that flag - I'll be happy to deprecate it anyways, it was misleading/confusing for private deployments.

cmd/cosign/cli/sign/sign.go
Hayden-IO: I thought that `ShouldUploadToTlog` wasn't called in attest/attest-blob, but no, it's that there's a different way that --tlog-upload is plumbed through, https://github.com/sigstore/cosign/blob/af30fe6fa838cd1f29f5ee03a90dcb781f0f3872/cmd/cosign/cli/signcommon/common.go#L606-L611. Sigh, too much duplication. Anywho, please ignore this!

> Correct, though attest has --no-upload, which is simply the reverse condition.

Ah, thanks. Yea, this is an inconsistency we can fix as part of the next Cosign.

cmd/cosign/cli/attest/attest.go
aaronlew02: I've updated `NewAttestationBundle` to instead return the original `PublicKey` object and the PEM-encoded version. The former is used exclusively for the `ddVerifier` in `sign` and `attest`, while the latter is used for legacy bundle construction.

test/e2e_test.go
aaronlew02: I've deprecated `--insecure-skip-verify`.

cmd/cosign/cli/sign/sign.go
aaronlew02: In `signDigest`, we now have a distinct sigstore-go signing event for each `payload`, which will result in multiple certificates, each embedded into their respective OCI signature and added to the bundle. Since `outputCertificate` only outputs a single certificate, I decided to only output the leaf.

cmd/cosign/cli/attest/attest.go
aaronlew02: I've replaced the hardcoded `crypto.SHA256` with the hash algorithm from the `keypair`.

I've confirmed that verification works with ECDSA-P384 in all available cases: key-based in all commands and keyless in `sign-blob`.

cmd/cosign/cli/sign/sign.go
aaronlew02: I've updated issue #4534 and linked the issue on relevant lines.

steiza (dismissed):
These updates LGTM!

test/e2e_test.go
steiza: This is fine as-is, and I think we should deprecate `--insecure-skip-verify` regardless.

But thinking about this more, the "problem" is that sigstore-go signing doesn't have a verification policy - it just looks at the trusted root, the bundle, and sort of infers what the verification policy should be. See things like https://github.com/sigstore/sigstore-go/blob/d149fde00a0c53d2bab889cf447aef7f62a1cc8c/pkg/sign/signer.go#L111 - "Oh, you have `timestampAuthorities`? You must be expecting a signed timestamp in the bundle".

It would be pretty easy to extend this logic in sigstore-go signing to check the trusted root for a `ctlogs` section, and if it exists to do SCT verification on signing. Is it worth it? I don't know - the verification checks on signing are more of a sanity check, and I'm not sure what the future of Fulcio SCT is. Of course, when a user explicitly asks us to verify (in the `verify` sub-commands) that's a different story.

cmd/cosign/cli/sign/sign.go
```diff
@@ -234,61 +256,89 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
...
+
+		bundleComponents, err := signcommon.ExtractComponentsFromProtoBundle(&pb)
 		if err != nil {
-			return err
+			return fmt.Errorf("extracting components from bundle: %w", err)
 		}
+
+		certPem, chainPem := signcommon.EncodeCertificatesToPEM(bundleComponents.Certificates)
+		if i == 0 {
+			firstCertPem = certPem
```
Hayden-IO: Can you clarify the reason we're pulling out the first certificate from the list of payloads?

Hayden-IO (dismissed):
Very happy with these changes! It's easy to see how we'll move forward with deprecating the older code paths now. I just left one small question. We'll get this PR in asap so we can start working on the followups, we can hold off on merging any other changes in the meantime to avoid merge conflicts.

cmd/cosign/cli/sign/sign.go
aaronlew02: This is to mimic the [current functionality](https://github.com/sigstore/cosign/blob/2b396bdfc8668e37debe77449fc1dc707a74696e/cmd/cosign/cli/sign/sign.go#L314), which is to output a leaf certificate from `sv.Bytes()`, as the `SignerVerifier` holds a certificate chain that is used for signing an image with multiple identities.

However, this only preserves functionality when signing an image with a single identity. Since the `SignerVerifier` is no longer going to be used, signing an image with multiple identities would require sigstore-go to fetch a new certificate for each identity.

To mitigate this (i.e., to use a single certificate to sign a single image with multiple identities), we could implement a `CertificateProvider` to cache the Fulcio certificate fetched for the first identity and pass that `CertificateProvider` to sigstore-go for any further signing events for the same image.

I've implemented this approach in the latest commit. WDYT?

pkg/cosign/bundle/sign.go
```diff
@@ -167,3 +168,42 @@ func (c *localCertProvider) GetCertificate(_ context.Context, _ sign.Keypair, _
...
+	}
+	fulcioOpts := &sign.FulcioOptions{
+		BaseURL: fulcioSvc.URL,
+		Timeout: 30 * time.Second,
+		Retries: 1,
+	}
+	return sign.NewFulcio(fulcioOpts), nil
+}
+
+func NewCachingFulcioProvider(signingConfig *root.SigningConfig) (sign.CertificateProvider, error) {
```
Hayden-IO, line 206: Let's add a comment for anything publicly exported. This function will not work once the certificate expires, so we need to call that out, that this shouldn't be used in long-running processes.

pkg/cosign/bundle/sign.go
```diff
@@ -167,3 +168,42 @@ func (c *localCertProvider) GetCertificate(_ context.Context, _ sign.Keypair, _
...
+func (c *cachingCertProvider) GetCertificate(ctx context.Context, keypair sign.Keypair, opts *sign.CertificateProviderOptions) ([]byte, error) {
+	c.once.Do(func() {
+		c.fetch = sync.OnceValues(func() ([]byte, error) {
+			return c.provider.GetCertificate(ctx, keypair, opts)
+		})
+	})
+	return c.fetch()
+}
+
+func NewFulcioProvider(signingConfig *root.SigningConfig) (sign.CertificateProvider, error) {
```
Hayden-IO: This isn't called outside of the bundle package, can it be not exported?

Hayden-IO (commented):
Were there any other changes made in commit 46219181277434c074564fc07a5359a7d13107ed besides changing "first" to "leaf"? Or do you have the commit hash from before the force push?

For any more commits until this is merged, can you avoid force pushing? Without the commit hash from before the force push, I don't know what changed and so this makes reviewing a large PR difficult. We'll squash commits on merge.

aaronlew02:
> Were there any other changes made in commit https://github.com/sigstore/cosign/commit/46219181277434c074564fc07a5359a7d13107ed besides changing "first" to "leaf"?

As of this comment, the change from "first" to "leaf" was the only modification to the first commit. My intention was to separate that minute change from the following commit. 

> For any more commits until this is merged, can you avoid force pushing?

Yes. I apologize for any inconvenience.

Hayden-IO (approved):
Will leave open til end of day in case there's any more comments, otherwise we'll ship it!

PR #4620, opened 2026-01-08
Support mTLS for TSA clients when using a signing config

aaronlew02:
Partially addresses #4570

#### Summary

This change adds support for using mTLS to connect to a TSA client via sigstore-go when signing with a signing config.

### DIFF
--- a/pkg/cosign/bundle/sign.go
+++ b/pkg/cosign/bundle/sign.go
@@ -20,6 +20,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"log"
+	"net/http"
 	"time"
 
 	"github.com/sigstore/cosign/v3/internal/ui"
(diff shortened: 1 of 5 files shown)


pkg/cosign/bundle/sign.go
```diff
@@ -29,7 +30,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
-func SignData(ctx context.Context, content sign.Content, keypair sign.Keypair, idToken string, cert []byte, signingConfig *root.SigningConfig, trustedMaterial root.TrustedMaterial) ([]byte, error) {
+func SignData(ctx context.Context, content sign.Content, keypair sign.Keypair, idToken string, cert []byte, signingConfig *root.SigningConfig, trustedMaterial root.TrustedMaterial, tsaClientTransport http.RoundTripper) ([]byte, error) {
```
Hayden-IO: @steiza Since you added this method, any opposition to changing its signature to accomadate this?

I wonder if we should instead create a SignOptions structure (I know we have one currently, but it has so many options that I think it's best to start fresh for bridging Cosign and sigstore-go) in case we need to further customize this function. WDYT @aaronlew02?

aaronlew02: I'm open to the idea of a new `SignOptions` to ease maintainability, though in terms of bridging with sigstore-go, my draft PR #4618 hasn't so far shown a need to expand the signature of `SignData`.

steiza: Yeah, `tsaClientTransport` definitely sticks out with respect to the other arguments as not quite like the others.

I also think we should have some sort of struct here to contain optional additional sigstore-go signing options. I have no idea what to call it. I supposed we could have `cmd/cosign/cli/options.SignOptions` and `pkg/cosign/bundle.SignOptions`, or maybe `SigningOptions`, or maybe something else?

Hayden-IO: `SignOptions` to mirror what's in Cosign already seems good.

aaronlew02: I've added a new `SignOptions` struct to be used as the last input to `SignData` with a single field `TSAClientTransport` .

PR #4622, opened 2026-01-09
Add hashedrekord support to attest-blob and attest

steiza:
#### Summary

For #3599, as an alternative to #4490

There are cases where people want to have large attestations (like SBOMs) which introduce a couple of issues:
- Rekor won't accept large dsse / intoto entries
- the resulting bundle is quite large (as it contains the large attestation)

`attest-blob` already has `--rekor-entry-type`, although previously it only supported `dsse` and `intoto`. This change adds `hashedrekord` support, and the resulting bundle has a `messageSignature` instead of a large `dsseEnvelope`.

Of course, you need a way to verify the resulting bundle. With this change you can do so with `verify-blob` (instead of `verify-blob-attestation`, where the blob you verify is the DSSE PAE.

We probably wouldn't land this as-is, due to two issues:

- currently the PAE is printed out for convenience, although for large attestations this is not very desirable. We could either have the command save the PAE to disk for you, or we could print out the hash and add hash support to `verify-blob` (`verify-blob-attestation` already has hash support).

- this only works when you are using signing config, although #4570 / #4618 are working on unifying codepaths so that in the future this would also work without signing config.

You can test this with something like:
```
$ go run cmd/cosign/main.go attest-blob --key=cosign.key --bundle=attest.sigstore.json --rekor-entry-type=hashedrekord --type=something --predicate=../sigstore-go/examples/sigstore-go-signing/intoto.txt ../sigstore-go/examples/sigstore-go-signing/hello_world.txt
```
```
$ echo -n "..." > pae.txt
```
```
$ go run cmd/cosign/main.go verify-blob --key=cosign.pub --bundle=attest.sigstore.json pae.txt
```

#### Release Note

- Added `--rekor-entry-type=hashedrekord` support to `attest` and `attest-blob` when signing config is used

#### Documentation

N/A

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -893,18 +893,29 @@ func TestSignVerifyWithTUFMirror(t *testing.T) {
 
 func prepareSigningConfig(t *testing.T, fulcioURL, rekorURL, oidcURL, tsaURL string) string { //nolint: unparam
 	startTime := "2024-01-01T00:00:00Z"
-	fulcioSpec := fmt.Sprintf("url=%s,api-version=1,operator=fulcio-op,start-time=%s", fulcioURL, startTime)
+	fulcioSpecs := []string{}
+	if fulcioURL != "unused" {
+		fulcioSpecs = []string{fmt.Sprintf("url=%s,api-version=1,operator=fulcio-op,start-time=%s", fulcioURL, startTime)}
+	}
+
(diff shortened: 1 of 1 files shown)


steiza:
After exploring this approach, I don't think this is a good idea. We should instead encourage people not to create large bundles that are 100s of MBs.

PR #4472, opened 2025-10-15
Fix segfault when no attestations are found

steiza:
#### Summary

Fixes #4468.

Note that this does not implement `cosign save` with the new attestations (as described in #4470), but at least it won't segfault.

#### Release Note

* Fixed segfault with subcommands `verify` and `verify-attestation` when called with `--local-image=true` and no attestations were found

#### Documentation

N/A

PR #4626, opened 2026-01-12
Allow --local-image with --new-bundle-format for v2 and v3 signatures

### LINKED ISSUE #4621: Incompatibility between cosign 3.0.3 and 2.3.4 when verifying images locally
**Description**

When validating the signed images locally (by extracting the tar) we see some incompatibilities with old versions of cosign. 

For instance when we are using cosign (2.4.3) and using the below command to verify the image, it succeeds.

**_bash$ /home/ubuntu/signing/cosign/243/cosign verify --ca-roots /home/ubuntu/signing/cosign/certs/verification_certs/staging/ca.crt --certificate-identity-regexp '.*' --certificate-oidc-issuer-regexp '.*' --private-infrastructure --insecure-ignore-sct --local-image /home/ubuntu/signing/cosign/vp/xyz_dir/xyz/_**

Verification for /home/ubuntu/signing/cosign/vp/xyz_dir/xyz/ --
  - The code-signing certificate was verified using trusted certificate authority certificates

[{"critical":{"identity":{"docker-reference":"172.17.0.1:5005/xyz"},"image":{"docker-manifest-digest":"sha256:b84d153fd42c8b6d27c82b59ed77ae591963b77533f1596e8a06c1da07c30ebe"},"type":"cosign container image signature"},"optional":{"Subject":"https://www.abc.com/en/about-us/security/pki","product":"Dummy"}}]

When we use the latest version of cosign (3.0.3), the same command fails.

**_bash$ /home/ubuntu/303/cosign verify --ca-roots /home/ubuntu/signing/cosign/certs/verification_certs/staging/ca.crt --certificate-identity-regexp '.*' --certificate-oidc-issuer-regexp '.*' --private-infrastructure --insecure-ignore-sct --local-image /home/ubuntu/signing/cosign/vp/xyz_dir/xyz/_**

Error: unsupported: CA roots/intermediates must be provided using --trusted-root when using --new-bundle-format
error during command execution: unsupported: CA roots/intermediates must be provided using --trusted-root when using --new-bundle-format 
**Version**

This breaks backward compatibility. Is this a bug? If not, could you suggest a working command for achieving the same with cosign (2.0.3)? I tried the options like trusted_root , setting the new-bundle-format to false etc. But none seems to work.

tuminoid:
## Summary

Remove the mutual exclusivity constraint between --local-image and --new-bundle-format flags in verify and verify-attestation commands. This enables verification of v2-signed images saved as local OCI layouts when using cosign v3. We also add support for cosign v3 signed images on disk by parsing OCI manifests for the bundles and performing full local verification.

## Problem

When users sign images with v2 format (attached signatures) and save them locally with cosign save, they cannot verify them in cosign v3 because:

- The --local-image and --new-bundle-format flags were mutually exclusive
- Cosign v3 defaults to --new-bundle-format=true
- V3 bundles stored as OCI referrer manifests could not be verified from local layouts

This affects workflows where images are signed and verified offline.

## Solution

Add HasLocalBundles() and HasLocalAttestationBundles() functions that auto-detect the signature format by checking for application/vnd.dev.sigstore.bundle media type layers in the local OCI layout. If no bundles are found (v2 attached signatures), verification falls back to v2 mode automatically.

For v3 bundles, add GetLocalBundles() to extract bundles from local OCI layouts and verifyLocalImageAttestationsSigstoreBundle() to perform full verification without requiring a registry round-trip.

This mirrors the existing auto-detection and verification patterns used for registry-based verification via GetBundles() and verifyImageAttestationsSigstoreBundle().

## Changes

 - Remove mutual exclusivity between --local-image and --new-bundle-format flags
 - Add HasLocalBundles() / HasLocalAttestationBundles() functions for format auto-detection and integrate auto-detection in verify.go and verify_attestation.go
 - Add GetLocalBundles() to extract v3 bundles from local OCI layouts
 - Add verifyLocalImageAttestationsSigstoreBundle() for local v3 bundle verification
 - Refactor hasLocalSigstoreBundles() to getLocalBundleDescriptors() for reuse
 - Update VerifyLocalImageAttestations() to use new v3 verification path

## Testing

 - Unit tests for HasLocalBundles covering v2 signatures, v3 bundles, no signatures, and error cases
 - Unit tests for GetLocalBundles covering valid layouts, missing blobs, and error cases
 - E2e tests for auto-detection and full verification of both v2 attached signatures and v3 bundles

Fixes #4621

### DIFF
--- a/pkg/cosign/verify_test.go
+++ b/pkg/cosign/verify_test.go
@@ -30,10 +30,12 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
@@ -46,7 +48,7 @@ import (
 	ggcrlayout "github.com/google/go-containerregistry/pkg/v1/layout"
 	gcrMutate "github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/random"
-	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/stream"
 	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/cosign/v3/internal/pkg/cosign/payload"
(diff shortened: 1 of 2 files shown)


haydentherapper (approved):
Thanks for the fix, this looks good!

tuminoid:
Reworked the implementation and tests a bit to fix a usage case in e2e that failed. PTAL.

Hayden-IO (dismissed):
Thanks for the comprehensive tests!

tuminoid:
I had missed a staticcheck issue while adding the tests. Fixed that, and rebased for easier merging.

steiza (commented):
Thanks for this pull request! This is on my list of things to review this week.

I'm trying to figure out the right `cosign sign` / `cosign save` / `cosign verify` commands to test this out, if you have that information available. I want to make sure `hasLocalSigstoreBundles()` is covering the various ways things might be stored.

tuminoid:
Thanks @steiza !

> I'm trying to figure out the right `cosign sign` / `cosign save` / `cosign verify` commands to test this out, if you have that information available. I want to make sure `hasLocalSigstoreBundles()` is covering the various ways things might be stored.

I also missed the difference in handling of attestations first time around, so its plausible there are more gaps. Added some tests and e2e is happy, but getting some more eyes to look for gaps is great!

For the commands part, I hope you don't mind me dropping my [signing playground](https://github.com/tuminoid/container-signing-poc) for BYO-PKI signing tests as reference.

- Signed and attached [here with cosign v2](https://github.com/tuminoid/container-signing-poc/blob/573a8e6b000019946fd12d63a5fff7d92f5d4d29/cosign-v2/Makefile#L53)
- Verifying local-image with [cosign v3 here](https://github.com/tuminoid/container-signing-poc/blob/573a8e6b000019946fd12d63a5fff7d92f5d4d29/cosign-v3/Makefile#L107)

My e2e for this is running `make local-image-test` from top-level, which:
- builds cosign from $COSIGN_SOURCE_DIR (switch between HEAD and my branch to see patch in effect)
- run cosign v2 signing + attach
- run cosign v3 verify with local image

steiza:
Okay, I had some more time to dig into this today - I don't think `hasLocalSigstoreBundles()` covers all the cases quite yet.

Here's what I'm using to test:

Sign an image with the new bundle format, a key, and no Rekor entry:
```
$ go run cmd/cosign/main.go sign --key cosign2.key --use-signing-config=false --tlog-upload=false localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a@sha256:0b64ad459d4a9c1281434eca87d7026e1375a2a886a66e9787db3abb847c96f9
```

Download signed image to disk:
```
$ go run cmd/cosign/main.go save --dir save localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a@sha256:0b64ad459d4a9c1281434eca87d7026e1375a2a886a66e9787db3abb847c96f9
```

Attempt to verify:
```
$ go run cmd/cosign/main.go verify --local-image=true --key cosign2.pub save/
Error: no signatures associated with the image saved in save/
error during command execution: no signatures associated with the image saved in save/
exit status 1
```

In Cosign, new bundles are saved to the OCI registry using the [OCI 1.1 referrers spec](https://github.com/sigstore/cosign/blob/8a78942c011dc6ea48f85483946f30addbadac3b/specs/BUNDLE_SPEC.md?plain=1#L93) with a fallback of using the referrers tag scheme.

It looks like our local registries support OCI 1.1 referrers, so the Sigstore bundle has its own manifest with a `subject` section that links it to the container image. When you query a registry that understands the referrer spec, it will return this manifest to you. But when you're looking at a bunch of files on disk (like with `--local-image=true`), if you're using the new protobuf bundle format, you have to manually look for files on disk that have a `subject` of whatever it is you're trying to verify (and probably `artifactType` set to `application/vnd.dev.sigstore.bundle.v0.3+json`).

This has come up in other recent pull request, like #4538 and #4539, so those could be helpful references (and / or we might be able to make a function that we could use to support these various use-cases - a shared function isn't a requirement though).

Sorry about that! Hope this explanation makes sense, and let me know if you have any questions.

tuminoid:
Thanks @steiza for taking time, good findings. I'll take another stab at it and try to find a place to test it in e2e as well.

tuminoid:
OK, I've added 2 more commits:
- first one implements what @steiza suggested for OCI referrers for v3 signed and saved images
- second one tries to cover all the new functions with unit tests and small change in e2e

They pass unit testing, cosign e2e and also in my POC environment (updates there not pushed yet).

While testing this change, I also managed to get stack trace out of cosign v3.0.2 doing v3 bundle verification, I need to see if I can reproduce that. I'll create separate issue on that, if I can reproduce it against HEAD.

tuminoid:
> While testing this change, I also managed to get stack trace out of cosign v3.0.2 doing v3 bundle verification, I need to see if I can reproduce that. I'll create separate issue on that, if I can reproduce it against HEAD.

Ah, it was already fixed in https://github.com/sigstore/cosign/pull/4472 so no worries there.

tuminoid:
@steiza @Hayden-IO could you trigger tests to see if CI agrees  with me ?

Issue #4634, opened 2026-01-13
Cosign sign with new_bundle_format = true creates an Intotov1 statement while not containing a predicate as the standard requires


dortam888:
**Description**

When using `cosign sign` command with `new-bundle-format=true`, it creates an Intotov1 statement (Rather than `new-bundle-format=false` simplesigning) without a predicate at all as required by in-toto (Reference: https://github.com/in-toto/attestation/blob/main/go/v1/statement.go#L20)
 
**Version**

GitVersion:    v3.0.4
GitCommit:     6832fba4928c1ad69400235bbc41212de5006176
GitTreeState:  "clean"
BuildDate:     2026-01-09T21:17:16Z
GoVersion:     go1.25.5
Compiler:      gc
Platform:      darwin/arm64

PR #4635, opened 2026-01-13
Add empty predicate to cosign sign when payload type is application/vnd.in-toto+json

dortam888:
#### Summary

Resolves issue https://github.com/sigstore/cosign/issues/4634.
Set a default empty predicate when using cosign sign with new_bundle_format to align with in-toto statement requirement to have a predicate

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -170,6 +170,7 @@ func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpt
 		Type:          intotov1.StatementTypeUri,
 		Subject:       []*intotov1.ResourceDescriptor{&subject},
 		PredicateType: types.CosignSignPredicateType,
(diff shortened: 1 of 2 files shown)


dortam888:
@Hayden-IO Thanks for the approval. 
It surfaced during one of JFrog internal tools verification process using the in-toto library and was triggered when customers upgraded to cosign V3 that enables the new_bundle_format flag by default.

steiza (approved):
Thanks for this change!

I see that the in-toto code you linked returns an error if there's no predicate, but confusingly https://github.com/in-toto/attestation/blob/main/spec/v1/predicate.md#fields says:

> A predicate has a required predicateType ([TypeURI](https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#TypeURI)) identifying what the predicate means, plus an optional predicate [JSON](https://www.json.org/) object containing additional, type-dependent parameters.

🤷

Issue #4617, opened 2026-01-05
Disallow both a key and identity during verification


Hayden-IO:
**Description**

We have inconsistency with usage of both `--key` and `--certificate-identity` when verifying signatures signed either by managed keys or using code signing certificates. When signed with a key, the identity is always ignored. When signed with a code signing cert, verification errs out with a key.

There are roughly three options for the UX:

1. When a key is provided, it takes precedent over Fulcio verification (what Cosign does now). Note that Cosign still uses other roots of trust, e.g. the log public key.
2. When a key is provided, it is merged into the trusted root, so that either a certificate or key is valid.
3. Disallow both a key and identity from being provided, to avoid the need to pick preferences.

I would recommend (3). When a user provides an identity, we should treat that as the user specifying that an identity is expected. By silently ignoring it, there may be instances where a user is tricked into verifying a signature signed by a managed key rather than with a code signing cert unexpected.

Copying from https://sigstore.slack.com/archives/C03SZ6SHU90/p1766169728333369:

```
OIDC signed
% cosign verify-blob --bundle ~/cosign-test/test.sigstore.json --certificate-identity piceri@github.com --certificate-oidc-issuer https://github.com/login/oauth a.txt
Verified OK

% cosign verify-blob --key ~/cosign-test/cosign.pub --bundle ~/cosign-test/test.sigstore.json --certificate-identity piceri@github.com --certificate-oidc-issuer https://github.com/login/oauth a.txt 
Error: expected key signature, not certificate
error during command execution: expected key signature, not certificate

Key signed
% cosign verify-blob --key ~/cosign-test/cosign.pub --bundle ~/cosign-test/bundle.sigstore.json a.txt
Verified OK

% cosign verify-blob --key ~/cosign-test/cosign.pub --bundle ~/cosign-test/bundle.sigstore.json --certificate-identity piceri@github.com --certificate-oidc-issuer https://github.com/login/oauth a.txt
Verified OK
```

cc @piceri

cmurphy:
Could the content of the bundle be used to automatically infer whether a key or certificate is supposed to be used?

PR #4636, opened 2026-01-13
disallow key and cert identity being used together during verification

piceri:
#### Summary

Closes https://github.com/sigstore/cosign/issues/4617

This pull request forces a failure when both the --key and --certificate-identity flags are used when trying to verify. This will avoid confusion as to which method was used to verify and prompt users to only provide one or the other. 

#### Release Note

Changed verify and verify-blob to not allow --key and --certificate-identity to be used together

### DIFF
--- a/test/e2e_tsa_test.go
+++ b/test/e2e_tsa_test.go
@@ -221,11 +221,7 @@ func TestSignBlobTSAMTLSWithSigningConfig(t *testing.T) {
 	}
 
 	verifyCmd := cliverify.VerifyBlobCmd{
-		KeyOpts: verifyKO,
-		CertVerifyOptions: options.CertVerifyOptions{
-			CertIdentityRegexp:   ".*",
-			CertOidcIssuerRegexp: ".*",
-		},
+		KeyOpts:    verifyKO,
 		IgnoreTlog: true,
 	}
 	must(verifyCmd.Exec(context.Background(), blobPath), t)


cmd/cosign/cli/verify/verify.go
```diff
@@ -98,6 +98,10 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		c.HashAlgorithm = crypto.SHA256
 	}
 
+	if c.KeyRef != "" && c.CertIdentity != "" {
+		return fmt.Errorf("provide either --key or --certificate-identity, not both")
```
Hayden-IO: Can you add a little bit more about what to do when you get this error? How should you know whether to provide a key or identity?

piceri:
> Thanks! Can we also update verify-attestation and verify-blob-attestation? Can we have a shared method for verifying this called by the four verify-* commands?

As I went to add the other verify commands, I found the NOf function being used to do similar mutually exclusive checks and copied that pattern.

I also was looking deeper at the options and noticed CertIdentityRegex and I believe that should be added to the check as well as it appears to be an alternative to CertIdentity. This did cause an e2e test to fail. I did my best to dig into that test and I believe it ok to remove the CertVerifyOptions as the validation is done by key, but please double check me here.

cmd/cosign/cli/options/errors.go
```diff
@@ -23,10 +23,21 @@ type KeyParseError struct{}
...
 }
 
 func (e *PubKeyParseError) Error() string {
 	return "exactly one of: key reference (--key), certificate (--cert) or hardware token (--sk) must be provided"
 }
+
+func (e *KeyAndIdentityParseError) Error() string {
+	return "exactly one of: key reference (--key) or certificate identity " +
+		"(--certificate-identity or --certificate-identity-regexp), must be provided. " +
+		"Bundles signed with identity will have a 'certificate' field under 'verificationMaterial.' " +
```
Hayden-IO: This will also cover the older Cosign bundle format - https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#properties - so we should mention a certificate could be under this as well.

PR #4642, opened 2026-01-16
docs(cosign): clarify RFC3161 revocation semantics

1seal:
#### Summary

Add a short note to `VerifyRFC3161Timestamp` clarifying that RFC3161 timestamp verification does not perform CRL/OCSP revocation checks, and that Sigstore's default workflow relies on TUF-distributed trust material + validity metadata for revocation.

#### Release Note

NONE

#### Documentation

N/A (doc comment only)

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1324,8 +1324,7 @@ func (s *sigContent) MessageSignatureContent() verify.MessageSignatureContent {
 // no root is provided with a timestamp.
 //
 // Note: This function does not perform CRL/OCSP certificate revocation checks.
-// Sigstore's default workflow relies on TUF-distributed trust material + validity metadata for revocation.
-// Private PKI deployments should enforce additional revocation policy as needed.
+// Callers are responsible for validating the TSA certificate and trusted material provided via CheckOpts according to their policy.
 func VerifyRFC3161Timestamp(sig oci.Signature, co *CheckOpts) (*timestamp.Timestamp, error) {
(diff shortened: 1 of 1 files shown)


pkg/cosign/verify.go
```diff
@@ -1322,6 +1322,10 @@ func (s *sigContent) MessageSignatureContent() verify.MessageSignatureContent {
 // returns the timestamp value.
 // It returns (nil, nil) if there is no timestamp, or (nil, err) if there is an invalid timestamp or if
 // no root is provided with a timestamp.
+//
+// Note: This function does not perform CRL/OCSP certificate revocation checks.
+// Sigstore's default workflow relies on TUF-distributed trust material + validity metadata for revocation.
```
Hayden-IO: Can this just be `Note: This function does not perform CRL/OCSP certificate revocation checks.`?

Hayden-IO (changes requested):

pkg/cosign/verify.go
kommendorkapten: We may also add that it's the client's responsibility to first verify the validity of the certificate _before_ adding it into the trusted material provided to the `CheckOpts` structure? I think that would more helpful to the client.

1seal:
updated per feedback: kept the first sentence as-is, removed the tuf-specific line, and added a short note that callers are responsible for validating the tsa certificate and trusted material in CheckOpts according to their policy (including revocation if desired).

thanks for the review.

PR #4643, opened 2026-01-16
Update conformance to latest

cmurphy:

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -44,6 +44,7 @@ jobs:
 
       - run: make cosign conformance
 
-      - uses: sigstore/sigstore-conformance@b7856cfca56fe3f957d4cefdc2c359cc36a84e14 # v0.0.24
+      - uses: sigstore/sigstore-conformance@eae6eb1f59e25c6d3d602c5dad3dc55767c2f1cb # v0.0.25
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance.yml
```diff
@@ -44,6 +44,7 @@ jobs:
 
       - run: make cosign conformance
 
-      - uses: sigstore/sigstore-conformance@b7856cfca56fe3f957d4cefdc2c359cc36a84e14 # v0.0.24
+      - uses: sigstore/sigstore-conformance@eae6eb1f59e25c6d3d602c5dad3dc55767c2f1cb # v0.0.25
         with:
           entrypoint: ${{ github.workspace }}/conformance
+          xfail: "test_verify*managed-key-happy-path] test_verify*managed-key-and-trusted-root] test_verify\[PATH-message-digest-mismatch_fail]"
```
Hayden-IO, line 50: does `]` also need to be escaped?

cmurphy: No, it never has. It's just the opening [ that breaks.

Hayden-IO: JW, why does this differ from sigstore-go?

cmurphy: I changed sigstore-go to look the same. It doesn't really matter, they both work.

PR #4648, opened 2026-01-20
Gracefully fail if bundle payload body is not a string

Hayden-IO:
Previously, `VerifyBundle` assumed `bundle.Payload.Body` was always a string and performed a direct type assertion. If the body was a different type (e.g., due to a malformed bundle), this would cause a panic.

This change adds a check to safely assert the type of `bundle.Payload.Body`.

Thanks to @morwn for reporting this!


1seal:
fyi i hit the same issue independently; makes sense. if you’d like, i can help add an additional regression case for non-string payload bodies (e.g. null/object/number) to ensure this never panics again.

PR #4649, opened 2026-01-20
fix: avoid panic on malformed tlog entries

1seal:
- avoid panic on malformed VerifyTLogEntryOffline inputs by returning errors
- add regression test to ensure malformed inputs do not panic


Hayden-IO (approved):
Thanks!

PR #4651, opened 2026-01-20
fix: avoid panic on malformed attestation payload

1seal:
- avoid panic on malformed AttestationToPayloadJSON inputs by returning errors
- add regression test to ensure non-string payloads do not panic


1seal:
fyi this switches the payload field assertion in AttestationToPayloadJSON to a checked assertion (returning an error on unexpected types) and adds a regression case for {"payload":null} so it won’t panic.

PR #4652, opened 2026-01-20
fix: avoid panic on malformed tlog entry body

1seal:
- avoid panic on malformed ComputeLeafHash inputs by returning errors
- add regression test to ensure non-string log entry bodies do not panic

### DIFF
--- a/pkg/cosign/tlog_test.go
+++ b/pkg/cosign/tlog_test.go
@@ -240,30 +240,60 @@ func TestVerifyTLogEntryOfflineFailsWithInvalidPublicKey(t *testing.T) {
 	}
 }
 
-func TestComputeLeafHashRejectsNonStringBodyWithoutPanic(t *testing.T) {
+func TestComputeLeafHash(t *testing.T) {
+	payload := []byte("hello")
+	expected := sha256.Sum256(append([]byte{0}, payload...))
+
 	tests := []struct {
-		name string
-		body interface{}
+		name          string
+		body          interface{}
+		wantHash      []byte
+		wantErrSubstr string
 	}{
-		{name: "nil body", body: nil},
-		{name: "bytes body", body: []byte("not a string")},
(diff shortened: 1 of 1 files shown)


Hayden-IO:
`ComputeLeafHash` parses an entry that comes directly from Rekor, so while being defensive here is fine, I don't see this is necessary.

1seal:
i agree this shouldn’t happen with a well-formed rekor response. i sent it as defense-in-depth to return an error instead of panicking; happy to close if you don’t want this hardening.

pkg/cosign/tlog_test.go
```diff
@@ -239,3 +239,32 @@ func TestVerifyTLogEntryOfflineFailsWithInvalidPublicKey(t *testing.T) {
 		t.Fatalf("Did not get expected error message, wanted 'is not type ecdsa.PublicKey' got: %v", err)
 	}
 }
+
+func TestComputeLeafHashRejectsNonStringBodyWithoutPanic(t *testing.T) {
```
Hayden-IO: Can this be renamed to just TestComputeLeafHash, and handle a successful test case as well?

pkg/cosign/tlog_test.go
```diff
@@ -239,3 +239,32 @@ func TestVerifyTLogEntryOfflineFailsWithInvalidPublicKey(t *testing.T) {
...
+		body interface{}
+	}{
+		{name: "nil body", body: nil},
+		{name: "bytes body", body: []byte("not a string")},
+		{name: "object body", body: map[string]interface{}{"k": "v"}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			defer func() {
```
Hayden-IO: Recovery shouldn't needed, no test should panic

Hayden-IO (commented):
Looks good to me, just needs a rebase

Hayden-IO (approved):

Issue #4655, opened 2026-01-20
oci 1.1 referrers selection is order-dependent (last element)


1seal:
hi, i noticed the oci 1.1 referrers discovery paths select the last referrer returned when multiple results exist, which makes behavior dependent on registry response order.

examples at commit 03be625f:
- `pkg/cosign/verify.go` (oci 1.1 signature discovery): warns on multiple results, then selects `results[numResults-1]`
- `pkg/oci/remote/remote.go` (`attachmentExperimentalOCI`): warns on multiple results, then selects `results[numResults-1]`

since referrers ordering is not guaranteed, this can lead to non-deterministic behavior across registries and over time when multiple referrers exist for the same `artifactType`.

question:
what behavior should cosign prefer when `len(index.Manifests) > 1` for a requested `artifactType`?
- fail closed as ambiguous?
- fetch/verify all matching referrers?
- deterministic selection using a specific key (e.g. validated annotations), and if so which key?

i can send a small PR that at least makes selection deterministic (by sorting descriptors before selecting), but wanted to confirm the intended semantics first.

PR #4656, opened 2026-01-20
chore: make oci11 referrer selection deterministic

1seal:
- make oci 1.1 referrer selection deterministic by sorting descriptors by digest before selecting the last result
- add unit tests to ensure selection is order-independent and does not mutate inputs
- note: this does not define the “correct” semantics for multiple matching referrers; please advise whether cosign should fail closed, verify all, or prefer a specific key (e.g. validated annotations)

refs #4655


cmurphy:
I see these TODOs about these referrers https://github.com/sigstore/cosign/blob/f4a0894d693421c8ba80ac65962304dce42eef8f/pkg/cosign/verify.go#L1644-L1647 - I wonder if that's a potential starting place for this? 

I'm curious if you ran into this in a real-world scenario? If so, what created those other referrers?

1seal:
up

PR #4657, opened 2026-01-21
fix(oci): bound bundle layer read

1seal:
- bound bundle layer reads to prevent unbounded allocation on untrusted inputs
- fail closed with an explicit error when the limit is exceeded
- add regression coverage for the size bound

### DIFF
--- a/pkg/oci/remote/signatures.go
+++ b/pkg/oci/remote/signatures.go
@@ -17,6 +17,7 @@ package remote
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
@@ -32,6 +33,7 @@ import (
 )
 
 const maxLayers = 1000
+const maxBundleLayerBytes = 10 << 20
 
 // Signatures fetches the signatures image represented by the named reference.
 // If the tag is not found, this returns an empty oci.Signatures.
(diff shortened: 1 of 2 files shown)


pkg/oci/remote/signatures.go
```diff
@@ -32,6 +33,7 @@ import (
 )
 
 const maxLayers = 1000
+const maxBundleLayerBytes = 10 << 20
```
Hayden-IO, line 36: How did you decide on this value? It depends what's being put into the bundles. Just a signature, 10 MB should be sufficient, but for an SBOM or attestation, this would be too low.

cc @steiza for your thoughts

1seal: good point, i picked 10 << 20 as a conservative default to prevent unbounded allocation when reading an untrusted layer, assuming the common case is small (signatures / bundle metadata). i agree it may be too low for sbom/attest-style payloads.

i’m happy to adjust this based on your guidance: either raise the default substantially, or make it configurable (e.g., option/env) so downstreams can set an appropriate limit for their bundle contents while still keeping a hard cap to avoid OOM.

steiza: We know that folks are [using cosign with 100MB+ bundles](https://github.com/sigstore/cosign/issues/3599), so 10 << 20 is definitely too small.

This just isn't a priority for Cosign right now, compared to other contributions looking for maintainer review that add protobuf bundle support or support OCI use-cases.

Architecturally, we want to move away from large bundles that include attestations that are many MBs, but where we are today is that's how some folks use Cosign.

Hayden-IO: Thanks @steiza, so it sounds like this isn't something that's considered a threat that needs mitigation. I'll close this out then.

PR #4663, opened 2026-01-23
Verify validity of chain rather than just certificate

Hayden-IO:
There was an assumption that a certificate would be issued during the validity period of all certificates in its chain. If a certificate was issued such that its NotAfter timestamp was past the NotAfter of another certificate in the chain, that would have not been detected. This is because during verification, we use the leaf certificate's NotBefore timestamp to verify the chain's validity, but check only the leaf certificate's validity against any provided timestamps (either from the Rekor timestamp, a signed timestamp, or the current time).

This scenario is not expected with a typical chain since a CA shouldn't issue a certificate that outlives an issuer.

This change enforces that the entire chain is valid at any of those provided timestamps.


Hayden-IO:
Yep, very much agreed. I verified this doesn't affect sigstore-go because of a more reasonable ordering - gather your timestamps, then verify the chain against those timestamps.

PR #4666, opened 2026-01-24
Automatically require signed timestamp with Rekor v2 entries

### LINKED ISSUE #4500: Improve the UX when verifying a Rekor v2 entry with a signed timestamp
**Description**

Currently, use-signed-timestamp must be specified when verifying a Rekor v2 entry. Since it’s always required, it’s unnecessary to be explicitly specified.

Hayden-IO:
Currently, when verifying a bundle with a short-lived certificate and a Rekor v2 entry, users must specify --use-signed-timestamps and provide a signed timestamp, because Rekor v2 entries do not contain integrated timestamps unlike Rekor v1. This isn't a great UX.

This change opportunistically sets --use-signed-timestamps when we detect that a user is verifying a short-lived certificate with a Rekor v2 entry. We won't automatically set this with a mixed Rekor v1 and Rekor v2 bundle (which the tooling will not produce, this would be for any custom clients) or when verifying with a managed key.

Fixes #4500

### DIFF
--- a/pkg/cosign/verify_bundle.go
+++ b/pkg/cosign/verify_bundle.go
@@ -21,8 +21,11 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/verify"
 )
 
-// VerifyNewBundle verifies a SigstoreBundle with the given parameters
+// VerifyNewBundle verifies a Sigstore bundle with the given parameters
 func VerifyNewBundle(_ context.Context, co *CheckOpts, artifactPolicyOption verify.ArtifactPolicyOption, bundle verify.SignedEntity) (*verify.VerificationResult, error) {
+	if err := rekorV2Bundle(bundle, co); err != nil {
+		return nil, err
+	}
 	trustedMaterial, verifierOptions, policyOptions, err := co.verificationOptions()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Hayden-IO:
Without the change, we see an error because the client assumes that an integrated timestamp from Rekor v1 is needed:

```
cosign initialize --staging
touch /tmp/blob
cosign sign-blob -y /tmp/blob --bundle /tmp/bundle.json 
....
Using payload from: /tmp/blob
Wrote bundle to file /tmp/bundle.json

cosign verify-blob --bundle /tmp/bundle.json --certificate-identity-regexp='.*' --certificate-oidc-issuer='https://accounts.google.com' /tmp/blob

Error: failed to verify timestamps: threshold not met for verified log entry integrated timestamps: 0 < 1
error during command execution: failed to verify timestamps: threshold not met for verified log entry integrated timestamps: 0 < 1
```

With the change, where the bundle contains a signed timestamp:

```
cosign verify-blob --bundle /tmp/bundle.json --certificate-identity-regexp='.*' --certificate-oidc-issuer='https://accounts.google.com' /tmp/blob

Verified OK
```

Issue #4553, opened 2025-11-28
cosign triangulate command generate a .sig tag even for referrers signatures


scraly:
With cosign 3.0.2 when we sign an image, it generates a signature in the OCI distribution spec 1.1 Referrers API format.
Before, it generate a signature in a .sig tag.

After signing an image, when we execute the cosign triangulate, it shows a digest ending by .sig but this .sig does not exists ...

```
# Sign an image
$ cosign sign -y --key cosign.key $IMAGE@sha256:$HASH
Enter password for private key: 

# Verify
$ cosign verify $IMAGE@sha256:$HASH --key cosign.pub -o text | jq
 
Verification for xxxxxx.c1.de1.container-registry.ovh.net/public/gophers-api@sha256:xxxxxxxxxxxxxxx --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The signatures were verified against the specified public key
{
  "critical": {
    "identity": {
      "docker-reference": "xxxxxx.c1.de1.container-registry.ovh.net/public/gophers-api@sha256:xxxxxxxxxxxxxxxxxxx"
    },
    "image": {
      "docker-manifest-digest": "sha256:xxxxxxxxxxxx"
    },
    "type": "https://sigstore.dev/cosign/sign/v1"
  },
  "optional": null
}

# Get a non existing .sig??
$ cosign triangulate $IMAGE@sha256:$HASH
xxxxxxxx.c1.de1.container-registry.ovh.net/public/gophers-api:sha256-xxxxxxxxxxxxxxxxxxxx.sig

# Testing it
$ crane manifest xxxxxxxx.c1.de1.container-registry.ovh.net/public/gophers-api:sha256-xxxxxxxxxxxxxxxxxxxx.sig
Error: fetching manifest xxxxxxxx.c1.de1.container-registry.ovh.net/public/gophers-api:sha256-xxxxxxxxxxxxxxxxxxxx.sig: GET https://xxxxxxxx.c1.de1.container-registry.ovh.net/v2/public/gophers-api/manifests/sha256-xxxxxxxxxxxxxxx.sig: NOT_FOUND: artifact public/gophers-api:sha256-xxxxxxxxxxxxxx.sig not found

$ skopeo inspect docker://xxxxxxx.c1.de1.container-registry.ovh.net/public/gophers-api:sha256-xxxxxxxxxxxxx.sig
[...]

PR #4667, opened 2026-01-26
Add protobuf bundle support to triangulate

steiza:
#### Summary

For #4553 (also making progress on #4470).

Before the OCI 1.1 referrer spec, `cosign triangulate` would tell you where you could expect to find attestations, signatures, and SBOMs in an OCI registry for a given container image. This command would return the expected location **even if those things did not exist**, since determining their expected location was deterministic.

That is **not** the case for Cosign signatures that use the OCI 1.1 referrer spec, so we only return the manifest of those signatures if they exist. 

#### Release Note

* `triangulate` supports protobuf bundles stored with the OCI 1.1 referrer specification
* `triangulate` defaults to `--type='referrer'`, to get the old behavior use `--type='attestation'`

#### Documentation

N/A

### DIFF
--- a/test/helpers.go
+++ b/test/helpers.go
@@ -311,6 +311,7 @@ func importECDSAPrivateKey(t *testing.T, privKey *ecdsa.PrivateKey, td, fname st
 	return cosignKeyPath
 }
 
+//nolint:unparam // KeyBytes incorrectly detected as not used
 func importSampleKeyPair(t *testing.T, td string) (*cosign.KeysBytes, string, string) {
 	//nolint: gosec
 	const validrsa1 = `-----BEGIN RSA PRIVATE KEY-----


cmd/cosign/cli/options/triangulate.go
```diff
@@ -31,6 +31,6 @@ var _ Interface = (*TriangulateOptions)(nil)
 func (o *TriangulateOptions) AddFlags(cmd *cobra.Command) {
 	o.Registry.AddFlags(cmd)
 
-	cmd.Flags().StringVar(&o.Type, "type", "signature",
-		"related attachment to triangulate (attestation|sbom|signature|digest), default signature (sbom is deprecated)")
+	cmd.Flags().StringVar(&o.Type, "type", "referrer",
```
cmurphy, line 34: I tried this on both an OCI 1.0 and 1.1 registry and found that changing the default means that triangulate doesn't find the signature on a 1.0 registry unless you set --type=signature. Could we brainstorm some way for this to figure out whether to look for a referrer or .sig?

steiza: `cosign triangulate --type=signature`'s existing implementation is weird in that it's telling you where an "old" Cosign signature would be, regardless of if it would exist or not. For example:

```
$ go run cmd/cosign/main.go triangulate --type signature localhost:1338/demo/this-image-does-not-exist@sha256:0000000000000000000000000000000000000000000000000000000000000000                        

localhost:1338/demo/this-image-does-not-exist:sha256-0000000000000000000000000000000000000000000000000000000000000000.sig
```

It's not actually making any calls to the specified OCI registry, it's just doing a transformation on the OCI path. This works because the old bundle location was deterministic, which is not the case for the referrer bundles.

It is a little weird! But I'm not quite sure what the behavior should be, without a large rewrite of the subcommand.

steiza: Thinking about this more, another option is that we could remove this command entirely and recommend folks use `cosign tree` instead.

The only use-case `triangulate` has over `tree` is if you haven't created an old bundle yet but need to know what its location would be, but I think we're going to start removing old bundle functionality pretty soon.

cmurphy: I agree we should deprecate `triangulate`. I'm looking closer at the command help description the existing code and I'm starting to think it doesn't make sense to try to make `triangulate` work at all. It's purely an interface to https://github.com/sigstore/cosign/blob/3b3ab24c7ce68dd63c0600b751d4f026f6da816a/pkg/oci/remote/remote.go#L116-L138 to get the hypothetical names of the image layers where the signature etc would be. If the name of the referrer is nondeterministic, this command just doesn't make sense to use. It's not the same as `tree`, but it just doesn't apply to modern cosign.

steiza:
See #4675 instead.

PR #4675, opened 2026-01-30
Remove cosign triangulate

steiza:
#### Summary

For https://github.com/sigstore/cosign/issues/4553 (also making progress on https://github.com/sigstore/cosign/issues/4470).

Instead of adding protobuf bundle support for `cosign triangulate`, users should use `cosign tree` instead to find SBOMs, signatures, and other OCI objects related to their container image.

#### Release Note

* Removed `cosign triangulate`; users should use `cosign tree` instead to find OCI objects (including SBOMs and signatures) related to a container image

#### Documentation

* https://github.com/sigstore/docs/pull/412


Hayden-IO:
Can we have a deprecation warning with instructions on what to do instead rather than removal, and then remove in a v4 release?

steiza:
Gah, sorry, I got a little gung-ho; see #4676 instead.

PR #4676, opened 2026-01-31
Deprecate cosign triangulate

steiza:
#### Summary

For https://github.com/sigstore/cosign/issues/4553 (also making progress on https://github.com/sigstore/cosign/issues/4470).

When you run `cosign triangulate` this will first print:

```
Command "triangulate" is deprecated, instead you can use:
  - "oras discover" will show referring artifacts that exist
  - "cosign tree" will show referring artifacts and old image references
```

#### Release Note

* Deprecated `cosign triangulate`; users should use `cosign tree` instead to find OCI objects (including SBOMs and signatures) related to a container image

#### Documentation

N/A


cmd/cosign/cli/triangulate.go
```diff
@@ -27,6 +27,7 @@ func Triangulate() *cobra.Command {
 	o := &options.TriangulateOptions{}
 
 	cmd := &cobra.Command{
+		Deprecated:       "use \"tree\" instead",
```
Hayden-IO: Should we give a little more information about why? Also, this is true for new bundles, but for old bundles, triangulate still works, correct?

steiza: I'm struggling with the "why" as I'm not totally sure why people were using `triangulate`.

As far as I can tell, it's only really referenced in the `Counter-Signing` section of the README, which is probably not a use-case we want to continue to recommend.

Confusingly, `triangulate` would always return a location of where an object should be, **regardless** of if that object exists or not. Unless you're planning on using the output of `triangulate` directly in another shell command, I think you're almost always better of using `tree` which shows you what **actually** exists in the registry.

So it does "work" for old bundles, but I really wouldn't recommend folks use this command.

I'm having a hard time figuring out how to say that succinctly though, but I'm open to ideas. Just riffing (remember all of these will be prefixed with "Command "triangulate" is deprecated, ...")

> use "tree" instead which will show you what objects actually exist

> instead use "tree" which supports all the object types Cosign understands

> as it only supports the old bundle format, use "tree" instead

Hayden-IO: https://blog.sigstore.dev/cosign-image-signatures-77bab238a93/ is the only place I can find references to this. I also don't have a good sense of why this existed over just tree, so yea, I think something along the lines of the first example you gave

steiza: In the spirit of #4681 I modified the deprecation message to be similar:

```
Command "triangulate" is deprecated, instead you can use:
  - "oras discover" will show referring artifacts that exist
  - "cosign tree" will show referring artifacts and old image references
```

Hayden-IO (approved):

jonjohnsonjr:
Don't deprecate the pun :(

PR #4683, opened 2026-02-03
Enforce TSA requirement for Rekor v2, Fuclio signing

Hayden-IO:
When using Rekor v2, the log entry does not provide a Signed Entry Timestamp that validates the time of inclusion. For keyless signing with short-lived certificates, a trusted timestamp is required to verify the certificate was valid at the time of signing.

This change updates the new sign path to enforce that a timestamp authority is configured when Rekor v2 is to be used with Fulcio and no TSA URL is provided.

This should not affect users signing using the public instance, since the signing configuration includes a public TSA.


Hayden-IO:
FYI @jku

PR #4691, opened 2026-02-09
Deprecate rekor-entry-type flag

aaronlew02:
#### Summary

This change deprecates the `--rekor-entry-type` flag for `attest` and `attest-blob`.


cmd/cosign/cli/options/attest.go
```diff
@@ -93,6 +93,7 @@ func (o *AttestOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.RekorEntryType, "rekor-entry-type", rekorEntryTypes[0],
 		"specifies the type to be used for a rekor entry upload ("+strings.Join(rekorEntryTypes, "|")+")")
 	_ = cmd.RegisterFlagCompletionFunc("rekor-entry-type", cobra.FixedCompletions(rekorEntryTypes, cobra.ShellCompDirectiveNoFileComp))
+	_ = cmd.Flags().MarkDeprecated("rekor-entry-type", "support for this flag will be removed in the future")
```
Hayden-IO: Can you add something along the lines of "it is strongly discouraged to rely on Rekor for attestation storage, and in future releases of Rekor, this functionality will be removed."?

PR #4698, opened 2026-02-13
Mark batch of flags for deprecation

### LINKED ISSUE #4696: Deprecation notices to get ready for cosign v4
**Description**

We're in the midst of some major Cosign changes as we move from v2 to v3 to v4, as described on #4221.

As we wrap up Cosign v3 releases, we want to go through and deprecate a lot of functionality, to prepare for removing it in v4. This issue will track all the flags we want to deprecate.

I think for now we will **not** deprecate flags that are an alternative to using signing config, but I'm happy to get feedback on that.

- [x] #4691
- [ ] `--attachment` verification looks for OCI 1.1 referrers instead
- [ ] `--attachment-tag-prefix` use OCI 1.1 referrers instead
- [x] `--b64` we're moving from detached signatures to signatures in the bundle
- [x] `--ca-intermediates` use a trusted root
- [x] `--ca-roots` use a trusted root
- [x] `--certificate` use a trusted root
- [x] `--certificate-chain` use a trusted root
- [ ] `--experimental-oci11` we always treat this as true with new bundle format
- [x] `--issue-certificate` we aren't supporting going forward
- [x] `--new-bundle-format` we're only supporting the new bundle format in v4+
- [x] `--oidc-issuer` we aren't supporting going forward
- [x] `--output` we're moving from detached signatures to signatures in the bundle
- [x] `--output-certificate` we're moving from detached certificates to certificates in the bundle
- [x] `--output-payload` use `--bundle` instead
- [x] `--output-signature` we're moving from detached signatures to signatures in the bundle
- [x] `--private-infrastructure` use `--insecure-ignore-tlog` instead
- [x] `--record-creation-timestamp` not used with sigstore-go
- [x] `--rekor-url` use transparency log entry in bundle
- [ ] `--replace` not needed with OCI 1.1 referrers
- [x] `--rfc3161-timestamp` we're moving from detached signed timestamps to signed timestamps in the bundle
- [x] `--sct` included in bundle format
- [x] `--signature` included in bundle format
- [x] `--signature-digest-algorithm` included in bundle format
- [ ] `--sign-container-identity` OCI 1.1 referrers do not contain this information
- [x] `--timestamp-certificate-chain` use a trusted root

piceri:
Resolves https://github.com/sigstore/cosign/issues/4696
Batch of flags for depreciation. Breaking this issue up into batches to make PR review a little easier.

#### Summary
The PR marks the following flags for depreciation:

- b64
- rfc3161-timestamp
- issue-certificate
- output
- output-certificate
- output-signature

I have left the `rfc3161-timestamp` flag on the `bundle-create `command undeprecated as it seems this flag would still be needed to create bundles in the new supported format, but I can depreciate it too if v4 will not support bundle creation.

(This PR also fixes some Fulcio typos)

#### Release Note
The following flags have been marked for depreciation:
- b64
- rfc3161-timestamp
- issue-certificate
- output
- output-certificate
- output-signature

### DIFF
--- a/cmd/cosign/cli/options/attest_blob.go
+++ b/cmd/cosign/cli/options/attest_blob.go
@@ -85,7 +85,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.OutputSignature, "output-signature", "",
 		"write the signature to FILE")
 	_ = cmd.MarkFlagFilename("output-signature", signatureExts...)
-	_ = cmd.Flags().MarkDeprecated("output-signature", "support for this flag will be removed in the future. please use the new bundle output format")
+	_ = cmd.Flags().MarkDeprecated("output-signature", "please use --bundle to provide the output bundle location, which will include the signature")
 
 	cmd.Flags().StringVar(&o.OutputAttestation, "output-attestation", "",
 		"write the attestation to FILE")
@@ -94,7 +94,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.OutputCertificate, "output-certificate", "",
 		"write the certificate to FILE")
 	_ = cmd.MarkFlagFilename("key", certificateExts...)
-	_ = cmd.Flags().MarkDeprecated("output-certificate", "support for this flag will be removed in the future. please use the new bundle output format")
+	_ = cmd.Flags().MarkDeprecated("output-certificate", "please use --bundle to provide the output bundle location, which will include the certificate")
 
 	cmd.Flags().StringVar(&o.BundlePath, "bundle", "",
 		"write everything required to verify the blob to a FILE")
@@ -149,7 +149,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.RFC3161TimestampPath, "rfc3161-timestamp-bundle", "",
 		"path to an RFC 3161 timestamp bundle FILE")
 	// _ = cmd.MarkFlagFilename("rfc3161-timestamp-bundle") // no typical extensions
-	_ = cmd.Flags().MarkDeprecated("rfc3161-timestamp-bundle", "support for this flag will be removed in the future. please use the new bundle output format")
+	_ = cmd.Flags().MarkDeprecated("rfc3161-timestamp-bundle", "please use --bundle to provide the output bundle location, which will include the signed timestamp")
 
 	cmd.Flags().BoolVar(&o.IssueCertificate, "issue-certificate", false,
 		"issue a code signing certificate from Fulcio, even if a key is provided")
--- a/cmd/cosign/cli/options/attest_blob.go
+++ b/cmd/cosign/cli/options/attest_blob.go
@@ -85,6 +85,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.OutputSignature, "output-signature", "",
 		"write the signature to FILE")
 	_ = cmd.MarkFlagFilename("output-signature", signatureExts...)
+	_ = cmd.Flags().MarkDeprecated("output-signature", "support for this flag will be removed in the future. please use the new bundle output format")
 
 	cmd.Flags().StringVar(&o.OutputAttestation, "output-attestation", "",
 		"write the attestation to FILE")
@@ -93,6 +94,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.OutputCertificate, "output-certificate", "",
 		"write the certificate to FILE")
 	_ = cmd.MarkFlagFilename("key", certificateExts...)
+	_ = cmd.Flags().MarkDeprecated("output-certificate", "support for this flag will be removed in the future. please use the new bundle output format")
 
 	cmd.Flags().StringVar(&o.BundlePath, "bundle", "",
 		"write everything required to verify the blob to a FILE")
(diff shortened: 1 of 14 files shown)


cmd/cosign/cli/options/attest_blob.go
```diff
@@ -85,6 +85,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.OutputSignature, "output-signature", "",
 		"write the signature to FILE")
 	_ = cmd.MarkFlagFilename("output-signature", signatureExts...)
+	_ = cmd.Flags().MarkDeprecated("output-signature", "support for this flag will be removed in the future. please use the new bundle output format")
```
steiza: Several nits!

I think we want to keep these flag deprecation messages as one sentence - maybe a `,` or (if you're feeling grammar-ly!) `;` instead of `.`?

... but these will already be preferred with "Flag --output-signature has been deprecated, ..." so I'm not sure we need to reiterate "support for this flag will be removed in the future".

Also I think it's nice for us to reference what flag folks should use, when it exists.

So maybe something like:

```
"please use --bundle to provide the output bundle location, which will include the signature"
```

It doesn't have to be exactly that; I'm open to feedback here.

cmd/cosign/cli/options/attest_blob.go
```diff
@@ -93,6 +94,7 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.OutputCertificate, "output-certificate", "",
 		"write the certificate to FILE")
 	_ = cmd.MarkFlagFilename("key", certificateExts...)
+	_ = cmd.Flags().MarkDeprecated("output-certificate", "support for this flag will be removed in the future. please use the new bundle output format")
```
steiza: Similarly:

```suggestion
	_ = cmd.Flags().MarkDeprecated("output-certificate", "please use --bundle to provide the output bundle location, which will include the certificate")
```

cmd/cosign/cli/options/attest_blob.go
```diff
@@ -147,7 +149,9 @@ func (o *AttestBlobOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.RFC3161TimestampPath, "rfc3161-timestamp-bundle", "",
 		"path to an RFC 3161 timestamp bundle FILE")
 	// _ = cmd.MarkFlagFilename("rfc3161-timestamp-bundle") // no typical extensions
+	_ = cmd.Flags().MarkDeprecated("rfc3161-timestamp-bundle", "support for this flag will be removed in the future. please use the new bundle output format")
```
steiza:
```suggestion
	_ = cmd.Flags().MarkDeprecated("rfc3161-timestamp-bundle", "please use --bundle to provide the output bundle location, which will include the signed timestamp")
```

cmd/cosign/cli/options/signblob.go
```diff
@@ -84,18 +84,22 @@ func (o *SignBlobOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&o.Base64Output, "b64", true,
 		"whether to base64 encode the output")
+	_ = cmd.Flags().MarkDeprecated("b64", "support for this flag will be removed in the future. please use the new bundle output format")
```
steiza:
```suggestion
	_ = cmd.Flags().MarkDeprecated("b64", "please use --bundle, which already base64 encodes content as appropriate")
```

cmd/cosign/cli/options/signblob.go
```diff
@@ -84,18 +84,22 @@ func (o *SignBlobOptions) AddFlags(cmd *cobra.Command) {
...
 
 	cmd.Flags().StringVar(&o.OutputSignature, "output-signature", "",
 		"write the signature to FILE")
 	_ = cmd.MarkFlagFilename("output-signature", signatureExts...)
+	_ = cmd.Flags().MarkDeprecated("output-signature", "support for this flag will be removed in the future. please use the new bundle output format")
 
 	// TODO: remove when output flag is fully deprecated
 	cmd.Flags().StringVar(&o.Output, "output", "", "write the signature to FILE")
 	_ = cmd.MarkFlagFilename("output", signatureExts...)
+	_ = cmd.Flags().MarkDeprecated("output", "support for this flag will be removed in the future. please use the new bundle output format")
```
steiza: Whatever we're using for `output-signature` will work here as well

```suggestion
	_ = cmd.Flags().MarkDeprecated("output", "please use --bundle to provide the output bundle location, which will include the signature")
```

steiza (commented):
A bunch of minor nits / suggestions - but you don't have to use the exact wording I suggested.

> I have left the rfc3161-timestamp flag on the bundle-create command undeprecated as it seems this flag would still be needed to create bundles in the new supported format

Exactly - I think we should leave this in, so people can still convert old bundles to new bundles.

cmd/cosign/cli/options/attest_blob.go
piceri: > Several nits!
> 
> I think we want to keep these flag deprecation messages as one sentence - maybe a `,` or (if you're feeling grammar-ly!) `;` instead of `.`?
> 
> ... but these will already be preferred with "Flag --output-signature has been deprecated, ..." so I'm not sure we need to reiterate "support for this flag will be removed in the future".
> 
> Also I think it's nice for us to reference what flag folks should use, when it exists.
> 
> So maybe something like:
> 
> ```
> "please use --bundle to provide the output bundle location, which will include the signature"
> ```
> 
> It doesn't have to be exactly that; I'm open to feedback here.

Good suggestions! I have adjusted the messages to this new style. I left the "support for this flag will be removed in the future" as supplying an empty string as the message seems to add the flag back to the docs, and it seemed worth it for the removal. Happy to remove those as well if we are fine adding the flags back to the help docs.

steiza (approved):
LGTM - thanks!

Issue #4707, opened 2026-02-18
WriteAttestationsReferrer drops per-layer annotations (certificate, chain, RFC3161 timestamp)


rdsharma:
## Summary

`WriteAttestationsReferrer` in `pkg/oci/remote/write.go` silently drops per-layer OCI descriptor annotations when pushing DSSE attestations via the OCI 1.1 Referrers API. This makes referrer attestation manifests unverifiable — cosign, Kyverno, and any policy engine reading attestation referrers cannot find the signing certificate, chain, or RFC3161 timestamp.

## Root Cause

`WriteAttestationsReferrer` (line 421) calls `atts.Layers()` which returns `[]v1.Layer` — raw layer content with no descriptor metadata. These layers are passed to `WriteReferrer`, which rebuilds descriptors (lines 357-361) using only `MediaType`, `Digest`, and `Size`, omitting the `Annotations` field entirely.

Cosign stores verification materials as per-layer descriptor annotations (not in layer content):
- `dev.sigstore.cosign/certificate`
- `dev.sigstore.cosign/chain`
- `dev.sigstore.cosign/rfc3161timestamp`
- `dev.cosignproject.cosign/signature`
- `predicateType`

All of these are silently lost.

The sibling function `WriteSignaturesExperimentalOCI()` does **not** have this bug because it calls `sigs.Get()` (which returns `[]oci.Signature` carrying annotations) rather than `sigs.Layers()`.

## Steps to Reproduce

This function is not called by cosign's CLI directly, but is exported public API in `pkg/oci/remote/`. We hit this as library consumers calling `WriteAttestationsReferrer` from our own signing tool.

1. Build an `oci.SignedEntity` with DSSE attestations that have per-layer descriptor annotations (certificate, chain, RFC3161 timestamp, predicateType):
   ```go
   // Annotations are set via mutate.AppendSignatures → mutate.Addendum{Annotations: ...}
   // Verified present via atts.Get() → sig.Annotations() before push
   ```
2. Push via the Referrers API:
   ```go
   d, err := ociremote.ResolveDigest(ref, remoteOpts...)
   ociremote.WriteAttestationsReferrer(d, signedEntity, remoteOpts...)
   ```
3. Fetch the resulting referrer manifest and inspect layer descriptors - annotations are missing:
   ```json
   "layers": [{
     "mediaType": "application/vnd.in-toto+json",
     "size": 1757,
     "digest": "sha256:..."
[...]

Issue #4708, opened 2026-02-18
VerifyImageAttestations has no OCI 1.1 referrer discovery path


rdsharma:
## Summary

`VerifyImageAttestations` in `pkg/cosign/verify.go` has no code path to discover attestations stored via the OCI 1.1 Referrers API. Unlike `VerifyImageSignatures` (which checks `co.ExperimentalOCI11` at line 619 and calls `verifyImageSignaturesExperimentalOCI`), the attestation verification always falls through to tag-based `.att` suffix lookup.

This means attestations pushed via `WriteAttestationsReferrer` (the Referrers API path) cannot be verified by `cosign verify-attestation --experimental-oci11`.

### VerifyImageSignatures (works)
```go
// verify.go line 619
if co.ExperimentalOCI11 {
    return verifyImageSignaturesExperimentalOCI(ctx, signedImageRef, co)
}
```

### VerifyImageAttestations (missing)
```go
// verify.go line 1024 — no ExperimentalOCI11 check
// Always goes to tag-based ociremote.AttestationTag (line 1044)
```

The `--experimental-oci11` flag is parsed by the CLI (`verify-attestation` command) but is never propagated to the verification logic for attestations. It is effectively dead code in the attestation path.

### Downstream Impact

Kyverno's `verifyImages` with `attestations` calls cosign's `VerifyImageAttestations` under the hood, so it inherits this limitation. Policy engines cannot verify referrer-based DSSE attestations until this is addressed.

Similarly, `cosign download attestation` has no referrer support — it only checks bundles and tag-based (`.att`) storage.

## Expected Behavior

`VerifyImageAttestations` should have an `ExperimentalOCI11` code path (analogous to the one in `VerifyImageSignatures`) that:
1. Queries the OCI Referrers API for attestation manifests matching the subject digest
2. Fetches and verifies each attestation layer using the same verification logic

`cosign download attestation` should also support `--experimental-oci11` or `COSIGN_REGISTRY_REFERRERS_MODE=oci-1-1`.

## Discovered During

End-to-end testing of a container signing pipeline using OCI 1.1 referrers for DSSE attestation storage. After fixing #4707 (per-layer annotations preserved), attestations are correctly pushed but cannot be verified.

saschagrunert:
[...]

PR #4709, opened 2026-02-18
fix: preserve per-layer annotations in WriteAttestationsReferrer

rdsharma:
## Summary

`WriteAttestationsReferrer` drops per-layer OCI descriptor annotations (certificate, chain, RFC3161 timestamp, predicateType) when pushing DSSE attestations via the OCI 1.1 Referrers API. This makes referrer attestation manifests unverifiable.

### Root Cause

- `WriteAttestationsReferrer` calls `atts.Layers()` → `[]v1.Layer` (raw content, no descriptor metadata)
- `WriteReferrer` rebuilds descriptors with only `MediaType`, `Digest`, `Size` — omitting `Annotations`
- The sibling function `WriteSignaturesExperimentalOCI()` does NOT have this bug (it uses `sigs.Get()`)

### Fix

1. **`WriteAttestationsReferrer`**: Use `atts.Get()` instead of `atts.Layers()` to retrieve `[]oci.Signature` objects that carry annotations via their descriptor.
2. **`WriteReferrer`**: After building each layer descriptor, check if the layer implements `Annotations() (map[string]string, error)` via structural interface assertion, and populate the descriptor's `Annotations` field when present.

The structural interface avoids importing `oci` (which would create a circular dependency) and keeps `WriteReferrer` generic.

### Testing

- Added `TestWriteAttestationsReferrerPreservesAnnotations` that creates an attestation with certificate, chain, and predicateType annotations, pushes via `WriteAttestationsReferrer`, and verifies all annotations appear in the referrer manifest's layer descriptors.
- Verified end-to-end: signed a container image with DSSE attestation using referrers storage, confirmed all 5 per-layer annotations present in the referrer manifest via direct OCI Referrers API inspection.

### Release Request

We would appreciate this being cherry-picked into a **v3.0.x bugfix release** if possible.

Resolves: #4707

Note: A related but separate gap exists where `VerifyImageAttestations` has no OCI 1.1 referrer discovery path (tracked in #4708).

Signed-off-by: Rajeev Sharma <443108+rdsharma@users.noreply.github.com>

### DIFF
--- a/pkg/oci/remote/write.go
+++ b/pkg/oci/remote/write.go
@@ -359,6 +359,15 @@ func WriteReferrer(d name.Digest, artifactType string, layers []v1.Layer, annota
 			Digest:    layerDigest,
 			Size:      layerSize,
 		}
+		// Preserve per-layer annotations when available (e.g. from oci.Signature
+		// objects that carry certificate, chain, and RFC3161 timestamp annotations).
+		if al, ok := layer.(interface {
+			Annotations() (map[string]string, error)
+		}); ok {
+			if ann, err := al.Annotations(); err == nil && len(ann) > 0 {
+				layerDescriptors[i].Annotations = ann
+			}
+		}
 	}
 
 	// Create a manifest that includes the blob as a layer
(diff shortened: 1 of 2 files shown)


steiza (approved):
As mentioned on #4707, at some point we will be deprecating `WriteAttestationsReferrer()` in favor of functions like `WriteAttestationNewBundleFormat()`.

But in the meantime, it sounds like this change would help you, and doesn't have any negative side-effects, so I think we should ship it.

PR #4714, opened 2026-02-22
fix: write output certificate in sign-blob with key-based signing

### LINKED ISSUE #4140: Issue with --output-certificate in cosign sign-blob

clarabennettdev:
Write certificate from `sv.Bytes()` when using key-based `sign-blob --output-certificate`.

Fixes #4140


Hayden-IO (changes requested):
See https://github.com/sigstore/cosign/pull/4224#pullrequestreview-2918953855 - Only a certificate should be output, not the key, since the key already should be known to the caller.

Hayden-IO:
Flag is now deprecated.

PR #4723, opened 2026-02-27
Fix parsing of in-toto for string predicates

### LINKED ISSUE #4699: possible regression on `cosign verify-attestation`
**Description**

When using cosign in its latest version (either main or v3.0.4), `cosign verify-attestation` tries to parse a protobuf payload for the in-toto attestations, while I would expect it to simply parse the JSON:
```
Error: no matching attestations: proto: syntax error (line 1:1): unexpected token "{\"spdxVersion\":%  # omitted for brevity
```

When using cosign `v3.0.2` - the issue does not exist and verify attestation works correctly.

This was seen while bumping cosign over in `external-secrets` repository -- thus we have an uploaded image that can work as a test reference:

```
cosign verify-attestation --new-bundle-format=false --type slsaprovenance ghcr.io/external-secrets/external-secrets@sha256:53d49968e16af2f5c1e1ee825692e5c149785b091ffd85f1b48491a6b5b05b7d --certificate-identity-regexp "https://github.com/external-secrets/external-secrets/.*" --certificate-oidc-issuer https://token.actions.githubusercontent.com 2>&1 | head -c 100
```

I did some analysis, and it seems the error is related to an in-toto lib bump (I think) 

https://github.com/sigstore/cosign/blob/39f05cdcf8a52ffaff3a2208acdb33b1b42875a0/pkg/cosign/verifiers.go#L75 

on v3.0.2 this parses successfully but in 3.0.4 forwards it doesn't anymore.

**Version**

v3.0.4 - regression happens
v3.0.2 - regression does not happen

Let me know if this is my mistake as well. While I have _some_ understanding, there is a high chance this is all confusion from my side :)

cmurphy:
ee3d9fe1 changed the parsing of an in-toto Statement from using the deprecated in-toto/in-toto-golang Statement[1] to the recommended in-toto/attestation Statement[2]. This type is more strict about the Predicate field. Attestations that were created with `--type spdxjson` correctly become a regular JSON object under the "predicate" field in the Statement object, but attestations that were created with `--type spdx` become a string, which cosign was unable to Unmarshal with the new stricter type. This change addresses the issue by pre-parsing of the predicate to ensure it is properly formatted before unmarshalling it.

[1] https://pkg.go.dev/github.com/in-toto/in-toto-golang/in_toto#Statement
[2] https://pkg.go.dev/github.com/in-toto/attestation/go/v1#Statement

Fixes #4699

### DIFF
--- a/pkg/cosign/verifiers.go
+++ b/pkg/cosign/verifiers.go
@@ -72,8 +72,8 @@ func IntotoSubjectClaimVerifier(sig oci.Signature, imageDigest v1.Hash, annotati
 		return err
 	}
 
-	st := in_toto.Statement{}
-	if err := json.Unmarshal(stBytes, &st); err != nil {
+	st, err := toStatement(stBytes)
+	if err != nil {
 		return err
 	}
 	for _, subj := range st.Subject {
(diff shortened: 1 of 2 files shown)


pkg/cosign/verifiers.go
```diff
@@ -92,3 +92,34 @@ func IntotoSubjectClaimVerifier(sig oci.Signature, imageDigest v1.Hash, annotati
...
+		// Convert it to a JSON object and try again.
+		stmtMap := make(map[string]any)
+		if err := json.Unmarshal(stBytes, &stmtMap); err != nil {
+			return nil, err
+		}
+		predicate, ok := stmtMap["predicate"]
+		if !ok {
+			return nil, fmt.Errorf("could not parse statement, could not find predicate")
+		}
+		if p, ok := predicate.(string); ok {
```
Hayden-IO: If it's not a string, should we throw an error?

cmurphy: Done

Issue #4684, opened 2026-02-04
CLI compatibility tracker WRT rekor v2


jku:
I wanted to track and document the "compatibility" of cosign CLI as rekor v2 and related changes start to happen. The main question being: _"will existing workflows that are based on calling cosign CLI break because the CLI changes?"_

Current state WRT CLI changes:
* cosign < 3.0:
  * existing workflows are unlikely to break when rekorv2 is enabled in signingconfig (as the client does not work with rekor v2 at all without some specific flags)
  * things will obviously start breaking for old cosign when the rekor v1 instance is made read-only but that's a future problem
* cosign 3.0 - 3.0.4: 
  * existing bundle signing flows (at least with public good instance) should keep working when signingconfig contains rekor v2 and a TSA: cosign will start signing with rekor 2 and will use a TSA -- @Hayden-IO was there anything to note here?
  * existing bundle verify flows will break when rekor v2 bundles appear because  an additional `--use-signed-timestamps` is needed (fixed in #4666)

I'm not very familiar with cosign and especially with the non-bundle flows so am not sure how they are affected: would appreciate insights on this.

Hayden-IO:
Nothing more to note for Rekor v2 + TSA - For the public instance, it should just work for signing without additional flags, but for private deployments. For the verification path, it will start to just work after v3.0.5 is cut.

jku:
I'm planning a PR to add a Troubleshooting section to README with some failure cases for SEO:
* the main case that needs documentation is _failed to verify timestamps: threshold not met for verified log entry integrated timestamps: 0 < 1_ -- this is very likely an older client verifying a bundle with a rekor 2 entry.
  * Fix: upgrade to latest release OR
  * Fix: in older releases `--use-signed-timestamps` may help
* _Error: no signatures found_ happens with `cosign verify` when cosign version < 3 and the image signature uses rekor v2  
  * Fix: upgrade to latest release
* I'd also like it if the supported versions are clearly listed, like maybe _"cosign project provides support for the latest release and security fixes for the latest 2.6.x release"_ -- but maybe a cosign maintainer should do that

PR #4724, opened 2026-02-27
Readme add troubleshooting section

jku:
Part of #4684:
* Move "What is not production ready?" section out of the Quickstart
* Add troubleshooting to the Quickstart

The goal is to have the slightly cryptic error messages in a place where people, search engines and AI can find them.

Happy to try another approach as well if there are suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,16 +196,16 @@ $ cosign verify-blob artifact \
 
 ### Troubleshooting
 
-If you encounter issues with cosign, first make sure you are using a recent release.
+If you encounter issues with Cosign, first make sure you are using a recent release: The Cosign project actively supports the most recent release as well as the last release in the v2 series.
 
 #### Common issues and remedies
 
-1. verify fails with `failed to verify timestamps: threshold not met for verified log entry integrated timestamps: 0 < 1`: You may be verifying a signature that requires RFC3161 timestamp support
-   * Upgrade to most recent cosign or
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -194,21 +194,22 @@ $ cosign verify-blob artifact \
   --certificate-oidc-issuer "https://token.actions.githubusercontent.com"
 ```
 
-### What ** is not ** production ready?
+### Troubleshooting
 
-While parts of `cosign` are stable, we are continuing to experiment and add new features.
-The following feature set is not considered stable yet, but we are committed to stabilizing it over time!
+If you encounter issues with cosign, first make sure you are using a recent release.
```
jku: would love to make this more concrete (like "cosign project provides support for latest release and last release in the 2.6 branch") but I don't want to put words into cosign maintainers mouths...

Hayden-IO: I was just mentioning this to @steiza - I'm OK with saying we support the last minor releases of Cosign v2 (2.6.x) and Cosign v3. We will drop v2 once Cosign v4 is releaesed.

There's a bit more we need around backporting but no need to bundle that here.

README.md
```diff
@@ -556,6 +557,22 @@ The following checks were performed on each of these signatures:
 
 ## FAQ
 
+### What ** is not ** production ready?
```
Hayden-IO: Can you delete this actually? All the formats are stable.

PR #4726, opened 2026-02-27
(fix): Fix necessary linter issues from gosec,errcheck and static check

ompushkara:


Hayden-IO:
Please also sign the DCO.

Issue #3168, opened 2023-08-07
How do you change the password of a generated private key?


jtackaberry:
**Question**
`cosign generate-key-pair` was certainly easy enough to run.  But it's not clear to me how to re-encrypt the private key with a different password.   This is a pretty obvious requirement but after some Googling and scanning TFM I couldn't find any mention of it, so, embarrassingly, I feel like I must be missing something rather basic.  Cluebatting most appreciated.

haydentherapper:
We don't have support for that, though it should be straightforward to implement it.

viveksahu26:
Hey @haydentherapper, I would like to work on it. 
I understand that we need to update the private key's password. So, to implement  that:
- It requires a `key` path
- and flag, `update-password`

So, the whole command would be like:
```
cosign generate-key-pair --key cosign.key --update-password=true`
```
So, the above command does the following procedure:
- First, read the file of the key,  then decode the pem format
- Secondly, Decrypt the ciphertext using the old password.
- Then, we can Encrypt it again, with a new password provided by the user.

My only concern is that `generate-key-pair` meaningfully doesn't seem to fit here, because we are updating the key passwords not generating keys.

The command would be more meaningful if it were something like:
`cosign generate-key-pair` -->  to generate key pair
`cosign update-key-pair  --key cosign.key`  --> to update the password of the private key.

haydentherapper:
A dedicated command like `update-key-pair` sounds good.

PR #4730, opened 2026-02-28
feat: add command update-key-pair (solves #3168)

Aex12:
#### Summary
This PR adds a new command `update-key-pair` that can be used to update the encryption password for existing keys generated by the command `generate-key-pair`. This solves issue #3168

#### Release Note

* Added new `cosign update-key-pair` CLI command for re-encrypting an existing
  encrypted private key with a new password. The command prompts interactively
  for the current password and the new password (with confirmation). The
  `COSIGN_PASSWORD` environment variable can be used to supply the current
  password non-interactively, and the new `COSIGN_NEW_PASSWORD` environment
  variable can be used to supply the new password non-interactively.
* Added exported function `cosign.UpdateKeyPair` to the `pkg/cosign` package, allowing programmatic re-encryption of an encrypted PEM private key with a new password.

#### Documentation

This change requires a documentation update. I'll open a PR as soon as the proposed API is confirmed to be okay.

### DIFF
--- a/cmd/cosign/cli/commands.go
+++ b/cmd/cosign/cli/commands.go
@@ -105,6 +105,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(Generate())
 	cmd.AddCommand(GenerateKeyPair())
 	cmd.AddCommand(ImportKeyPair())
+	cmd.AddCommand(UpdateKeyPair())
 	cmd.AddCommand(Initialize())
 	cmd.AddCommand(Load())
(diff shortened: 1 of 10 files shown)


Hayden-IO:
Hey, thanks for creating this PR. We've been reviewing commands in Cosign, and maintainers are pushing for fewer rather than more commands. Looking over this PR, there's a lot of duplication with generate-key-pair. I know the issue suggested a new command, but could have an --update flag on the existing command instead?

PR #4737, opened 2026-03-04
fix(verify): Attach detached certificates to static signatures via wrapped verifier

aaronlew02:
#### Summary

This change updates `loadSignatureFromFile` to properly bind the provided `--certificate` and `--certificate-chain` to the constructed signature object. Previously, verification using detached materials ignored these flags during object initialization, which caused transparency log lookups to incorrectly fall back to querying with a raw public key instead of the full certificate PEM, preventing the signature from being found in the log.

#### Release Note

Previously, running (1) and then verifying using the requested detached materials by running (2) (with or without the `payload`) would result in a `signature not found in transparency log` error. With this change, verification will succeed.

(1) `cosign sign --new-bundle-format=false --output-signature sig.b64 --output-certificate cert.pem --output-payload payload.json --bundle sign.bundle --use-signing-config=false [IMAGE_URI]`
(2) `cosign verify --new-bundle-format=false --signature sig.b64 --certificate cert.pem --payload payload.json --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]`

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -4996,3 +4996,85 @@ func selfSignedCertificate() (*x509.Certificate, *ecdsa.PrivateKey, error) {
 	}
 	return cert, priv, nil
 }
+
+func TestSignVerifyDetachedKeyless(t *testing.T) {
+	td := t.TempDir()
+	err := setLocalEnv(t, td)
+	must(err, t)
+	must(fulcioroots.ReInit(), t)
+
+	repo, stop := reg(t)
+	defer stop()
+	imgName := path.Join(repo, "cosign-e2e-detached-keyless")
+
+	_, _, cleanup := mkimage(t, imgName)
+	defer cleanup()
+
+	identityToken, err := getOIDCToken()
+	must(err, t)
+
+	ctx := context.Background()
+	sigFile := filepath.Join(td, "sig.out")
+	certFile := filepath.Join(td, "cert.out")
+
+	// Verify should fail before signing
+	failCmd1 := cliverify.VerifyCommand{
+		RekorURL: rekorURL,
+		CertVerifyOptions: options.CertVerifyOptions{
+			CertOidcIssuer: os.Getenv("ISSUER_URL"),
+			CertIdentity:   certID,
+		},
+	}
+	mustErr(failCmd1.Exec(ctx, []string{imgName}), t)
+
+	ko := options.KeyOpts{
+		FulcioURL:        fulcioURL,
+		RekorURL:         rekorURL,
+		IDToken:          identityToken,
+		SkipConfirmation: true,
+	}
+	so := options.SignOptions{
+		Upload:            true,
+		TlogUpload:        true,
+		OutputSignature:   sigFile,
+		OutputCertificate: certFile,
+		UseSigningConfig:  false,
+	}
+	must(sign.SignCmd(ctx, ro, ko, so, []string{imgName}), t)
+
+	// Verify should fail with a detached signature but no certificate
+	failCmd2 := cliverify.VerifyCommand{
+		RekorURL:     rekorURL,
+		SignatureRef: sigFile,
+		CertVerifyOptions: options.CertVerifyOptions{
+			CertOidcIssuer: os.Getenv("ISSUER_URL"),
+			CertIdentity:   certID,
+		},
+	}
+	mustErr(failCmd2.Exec(ctx, []string{imgName}), t)
+
+	// Now verify should work using the certificate
+	cmd := cliverify.VerifyCommand{
+		RekorURL:     rekorURL,
+		SignatureRef: sigFile,
+		CertRef:      certFile,
+		CertVerifyOptions: options.CertVerifyOptions{
+			CertOidcIssuer: os.Getenv("ISSUER_URL"),
+			CertIdentity:   certID,
+		},
+	}
+	must(cmd.Exec(ctx, []string{imgName}), t)
+
+	// Verify should also work with the certificate chain
+	cmdWithChain := cliverify.VerifyCommand{
+		RekorURL:     rekorURL,
+		SignatureRef: sigFile,
+		CertRef:      certFile,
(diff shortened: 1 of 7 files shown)


Hayden-IO (commented):
I'm not sure this is the right way to resolve this. The certificate and chain were already [processed to create a signature verifier](https://github.com/aaronlew02/cosign/blob/64e5b86573888d5ee855c29f371c5de4d8dc2769/cmd/cosign/cli/verify/verify.go#L188), so they shouldn't need to be passed along ideally.

Have you looked at where the [certificate is extracted from the sig object](https://github.com/sigstore/cosign/blob/01e227a7d0863380f01cee484d90b18288aa8f64/pkg/cosign/verify.go#L951-L973)? A couple of things seem suspicious to me:

* sig.Cert() isn't erring out when empty - why isn't it?
* Can the certificate be pulled from co.SigVerifier instead?

I'd also look at https://github.com/sigstore/cosign/pull/4525 where this function was updated to see if it's related.

aaronlew02:
> sig.Cert() isn't erring out when empty - why isn't it?

This is to support cases where the constructed signature should not contain a certificate, such as key-based signing, in which the signature is indexed in Rekor by its public key.

> Can the certificate be pulled from co.SigVerifier instead?

Currently, no. `SigVerifier`'s type interface, [`Verifier`](https://github.com/sigstore/sigstore/blob/b56c8664d026358a77a52f88832759047eba7e85/pkg/signature/verifier.go#L33-L36), does not include a field for the certificate, only the public key, which seems in line with the interface's purpose.

An alternative approach, so that the provided certificate and chain do not have to be loaded from their respective files multiple times in the same verification workflow, is to return the parsed certificate with `SigVerifier` from `LoadVerifierFromKeyOrCert`, provide the parsed certificate to `keyBytes` via `CheckOpts`, and use the parsed certificate as a fallback if `sig.Cert()` returns `nil` but the parsed certificate is available.

I've pushed an implementation of this approach in the latest commit. WDYT?

cmurphy (commented):
Could you add a unit test or e2e test for this?

Hayden-IO (commented):
I'm hesitant to add more fields to CheckOpts, especially when we have related fields like RootCerts/IntermediatesCerts already. I could see an argument for a Certificate field that's documented to only be used for constructing a Rekor entry, though I'm a little concerned that it gets misused and assumed to be validated.

Part of the issue is we do verification at multiple points in the code - either when you provide a flag to initialize a verifier, or deeper in the call graph when there is no verifier and it's assumed a certificate existed on the OCI image. Right now, we either have a `co.SigVerifier` that contains a provided key or key from a validated certificate, or we have an unvalidated certificate and its chain (optionally, it can also be initialized through TUF for the relevant CheckOpts fields) passed through the OCI structure.

Do you have thoughts on other ways we could approach this? Could we initialize `static.NewSignature` closer to where we fetch a certificate and its chain? Any other ideas?

aaronlew02:
@Hayden-IO There's an alternative implementation (pushed in the latest commit) which attempts to address your previous question of whether the certificate can be pulled from `co.SigVerifier`. While the `Verifier` type from sigstore/sigstore does not contain a certificate or chain, we can wrap the type in a new struct, `verifierWithCertChain`, that contains the `Verifier` and the provided certificate and chain, and pull the certificate from that struct. WDYT?

aaronlew02:
@cmurphy I've added a unit test for `LoadSignatureFromFile` that verifies that detached certificates are attached to a loaded signature.

cmurphy (commented):
> @cmurphy I've added a unit test for LoadSignatureFromFile that verifies that detached certificates are attached to a loaded signature.

Thanks for adding a test. The new test validates the new properties of the signature that you added, but it's not actually validating that the issue you reported is fixed. I'd like to see something that actually runs the full Verify* flow to make sure that, given a detached certificate, the signature will verify. A unit test for this would probably look something like `TestVerifyImageSignatureWithKeyAndCert`, or often an e2e test is easier for verifying the full functionality. It would also be good to make sure it's testing both the cert-only and cert-chain use cases.

aaronlew02:
@cmurphy I've replaced the unit test with an e2e test, `TestSignVerifyDetachedKeyless`, that targets a keyless signing and verification workflow in which a detached signature is verified using a certificate and chain.

test/e2e_test.go
```diff
@@ -4996,3 +4996,85 @@ func selfSignedCertificate() (*x509.Certificate, *ecdsa.PrivateKey, error) {
...
+		},
+	}
+	must(cmd.Exec(ctx, []string{imgName}), t)
+
+	// Verify should also work with the certificate chain
+	cmdWithChain := cliverify.VerifyCommand{
+		RekorURL:     rekorURL,
+		SignatureRef: sigFile,
+		CertRef:      certFile,
+		CertChain:    os.Getenv("SIGSTORE_ROOT_FILE"),
```
cmurphy: Is this meaningfully different from the above command with only CertRef set? If the same signature was verifiable with only the cert and not the chain before, then setting CertChain here doesn't seem to be demonstrating that this fixes the issue of --certificate-chain being ignored as you describe in the PR description. Do you have thoughts on that?

aaronlew02: This is a great point. Since the e2e testing environment sets `SIGSTORE_ROOT_FILE`, that chain is [already used](https://github.com/sigstore/cosign/blob/8f34218544e1e9324780ea14438a7cae72cc2dbf/cmd/cosign/cli/verify/common.go#L464-L475) for verification via the first command.

I've modified the test so that, after the first command succeeds, `SIGSTORE_ROOT_FILE` is invalidated (enforcing the need for a certificate chain to be provided explicitly), the first command fails expectedly, and the second command passes with the root file provided explicitly.

cmurphy (approved):
This looks good to me but I'll let @Hayden-IO weigh back in

Hayden-IO (approved):
LGTM!

PR #3891, opened 2024-09-27
remove previously deprecated attach/download sbom commands

hectorj2f:
#### Summary
The attach/download SBOM commands have been declared as deprecated long time ago. I believe it is a good time to cleanup these not recommended actions.

#### Release Note
Remove deprecated attach/download SBOM.

PR #4742, opened 2026-03-06
Stream SBOM download

### LINKED ISSUE #4741: "download sbom" reads whole blob to memory
**Description**

Whole SBOM is read to memory just to be dumped into the file. For various values of COSIGN_MAX_ATTACHMENT_SIZE it could be quite costly, so streaming variant would be nice. Replacing ``Payload`` with something like ``StreamingPayload`` could be useful also in different places. There is https://github.com/sigstore/cosign/issues/4112 for similar issue.

tkopecek:
Downloading SBOM can be memory-intensive as whole SBOM is read into memory, while it can be streamed.

Closes #4741

#### Summary
Downloading large SBOMs requires a lot of memory. Using streaming approach would limit it. For testing improvements just monitor used memory e.g. via ``GODEBUG=gctrace=1`` Otherwise there should be no changes in behaviour, so all current tests should pass.

#### Release Note
NONE

IMPORTANT: This MR is heavily supported by Claude as I'm not a go developer. So, if you're willing to go with this apprach, please a) doublecheck the code changes b) let me know if there is AI-assistance policy which I should acknowledge.

### DIFF
--- a/cmd/cosign/cli/download/sbom.go
+++ b/cmd/cosign/cli/download/sbom.go
@@ -85,24 +85,31 @@ func SBOMCmd(
 		return nil, fmt.Errorf("getting sbom attachment: %w", err)
 	}
 
-	// "attach sbom" attaches a single static.NewFile
-	sboms := make([]string, 0, 1)
-
 	mt, err := file.FileMediaType()
 	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


Hayden-IO (commented):
This command is deprecated and should be removed in an upcoming release - https://github.com/sigstore/cosign/pull/3891 - so I don't think we need to make any updates here.

tkopecek:
ok, closing MR/issue

PR #4745, opened 2026-03-07
fix: write output certificate in sign-blob when signing with key

### LINKED ISSUE #4140: Issue with --output-certificate in cosign sign-blob

Onyx2406:
## Summary

- Fix `--output-certificate` flag not writing any file when using `cosign sign-blob --key`
- Align `sign-blob`'s certificate output behavior with `sign`'s implementation

## Motivation

As reported in #4140, `cosign sign-blob --key cosign.key --output-certificate certificate.crt` silently produces no certificate file, while `cosign sign --key cosign.key --output-certificate certificate.crt` correctly writes the public key to the output file.

## Root Cause

The `sign-blob` command used `extractCertificate()` to handle `--output-certificate`, which calls `sv.Bytes()` and then attempts to parse the result as a PEM certificate via `UnmarshalCertificatesFromPEM`. When signing with `--key`, `sv.Bytes()` returns a PEM-encoded **public key** (not a certificate), so `UnmarshalCertificatesFromPEM` fails and `extractCertificate` returns `nil` — causing nothing to be written.

In contrast, the `sign` command calls `sv.Bytes()` directly and writes whatever it returns (certificate or public key) to the output file.

## Changes

- **`cmd/cosign/cli/sign/sign_blob.go`**: Replace `extractCertificate()` call in the `--output-certificate` handler with a direct `sv.Bytes()` call, consistent with how `sign` handles it. The `extractCertificate` function is retained as it is still used in the old bundle format path where certificate-only filtering is appropriate.
- **`cmd/cosign/cli/sign/sign_blob_test.go`**: Add assertions verifying that the output certificate file is created and contains valid PEM public key data when signing with `--key`.

## Test Plan

- [x] Existing `TestSignBlobCmd` passes
- [x] New test assertions verify the output certificate file exists, is non-empty, and contains a valid `PUBLIC KEY` PEM block
- [x] `go test ./cmd/cosign/cli/sign/ -run TestSignBlobCmd -v` passes

Fixes #4140

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -243,20 +243,15 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
 	}
 
 	if outputCertificate != "" {
-		certBytes, err := extractCertificate(ctx, sv)
+		rekorBytes, err := sv.Bytes(ctx)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("create certificate file: %w", err)
 		}
-		if certBytes != nil {
-			bts := certBytes
-			if b64 {
-				bts = []byte(base64.StdEncoding.EncodeToString(certBytes))
-			}
-			if err := os.WriteFile(outputCertificate, bts, 0600); err != nil {
-				return nil, fmt.Errorf("create certificate file: %w", err)
-			}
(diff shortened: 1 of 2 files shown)


Hayden-IO:
I've mentioned this before (https://github.com/sigstore/cosign/pull/4224#pullrequestreview-2918953855), but only the certificate should be output, not the key. 

This flag is now deprecated nonetheless, so I'm going to close this and the attached issue.

PR #4746, opened 2026-03-08
fix: write output certificate in sign-blob with key-based signing

### LINKED ISSUE #4140: Issue with --output-certificate in cosign sign-blob

ogulcanaydogan:
## Summary

`sign-blob --output-certificate` silently skips writing the certificate file when signing with a private key (`--key`), while `sign --output-certificate` works fine in the same scenario.

The root cause is that `sign-blob` uses `extractCertificate()` which returns nil for non-Fulcio signers (i.e. private key based), so the file never gets written. Meanwhile, `sign` just calls `sv.Bytes()` directly and writes whatever it gets.

This patch aligns `sign-blob` with `sign` by calling `sv.Bytes()` directly, so both commands write the public key PEM when there is no Fulcio certificate.

## Changes

- `sign_blob.go`: replace `extractCertificate()` call with `sv.Bytes()` in the `--output-certificate` code path
- `sign_blob_test.go`: add assertion to verify the certificate file is actually written

## Testing

- Existing unit test `TestSignBlobCmd` passes
- Added verification that the output certificate file is non-empty when signing with a key

Fixes #4140

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -243,20 +243,18 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
 	}
 
 	if outputCertificate != "" {
-		certBytes, err := extractCertificate(ctx, sv)
+		certBytes, err := sv.Bytes(ctx)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error getting certificate: %w", err)
 		}
-		if certBytes != nil {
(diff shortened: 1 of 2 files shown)


Hayden-IO:
I've mentioned this before (https://github.com/sigstore/cosign/pull/4224#pullrequestreview-2918953855), but only the certificate should be output, not the key. 

This flag is now deprecated nonetheless, so I'm going to close this and the attached issue.

ogulcanaydogan:
Makes sense, thanks for the context

PR #4747, opened 2026-03-08
feat: add retry with exponential backoff for Fulcio and Rekor RPCs

### LINKED ISSUE #2198: Cosign should retry Fulcio and Rekor RPCs
**Description**

@vaikas mentioned to me that cosign commands occasionally fail due to an unexpected error either from the network or Sigstore backends.  These errors typically will come at the very end of a CI job.  To avoid failing expensive long running jobs with a transient cosign error we should retry failed Sigstore calls. 

We should do this in all Sigstore clients. 

Ideally this would be implemented with configurable exponential backoff and timeout. 

**Version**
all versions

ogulcanaydogan:
## Summary

Transient network errors or server-side 5xx responses during signing can fail expensive CI jobs at the very end. This adds retry logic with exponential backoff around Fulcio and Rekor API calls so that temporary failures don't break entire pipelines.

## Changes

New `internal/pkg/retry` package using `cenkalti/backoff/v4` (already an indirect dep):
- `IsRetriable()` checks if an error should be retried (5xx, 429, network errors)
- `Do()` wraps a function with exponential backoff, max 3 retries, 30s timeout
- Non-retriable errors (400, 403, 404) fail immediately

Wrapped call sites:
- **Fulcio**: `GetCert()` - certificate signing request
- **Rekor**: `doUpload()` - create log entry (conflict errors skip retry)
- **Rekor**: `GetTlogEntry()` - fetch entry by UUID
- **Rekor**: `FindTlogEntry()` - search log query
- **Rekor**: `rekorPubsFromClient()` - fetch public key

## Testing

- Unit tests for retry package cover: success, retry-then-success, non-retriable errors, context cancellation
- All existing tests pass

```
go test ./internal/pkg/retry/ -v
go build ./cmd/cosign/cli/fulcio/
go build ./pkg/cosign/
```

Fixes #2198

### DIFF
--- a/cmd/cosign/cli/fulcio/fulcio.go
+++ b/cmd/cosign/cli/fulcio/fulcio.go
@@ -27,6 +27,7 @@ import (
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v3/internal/auth"
 	"github.com/sigstore/cosign/v3/internal/pkg/cosign/fulcio/fulcioroots"
+	"github.com/sigstore/cosign/v3/internal/pkg/retry"
 	"github.com/sigstore/fulcio/pkg/api"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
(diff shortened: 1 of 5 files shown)


ogulcanaydogan:
The e2e test failure appears to be an infrastructure issue unrelated to this change:

```
failed to execute bake: read |0: file already closed
```

All other 28 checks pass, including unit tests on all platforms, lint, conformance, e2e-cross, e2e-kms, e2e-registry, and validate-release-job.

ogulcanaydogan:
Thanks for the pointer. I'll look into adding retry logic in sigstore-go instead

PR #4761, opened 2026-03-09
fix(sign): closing SignerVerifier too early when signing with a security key

epelip:
#### Summary

Fixes an error occurring when using a security key, such as a Yubikey or PKCS11 key for signing. The SignerVerifier was wrapped in a SignerVerifierKeypair and the underlying SignerVerifier closed before consuming the keypair to create the signature. This fix moves the deferred SignerVerifier.Close() call to the same function the sign method is called with the keypair.

Triggering the bug and testing the fix can be performed e.g. with the following command:

```
cosign sign --sk <IMAGE DIGEST>
```

Note that this requires compiling via `cosign-pivkey-pkcs11key` make target and a Yubikey initialized using the following commands (note that these commands will reset the Yubikey PIV application and wipe the existing keys):

```
cosign piv-tool reset
cosign piv-tool generate-key
```

#### Release Note

Fixed an error occurring when using a security key, such as a Yubikey or PKCS11 key for signing caused by early closing of the HSM.

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -77,6 +77,11 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
 	}
 
 	keypair, sv, certBytes, idToken, err := signcommon.GetKeypairAndToken(ctx, ko, certPath, certChainPath)
+	defer func() {
+		if sv != nil {
+			sv.Close()
+		}
+	}()
 	if err != nil {
 		return nil, fmt.Errorf("getting keypair and token: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 The Sigstore Authors.
+// Copyright 2026 The Sigstore Authors.
```
Hayden-IO: We can leave this as-is, copyright year only should be updated for new files.

epelip: Reverted and amended the commit

PR #4762, opened 2026-03-09
Disallow --new-bundle-format and --rfc3161-timestamp

### LINKED ISSUE #4594: cosign does not generate signed timestamp anymore

Hayden-IO:
With the modern bundle format, the signed timestamp output flag will be ignored. We should log a warning like with the other output flags.

Fixes #4594

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -130,7 +130,7 @@ race conditions or (worse) malicious tampering.
 			}
 			if err := signcommon.LoadTrustedMaterialAndSigningConfig(cmd.Context(), &ko, o.UseSigningConfig, o.SigningConfigPath,
 				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL, o.TrustedRootPath, o.TlogUpload,
-				o.NewBundleFormat, "", o.Key, o.IssueCertificate, o.Output, "", o.OutputCertificate, o.OutputPayload, o.OutputSignature); err != nil {
+				o.NewBundleFormat, "", o.Key, o.IssueCertificate, o.Output, "", o.OutputCertificate, o.OutputPayload, o.OutputSignature, ""); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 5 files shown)


cmd/cosign/cli/sign.go
```diff
@@ -130,7 +130,7 @@ race conditions or (worse) malicious tampering.
 			}
 			if err := signcommon.LoadTrustedMaterialAndSigningConfig(cmd.Context(), &ko, o.UseSigningConfig, o.SigningConfigPath,
 				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL, o.TrustedRootPath, o.TlogUpload,
-				o.NewBundleFormat, "", o.Key, o.IssueCertificate, o.Output, "", o.OutputCertificate, o.OutputPayload, o.OutputSignature); err != nil {
+				o.NewBundleFormat, "", o.Key, o.IssueCertificate, o.Output, "", o.OutputCertificate, o.OutputPayload, o.OutputSignature, ""); err != nil {
```
steiza, line 133: I guess when the pre-protobuf bundle timestamp support was added, it was just added to the `*-blob` subcommands? Just another reason I'm glad we're in the process of deprecating these flags...

PR #4764, opened 2026-03-10
fix(deps): CVE-2026-2303 / CVE-2026-2303 go.mongodb.org/mongo-driver

matthiasbruns:
The mongo-go-driver repository contains CGo bindings for GSSAPI (Kerberos) authentication on Linux and macOS. The C wrapper implementation contains a heap out-of-bounds read vulnerability due to incorrect assumptions about string termination in the GSSAPI standard. Since GSSAPI buffers are not guaranteed to be null-terminated or have extra padding, this results in reading one byte past the allocated heap buffer.

For more details see: https://www.cve.org/CVERecord?id=CVE-2026-1849

#### Summary
Our BlackDuck scans found a vulnerability in the latest cosign release, which would also persist in the current [main](https://github.com/sigstore/cosign/blob/main/go.mod#L274C2-L274C37)
go.mongodb.org/mongo-driver v1.17.6

> The mongo-go-driver is vulnerable to a heap out-of-bounds read due to incorrect handling of string termination in the GSSAPI (Kerberos) authentication implementation. This could allow an attacker to read memory beyond the allocated heap buffer, potentially exposing sensitive information or causing unexpected behavior.

#### Release Note

* Updated go.mongodb.org/mongo-driver  to v1.17.9

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -271,7 +271,7 @@ require (
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
 	github.com/yashtewari/glob-intersection v0.2.0 // indirect
-	go.mongodb.org/mongo-driver v1.17.6 // indirect
+	go.mongodb.org/mongo-driver v1.17.9 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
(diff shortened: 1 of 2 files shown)


Hayden-IO:
We don't use mongo in any way, this is pulled in through openapi I believe. Can you demonstrate how Cosign would be affected by this vulnerability?

cmurphy (approved):
I don't mind updating the dependency to make scanners happy even if we're not affected by this CVE

PR #4785, opened 2026-03-30
Require bundle output or registry upload 

### LINKED ISSUE #4534: Require bundle output or registry upload

aaronlew02:
Closes #4534

#### Summary

This change requires that the user enable OCI registry upload (where applicable) or specify a bundle output path so that all signing events provide material with which the user can verify the signing event.

#### Release Note

This is a **breaking change**. When using `--new-bundle-format` for signing and attestation, it is now strictly required that the resulting verification material is either uploaded to an OCI registry or written to a local file. 

Previously, there was a loophole where users could run `sign`, `attest`, `sign-blob`, or `attest-blob` with `--new-bundle-format`, but without uploading to a registry and without specifying a `--bundle` output path. This would result in a successful command but the verification material would be silently discarded, leaving nothing to verify against.

Workflows using these commands, with the new bundle format, and with both of the following behaviors, will now fail:
* Not uploading to a registry (i.e., setting `--upload=false` with `sign`, using `--no-upload` with `attest`, or using the `sign-blob` and `attest-blob` commands)
* Not saving the bundle locally (i.e., not providing the `--bundle <path>` flag) 

To fix this, such workflows must do at least one of the following:
* Upload to a registry (enabled by default)
* Save the bundle locally (i.e., provide the `--bundle <path>` flag)

This change applies only to users requesting a new-format bundle.

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -68,6 +68,12 @@ func Attest() *cobra.Command {
 
 		Args:             cobra.MinimumNArgs(1),
 		PersistentPreRun: options.BindViper,
+		PreRunE: func(_ *cobra.Command, _ []string) error {
+			if o.NewBundleFormat && o.NoUpload && o.BundlePath == "" {
+				return fmt.Errorf("must enable upload to the OCI registry or specify a local --bundle path with --new-bundle-format")
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			oidcClientSecret, err := o.OIDC.ClientSecret()
 			if err != nil {
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -147,7 +147,6 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 	}
 
 	if c.NewBundleFormat {
-		// TODO(#4534): Require bundle output or registry upload
 		if c.BundlePath != "" {
 			if err := os.WriteFile(c.BundlePath, bundleBytes, 0600); err != nil {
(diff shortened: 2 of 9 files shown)


cmurphy (commented):
This seems sensible. I think may technically count as a breaking change, although I don't think it warrants waiting until v4. Can you add more detail to the release note to explain the change so that anyone who might have a workflow that falls into this strange loophole knows what exactly is changing and how to fix it?

aaronlew02:
@cmurphy I've updated the release note to include more detail for breaking workflows.

steiza (approved):
While technically this may be a breaking change, the old behavior of discarding the signed material wasn't very useful!

Hayden-IO:
+1, the purpose of this is to avoid doing throwaway work. It is technically a breaking change, but I'd argue it's more of a bug in that you don't actually persist the signature and have no way to look it up later (since we removed online lookups from the new bundle path).

PR #4809, opened 2026-04-13
fix: update dependencies to use new azure sdk components

### LINKED ISSUE #3913: Request to move away from `github.com/chrismellard/docker-credential-acr-env`
**Description**

Cosign currently uses `github.com/chrismellard/docker-credential-acr-env v0.0.0-20230304212654-82a0ddb27589` in [go.mod](https://github.com/sigstore/cosign/blob/main/go.mod#L10).

This package (https://github.com/chrismellard/docker-credential-acr-env) is inactive and has received no new commits since March 2023. The version `v0.0.0-20230304212654-82a0ddb27589` used in cosign has also not been updated since March 2023.

This package brings in some End of Life packages for Azure Authentication, specifically https://pkg.go.dev/github.com/Azure/go-autorest/autorest/adal and https://pkg.go.dev/github.com/Azure/go-autorest/autorest/azure/auth which reached End of Life on 31 March 2023. This means cosign brings in these end of life modules as indirect dependencies: https://github.com/sigstore/cosign/blob/main/go.mod#L86-L87

Running a `go mod why` shows:
```
# github.com/Azure/go-autorest/autorest/adal
github.com/sigstore/cosign/v2/cmd/cosign
github.com/sigstore/cosign/v2/cmd/cosign/cli
github.com/sigstore/cosign/v2/cmd/cosign/cli/options
github.com/chrismellard/docker-credential-acr-env/pkg/credhelper
github.com/Azure/go-autorest/autorest/azure/auth
github.com/Azure/go-autorest/autorest/adal
```

**Suggested Remediation**
I recommend that you remove the use of `github.com/chrismellard/docker-credential-acr-env` as it does not appear to be actively maintained and introduces end of life packages into cosign, and instead use the azure packages directly.

gaganhr94:
Replace the unmaintained `github.com/chrismellard/docker-credential-acr-env` with an
  inline ACR credential helper that uses the modern Azure SDK (`azidentity` +
  `azcontainerregistry`) directly.

  The old library depends on the deprecated `Azure/go-autorest` SDK which is no longer
  maintained. The new implementation uses `azidentity.NewDefaultAzureCredential` to
  obtain an AAD token and exchanges it for an ACR refresh token via the
  `azcontainerregistry` authentication client. It implements `authn.Helper` with
  silent fallthrough for non-ACR registries, matching the existing pattern used by the
  ECR and Alibaba ACR helpers in the `MultiKeychain`.

  This drops the entire `go-autorest` transitive dependency tree
  (`Azure/go-autorest`, `autorest/adal`, `autorest/azure/auth`, `autorest/azure/cli`,
  `autorest/date`, `logger`, `tracing`, `dimchansky/utfbom`, `golang-jwt/jwt/v4`) and
  the legacy `Azure/azure-sdk-for-go v68` indirect dependency.

  Closes #3913

  #### Release Note

  Replaced the unmaintained `docker-credential-acr-env` Azure Container Registry
  credential helper with a built-in implementation using the modern Azure Identity SDK.
  No change in user-facing behavior — ACR authentication continues to work via
  `DefaultAzureCredential` when `--k8s-keychain` is enabled.

  #### Documentation

  NONE — no user-facing configuration or CLI changes. ACR authentication behavior is
  unchanged.


gaganhr94:
@Hayden-IO made changes to use the Azure libraries instead. Please review.

Hayden-IO:
@steiza Do you have a way to verify this? I don't have access to an azure environment.

steiza:
This code is from way before I was a Cosign maintainer or contributor. Can you describe the cosign command / situation where you're running into this?

As far as I can tell, the function you're updating (`GetRegistryClientOpts()`) is only called from cosign commands like `clean`, `attach sbom`, `copy`, `upload wasm`, and `upload blob`. Many of those commands are probably going to be deprecated, we'd recommend you use a general-purpose OCI tool (like [oras](https://oras.land/)) to do those actions instead.

Issue #3794, opened 2024-07-23
Bundle inspection and generation utilities


haydentherapper:
**Description**

Inspired by @codysoyland's https://github.com/codysoyland/sigstore-bundle-upgrade and building on the upcoming bundle support (https://github.com/sigstore/cosign/issues/3139), it would be great to have utilities to work with or create bundles outside of signing and verification workflows. This could include:

```
cosign bundle upgrade --version 0.3 --in-place foo.sigstore.json
cosign bundle synthesize --certificate foo.crt --signature foo.sig --attestation foo.json --out foo.sigstore.json
cosign bundle verify --identity foo foo.sigstore.json
cosign bundle inspect foo.sigstore.json
cosign bundle export --format protobuf --output foo.sigstore.pb foo.sigstore.json
cosign bundle export-certificate --output foo.crt foo.sigstore.json
```

In order:
* Upgrading a bundle version, resolving new requirements
* Merging detached verification material into a bundle
* Verifying that a bundle contains a specified identity (@codysoyland, did you have something else in mind for this? If this is effectively `verify-blob` without artifact verification, I'd skip this one, or at least rename)
* Provide user-readable bundle output (machine readable should just do `cat foo.sigstore.json | jq`)
* Export json bundle as protobuf (and should support proto -> json as well)
* Decompose bundle into detached verification material

Discussion: https://sigstore.slack.com/archives/C0440BFT43H/p1721088432507969

codysoyland:
(re: `cosign bundle verify`)
> @codysoyland, did you have something else in mind for this? If this is effectively verify-blob without artifact verification, I'd skip this one, or at least rename

Yeah, this one is probably redundant with existing verify commands, so it can be omitted.

steiza:
Okay, so merging in the content from #3855, what about if we start with:

```
cosign bundle create

Assembles a protobuf ("new") bundle. Has flags similar to `verify-blob` and
`verify-blob-attestation`. Includes whatever information is not empty
strings.

    --attestations='':
        path to attestation FILE. Conflicts with --signature.

    --bundle='':
	path to old bundle FILE to convert to new bundle

    --certificate='':
	path to the public certificate. The certificate will be verified
	against the Fulcio roots.

    --out='':
	path to bundle FILE to output

    --rfc3161-timestamp='':
	path to RFC3161 timestamp FILE

    --sct='':
	path to a detached Signed Certificate Timestamp, formatted as a
[...]

PR #4820, opened 2026-04-15
Add bundle upgrade command

aaronlew02:
Partially addresses #3794 

#### Summary

This change adds a `bundle upgrade` command which upgrades a Sigstore protobuf bundle to the latest version.

### DIFF
--- a/cmd/cosign/cli/bundle/upgrade.go
+++ b/cmd/cosign/cli/bundle/upgrade.go
@@ -1,0 +1,165 @@
+//
+// Copyright 2026 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundle
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	rekor "github.com/sigstore/rekor/pkg/client"
+	"github.com/sigstore/rekor/pkg/generated/client"
+	"github.com/sigstore/rekor/pkg/generated/client/entries"
+	"github.com/sigstore/rekor/pkg/tle"
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/sigstore/cosign/v3/internal/ui"
+)
+
+type UpgradeCmd struct {
+	In       string
+	Out      string
+	InPlace  string
+	RekorURL string
+}
+
+func (c *UpgradeCmd) Exec(ctx context.Context) error {
+	inputPath := c.In
+	if c.InPlace != "" {
+		inputPath = c.InPlace
+	}
+
+	data, err := os.ReadFile(inputPath)
+	if err != nil {
+		return fmt.Errorf("reading input file: %w", err)
+	}
+
+	if err := detectFormat(data); err != nil {
+		return fmt.Errorf("detecting bundle format: %w", err)
+	}
+
+	rekorClient, err := rekor.GetRekorClient(c.RekorURL)
+	if err != nil {
+		return fmt.Errorf("creating rekor client: %w", err)
+	}
+
+	upgradedBundle, err := upgradeBundle(ctx, data, rekorClient)
+	if err != nil {
+		return fmt.Errorf("upgrading bundle: %w", err)
+	}
+
+	outputPath := c.Out
+	if c.InPlace != "" {
+		outputPath = c.InPlace
+	}
+
+	err = os.WriteFile(outputPath, upgradedBundle, 0600)
+	if err != nil {
+		return fmt.Errorf("writing upgraded bundle: %w", err)
+	}
+
+	ui.Infof(ctx, "Successfully upgraded bundle written to %s", outputPath)
+	return nil
+}
+
+func detectFormat(data []byte) error {
+	var m map[string]interface{}
+	if err := json.Unmarshal(data, &m); err != nil {
+		return fmt.Errorf("unmarshaling JSON for detection: %w", err)
+	}
+
+	if _, hasMediaType := m["mediaType"]; hasMediaType {
+		return nil
+	}
+
+	if _, hasBase64Signature := m["base64Signature"]; hasBase64Signature {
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/options/bundle.go
```diff
@@ -90,3 +90,29 @@ func (o *BundleCreateOptions) AddFlags(cmd *cobra.Command) {
...
+	cmd.Flags().StringVar(&o.In, "in", "", "path to the bundle file to upgrade")
+	_ = cmd.MarkFlagFilename("in", bundleExts...)
+
+	cmd.Flags().StringVar(&o.Out, "out", "", "path to the output upgraded bundle file")
+	_ = cmd.MarkFlagFilename("out", bundleExts...)
+
+	cmd.Flags().StringVar(&o.InPlace, "in-place", "", "path to the bundle file to upgrade in place")
+	_ = cmd.MarkFlagFilename("in-place", bundleExts...)
+
+	cmd.Flags().StringVar(&o.RekorURL, "rekor-url", "https://rekor.sigstore.dev", "address of rekor STL server")
```
Hayden-IO: `stl` isn't a well-known acronym, can you specify transparency log?

cmd/cosign/cli/options/bundle.go
```diff
@@ -90,3 +90,29 @@ func (o *BundleCreateOptions) AddFlags(cmd *cobra.Command) {
...
+var _ Interface = (*BundleUpgradeOptions)(nil)
+
+func (o *BundleUpgradeOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&o.In, "in", "", "path to the bundle file to upgrade")
+	_ = cmd.MarkFlagFilename("in", bundleExts...)
+
+	cmd.Flags().StringVar(&o.Out, "out", "", "path to the output upgraded bundle file")
+	_ = cmd.MarkFlagFilename("out", bundleExts...)
+
+	cmd.Flags().StringVar(&o.InPlace, "in-place", "", "path to the bundle file to upgrade in place")
```
Hayden-IO: One alternative to `--in-place` is to not require `--out`, and when it's not provided, then print to std out. Then you could replicate `in-place` with something like `cosign upgrade --in bundle.json > bundle.json`. Thoughts?

cmd/cosign/cli/bundle/upgrade.go
```diff
@@ -0,0 +1,165 @@
...
+	err = os.WriteFile(outputPath, upgradedBundle, 0600)
+	if err != nil {
+		return fmt.Errorf("writing upgraded bundle: %w", err)
+	}
+
+	ui.Infof(ctx, "Successfully upgraded bundle written to %s", outputPath)
+	return nil
+}
+
+func detectFormat(data []byte) error {
```
Hayden-IO: While the error below is helpful, we can also document in the help text that this command is for the standardized bundle format only and skip the additional parsing.

cmd/cosign/cli/bundle/upgrade.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if _, hasBase64Signature := m["base64Signature"]; hasBase64Signature {
+		return fmt.Errorf("cannot upgrade legacy bundles; use `cosign bundle create` to create a new bundle from your legacy bundle and artifact")
+	}
+
+	return fmt.Errorf("unrecognized bundle format")
+}
+
+func upgradeBundle(ctx context.Context, data []byte, rekorClient *client.Rekor) ([]byte, error) {
+	var bundle protobundle.Bundle
```
Hayden-IO, line 69: Can we use `bundle.LoadJSONFromPath`?

cmd/cosign/cli/bundle/upgrade.go
```diff
@@ -0,0 +1,165 @@
...
+		return data, nil
+	case "application/vnd.dev.sigstore.bundle+json;version=0.1":
+		ui.Infof(ctx, "Upgrading from v0.1 to v0.3...")
+	case "application/vnd.dev.sigstore.bundle+json;version=0.2":
+		ui.Infof(ctx, "Upgrading from v0.2 to v0.3...")
+	default:
+		return nil, fmt.Errorf("unsupported bundle version: %s", bundle.MediaType)
+	}
+
+	if chainContent, ok := bundle.VerificationMaterial.Content.(*protobundle.VerificationMaterial_X509CertificateChain); ok {
```
Hayden-IO, line 93: Can we add a log message that we're truncating the certificate chain and log the certificates' subjects?

cmd/cosign/cli/options/bundle.go
aaronlew02: I've replaced the description with "URL of the transparency log" to [match](https://github.com/sigstore/cosign/blob/39940de028616e038b8262f11315aedc2f81806e/cmd/cosign/cli/options/trustedroot.go#L120) `TrustedRootCreateOptions`. 

(The description "address of rekor STL server" is used in all other instances of `--rekor-url`.)

cmd/cosign/cli/options/bundle.go
aaronlew02: This is a great idea. WDYT about instead removing `--in` and `--out`, having the bundle path be the sole argument, and printing to `stdout`?

cmd/cosign/cli/bundle/upgrade.go
aaronlew02: Good idea. I've removed `detectFormat` and added "This command only supports standardized bundles" to the help text.

cmd/cosign/cli/bundle/upgrade.go
aaronlew02: No, because `LoadJSONFromPath` calls `UnmarshalJSON`, which calls `validate`, which returns an error if a v0.1 bundle does not contain an inclusion promise ([source](https://github.com/sigstore/sigstore-go/blob/952825863bf739f4ec36652c4b38c4564eeb8330/pkg/bundle/bundle.go#L207)).

cmd/cosign/cli/bundle/upgrade.go
aaronlew02: I've added a log message to announce certificate chain truncation and a log for each certificate's subject, including the certificate parsing error if applicable.

cmd/cosign/cli/options/bundle.go
Hayden-IO: To keep uniformity with the other subcommands, let's still have an --out flag. Either would work, I'd just err on the side of keeping flags consistent.

aaronlew02: I've removed the `--in` and `--in-place` flags, making the bundle path the sole argument. I've kept the `--out` flag for file outputs; if not provided, the result prints to `stdout`.

cmd/cosign/cli/bundle/upgrade.go
aaronlew02: I've updated the subject logging to be a logging of the broader identity, with a preference of an email address or URI over the subject and `<none>` as a fallback.

This is especially to support Fulcio certificates, which contain one of the former two but not the latter.

Hayden-IO (approved):
Great!

PR #4821, opened 2026-04-15
fix(layout): preserve manifest artifactType in index.json when saving

### LINKED ISSUE #4694: Manifest's `.config.mediaType` being prefered as "artifact type" in OCI-layout `index.json` over `.artifactType`
**Description**

Saving an OCI-artifact locally with `cosign save --dir` results in an `index.json` in which the `manifests` entry of the artifact's manifest contains the value of `.config.mediaType`, even when `.artifactType` is being provided by that manifest.

The OCI-specification is clear on [how to assign an artifact's type](https://github.com/opencontainers/image-spec/blob/v1.1.1/manifest.md#guidelines-for-artifact-usage) and therefore the sequence to identify it should be:

1. Whatever `.artifactType` provides
2. Whatever `.config.mediaType` provides (only when no `.artifactType` has been provided).

For example, we have an artifact with `artifactType` and an empty config:

``` json
{
  "schemaVersion": 2,
  "artifactType": "application/vnd.my-artifact-type",
  "config": {
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2,
    "data": "e30="
  },
  "layers": [
    {
      "mediaType": "application/octet-stream",
      "digest": "sha256:36c0cd9e4e37bc3db0f9186a0d5ba61f05acafeaba8f1b96f1db02f584fd2cc6",
      "size": 10485760,
      "annotations": {
        "org.opencontainers.image.title": "./random.binary"
      }
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2026-02-12T09:52:32Z"
  }
}
```

Notice the values `artifactType: application/vnd.my-artifact-type` and `config.mediaType: application/vnd.oci.empty.v1+json`.

The resulting `index.json` in the OCI layout looks like this:

``` json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "size": 641,
      "digest": "sha256:5c1b52a23d98e5ef0a00866a0b0a18b79f625110def619fcc8393c48938125da",
      "annotations": {
        "kind": "dev.cosignproject.cosign/image"
      },
      "artifactType": "application/vnd.oci.empty.v1+json"
    }
  ]
}
```

Notice that `artifactType` for the manifest is not the `artifactType` provided by the manifest, but the `config.mediaType` of that manifest!

**Version**

GitVersion:    v3.0.4
GitCommit:     6832fba4928c1ad69400235bbc41212de5006176
GitTreeState:  "clean"
BuildDate:     2026-01-09T21:17:16Z
GoVersion:     go1.25.5
Compiler:      gc
Platform:      darwin/arm64

alliasgher:
## Problem

`cosign save --dir` writes the OCI artifact's `config.mediaType` as the `artifactType` field in `index.json` instead of the manifest-level `artifactType`, even when the manifest explicitly provides one.

**Root cause:** `partial.Descriptor` (go-containerregistry) falls back to `mf.Config.MediaType` when the image does not implement `ArtifactType()`. The current version of go-containerregistry's `v1.Manifest` struct does not expose the manifest-level `artifactType` field, so the correct value is never used.

**OCI spec:** the `manifests` entry in `index.json` [must use `artifactType`](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage) from the manifest when present, not `config.mediaType`.

Fixes #4694

## Fix

In `appendImage`, after obtaining the descriptor from `partial.Descriptor`, parse the raw manifest JSON to extract a top-level `artifactType` and override the descriptor's field when found. The existing `config.mediaType` fallback still applies for manifests without an explicit `artifactType`.

## Test

`TestAppendImagePreservesArtifactType` creates a wrapped image whose `RawManifest()` injects an `artifactType` field, writes it with `WriteSignedImage`, reads back the `index.json`, and asserts the descriptor carries the correct `artifactType`.


steiza (approved):
Sorry for the delay! Can you push an empty change, or rebase these changes off of main, in order to trigger the automated tests to run? Then I think we can approve and land this change.

Generally, it doesn't look like by default Cosign does not set custom artifact types. However, I can see how if your image uses a custom artifact type that we should honor this.

Also, FYI we are working towards getting Cosign out of the business of doing OCI things, and instead recommending people use OCI-specific tooling like [oras](https://oras.land/). But [at least as of 2 weeks ago we aren't deprecated `cosign save` yet](https://github.com/sigstore/cosign/issues/4696#issuecomment-4509951935), so it seems like we should make this update.

alliasgher:
Rebased onto main (e0b7cfa).

PR #4822, opened 2026-04-15
Deprecate Flags for v4: Certificates

piceri:

#### Summary
This PR deprecates the following flags in preparation for v4:

 - certificate
 - ca-intermediates
 - ca-roots
 - certificate-chain
 - timestamp-certificate-chain
 
These flags all relate to certificates that can be replaced using a trusted root.

`certificate` and `certificate-chain` are also used in signing operations. This PR only covers deprecating these flags in the verification subcommands. 

#### Release Note
The following flags were deprecated in validation subcommands:

 - certificate
 - ca-intermediates
 - ca-roots
 - certificate-chain
 - timestamp-certificate-chain


cmd/cosign/cli/verify.go
```diff
@@ -57,16 +57,8 @@ against the transparency log.`,
...
-
-  # verify image with local certificate and certificate bundles of CA roots
-  # and (optionally) CA intermediates
-  cosign verify --cert cosign.crt --ca-roots ca-roots.pem --ca-intermediates ca-intermediates.pem <IMAGE>
-
-  # verify image using keyless verification with the given certificate
-  # chain and identity parameters, without Fulcio roots (for BYO PKI):
-  cosign verify --cert-chain chain.crt --certificate-oidc-issuer https://issuer.example.com --certificate-identity foo@example.com <IMAGE>
+  # verify image with a trusted root
+  cosign verify --trusted-root trusted_root.json <IMAGE>
```
steiza, line 61: Nice!

cmd/cosign/cli/verify.go
```diff
@@ -282,21 +274,18 @@ func VerifyBlob() *cobra.Command {
 		Use:   "verify-blob",
 		Short: "Verify a signature on the supplied blob",
 		Long: `Verify a signature on the supplied blob input using the specified key reference.
-You may specify either a key, a certificate or a kms reference to verify against.
-	If you use a key or a certificate, you must specify the path to them on disk.
+You may specify either a key, a bundle with trusted root, or a kms reference to verify against.
```
steiza:
```suggestion
You may specify either a key, a bundle (optionally with trusted root), or a kms reference to verify against.
```
Minor: if you don't provide a trusted root, we will assume you're using the Sigstore Public Good Instance and use its associated trusted root for you.

cmd/cosign/cli/verify.go
```diff
@@ -282,21 +274,18 @@ func VerifyBlob() *cobra.Command {
...
+	If you use a key, bundle, or trusted root, you must specify the path to them on disk.
 
 The signature may be specified as a path to a file or a base64 encoded string.
 The blob may be specified as a path to a file or - for stdin.`,
-		Example: ` cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--certificate <cert>) --signature <sig> <blob>
+		Example: ` cosign verify-blob (--key <key path>|<key url>|<kms uri>)|(--bundle <bundle> --trusted-root <trusted root>) --signature <sig> <blob>
```
steiza:
```suggestion
		Example: ` cosign verify-blob --bundle <bundle> (--key <key path>|<key url>|<kms uri>) (--trusted-root <trusted root>) <blob>
```
I think for this example we want to suggest people alway use a bundle, even though technically we haven't deprecated `--signature` yet.

Additionally, you could specify both a key and a trusted root - say you signed with a key but then got a signed timestamp from GitHub's (or someone else's) timestamp authority.

cmd/cosign/cli/verify.go
```diff
@@ -282,21 +274,18 @@ func VerifyBlob() *cobra.Command {
...
   # Verify a simple blob and message
   cosign verify-blob --key cosign.pub (--signature <sig path>|<sig url> msg)
 
-  # Verify a signature with certificate and CA certificate chain
-  cosign verify-blob --certificate cert.pem --certificate-chain certchain.pem --signature $sig <blob>
-
-  # Verify a signature with CA roots and optional intermediate certificates
-  cosign verify-blob --certificate cert.pem --ca-roots caroots.pem [--ca-intermediates caintermediates.pem] --signature $sig <blob>
+  # Verify a signature with a bundle and trusted root
+  cosign verify-blob --bundle <bundle> --trusted-root trusted_root.json --signature $sig <blob>
```
steiza:
```suggestion
  cosign verify-blob --bundle <bundle> --trusted-root trusted_root.json <blob>
```
I don't think you would specify a bundle and a signature - those should be mutually exclusive.

steiza (commented):
Generally LGTM! A few minor comments on the subcommand comments.

cmd/cosign/cli/verify.go
piceri: Good comments, thanks!

steiza (approved):
LGTM! Thanks!

PR #4829, opened 2026-04-20
bundle create: Prevent IgnoreTlog when bundle contains SET

aaronlew02:
#### Summary

This change updates the `bundle create` command to prevent the user from enabling `--ignore-tlog` if the provided bundle contains a Signed Entry Timestamp (SET).

The presence of a SET in an old-format bundle indicates that there should be a corresponding Rekor entry, which should be fetched for the creation of a new-format bundle.


Hayden-IO (commented):
Looks good. Can you add a test?

Hayden-IO (approved):

PR #4838, opened 2026-04-24
Deprecate flags bundle

piceri:
#### Summary

This change deprecates the following flags:
| Flag | Deprecated in | Still exists (not deprecated) in |
   |------|---------------|----------------------------------|
   | `--sct` | everywhere | *(none)* |
   | `--signature` | verify, verify-blob, verify-blob-attestation | attach signature, bundle create |
   | `--signature-digest-algorithm` | everywhere | *(none)* |
   | `--output-payload` | everywhere | *(none)* |
   | `--new-bundle-format` | everywhere | *(none)* |

These flags are all now covered by the new bundle format.

I do not think cosign verify supports --bundle yet so I avoided deprecating some flags there for the moment. 

This change also changes the example commands in the verify-* subcommands that used `--signature` to use `--bundle`.

#### Release Note
The following flags have been deprecated:
- sct
- signature
- signature-digest-algorithm
- output-payload
- new-bundle-format


cmd/cosign/cli/options/verify.go
```diff
@@ -192,6 +194,7 @@ func (o *VerifyBlobOptions) AddFlags(cmd *cobra.Command) {
 	o.SecurityKey.AddFlags(cmd)
 	o.Rekor.AddFlags(cmd)
 	o.CertVerify.AddFlags(cmd)
+	_ = cmd.Flags().MarkDeprecated("sct", "please use --bundle to provide Signed Certificate Timestamp")
```
steiza: This should probably be done in cmd/cosign/cli/options/certificate.go to cover all the commands that use this flag, instead of marking it as deprecated on a command-by-command basis?

cmd/cosign/cli/verify.go
```diff
@@ -285,48 +281,34 @@ func VerifyBlob() *cobra.Command {
...
-  cosign verify-blob --key cosign.pub --signature $sig msg
-
-  # verify a signature with public key provided by URL
-  cosign verify-blob --key https://host.for/<FILE> --signature $sig msg
+		Example: ` cosign verify-blob --bundle <bundle path> --certificate-identity <identity> --certificate-oidc-issuer <issuer> <blob>
 
-  # verify a signature with signature and key provided by URL
-  cosign verify-blob --key https://host.for/<FILE> --signature https://example.com/<SIG>
+  # Verify a blob (keyless, Fulcio-issued certificate)
+  cosign verify-blob --bundle artifact.sigstore.json --certificate-identity foo@example.com --certificate-oidc-issuer https://token.actions.githubusercontent.com <blob>
```
steiza:
```suggestion
  cosign verify-blob --bundle artifact.sigstore.json --certificate-identity foo@example.com --certificate-oidc-issuer https://accounts.google.com <blob>
```
Minor: if your identity is an email address, your issuer is likely going to be for human identity and not workload identity (like GitHub Actions).

cmd/cosign/cli/verify.go
```diff
@@ -395,14 +377,29 @@ func VerifyBlobAttestation() *cobra.Command {
 		Use:   "verify-blob-attestation",
 		Short: "Verify an attestation on the supplied blob",
 		Long: `Verify an attestation on the supplied blob input using the specified key reference.
-You may specify either a key or a kms reference to verify against.
+You may specify either a key, a kms reference, or a bundle to verify against.
```
steiza:
```suggestion
You may specify either a key or a kms reference to verify against.
```

Minor: I would put this back to how it was.

Keys and KMS references are verification material - like a trusted root. They are separate from signed material, like a Sigstore protobuf bundle. Put another way, you can verify a bundle with a key and/or trusted root.

cmd/cosign/cli/verify.go
```diff
@@ -395,14 +377,29 @@ func VerifyBlobAttestation() *cobra.Command {
...
 
-The signature may be specified as a path to a file or a base64 encoded string.
+The --bundle flag is the preferred way to provide the attestation and verification material.
```
steiza:
```suggestion
Signed material is provided with the --bundle flag.
```

Words are tricky, but to me the verification material is the trusted root, not the protobuf bundle.

cmd/cosign/cli/verify.go
```diff
@@ -395,14 +377,29 @@ func VerifyBlobAttestation() *cobra.Command {
...
 The blob may be specified as a path to a file.`,
-		Example: ` cosign verify-blob-attestation (--key <key path>|<key url>|<kms uri>) --signature <sig> [path to BLOB]
+		Example: ` cosign verify-blob-attestation --bundle <path> --certificate-identity <identity> --certificate-oidc-issuer <issuer> [path to BLOB]
+
+  # Verify a blob attestation (keyless)
+  cosign verify-blob-attestation --bundle artifact.sigstore.json --certificate-identity foo@example.com --certificate-oidc-issuer https://token.actions.githubusercontent.com <blob>
```
steiza: Same as before - we should either use a workload identity or change the issuer to be one that issues human identity.

steiza (commented):
Nice work! Every time I sit down to start to review one of these, I think "we're just deprecating some flags, how hard can it be?" :lolsob:

I have several minor suggestions on documentation wording and about how we're marking the `--sct` flag for deprecation.

cmd/cosign/cli/verify.go
```diff
@@ -397,12 +379,27 @@ func VerifyBlobAttestation() *cobra.Command {
...
 You may specify either a key or a kms reference to verify against.
 
-The signature may be specified as a path to a file or a base64 encoded string.
+Signed material is provided with the --bundle flag.
 The blob may be specified as a path to a file.`,
-		Example: ` cosign verify-blob-attestation (--key <key path>|<key url>|<kms uri>) --signature <sig> [path to BLOB]
+		Example: ` cosign verify-blob-attestation --bundle <path> --certificate-identity <identity> --certificate-oidc-issuer <issuer> [path to BLOB]
+
+  # Verify a blob attestation (keyless)
+  cosign verify-blob-attestation --bundle artifact.sigstore.json --certificate-identity foo@example.com --certificate-oidc-issuer https://token.actions.githubusercontent.com <blob>
```
steiza:
```suggestion
  cosign verify-blob-attestation --bundle artifact.sigstore.json --certificate-identity foo@example.com --certificate-oidc-issuer https://accounts.google.com <blob>
```

steiza (commented):
Since I landed #4844 this will have to be rebased - and I have one other minor comment, otherwise LGTM!

piceri:
> Nice work! Every time I sit down to start to review one of these, I think "we're just deprecating some flags, how hard can it be?" :lolsob:

Same, I keep thinking they will be quick but find myself second guessing working and learning new differences between commands. I think the rebase should be done now and I added one change try and unify the examples for verify-blob and verify-blob-attestation a little more.

steiza (approved):

PR #4841, opened 2026-04-28
Add OCI 1.1 referrer discovery for attestation verification

### LINKED ISSUE #4708: VerifyImageAttestations has no OCI 1.1 referrer discovery path

saschagrunert:
Closes #4708

The `--experimental-oci11` flag was parsed by the CLI but never propagated to `VerifyImageAttestations`, making it dead code for the attestation path.

This adds `verifyImageAttestationsExperimentalOCI` (analogous to `verifyImageSignaturesExperimentalOCI`) which queries the Referrers API using `application/vnd.in-toto+json` as the artifact type (matching `WriteAttestationsReferrer`), then verifies each result through `VerifyImageAttestation`. Unlike the signature variant, this iterates all referrer results since multiple attestations per image are expected.

A full e2e round-trip test is not possible yet because `cosign attest` has no `--registry-referrers-mode` option. Adding write-side support would be a natural follow-up.

Prerequisite for CRI-O's container image attestation verification (https://github.com/cri-o/cri-o/issues/9766).

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1029,11 +1029,21 @@ func loadSignatureFromFile(ctx context.Context, sigRef string, signedImgRef name
 
 // VerifyImageAttestations does all the main cosign checks in a loop, returning the verified attestations.
 // If there were no valid attestations, we return an error.
+// If co.ExperimentalOCI11 is set, attestations are discovered via the OCI 1.1 Referrers API.
 func VerifyImageAttestations(ctx context.Context, signedImgRef name.Reference, co *CheckOpts, nameOpts ...name.Option) (checkedAttestations []oci.Signature, bundleVerified bool, err error) {
 	// Enforce this up front.
 	if co.RootCerts == nil && co.SigVerifier == nil && co.TrustedMaterial == nil {
 		return nil, false, errors.New("one of verifier, root certs, or TrustedMaterial is required")
 	}
+
+	// Try first using OCI 1.1 behavior if experimental flag is set.
+	if co.ExperimentalOCI11 {
+		verified, bundleVerified, err := verifyImageAttestationsExperimentalOCI(ctx, signedImgRef, co)
+		if err == nil {
+			return verified, bundleVerified, nil
+		}
+	}
+
 	if co.NewBundleFormat {
 		return verifyImageAttestationsSigstoreBundle(ctx, signedImgRef, co, nameOpts...)
 	}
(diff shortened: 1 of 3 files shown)


saschagrunert:
@Hayden-IO @steiza @cmurphy @cpanato @bobcallaway - would appreciate a review when you get a chance. CI probably needs a retest.

pkg/cosign/verify.go
```diff
@@ -1667,6 +1677,55 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+	artifactType := types.IntotoPayloadType
+	index, err := ociremote.Referrers(digest, artifactType, co.RegistryClientOpts...)
+	if err != nil {
+		return nil, false, err
+	}
+	results := index.Manifests
+	if len(results) == 0 {
+		return nil, false, fmt.Errorf("unable to locate reference with artifactType %s", artifactType)
+	} else if len(results) > 1 {
+		ui.Warnf(ctx, "there were a total of %d references with artifactType %s\n", len(results), artifactType)
```
cmurphy, line 1700: You said multiple attestations were expected, is this warning necessary?

pkg/cosign/verify.go
```diff
@@ -1667,6 +1677,55 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
...
+		if err != nil {
+			return nil, false, err
+		}
+		atts, err := ociremote.Signatures(st, co.RegistryClientOpts...)
+		if err != nil {
+			return nil, false, err
+		}
+		verified, bv, err := VerifyImageAttestation(ctx, atts, h, co)
+		if err != nil {
+			continue
```
cmurphy, line 1716: Should these errors be accumulated and reported with `errors.Join` instead of ignored?

pkg/cosign/verify_oci_test.go
```diff
@@ -236,3 +243,97 @@ func TestVerifyImageAttestationsSigstoreBundle(t *testing.T) {
 	assert.False(t, bundleVerified)
 	assert.Len(t, atts, 0)
 }
+
```
cmurphy, line 246: I found these tests all passed for me even when I reverted the other code changes, so I'm not sure they're properly exercising your changes.

pkg/cosign/verify.go
```diff
@@ -1029,11 +1029,21 @@ func loadSignatureFromFile(ctx context.Context, sigRef string, signedImgRef name
...
+	// Try first using OCI 1.1 behavior if experimental flag is set.
+	if co.ExperimentalOCI11 {
+		verified, bundleVerified, err := verifyImageAttestationsExperimentalOCI(ctx, signedImgRef, co)
+		if err == nil {
+			return verified, bundleVerified, nil
+		}
+	}
+
 	if co.NewBundleFormat {
 		return verifyImageAttestationsSigstoreBundle(ctx, signedImgRef, co, nameOpts...)
```
cmurphy, line 1048: This, I think, should already be using the OCI 1.1 behavior as long as you are using the new bundle format, which we strongly encourage. Is that behavior not working correctly, or is there a reason you're not able to transition to the new bundle format?

cmurphy (commented):
Good to see you, @saschagrunert !

saschagrunert:
Thanks for the review @cmurphy! You're right that the `NewBundleFormat` path already handles OCI 1.1 referrer discovery for attestations via `GetBundles()`. Since the ecosystem is converging on sigstore bundles as the standard format, this separate `ExperimentalOCI11` path for attestations isn't needed. Closing in favor of using `NewBundleFormat`.

PR #4848, opened 2026-05-04
fix(verify): return error instead of panicking when no timestamps verify

### LINKED ISSUE #4846: VerifyRFC3161Timestamp panics on verifiedTimestamps[0] when threshold is not met
**Description**

`VerifyRFC3161Timestamp` panics with `runtime error: index out of range [0] with length 0` when
the `TrustedMaterial` branch's call to `verify.VerifySignedTimestamp` returns zero verified
timestamps. The function logs the per-attempt errors as a warning but never checks the result
length before indexing `verifiedTimestamps[0].Time`.

### Reproduction

Trigger condition: call `VerifyRFC3161Timestamp` with a `CheckOpts` whose `TrustedMaterial`
is non-nil but whose `TimestampingAuthorities()` doesn't include a TSA matching the timestamp
being verified. `verify.VerifySignedTimestamp` returns `(emptySlice, errs, nil)` — no
top-level error, just per-attempt errors. The function falls through to:

```go
return &timestamp.Timestamp{Time: verifiedTimestamps[0].Time}, nil
```

…which panics.

Stack trace:

```
2026/05/04 11:29:32 Warning: subset of signed timestamps failed to verify: [unable to verify signed timestamps: %!w(<nil>)]
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/sigstore/cosign/v3/pkg/cosign.VerifyRFC3161Timestamp({...}, ...)
	github.com/sigstore/cosign/v3@v3.0.6/pkg/cosign/verify.go:1406 +0x5b6
```

Note the warning log shows `%!w(<nil>)` — the underlying error is wrapped without a value, which
is itself worth a separate look but isn't the cause of the panic.

### Root cause

[`pkg/cosign/verify.go:1399-1406`](https://github.com/sigstore/cosign/blob/v3.0.6/pkg/cosign/verify.go#L1399-L1406):

```go
verifiedTimestamps, verifyErrs, err := verify.VerifySignedTimestamp(entity, co.TrustedMaterial)
if err != nil {
    return nil, fmt.Errorf("unable to verify signed timestamps with trusted root: %w", err)
}
if len(verifyErrs) > 0 {
    log.Printf("Warning: subset of signed timestamps failed to verify: %v", verifyErrs)
}
return &timestamp.Timestamp{Time: verifiedTimestamps[0].Time}, nil
```

`verify.VerifySignedTimestamp` is documented to return verified timestamps + per-attempt errors,
where an empty result with non-empty errors means "all attempts failed" rather than "function
errored". This handler treats `verifyErrs` as an informational warning rather than a verification
failure.

### Proposed fix

```go
verifiedTimestamps, verifyErrs, err := verify.VerifySignedTimestamp(entity, co.TrustedMaterial)
if err != nil {
    return nil, fmt.Errorf("unable to verify signed timestamps with trusted root: %w", err)
}
if len(verifiedTimestamps) == 0 {
    return nil, fmt.Errorf("no signed timestamps verified: %v", verifyErrs)
}
if len(verifyErrs) > 0 {
    log.Printf("Warning: subset of signed timestamps failed to verify: %v", verifyErrs)
}
return &timestamp.Timestamp{Time: verifiedTimestamps[0].Time}, nil
```

**Version**

- `cosign` v3.0.6
- `sigstore-go` v1.1.4
- `timestamp-authority/v2` v2.0.6

jimgus:
Closes #4846

#### Summary

`VerifyRFC3161Timestamp`'s `TrustedMaterial` branch indexes `verifiedTimestamps[0].Time`
after only logging `verifyErrs` as a warning. When `verify.VerifySignedTimestamp` returns an
empty slice with non-empty `verifyErrs` (i.e. all attempts failed but no top-level error), the
function panics with `runtime error: index out of range [0] with length 0` instead of
returning an error to the caller.

Three-line fix: guard on `len(verifiedTimestamps) == 0` and return an error wrapping
`verifyErrs`. Reviewers can verify by constructing a `CheckOpts` whose `TrustedMaterial`
contains no `TimestampingAuthority` matching the timestamp under verification — the same
shape that triggers the panic in the linked issue.

`go test ./pkg/cosign/...` passes with no regressions.

#### Release Note

```release-note
Fix a panic in `VerifyRFC3161Timestamp` that occurred when zero signed timestamps could be
verified against the provided `TrustedMaterial`. The function now returns an error wrapping
the per-attempt verification errors instead of indexing an empty slice.
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1400,6 +1400,9 @@ func VerifyRFC3161Timestamp(sig oci.Signature, co *CheckOpts) (*timestamp.Timest
 		if err != nil {
 			return nil, fmt.Errorf("unable to verify signed timestamps with trusted root: %w", err)
 		}
+		if len(verifiedTimestamps) == 0 {
+			return nil, fmt.Errorf("no signed timestamps verified: %v", verifyErrs)
+		}
 		if len(verifyErrs) > 0 {
 			log.Printf("Warning: subset of signed timestamps failed to verify: %v", verifyErrs)
 		}
--- a/pkg/cosign/verify_test.go
+++ b/pkg/cosign/verify_test.go
@@ -1843,6 +1843,55 @@ func TestVerifyRFC3161Timestamp(t *testing.T) {
 	}
 }
 
+// emptyTSATrustedMaterial is a TrustedMaterial whose TimestampingAuthorities()
+// returns an empty slice. Used to exercise the path where
+// verify.VerifySignedTimestamp returns (emptySlice, verifyErrs, nil) — i.e.
+// no top-level error, but no timestamps verified either.
+type emptyTSATrustedMaterial struct {
+	root.BaseTrustedMaterial
+}
+
+// Regression test for the panic where VerifyRFC3161Timestamp's TrustedMaterial
+// branch indexed verifiedTimestamps[0] without checking len(). When no
+// TimestampingAuthority in TrustedMaterial matches the timestamp under
+// verification, sigstore-go returns an empty slice with non-empty per-attempt
+// errors and a nil top-level error. The function should return that as a
+// regular verification error, not panic.
+func TestVerifyRFC3161Timestamp_NoMatchingTSAInTrustedMaterial(t *testing.T) {
+	rootCert, rootKey, _ := test.GenerateRootCa()
+	leafCert, privKey, _ := test.GenerateLeafCert("subject", "oidc-issuer", rootCert, rootKey)
+	pemRoot := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: rootCert.Raw})
+	pemLeaf := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: leafCert.Raw})
+	payload := []byte{1, 2, 3, 4}
+	h := sha256.Sum256(payload)
+	signature, _ := privKey.Sign(rand.Reader, h[:], crypto.SHA256)
+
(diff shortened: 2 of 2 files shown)


steiza (commented):
This looks good to me!

Do note that this codepath is using Cosign's "old" signature format where the signed timestamp was separate from the signature. We are migrating to using the Sigstore protobuf bundle format, and verification of that format does not use this codepath.

Also, you'll need to rebase these changes off of `main` to pick up the `e2e-kms` fix that landed in https://github.com/sigstore/cosign/pull/4824 (sorry about that!)

pkg/cosign/verify.go
```diff
@@ -1400,6 +1400,9 @@ func VerifyRFC3161Timestamp(sig oci.Signature, co *CheckOpts) (*timestamp.Timest
 		if err != nil {
 			return nil, fmt.Errorf("unable to verify signed timestamps with trusted root: %w", err)
 		}
+		if len(verifiedTimestamps) == 0 {
+			return nil, fmt.Errorf("no signed timestamps verified: %v", verifyErrs)
+		}
 		if len(verifyErrs) > 0 {
 			log.Printf("Warning: subset of signed timestamps failed to verify: %v", verifyErrs)
```
cmurphy, line 1407: Couldn't we just return here instead of adding a second conditional?

jimgus: Collapsing into the verifyErrs > 0 branch would turn partial failures (some timestamps verified, others didn't) from a warning into a hard error. Happy to change if you think that is a better solution.

steiza:
Thanks for this contribution - we ended up landing this change in #4881

PR #4849, opened 2026-05-04
fix(verify): honour user-provided TSA chain when TrustedMaterial is set

### LINKED ISSUE #4847: VerifyRFC3161Timestamp silently drops user-provided TSA chain when TrustedMaterial is set (regression from v2)
**Description**

`VerifyRFC3161Timestamp` in `pkg/cosign/verify.go` silently drops the user-provided TSA chain
(`co.TSACertificate`, `co.TSAIntermediateCertificates`, `co.TSARootCertificates`) whenever
`co.TrustedMaterial` is non-nil. This is a regression from cosign v2, where the same call always
honoured the explicit TSA fields. As a result, callers who set `TrustedMaterial` for Fulcio/Rekor
trust *and* an explicit TSA chain for an out-of-tree TSA (e.g. GitHub's private TSA used by
`actions/attest-build-provenance` on private repositories) get verification failures — and in
some cases a runtime panic — with no indication that their TSA chain was ignored.

### Background

GitHub Actions Artifact Attestations on private repositories use:

- GitHub's private TUF root (`https://tuf-repo.github.com`) for trust material
- GitHub's private TSA — no public Rekor entry by design
- The Sigstore Bundle format (OCI 1.1 referrer, predicate `slsa.dev/provenance/v1`)

To verify these in a third-party admission controller, callers populate `cosign.CheckOpts` with:

- `TrustedMaterial` from the public sigstore TUF root (for Fulcio CA + Rekor pubkeys)
- `TSACertificate` / `TSAIntermediateCertificates` / `TSARootCertificates` from GitHub's TSA
  cert chain — GitHub's private TSA isn't in any public sigstore TUF root, so the legacy fields
  are the only place to put it

This is exactly how Kyverno's `ImageValidatingPolicy` (IVPOL) builds CheckOpts; see
[kyverno/kyverno pkg/image/verifiers/ivpol/cosign/opts.go:112-148](https://github.com/kyverno/kyverno/blob/v1.18.0/pkg/image/verifiers/ivpol/cosign/opts.go#L112-L148).

### The bug

[`pkg/cosign/verify.go:1394-1407`](https://github.com/sigstore/cosign/blob/v3.0.6/pkg/cosign/verify.go#L1394-L1407):

```go
if co.TrustedMaterial != nil {
    entity := &signedEntityForTimestamp{
        timestamp:  ts,
        sigContent: &sigContent{rawSig: tsBytes},
    }
    verifiedTimestamps, verifyErrs, err := verify.VerifySignedTimestamp(entity, co.TrustedMaterial)
    if err != nil {
        return nil, fmt.Errorf("unable to verify signed timestamps with trusted root: %w", err)
    }
    if len(verifyErrs) > 0 {
        log.Printf("Warning: subset of signed timestamps failed to verify: %v", verifyErrs)
    }
    return &timestamp.Timestamp{Time: verifiedTimestamps[0].Time}, nil
}

return tsaverification.VerifyTimestampResponse(ts.SignedRFC3161Timestamp, bytes.NewReader(tsBytes),
    tsaverification.VerifyOpts{
        TSACertificate: co.TSACertificate,
        Intermediates:  co.TSAIntermediateCertificates,
        Roots:          co.TSARootCertificates,
    })
```

The `if co.TrustedMaterial != nil` branch never consults `co.TSACertificate` / `TSARootCertificates` /
`TSAIntermediateCertificates`. If `TrustedMaterial.TimestampingAuthorities()` doesn't contain a TSA
matching the timestamp being verified, the verification silently fails — even though the caller
provided exactly the right TSA chain in the legacy fields.

### v2 / v3 comparison

In [cosign v2.5.0 `VerifyRFC3161Timestamp`](https://github.com/sigstore/cosign/blob/v2.5.0/pkg/cosign/verify.go#L1227-L1267),
this branch did not exist — the function unconditionally used the explicit TSA fields:

```go
return tsaverification.VerifyTimestampResponse(ts.SignedRFC3161Timestamp, bytes.NewReader(tsBytes),
    tsaverification.VerifyOpts{
        TSACertificate: co.TSACertificate,
        Intermediates:  co.TSAIntermediateCertificates,
        Roots:          co.TSARootCertificates,
    })
```

The v3 branch was introduced in [`32a2d62a` "Upgrade to TUF v2 client with trusted root"](https://github.com/sigstore/cosign/commit/32a2d62a).
The commit message states the design intent:

> Where possible, use sigstore-go's verifiers which natively accept the trusted root as its trusted material.

The case where a caller sets *both* `TrustedMaterial` (for keyless trust) **and** the legacy TSA
fields (for an out-of-tree TSA) doesn't appear to have been considered.

### Reproduction

Self-contained Go program (90 lines including helper). It calls `cosign.VerifyRFC3161Timestamp`
twice with identical inputs except for one bit: `co.TrustedMaterial` set vs. nil. The toggle
between fail and pass demonstrates the bug.

```go
package main

import (
    "bytes"
    "crypto/x509"
    "encoding/pem"
    "fmt"
    "log"
    "os"

    "github.com/sigstore/cosign/v3/pkg/cosign"
    cosignbundle "github.com/sigstore/cosign/v3/pkg/cosign/bundle"
    "github.com/sigstore/cosign/v3/pkg/oci/static"
    "github.com/sigstore/sigstore-go/pkg/root"
)

// emptyButNonNilTrustedMaterial mimics the state a caller ends up with after
// loading a trusted root that doesn't happen to contain the TSA they need
// (for example: public sigstore TUF root used to verify a GitHub-TSA-signed
// attestation).
type emptyButNonNilTrustedMaterial struct {
    root.BaseTrustedMaterial
}

func main() {
    if len(os.Args) != 4 {
        log.Fatalf("usage: %s <tsr.bin> <signed-payload.bin> <tsa-chain.pem>", os.Args[0])
    }
    tsr, err := os.ReadFile(os.Args[1])
    if err != nil { log.Fatal(err) }
    sigBytes, err := os.ReadFile(os.Args[2])
    if err != nil { log.Fatal(err) }
    chainPEM, err := os.ReadFile(os.Args[3])
    if err != nil { log.Fatal(err) }

    leaf, intermediates, roots := splitChain(chainPEM)
    rfc3161 := &cosignbundle.RFC3161Timestamp{SignedRFC3161Timestamp: tsr}
    sig, err := static.NewAttestation(sigBytes, static.WithRFC3161Timestamp(rfc3161))
    if err != nil { log.Fatal(err) }

    co := &cosign.CheckOpts{
        TrustedMaterial:             &emptyButNonNilTrustedMaterial{},
        TSACertificate:              leaf,
        TSAIntermediateCertificates: intermediates,
        TSARootCertificates:         roots,
        UseSignedTimestamps:         true,
    }

    fmt.Println("Case 1: TrustedMaterial != nil + TSA chain populated")
    if _, err := cosign.VerifyRFC3161Timestamp(sig, co); err != nil {
        fmt.Printf("  ✗ %v\n", err)
    } else {
        fmt.Println("  ✓ verified")
    }

    fmt.Println("Case 2: TrustedMaterial == nil + same TSA chain")
    co.TrustedMaterial = nil
    if _, err := cosign.VerifyRFC3161Timestamp(sig, co); err != nil {
        fmt.Printf("  ✗ %v\n", err)
    } else {
        fmt.Println("  ✓ verified")
    }
}

func splitChain(b []byte) (leaf *x509.Certificate, ints, roots []*x509.Certificate) {
    rest := b
    for {
        block, r := pem.Decode(rest)
        if block == nil { break }
        rest = r
        c, err := x509.ParseCertificate(block.Bytes)
        if err != nil { log.Fatal(err) }
        switch {
        case !c.IsCA:
            leaf = c
        case bytes.Equal(c.RawIssuer, c.RawSubject):
            roots = append(roots, c)
        default:
            ints = append(ints, c)
        }
    }
    return
}
```

Run against any image attested via `actions/attest-build-provenance` plus the matching TSA chain
(e.g. GitHub's TSA chain from `https://tuf-repo.github.com/.../trusted_root.json`):

```
Case 1: TrustedMaterial != nil + TSA chain populated
  ✗ runtime error: index out of range [0] with length 0   <-- separate panic bug, see #4846
Case 2: TrustedMaterial == nil + same TSA chain
  ✓ verified
```

The single bit `co.TrustedMaterial == nil` decides whether the caller's TSA chain matters. Inputs
otherwise identical. (Case 1 currently surfaces as a panic from the same function — filed
separately as #4846 because it's an independent issue. With that panic fixed but nothing else
changed, Case 1 returns a verification error instead of panicking; the chain is still ignored.)

### Real-world impact

Anyone migrating an admission controller from `sigstore/cosign/v2` to `v3` while verifying
GitHub Actions Artifact Attestations on private repositories will hit this:

- A `cosign.CheckOpts` shape that worked on v2 silently regresses on v3
- `sigstore/policy-controller` (still on cosign v2) avoids the bug structurally because it
  mutually-excludes `TrustedMaterial` and the legacy TSA fields, putting custom TSAs into a
  `TrustRoot` CR. Anyone whose admission controller mixes the two patterns (Kyverno IVPOL is
  the case I encountered) hits it.

### Proposed fix

I have two candidate shapes; happy to PR whichever you prefer.

**Option A (smaller, preserves v2 contract):** prefer the legacy TSA fields when explicitly set —
treat them as a directive that the caller wants to verify against a specific TSA, regardless of
whether `TrustedMaterial` happens to also contain a TSA:

```go
if co.TSARootCertificates != nil {
    return tsaverification.VerifyTimestampResponse(...)  // honour caller's chain
}
if co.TrustedMaterial != nil {
    // existing branch, unchanged
}
```

**Option B (more idiomatic for the v3 sigstore-go path):** when both are set, merge the legacy
fields into `TrustedMaterial` as a `SigstoreTimestampingAuthority` and combine via
`TrustedMaterialCollection`, then use the existing branch.

Option A is what I implemented and verified locally — the fix passes `go test ./pkg/cosign/...`
with no regressions. Option B is closer to the v3 architectural direction but needs a small
wrapper type. Happy to PR either.

**Version**

- `cosign` v3.0.6
- `sigstore-go` v1.1.4
- `timestamp-authority/v2` v2.0.6

jimgus:
Closes #4847

#### Summary

`VerifyRFC3161Timestamp` silently dropped `co.TSACertificate` /
`co.TSAIntermediateCertificates` / `co.TSARootCertificates` whenever `co.TrustedMaterial`
was non-nil — a regression from cosign v2 that breaks the use case where a caller legitimately
wants to combine `TrustedMaterial` (Fulcio CA / Rekor pubkeys from a public sigstore TUF root)
with a TSA chain for an out-of-tree TSA not present in any TUF root (e.g. GitHub's private TSA
used by `actions/attest-build-provenance` on private repositories). See #4847 for the full
diagnosis, including reproduction toggling between fail and pass on the single bit
`co.TrustedMaterial == nil`.

This PR implements **Option A** from the issue: when `co.TSARootCertificates` is set, treat
it as an explicit directive and use the legacy `tsaverification.VerifyTimestampResponse` path,
even when `TrustedMaterial` is also set. The `TrustedMaterial` branch is unchanged for
callers who don't set the legacy fields. Net effect: any v2 CheckOpts shape that previously
worked continues to work; new CheckOpts that carry both fields now work as the caller
clearly intends.

Option B in the issue (merge the legacy fields into `TrustedMaterial` as a
`SigstoreTimestampingAuthority`) is also viable; happy to switch to that shape if you'd
prefer the v3-architectural direction. Option A picked here because it's strictly additive
(16 lines, no deletions, no new types) and easier to backport.

Reviewers can verify with the reproduction program in #4847: it should output
`✓ verified` for both Case 1 and Case 2 with this patch applied. `go test ./pkg/cosign/...`
passes with no regressions.

#### Release Note

```release-note
Fix `VerifyRFC3161Timestamp` silently dropping the user-provided TSA chain
(`TSACertificate` / `TSAIntermediateCertificates` / `TSARootCertificates`) when
`TrustedMaterial` is also set on `CheckOpts`. This restores the v2 contract for callers
who combine a public sigstore trusted root with a TSA chain for an out-of-tree timestamp
authority (e.g. GitHub's private TSA used by Artifact Attestations on private
repositories).
```

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1391,6 +1391,17 @@ func VerifyRFC3161Timestamp(sig oci.Signature, co *CheckOpts) (*timestamp.Timest
 		tsBytes = rawSig
 	}
 
+	// Honour the explicit TSA chain when set — TrustedMaterial may not
+	// contain the TSA used to sign this timestamp.
+	if co.TSARootCertificates != nil {
+		return tsaverification.VerifyTimestampResponse(ts.SignedRFC3161Timestamp, bytes.NewReader(tsBytes),
+			tsaverification.VerifyOpts{
+				TSACertificate: co.TSACertificate,
+				Intermediates:  co.TSAIntermediateCertificates,
+				Roots:          co.TSARootCertificates,
+			})
+	}
+
 	if co.TrustedMaterial != nil {
 		entity := &signedEntityForTimestamp{
 			timestamp:  ts,
--- a/pkg/cosign/verify_test.go
+++ b/pkg/cosign/verify_test.go
@@ -1843,6 +1843,65 @@ func TestVerifyRFC3161Timestamp(t *testing.T) {
 	}
 }
 
+// TrustedMaterial with no TSAs — exercises the case where the caller's
+// TSA chain is the only place the signing TSA is configured.
+type noMatchingTSAsMaterial struct {
+	root.BaseTrustedMaterial
+}
+
+func TestVerifyRFC3161Timestamp_LegacyChainHonouredWithTrustedMaterial(t *testing.T) {
+	rootCert, rootKey, _ := test.GenerateRootCa()
+	leafCert, privKey, _ := test.GenerateLeafCert("subject", "oidc-issuer", rootCert, rootKey)
+	pemRoot := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: rootCert.Raw})
+	pemLeaf := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: leafCert.Raw})
+	payload := []byte{1, 2, 3, 4}
+	h := sha256.Sum256(payload)
+	signature, _ := privKey.Sign(rand.Reader, h[:], crypto.SHA256)
+
+	client, err := tsaMock.NewTSAClient(tsaMock.TSAClientOptions{Time: time.Now()})
+	if err != nil {
+		t.Fatal(err)
+	}
+	tsBytes, err := tsa.GetTimestampedSignature(signature, client)
+	if err != nil {
+		t.Fatalf("unexpected error creating timestamp: %v", err)
+	}
+	rfc3161TS := bundle.RFC3161Timestamp{SignedRFC3161Timestamp: tsBytes}
+
+	certChainPEM, err := cryptoutils.MarshalCertificatesToPEM(client.CertChain)
+	if err != nil {
+		t.Fatalf("unexpected error marshalling cert chain: %v", err)
+	}
+	leaves, intermediates, roots, err := tsa.SplitPEMCertificateChain(certChainPEM)
+	if err != nil {
+		t.Fatal("error splitting response into certificate chain")
+	}
+
+	ociSig, _ := static.NewSignature(payload,
+		base64.StdEncoding.EncodeToString(signature),
+		static.WithCertChain(pemLeaf, appendSlices([][]byte{pemRoot})),
+		static.WithRFC3161Timestamp(&rfc3161TS))
+
+	// Both TrustedMaterial AND the legacy TSA fields are set. TrustedMaterial
+	// has no matching TSA; the legacy fields contain the chain that actually
+	// signed the timestamp. Verification must use the legacy chain.
+	ts, err := VerifyRFC3161Timestamp(ociSig, &CheckOpts{
+		TrustedMaterial:             &noMatchingTSAsMaterial{},
+		TSACertificate:              leaves[0],
+		TSAIntermediateCertificates: intermediates,
+		TSARootCertificates:         roots,
+	})
+	if err != nil {
+		t.Fatalf("expected verification to succeed when legacy TSA chain is set alongside TrustedMaterial, got: %v", err)
+	}
+	if ts == nil {
+		t.Fatalf("expected non-nil verified timestamp")
+	}
+	if err := CheckExpiry(leafCert, nil, ts.Time); err != nil {
+		t.Fatalf("unexpected error using time from timestamp to verify certificate: %v", err)
+	}
+}
+
 // This test verifies that artifact verification rejects signatures
 // where a CA certificate in the issuing chain is expired. This is a contrived
 // example because CAs shouldn't issue certificates where the leaf's validity


jimgus:
@steiza — flagging a question that affects whether this PR shape is the right one, since on #4848 you noted that bundle-format verification doesn't use the legacy signed-timestamp codepath.

The reproducer that surfaced both bugs (`jimgus/kyverno-tsa-debug`) runs Kyverno IVPOL with a `ClusterImagePolicy` declaring `signatureFormat: bundle`, and *does* land in `VerifyRFC3161Timestamp` here. Kyverno IVPOL builds `CheckOpts` with both `TrustedMaterial` (TUF root, no GitHub TSA) *and* the user-provided `TSACertificate / TSARootCertificates / TSAIntermediateCertificates` (`tsaCertChain` from the policy), and verification goes through `verify.go:1394`.

So either:

- (a) bundle-format does still reach `VerifyRFC3161Timestamp` when `TSACertificate` is non-nil, in which case Option A in this PR (prefer the legacy fields when explicitly set) is the right fix; or
- (b) Kyverno IVPOL's invocation is effectively running the legacy path despite the `signatureFormat: bundle` declaration, and the right answer upstream is something else (e.g. push IVPOL toward a bundle-native verifier and let this codepath continue to ignore TSA fields when `TrustedMaterial` is set).

Which way would you frame it? Happy to take this in either direction; I just want to make sure the right code is being patched if bundle is what users actually intend.

pkg/cosign/verify.go
```diff
@@ -1391,6 +1391,17 @@ func VerifyRFC3161Timestamp(sig oci.Signature, co *CheckOpts) (*timestamp.Timest
 		tsBytes = rawSig
 	}
 
+	// Honour the explicit TSA chain when set — TrustedMaterial may not
+	// contain the TSA used to sign this timestamp.
```
cmurphy, line 1395: This does seem like a regression but I'm curious why you wouldn't just put the TSA cert in your own trusted_root.json if you're using a trusted root?

steiza: I really want to discourage people from calling this function from outside the Cosign codebase.

Like I mentioned on https://github.com/sigstore/cosign/issues/4847#issuecomment-4399571885, you could call `VerifyNewBundle()`, but you're probably better off calling sigstore-go directly.

When we use this function in Cosign (like in https://github.com/sigstore/cosign/blob/f15ac759ffe73b774b6ab521cd181d57d58a8bfd/cmd/cosign/cli/verify/verify_blob.go#L287) we've already passed the point where we handle new bundles, and we've already checked that there [isn't a trusted root](https://github.com/sigstore/cosign/blob/f15ac759ffe73b774b6ab521cd181d57d58a8bfd/cmd/cosign/cli/verify/verify_blob.go#L172-L174).

These functions simply were not designed to be called directly, and as soon as we finish https://github.com/sigstore/cosign/issues/4696 we're going to be deleting these functions (but not delete things like `VerifyNewBundle()`, of course).

jimgus:
I will use `sigstore-go` as suggested in https://github.com/sigstore/cosign/issues/4847#issuecomment-4399571885

Issue #3356, opened 2023-11-10
Signing with tlog upload is not idempotent


reedloden:
**Description**

We sign our images during the final stage of our release. However, because of the sheer amount of things that happen in that "promotion" stage, it is sometimes necessary to re-run it. cosign 1.x didn't have an issue with this, as signatures were not uploaded automatically to the transparency log. However, with cosign 2.x, we're running into an issue due to a signature already existing:

$ `cosign sign --yes --key awskms:///... --recursive public.ecr.aws/foo/bar@sha256:abcdef...`

...
> Signature already exists. Displaying proof
> Error: signing [public.ecr.aws/foo/bar@sha256:abcdef...]: recursively signing: signing digest: unexpected entry returned from rekor server: expected EntryUUID 24296fb24b8ad77aa8f736fc36ff0184b30419c0347d81b22f5d3328592015e2369a8f71bc741bac got UUID 24296fb24b8ad77aa8f736fc36ff0184b30419c0347d81b22f5d3328592015e2369a8f71bc741bac from Tree c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d
> main.go:74: error during command execution: signing [public.ecr.aws/foo/bar@sha256:abcdef...]: recursively signing: signing digest: unexpected entry returned from rekor server: expected EntryUUID 24296fb24b8ad77aa8f736fc36ff0184b30419c0347d81b22f5d3328592015e2369a8f71bc741bac got UUID 24296fb24b8ad77aa8f736fc36ff0184b30419c0347d81b22f5d3328592015e2369a8f71bc741bac from Tree c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d

https://github.com/sigstore/cosign/blob/aaa354d0dfe2ed315d85ca3737cdbb34fe02b95a/pkg/cosign/tlog.go#L207-L224 makes it seem like signing an image multiple times is fine (and it's what I would expect).

**Version**

cosign 2.2.1

haydentherapper:
Lemme dig in, this is definitely a bug.

haydentherapper:
@reedloden Fixed in https://github.com/sigstore/cosign/pull/3371

PR #4856, opened 2026-05-10
Treat Rekor v1 409 conflict as success in v2 bundle sign path

### LINKED ISSUE #4851: cosign attest retries mishandled
**Description**

 Observed in CI on :`cosign attest --yes --predicate <file> --type <type> <image>@sha256:…:`

```
  Error: signing ...: signing bundle: error signing bundle: [POST /api/v1/log/entries][409]
  createLogEntryConflict {"code":409,"message":"an equivalent entry already exists in the transparency log with UUID 108e9186e8c5677a…"}
```

#3356 solved it but my guess is that this is v2 signing and it's a different branch that doesn't handle the idempotency the same way

Looks like the error is coming from: https://github.com/sigstore/cosign/blob/main/pkg/cosign/bundle/sign.go#L122-L146

**Version**

```
Run sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003
  with:
    cosign-release: v3.0.5
    install-dir: $HOME/.cosign
    use-sudo: false
  env:
    REGISTRY_IMAGE: kong/kong-event-gateway-dev
```

ChrisJr404:
## Summary

When `cosign attest --yes` is re-run against the same image and predicate (a common CI retry pattern), Rekor v1 returns `409 createLogEntryConflict` with the existing entry's UUID. The v1 signing path in `pkg/cosign/tlog.go` (see `doUpload`) has handled this for a long time by fetching the existing entry and continuing. The newer v2 bundle path in `pkg/cosign/bundle/sign.go` does not, and surfaces the error verbatim:

```
Error: signing ...: signing bundle: error signing bundle: [POST /api/v1/log/entries][409] createLogEntryConflict {"code":409,"message":"an equivalent entry already exists in the transparency log with UUID 108e9186e8c5677a..."}
```

This wraps the `sign.Transparency` returned by `sigstore-go`'s `sign.NewRekor` with an idempotent variant. The wrapper:

1. Calls the inner Transparency normally.
2. If the returned error is `*entries.CreateLogEntryConflict`, parses the UUID out of the `Location` header.
3. Builds a Rekor v1 read client, calls `GetLogEntryByUUID`, converts the existing entry into a `TransparencyLogEntry` via `tle.GenerateTransparencyLogEntry`, and appends it to the bundle.
4. Any non-conflict error propagates untouched.

The wrapper is only installed when `MajorAPIVersion == 1`. Rekor v2 (`rekor-tiles`) has no lookup-by-UUID API on its read client today, so fixing the v2 case requires upstream support and is out of scope for this PR.

Fixes #4851
Refs #3356

## Test plan

- [x] Unit tests added in `pkg/cosign/bundle/rekor_idempotent_test.go` covering: pass-through on no error, non-conflict error propagation, 409 with successful fetch (entry appended to a fresh and an existing `TlogEntries`), factory error, fetch error, TLE conversion error, and the `Location`-to-UUID parser.
- [x] `go test ./pkg/cosign/bundle/...`
- [x] `go test ./pkg/cosign/...`
- [x] `go vet` and `golangci-lint run ./pkg/cosign/bundle/...` clean.

### DIFF
--- a/pkg/cosign/bundle/rekor_idempotent.go
+++ b/pkg/cosign/bundle/rekor_idempotent.go
@@ -1,0 +1,132 @@
+// Copyright 2026 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundle
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	protorekor "github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1"
+	rekorclient "github.com/sigstore/rekor/pkg/client"
+	"github.com/sigstore/rekor/pkg/generated/client"
+	"github.com/sigstore/rekor/pkg/generated/client/entries"
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/tle"
+
+	"github.com/sigstore/cosign/v3/internal/ui"
+	"github.com/sigstore/sigstore-go/pkg/sign"
+)
+
+// rekorClientFactory builds a Rekor v1 read client lazily. Tests inject a stub.
+type rekorClientFactory func() (*client.Rekor, error)
+
+// rekorEntryFetcher abstracts the v1 GetLogEntryByUUID lookup so tests can stub it.
+type rekorEntryFetcher func(ctx context.Context, c *client.Rekor, uuid string) (*models.LogEntryAnon, error)
+
+// rekorTLEGenerator abstracts entry-to-TLE conversion so tests can stub it.
+type rekorTLEGenerator func(anon models.LogEntryAnon) (*protorekor.TransparencyLogEntry, error)
+
+// idempotentRekor wraps an upstream sigstore-go Transparency implementation and
+// treats a Rekor v1 409 conflict ("an equivalent entry already exists") as a
+// success by fetching the existing entry and appending it to the bundle, matching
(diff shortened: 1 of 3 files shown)


Hayden-IO (commented):
Thanks for creating this. I might have an opinion on where this should live later on, whether this makes more sense here or in Sigstore-go, but we can discuss that later on.

For v2, we do include the log index in a response header which can be used to query for the entry. Can you add support for this as well? There should be some examples of using the Read client in either the end to end tests in Rekor-tiles or in Sigstore/rekor-monitor. https://github.com/sigstore/rekor-tiles/pull/621 for a ref

Hayden-IO:
@steiza do you have an opinion on where this should live? I would lean towards sigstore-go as it would be beneficial for any application building on the Go SDK.

steiza:
Yeah, this can definitely live in sigstore-go. Just to make sure we're all on the same page for the implementation:

Modify both the Rekor v1 and v2 signing path:
- Attempt to create a Rekor log entry
- If you get an error that the entry already exists, get the UUID (for Rekor v1) or log index (for Rekor v2) from the error response
- Query the log for that entry
- Include that entry in the bundle
- Continue with signing like you would if creating the Rekor log entry succeeded

ChrisJr404:
Got it, sigstore-go makes sense as the home — none of the wrapper logic is cosign-specific (it's rekor client work plus bundle conversion).

I'll port it across and extend it to cover the v2 path the same way: parse the log index from the conflict response, look up the entry, append the TLE.

Worth flagging that the rekor-tiles v2 read client didn't expose a clean lookup-by-index path when I checked. If that's still the case, the v2 branch may need a small upstream change in rekor-tiles first. I'll confirm against the current sigstore-go v2 client and surface it if so.

What's your preference for this PR — close it now and I'll cross-link from the sigstore-go one, or leave it open as a placeholder that bumps the sigstore-go version after the new logic lands there?

Hayden-IO:
@steiza +1, those are the changes we'll need.

For Rekor v2: You'll need to convert an entry index to a tile index, using one of the methods from https://github.com/transparency-dev/tessera/blob/main/api/layout/paths.go. The reader will then fetch a whole tile and extract the relevant entry from the entry bundle.

It's fine to close this now and open a new PR in sigstore-go.

ChrisJr404:
Closing this and porting to sigstore-go per Hayden's guidance. The Tessera path pointer is the missing piece I needed for the v2 branch (`api/layout/paths.go` for entry-index to tile-index conversion, then fetch the tile and pull the entry from the bundle). Will link the sigstore-go PR back here once it's up. Thanks both for the redirect.

PR #4867, opened 2026-05-14
conformance: xfail rekor2-dsse-happy-path

codysoyland:
## Summary

sigstore/sigstore-conformance#371 regenerates the `rekor2-dsse-*` bundle-verify fixtures to encode DSSE envelopes as `hashedrekord/0.0.2` entries per [rekor-v2-spec §6.1.4](https://github.com/sigstore/architecture-docs/pull/63), where `digest = Hash(PAE(payloadType, payload))` and `signature.content = envelope.signatures[0].sig`. Cosign does not yet recognize this encoding for DSSE bundles, so `rekor2-dsse-happy-path` will start failing once #371 lands.

xfails the happy-path case in both PR and nightly conformance workflows per @Hayden-IO's request on #371. The three `rekor2-dsse-*_fail` cases keep passing because cosign rejects the unrecognized entry, which is what those tests expect.

## Related

- Spec change: sigstore/architecture-docs#63
- Conformance tests: sigstore/sigstore-conformance#371

### DIFF
--- a/.github/workflows/conformance-nightly.yml
+++ b/.github/workflows/conformance-nightly.yml
@@ -45,7 +45,7 @@ jobs:
       - uses: sigstore/sigstore-conformance@main
         with:
           entrypoint: ${{ github.workspace }}/conformance
-          xfail: "test_verify*PATH-message-digest-mismatch_fail]"
+          xfail: "test_verify*PATH-message-digest-mismatch_fail] test_verify*rekor2-dsse-happy-path]"
 
       - name: Create Issue on Failure
         if: failure()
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):

Hayden-IO:
Oh, is it failing because the test doesn't exist? We might have to wait until the PR is merged then.

codysoyland:
More accurately, the test does exist, but uses the DSSE type so it passes when it is expected to fail. But yes, the solution is to sequence the conformance merge first.

codysoyland:
No longer applicable

PR #4870, opened 2026-05-18
fix: use Header.Set to prevent duplicate Authorization on retry

SebTardif:
## Summary

The GitHub Actions OIDC token provider (`pkg/providers/github/github.go`) uses `Header.Add` inside a retry loop. On each retry iteration, an additional `Authorization` header is appended to the request. By the third attempt, three identical headers are sent. Some servers and proxies reject requests with duplicate `Authorization` headers, causing retries to fail when they should succeed.

## Fix

Replace `Header.Add` with `Header.Set` so only one `Authorization` header is ever present, regardless of how many retries occur.

## Testing

Added `TestProvide_NoDuplicateAuthHeaders` which uses `httptest` to force two retries via connection hijack and asserts exactly one `Authorization` header is received on every attempt. The test fails on the old code (2 headers on attempt 2, 3 on attempt 3) and passes with the fix.

### DIFF
--- a/pkg/providers/github/github.go
+++ b/pkg/providers/github/github.go
@@ -68,7 +68,7 @@ func (ga *githubActions) Provide(ctx context.Context, audience string) (string, 
 
 	// Retry up to 3 times.
 	for i := 0; ; i++ {
-		req.Header.Add("Authorization", "bearer "+env.Getenv(env.VariableGitHubRequestToken))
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
Thanks for catching this!

steiza (approved):

PR #4836, opened 2026-04-23
Use the configured Target Repository more consistently.

andrew-womeldorf:
Resolves #4591 

#### Summary

`COSIGN_REPOSITORY=signature-repository cosign verify image-repository/my-image:latest` does not list referrers from the repository designated in the `COSIGN_REPOSITORY` environment variable. This also affects Kyverno's ImageValidatingPolicy when using the new bundle format in a separate signature repository. See https://github.com/kyverno/kyverno/issues/15950.

This can be verified by using cosign to create and attach a signature first, then trying to verify it.

```sh
$ export COSIGN_REPOSITORY=signature-repository
$ cosign sign image:tag ...
$ cosign verify image:tag
```

Note that there is an existing PR that also addresses #4591, but it's several months old, many commits behind, and inactive. I took a slightly different approach here, too. See https://github.com/sigstore/cosign/pull/4593

#### Release Note

Fixed `COSIGN_REPOSITORY` and `WithTargetRepository` to be respected when fetching OCI 1.1 referrer-based bundles during verification. Previously, `GetBundles` and `Referrers` ignored the configured target repository, causing verification to fail when bundles were stored in an alternate registry.

PR #4871, opened 2026-05-18
Honor COSIGN_REPOSITORY when verifying new-bundle attestations

### LINKED ISSUE #4872: COSIGN_REPOSITORY silently ignored when verifying attestations with --new-bundle-format
## Description

When `COSIGN_REPOSITORY` is set, `cosign attest --new-bundle-format` correctly writes the attestation to the override repository, but `cosign verify-attestation --new-bundle-format` reads referrers from the *subject's* repository and never consults the override. Verification therefore fails with `no valid bundles exist in registry`, even though the bundle was just written successfully.

The legacy `.att` tag path honors the override via [`ociremote.AttestationTag`](https://github.com/sigstore/cosign/blob/main/pkg/oci/remote/remote.go#L122-L126), which routes through `o.TargetRepository`. The new-bundle path bypasses that and was never wired up.

## Reproduction

```sh
export COSIGN_REPOSITORY=registry.example.com/sigstore-store
IMAGE=registry.example.com/app@sha256:...

# Writes the bundle into $COSIGN_REPOSITORY — succeeds.
cosign attest --new-bundle-format --predicate pred.json --type custom "$IMAGE"

# Looks for referrers under registry.example.com/app — fails.
cosign verify-attestation --new-bundle-format \
  --certificate-identity ... --certificate-oidc-issuer ... "$IMAGE"
# Error: no matching attestations: no valid bundles exist in registry
```

## Root cause

In `GetBundles` ([pkg/cosign/verify.go](https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go)), both the `Referrers` query and the per-result `Bundle` fetch are constructed from the subject's digest:

```go
index, err := ociremote.Referrers(digest, "", registryClientOpts...)
// ...
st, err := name.ParseReference(fmt.Sprintf("%s@%s", digest.Repository, result.Digest.String()), nameOpts...)
```

`digest.Repository` is the subject's repository, regardless of `WithTargetRepository` (i.e. `COSIGN_REPOSITORY`). The same code path is hit by the auto-detect probe in `cmd/cosign/cli/verify/{verify,verify_attestation}.go`, so even falling back to the legacy format requires the new-bundle probe to have looked in the wrong place first.

## Impact

Any user who configures `COSIGN_REPOSITORY` (e.g. to separate signing artifacts from application images, or because they don't have write access to the subject's registry) cannot verify attestations produced in the new bundle format. Sign succeeds; verify always fails.

codysoyland:
## Summary

`COSIGN_REPOSITORY` is silently ignored on verify when `--new-bundle-format=true`.
`cosign attest` writes to the override repo, but `cosign verify-attestation`
reads referrers from the *subject's* repo, so verification always fails.

The legacy `.att` tag path honors the override (via `ociremote.AttestationTag`).
The new-bundle path bypassed that and was never wired up.

## Fix

- New `ociremote.ResolveTargetDigest` helper translates a `name.Digest` into the
  target repository configured by `WithTargetRepository` (set from
  `COSIGN_REPOSITORY`), mirroring `AttestationTag`.
- `GetBundles` resolves once and uses that target for both the `Referrers`
  query and the per-bundle fetch.
- Drop the now-unused `nameOpts` parameter from `GetBundles`. The old code
  passed it to `name.ParseReference` to reapply `name.Insecure` across a
  string round-trip; `target.Digest(...)` reuses the existing `Repository`
  directly, so the insecure flag is preserved without it. Exported
  `VerifyImageAttestations` keeps its variadic signature for backwards
  compatibility.

## Tests

- Unit test in `pkg/cosign/verify_oci_test.go` pushes an attestation to a
  separate repo on the in-process registry and asserts `GetBundles` finds it
  only when the same `WithTargetRepository` option is passed on read.
- Pass `TrustedRootPath` via `CommonVerifyOptions` in
  `TestSignVerifyWithRepoOverride`; the new-bundle verify path requires
  trusted material.

Fixes https://github.com/sigstore/cosign/issues/4872

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1029,13 +1029,13 @@ func loadSignatureFromFile(ctx context.Context, sigRef string, signedImgRef name
 
 // VerifyImageAttestations does all the main cosign checks in a loop, returning the verified attestations.
 // If there were no valid attestations, we return an error.
-func VerifyImageAttestations(ctx context.Context, signedImgRef name.Reference, co *CheckOpts, nameOpts ...name.Option) (checkedAttestations []oci.Signature, bundleVerified bool, err error) {
+func VerifyImageAttestations(ctx context.Context, signedImgRef name.Reference, co *CheckOpts, _ ...name.Option) (checkedAttestations []oci.Signature, bundleVerified bool, err error) {
 	// Enforce this up front.
 	if co.RootCerts == nil && co.SigVerifier == nil && co.TrustedMaterial == nil {
 		return nil, false, errors.New("one of verifier, root certs, or TrustedMaterial is required")
 	}
 	if co.NewBundleFormat {
-		return verifyImageAttestationsSigstoreBundle(ctx, signedImgRef, co, nameOpts...)
+		return verifyImageAttestationsSigstoreBundle(ctx, signedImgRef, co)
 	}
 
 	// This is a carefully optimized sequence for fetching the attestations of
@@ -1667,7 +1667,7 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
 	return verifySignatures(ctx, sigs, h, co)
 }
 
-func GetBundles(_ context.Context, signedImgRef name.Reference, registryClientOpts []ociremote.Option, nameOpts ...name.Option) ([]*sgbundle.Bundle, *v1.Hash, error) {
+func GetBundles(_ context.Context, signedImgRef name.Reference, registryClientOpts []ociremote.Option) ([]*sgbundle.Bundle, *v1.Hash, error) {
 	// This is a carefully optimized sequence for fetching the signatures of the
 	// entity that minimizes registry requests when supplied with a digest input
 	digest, err := ociremote.ResolveDigest(signedImgRef, registryClientOpts...)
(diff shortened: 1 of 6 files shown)


pkg/cosign/verify.go
```diff
@@ -1667,7 +1667,7 @@ func verifyImageSignaturesExperimentalOCI(ctx context.Context, signedImgRef name
 	return verifySignatures(ctx, sigs, h, co)
 }
 
-func GetBundles(_ context.Context, signedImgRef name.Reference, registryClientOpts []ociremote.Option, nameOpts ...name.Option) ([]*sgbundle.Bundle, *v1.Hash, error) {
```
codysoyland, line 1670: I *think* it's safe to remove nameOpts here from the public api even though it's technically a breaking change. The passed in `name.Reference` already carries the insecure flag that is captured by name.Option. Because it's a variadic argument list, go callers are unlikely to pass it in. We can add it back with `_ ...name.Option` if we would like to guarantee API stability.

steiza (commented):
From a quick review, would #4836 work for you? I was leaving that open for a few days in case other maintainers wanted to weigh in, but maybe we should just land that?

cpanato (approved):

codysoyland:
> From a quick review, would #4836 work for you? I was leaving that open for a few days in case other maintainers wanted to weigh in, but maybe we should just land that?

Yeah, looks like #4836 solves it pretty well and introduces some improved test coverage. So let's land that one.

PR #4875, opened 2026-05-19
feat(cli): add first-class WebAssembly signing and embedded signature management

MacroModel:
## Summary

This PR adds first-class WebAssembly support to cosign so `.wasm` modules can be signed, verified, inspected, stripped, and re-signed as self-contained artifacts.

The new workflow stores Sigstore verification material inside a WebAssembly custom section named `wasm-cosign`, while preserving normal runtime behavior. This makes cosign usable as a module-level signing system for wasm in the same way code-signing tooling is used for PE artifacts.

## What changed

- Added a new `cosign wasm` command group.
- Added subcommands for:
  - `sign`
  - `verify`
  - `list-signatures`
  - `extract-signatures`
  - `strip-signatures`
- Added `--wasm` / `--wasm-output` support to `sign-blob` and `verify-blob`.
- Added a new `pkg/wasm` helper package for parsing, stripping, appending, and inspecting embedded signature sections.
- Added docs for the wasm command group and a detailed manual for the embedded signature model.
- Added tests covering multi-signature behavior, stripping, inspection, verification, and tamper cases.

## How it works

Cosign treats a wasm module as signed by computing the canonical payload from the module with every `wasm-cosign` custom section removed.

That means:

- The signature section itself is excluded from the digest.
- All other wasm bytes remain protected.
- Adding another `wasm-cosign` section does not break existing signatures.
- Any mutation to non-signature sections invalidates verification.

This design supports multiple embedded signatures from different vendors or from the same vendor at different stages of a release pipeline.

## User workflow

Typical usage looks like this:

```bash
cosign wasm sign --key cosign.key --wasm-output signed.wasm module.wasm
cosign wasm verify --key cosign.pub signed.wasm
cosign wasm list-signatures signed.wasm
cosign wasm extract-signatures signed.wasm
cosign wasm strip-signatures -o unsigned.wasm signed.wasm
```

The blob commands also support wasm mode for compatibility with existing automation:

```bash
cosign sign-blob --wasm --wasm-output signed.wasm --key cosign.key module.wasm
cosign verify-blob --wasm --key cosign.pub signed.wasm
```

## Security model

This PR is designed to make wasm signing safe by default:

- Only exact `wasm-cosign` custom sections are treated as embedded signatures.
- Verification strips all embedded signature sections before checking the protected payload.
- Listing and extraction are inspection tools, not trust decisions.
- Non-signature wasm bytes, including custom sections like `producers`, remain covered by the signature.
- Multi-signature modules are verified section-by-section, so tampering with one embedded bundle is detected.

## Validation

I validated the implementation with controlled wasm fixtures generated via `wat2wasm`, including:

- happy-path signing and verification
- multiple embedded signatures
- moving signature sections to different positions
- stripping embedded signatures back to the original module
- tampering with code, exports, and custom sections
- tampering with the embedded `wasm-cosign` payload itself
- malformed wasm and unsupported version handling

## Notes

This work turns cosign into a practical wasm signing system rather than a generic blob signer with a wasm flag. It gives wasm artifacts a complete lifecycle for signing, verification, inspection, and cleanup.

### DIFF
--- a/cmd/cosign/cli/commands.go
+++ b/cmd/cosign/cli/commands.go
@@ -114,6 +114,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(Save())
 	cmd.AddCommand(Sign())
 	cmd.AddCommand(SignBlob())
+	cmd.AddCommand(Wasm())
 	cmd.AddCommand(Upload())
 	cmd.AddCommand(Verify())
 	cmd.AddCommand(VerifyAttestation())
--- a/cmd/cosign/cli/options/signblob.go
+++ b/cmd/cosign/cli/options/signblob.go
@@ -35,6 +35,8 @@ type SignBlobOptions struct {
 	Output               string // deprecated: TODO remove when the output flag is fully deprecated
 	OutputSignature      string // TODO: this should be the root output file arg.
 	OutputCertificate    string
+	Wasm                 bool
+	WasmOutput           string
 	SecurityKey          SecurityKeyOptions
 	Fulcio               FulcioOptions
 	Rekor                RekorOptions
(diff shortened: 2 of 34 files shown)


Hayden-IO:
Hey, thanks for reaching out. We aren't planning to extend Cosign for additional types of binaries, beyond blobs, attestations and containers. I'd recommend using `sign-blob` and building separate tooling that canonicalizes a wasm module into bytes to sign, and if applicable embedding the signature bundle into the package. Alternatively you can build your own application using one of the Sigstore SDKs like https://github.com/sigstore/sigstore-go, which would canonicalize the wasm module into bytes, sign and verify with sigstore-go, and handle storage of the signature metadata.

PR #4878, opened 2026-05-20
verify: return TUF errors for new bundle trusted roots

crazy-max:
#### Summary

This change fixes new bundle verification so a failure to load the live trusted root from TUF is returned at the point where it happens. New bundle verification cannot fall back to legacy TUF targets, so returning the wrapped TUF error makes the failure actionable instead of later reporting that trusted material is nil: https://github.com/docker/github-builder/actions/runs/25808830274/job/75819126449?pr=201#step:4:54

```
cosign verify --experimental-oci11 --new-bundle-format --certificate-oidc-issuer https://token.actions.githubusercontent.com/ --certificate-identity-regexp ^https://github.com/docker/github-builder(-experimental)?/.github/workflows/bake.yml.*$ tonistiigi/binfmt@sha256:c7f7f49f896505f3a38e11768b25721a267554b573ad86efa481a8c7e7863a15
  Error: Error: Cosign verify command failed with: error during command execution: trusted root is required when using new bundle format
```

#### Release Note

Fixed new bundle verification to report trusted root loading failures from TUF instead of a later generic missing trusted root error.

#### Documentation

This change does not require documentation updates because it only changes error handling for an existing verification path.

cc @tonistiigi @Hayden-IO

### DIFF
--- a/cmd/cosign/cli/verify/common.go
+++ b/cmd/cosign/cli/verify/common.go
@@ -190,6 +190,9 @@ func SetTrustedMaterial(ctx context.Context, trustedRootPath, certChain, caRoots
 		env.Getenv(env.VariableSigstoreTSACertificateFile) == "" {
 		co.TrustedMaterial, err = cosign.TrustedRoot()
 		if err != nil {
+			if co.NewBundleFormat {
+				return fmt.Errorf("getting trusted root from TUF for new bundle verification: %w", err)
+			}
 			ui.Warnf(ctx, "Could not fetch trusted_root.json from the TUF repository. Continuing with individual targets. Error from TUF: %v", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


Hayden-IO:
I added another maintainer to take a look at this, as I think this was intentional to continue to support an older way to provide trust root material. Unfortunately this has become a bit complex and we will be removing older paths.

cmurphy:
@Hayden-IO this is fine, new bundle format always requires `TrustedMaterial` to be set, and they would hit another error in `CheckSigstoreBundleUnsupportedOptions` anyway.

@crazy-max could you rebase to adopt the fix for the e2e-kms failure?

crazy-max:
> @Hayden-IO this is fine, new bundle format always requires `TrustedMaterial` to be set, and they would hit another error in `CheckSigstoreBundleUnsupportedOptions` anyway.
> 
> @crazy-max could you rebase to adopt the fix for the e2e-kms failure?

Rebased, thanks

Hayden-IO (approved):

PR #4880, opened 2026-05-20
Fix Ed25519ph check to respect custom signing configs in sign-blob

aaronlew02:
#### Summary

This change updates the logic in `sign-blob` used to determine when Ed25519ph is disabled for Ed25519 keys (i.e., when transparency log upload is disabled) to take into account a user-provided signing config.

### DIFF
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -59,14 +59,6 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
 	ctx, cancel := context.WithTimeout(ctx, ro.Timeout)
 	defer cancel()
 
-	// TODO - this does not take ko.SigningConfig into account
-	if !tlogUpload {
-		// To maintain backwards compatibility with older cosign versions,
-		// we do not use ed25519ph for ed25519 keys when the signatures are not
-		// uploaded to the Tlog.
-		ko.DefaultLoadOptions = &[]signature.LoadOption{}
-	}
-
 	var shouldUpload bool
 	var err error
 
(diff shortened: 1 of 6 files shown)


Hayden-IO (commented):
LGTM. Do you have a test or command we can run to verify this?

aaronlew02:
I added test cases for Ed25519 signing in `sign-blob`, which demonstrated that Ed25519 signing caused a panic ("`crypto: requested hash function #0 is unavailable`") on the call to `GetHashType().New()`, since the hash algorithm for standard Ed25519 is unspecified (`crypto.Hash(0)`).

To enable these test cases to pass, the scope of this PR has expanded to fix Ed25519 signing without pre-hashing in general, since I:
1. added unit tests for Ed25519 methods in `KMSKeypair` which further demonstrated the panic,
2. fixed the panic by adding handling of Ed25519 signing without pre-hashing in `SignData` and `HashReader`, and
3. added a test case for an unspecified hash algorithm in `HashReader`.

Should I extract the last three commits into a separate PR upon which this PR can depend?

Hayden-IO (approved):
Wow this is some deja vu from trying to deal with this in https://github.com/sigstore/cosign/pull/4414/. Supporting ed25519 has been a painpoint :/

Thanks for the additional updates.

Where is `HashReader` used such that we needed to update it?

aaronlew02:
`HashReader` is used by `sign-blob`, specifically in [`getPayload`](https://github.com/sigstore/cosign/blob/0bfb59e07483a20d3570f7ee8370a17147f51b69/cmd/cosign/cli/sign/sign_blob.go#L43-L53). If the `hashFunction` passed to `NewHashReader` is `crypto.Hash(0)`, the original would call `.New()` on the hash function, leading to a panic, but this is now handled by returning a `HashReader` with a `Hash` of `nil`.

PR #4881, opened 2026-05-20
Fix crash verifying timestamps when no timestamp was verified

Hayden-IO:
If no TSA chain in the trust root can verify a signed timestamp, then a crash would occur when Cosign tries to read one of the verified timestamps. We now throw an error if no timestamp was verified.


steiza (approved):
I cannot wait to delete this function. Why aren't you using the new bundle format? /s

Seriously though, thanks for fixing this.

PR #4894, opened 2026-05-26
Deprecate subcommands

piceri:

#### Summary
The change deprecates the following subcommands following the guidance in VERSIONING.md:
- attach
- dockerfile
- generate
- manifest
- triangulate (already deprecated but changed the message)
- upload

#### Release Note
The following subcommands have been deprecated and will be removed in v4.0.0:
- attach
- dockerfile
- generate
- manifest
- triangulate
- upload


cmd/cosign/cli/generate.go
```diff
@@ -30,6 +30,7 @@ func Generate() *cobra.Command {
 		Long: `Generates an unsigned payload from the supplied container image and flags.
 This payload matches the one generated by the "cosign sign" command and can be used if you need
 to sign payloads with your own tooling or algorithms.`,
+		Deprecated: "generate will be removed in v4.0.0 (see https://github.com/sigstore/cosign/issues/4696). Instead, please use in-toto tooling",
```
piceri, line 33: @steiza I looked into in-toto tooling as mentioned in the issue, but I am unsure how to use it to be compatible with this command. Do you have any suggestions for further guidance here?

Hayden-IO:
Will this also handle https://github.com/sigstore/cosign/pull/3891?

cmd/cosign/cli/generate.go
steiza: No, I think this was aspirational - we hope that in-toto will produce tooling for producing statements. If nothing else, we could point people at https://github.com/in-toto/attestation/blob/main/examples/go/main.go? But also this is fine to leave as-is.

cmd/cosign/cli/attach.go
```diff
@@ -26,8 +26,9 @@ import (
 
 func Attach() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "attach",
-		Short: "Provides utilities for attaching artifacts to other artifacts in a registry",
+		Use:        "attach",
+		Short:      "Provides utilities for attaching artifacts to other artifacts in a registry",
+		Deprecated: "attach will be removed in v4.0.0 (see https://github.com/sigstore/cosign/issues/4696). Instead, please use oras for attaching artifacts to other artifacts in a registry; https://docs.sigstore.dev/cosign/signing/signing_with_containers/#upload-signature-to-a-container-registry-later",
```
steiza:
```suggestion
		Deprecated: "attach will be removed in v4.0.0 (see https://github.com/sigstore/cosign/issues/4696). Instead, please use `oras attach` for attaching artifacts to other artifacts in a registry",
```

In the other deprecation strings, we just provide people the command they should use instead - I think that's fine to do here as well.

steiza (commented):
Minor comments, and it looks like we need to run `make docgen`, otherwise LGTM!

steiza:
> Will this also handle https://github.com/sigstore/cosign/pull/3891?

I think the plan is to:
- Mark everything deprecated (of which this is the last pull request, unless we start going into public functions in `pkg/`)
- Make a Cosign v3.1.0 release
- **Then** start deleting things (like as suggested in #3891)

To be fair, the subcommands deleted in #3891 have been marked as deprecated for awhile, so I wouldn't feel bad about deleting them earlier. But we weren't planning on doing any deletions until after the Cosign v3.1.0 release.

steiza (approved):
LGTM!

Issue #4814, opened 2026-04-14
Smaller cosign-sign binary


Hayden-IO:
**Description**

Building on https://github.com/sigstore/cosign/issues/3961 and splitting https://github.com/sigstore/cosign/issues/4221#issuecomment-3398131745 out into its own issue, let's build a smaller sign binary. Smaller would mean both smaller in scope, stripping away as much functionality as possible while still retaining core signing features, minimal UX (less flags), and smaller in the dependency tree (driven through a reduction in dependencies in sigstore-go).

I propose we support:

* Well-lit signing path for the public instance, with Fulcio, Rekor, and a TSA
* TUF for retrieving roots of trust
* OpenID Connect authentication, though maybe look for a minimal dependency OIDC library that doesn't support the entirety of the spec
* On-disk keys

And we would not support:

* Anything with containers, as cloud SDKs are large
* Anything with KMS, for the same reason
* PKCS11 support
* Any complexity from supporting private PKIs

This should be a single command, something like `cosign-lite sign` that generates a signature bundle. The flag names should match `cosign`.

Open questions:

* What about attestations? Have that be another tool? Support providing a payload, so we only have to construct the DSSE? Or also support providing an in-toto predicate?

aaronlew02:
Essentially, `cosign-lite sign` is a dependency-reduced version of `cosign sign-blob`.

We should remove any [functionality](https://github.com/sigstore/cosign/blob/8a86a7cf843872c79a33401de60a8c77aac23b45/cmd/cosign/cli/sign/sign_blob.go#L138-L214) related to the old bundle format or detached verification material.

To address the list of items we should support:

> Well-lit signing path for the public instance, with Fulcio, Rekor, and a TSA
* We should continue to allow users to provide a signing config and trusted root for custom service inputs, with a fallback to TUF otherwise, leaving the public instance as the default. 
* We should remove service input flags for the sake of simplicity.
> TUF for retrieving roots of trust
* We should include an `initialize` command to match the current one, so that users can continue to use the staging instance or other deployments. 
* TUF-related environment variables should work as well.
> OpenID Connect authentication, though maybe look for a minimal dependency OIDC library that doesn't support the entirety of the spec
* We should continue to support the web-based OIDC flow as well as the `envvar`, `filesystem`, and `github` identity providers.
[...]

PR #4893, opened 2026-05-26
Add cosign-lite

aaronlew02:
#### Summary

This change adds `cosign-lite`, a smaller, dependency-reduced binary for signing and verification. 

`cosign-lite` has six subcommands: `sign`, `attest`, `verify`, `verify-attestation`, `initialize`, and `generate-key-pair`.

It compiles to a binary that is over 75% smaller than standard `cosign`, stripping away heavy functionality and transitive dependencies related to container registries, cloud KMS SDKs, and hardware tokens. It retains support for local, password-encrypted private keys, and keyless signing via OIDC (including interactive browser-based authentication and ambient credential detection in environments like GitHub Actions).

The `sign` command is dedicated to signing raw blobs (much like `cosign sign-blob`), and the `attest` command is dedicated to signing in-toto payloads (much like `cosign attest-blob`, except that only the pre-constructed statement may be provided, bypassing predicate generation).

PR #4895, opened 2026-05-27
lite: Add binary and initialize command

aaronlew02:
Partially addresses #4814 

#### Summary

This change adds the `cosign-lite` binary and its `initialize` command.

#### Release Note

This series of changes adds `cosign-lite`, a smaller, dependency-reduced binary for signing and verification. 

`cosign-lite` has six subcommands: `sign`, `attest`, `verify`, `verify-attestation`, `initialize`, and `generate-key-pair`.

It compiles to a binary that is over 75% smaller than standard `cosign`, stripping away heavy functionality and transitive dependencies related to container registries, cloud KMS SDKs, and hardware tokens. It retains support for local, password-encrypted private keys, and keyless signing via OIDC (including interactive browser-based authentication and ambient credential detection in environments like GitHub Actions).

The `sign` command is dedicated to signing raw blobs (much like `cosign sign-blob`), and the `attest` command is dedicated to signing in-toto payloads (much like `cosign attest-blob`, except that only the pre-constructed statement may be provided, bypassing predicate generation).


Hayden-IO:
We'll need to do a little bikeshedding here and decide on a name. Originally, I was thinking `cosign-lite`, which would work well if the binary supported both signing and verification - `cosign-lite sign` is less repetitive than `cosign-sign sign`. We had proposed a binary each for signing and for verification though, and so it'll need a distinct module name so that the signer and verifier binaries can be installed on the same system.

@aaronlew02 Have you done any testing around verification yet, for what the dependency tree will look like for a verifier binary? I'm wondering if it's not that different from `cosign-sign`, and if so, then we could produce just one binary.

cc @steiza @cmurphy

aaronlew02:
> We had proposed a binary each for signing and for verification though, and so it'll need a distinct module name so that the signer and verifier binaries can be installed on the same system.

I agree. That's why I went with `cosign-sign` for now (and `cosign-verify` for the verification binary). We could land both and, if feasible, merge them into a single `cosign-lite` binary.

> Have you done any testing around verification yet, for what the dependency tree will look like for a verifier binary? I'm wondering if it's not that different from `cosign-sign`, and if so, then we could produce just one binary.

No, but I can.

aaronlew02:
I've implemented a combined `cosign-lite` binary for signing and verification (#4893) and determined that `cosign-lite` remains over 75% smaller than `cosign` (see [data](https://github.com/sigstore/cosign/pull/4893#issuecomment-4596462713)). We should therefore shift to developing a single `cosign-lite` binary.

PR #4902, opened 2026-05-28
fix: wire insecure-skip-verify to bundle generation

### LINKED ISSUE #4865: cosign v3 fails to sign image with certificate
**Description**

Cosign v2 works fine when you are trying to sign image with your own certificate by using the next command:

`cosign sign $IMAGE --insecure-skip-verify=true --tlog-update=false --certificate leaf.cer --certificate-chain ca_bundle.pem --key imported-cosign.key`

But when i'm trying to sign the same way with cosign v3 it fails:

`cosign sign $IMAGE --insecure-skip-verify=true --certificate leaf.cer --certificate-chain ca_bundle.pem --key imported-cosign.key --signing-config my-config.json`

Error message: 

`error: signing [$IMAGE] recursively signing: signing digest: signing bundle: error signing bundle: fqailed to verify leaf certificate: leaf certificate verification failed`

`my-config.json` contains default config generated by `cosign signing-config create`, leaf certificate contains digital signature and code signing attributes in keyUsage/extendedKeyUsage.

Is it bug or cosign v3 requires something additional to sign image by your own certificate?  

**Version**

v2.5.0 (38bb98697...), go1.24.1
v3.0.6 (f1ad3ee952...), go1.25.7

agustinushanssen:
Fixes #4865

When the \`--insecure-skip-verify\` flag was marked deprecated due to the removal of SCT verification, it inadvertently broke BYOC (Bring Your Own Certificate) workflows. The bundle generator in v3 strictly verified custom leaf certificates against public roots because the skip flag was dropped.

This commit:
- Un-deprecates the flag and clarifies its requirement for BYOC.
- Adds an InsecureSkipVerify field to bundle.SignOptions.
- Prevents assigning a TrustedRoot in bundle.SignData if skipped.
- Wires the CLI flag down to the bundle options for both blobs and images.

#### Summary
**Motivation & Problem:**
When the `--insecure-skip-verify` flag was deprecated in v3 (due to the removal of SCT verification), it inadvertently broke the BYOC (Bring Your Own Certificate) workflow. Enterprise users relying on custom certificates were hitting a `failed to verify leaf certificate` error during bundle generation. The CLI correctly parsed the flag but dropped it before handing off to the bundle builder. As a result, `bundleOpts.TrustedRoot` was being populated unconditionally in `pkg/cosign/bundle/sign.go`, forcing the underlying `sigstore-go` library to strictly verify custom leaf certs against public roots.

This PR fixes this regression by:
1. Un-deprecating the `--insecure-skip-verify` flag and clarifying its requirement for BYOC workflows.
2. Adding an `InsecureSkipVerify` boolean to `bundle.SignOptions`.
3. Updating `bundle.SignData` to block the assignment of `TrustedRoot` if skipped.
4. Wiring the CLI flag down to the bundle options for both blobs and images.

**How reviewers can test:**
1. Generate a local Root CA and Leaf Certificate with `codeSigning` extensions.
2. Attempt to sign a blob or image using `--certificate`, `--key`, and `--certificate-chain` alongside `--insecure-skip-verify=true`.
3. Verify that the bundle signs successfully without triggering the leaf certificate verification crash.

#### Release Note
Fixed a regression where signing with custom certificates (BYOC) failed with a `leaf certificate verification failed` error during bundle generation. Un-deprecated the `--insecure-skip-verify` CLI flag and properly wired it to bypass strict root verification when generating bundles with custom certificates.

#### Documentation
CLI help text and struct documentation for `--insecure-skip-verify` have been updated inline to clarify its active role in bypassing strict verification for BYOC workflows. No external documentation PR is currently linked, but the CLI reference will automatically reflect the un-deprecation.

### DIFF
--- a/cmd/cosign/cli/options/fulcio.go
+++ b/cmd/cosign/cli/options/fulcio.go
@@ -46,6 +46,5 @@ func (o *FulcioOptions) AddFlags(cmd *cobra.Command) {
 		"fulcio interactive oauth2 flow to use for certificate from fulcio. Defaults to determining the flow based on the runtime environment. (options) normal|device|token|client_credentials")
 
 	cmd.Flags().BoolVar(&o.InsecureSkipFulcioVerify, "insecure-skip-verify", false,
-		"skip verifying fulcio published to the SCT (this should only be used for testing).")
-	_ = cmd.Flags().MarkDeprecated("insecure-skip-verify", "SCT verification is no longer performed during signing/attestation.")
+		"skip strict certificate verification, useful for BYOC workflows.")
 }
--- a/cmd/cosign/cli/options/key.go
+++ b/cmd/cosign/cli/options/key.go
@@ -54,9 +54,9 @@ type KeyOpts struct {
 	// for valid values.
 	FulcioAuthFlow string
 
-	// Deprecated: SCT verification is no longer performed during signing/attestation.
 	// Modeled after InsecureSkipVerify in tls.Config, this disables
-	// verifying the SCT.
+	// strict leaf certificate verification, which is necessary for BYOC
+	// (Bring Your Own Certificate) workflows.
 	InsecureSkipFulcioVerify bool
 
 	// TrustedMaterial contains trusted metadata for all Sigstore services. It is exclusive with RekorPubKeys, RootCerts, IntermediateCerts, CTLogPubKeys, and the TSA* cert fields.
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -292,6 +292,7 @@ func signDigest(ctx context.Context, digest name.Digest, payload []byte, ko opti
 	cbundleOpts := cbundle.SignOptions{
 		TSAClientTransport:  tsaClientTransport,
 		CertificateProvider: certProvider,
+		InsecureSkipVerify:  ko.InsecureSkipFulcioVerify,
 	}
 
 	ociSigs := make([]oci.Signature, len(payloads))
--- a/cmd/cosign/cli/sign/sign_blob.go
+++ b/cmd/cosign/cli/sign/sign_blob.go
@@ -122,7 +122,11 @@ func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpt
 			return nil, fmt.Errorf("getting TSA client transport: %w", err)
 		}
 	}
-	signOpts := cbundle.SignOptions{TSAClientTransport: tsaClientTransport}
+	signOpts := cbundle.SignOptions{
+		TSAClientTransport: tsaClientTransport,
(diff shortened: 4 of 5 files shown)


Hayden-IO:
Can you please provide a reproduction for the original issue? Can you provide an example cert chain that triggers this?

This flag should be able to be deprecated now, it's no longer needed, so if this fix works, then something is incorrectly wired up.

agustinushanssen:
Thanks for taking a look! Your intuition makes perfect sense—if BYOC (Bring Your Own Certificate) is supposed to work natively with a custom chain, then wiring up the skip flag is definitely just masking a deeper issue with how the trust anchors are established.

It appears that when a user provides a custom `--certificate-chain`, it is not being properly respected or passed down as the `TrustedRoot` when `sigstore-go` attempts to build and verify the bundle, causing it to fail strict verification against the public roots.

Here is the exact minimal reproduction and cert chain I used to trigger the issue on `main`:

### 1. Generate the local Root CA & Leaf Cert
```bash
# Generate Root CA
openssl req -x509 -newkey rsa:4096 -days 365 -nodes -keyout ca.key -out ca_bundle.pem -subj "/CN=Fake Cosign Root CA"

# Generate Leaf CSR
openssl req -newkey rsa:4096 -nodes -keyout imported-cosign.key -out leaf.csr -subj "/CN=Fake Cosign Leaf"

# Create extensions file required by Cosign v3
cat <<EOF > leaf.ext
authorityKeyIdentifier=keyid,issuer
basicConstraints=CA:FALSE
keyUsage = digitalSignature
extendedKeyUsage = codeSigning
subjectAltName = email:test@example.com
EOF

# Sign the Leaf with the Root CA
openssl x509 -req -in leaf.csr -CA ca_bundle.pem -CAkey ca.key -CAcreateserial -out leaf.cer -days 365 -extfile leaf.ext
```

### 2. Import the key into Cosign:

```bash
cosign import-key-pair --key imported-cosign.key --output-key-prefix cosign
# (Enter a dummy password when prompted)
```

### 3. Attempt to sign a blob with the custom chain:

```bash
echo "hello world" > dummy.txt
cosign signing-config create > my-config.json

cosign sign-blob dummy.txt \
  --bundle bundle.json \
  --certificate leaf.cer \
  --certificate-chain ca_bundle.pem \
  --key cosign.key \
  --signing-config my-config.json
```

### The Result on `main`:
`Error: signing dummy.txt: signing bundle: error signing bundle: failed to verify leaf certificate: leaf certificate verification failed`

Because the original author of #4865 was heavily relying on `--insecure-skip-verify` in v2 to bypass this exact crash, they hit a wall in v3 when the flag was dropped.

If the consensus is that the flag should remain deprecated and we should fix the underlying trust-root wiring instead, I completely agree. I'm happy to close this PR and pivot to investigating why `--certificate-chain` isn't satisfying the `TrustedRoot` requirement during `sign.Bundle()`. Let me know how you'd prefer to proceed!

Issue #4817, opened 2026-04-14
`verify` output missing success indicators for BYO-PKI certificate chain validation and RFC3161 timestamp verification


anshuman-mor:
## Summary

When using `cosign verify` in BYO-PKI mode, three security checks run and fail-hard on error — but on success they produce **no output bullet** and/or have no corresponding flag. This creates an audit/compliance gap and a completeness gap for enterprise PKI users.

---

## Affected flags / flow

- `--cert <leaf.pem> --ca-roots <ca-bundle.pem>` (BYO-PKI, offline, no Fulcio)
- `--use-signed-timestamps --timestamp-certificate-chain <tsa-chain.pem>`
- SCT verification with a custom CT log (no flag exists)

---

## Issue 1 — BYO-PKI cert chain validation is silent on success

When `--cert` and `--ca-roots` are provided, cosign verifies the signing certificate against the supplied CA bundle. This check fails hard when the chain is invalid (e.g. wrong root → clear error). But on success, no bullet is printed.

The "Any certificates were verified against the Fulcio roots" bullet only fires in the Fulcio/keyless flow. There is no equivalent bullet for the BYO-PKI cert chain path.

**Suggested bullet:** `The signing certificate was verified against the provided CA roots`

---

## Issue 2 — RFC3161 timestamp verification is silent on success

When `--use-signed-timestamps` and `--timestamp-certificate-chain` are provided, cosign verifies the `dev.sigstore.cosign/rfc3161timestamp` annotation against the TSA certificate chain. Again, this fails hard on invalid input (e.g. wrong TSA chain root → `TSA certificate chain must contain at least one root certificate`). But on success, no bullet appears in the output.

**Suggested bullet:** `The RFC3161 timestamp was verified against the TSA certificate chain`

---

## Issue 3 — No way to supply a custom CT log for BYO-PKI SCT verification

In the keyless/Fulcio flow, cosign verifies the SCT (Signed Certificate Timestamp) embedded in Fulcio-issued certificates against Sigstore's own CT log key (embedded via TUF). In BYO-PKI mode, users must always pass `--insecure-ignore-sct` because:

1. Their enterprise CA certificates were not issued by Fulcio and are therefore never submitted to Sigstore's CT log.
2. Even if the enterprise CA runs its own RFC 6962 CT log, there is no `--ct-log-public-key` (or equivalent) flag to point cosign at it.

This is inconsistent with what exists for other BYO-PKI components:
[...]

PR #4904, opened 2026-05-28
fix: report certificate and timestamp verification checks

bmendonca3:
## Summary
- report certificate-chain verification in `cosign verify` output when verification used a supplied certificate
- report RFC3161 timestamp verification in the success header when signed timestamp verification was enabled
- cover the new success lines with a focused header-output test

This addresses the output-precision parts of #4817. Custom CT log support is intentionally left out of scope.

## Testing
- `PATH=/tmp/go1.26.3/bin:$PATH go test ./cmd/cosign/cli/verify -run TestPrintVerificationHeaderIncludesCertificateAndTimestampChecks -count=1`
- `PATH=/tmp/go1.26.3/bin:$PATH go test ./cmd/cosign/cli/verify -count=1`
- `git diff --check`

### DIFF
--- a/cmd/cosign/cli/verify/common.go
+++ b/cmd/cosign/cli/verify/common.go
@@ -200,7 +200,7 @@ func SetTrustedMaterial(ctx context.Context, trustedRootPath, certChain, caRoots
 }
 
 // PrintVerificationHeader prints boilerplate information after successful verification.
-func PrintVerificationHeader(ctx context.Context, imgRef string, co *cosign.CheckOpts, bundleVerified, fulcioVerified bool) {
+func PrintVerificationHeader(ctx context.Context, imgRef string, co *cosign.CheckOpts, bundleVerified, fulcioVerified, certVerified bool) {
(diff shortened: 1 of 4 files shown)


steiza (commented):
Hello there! Thanks for submitting this change.

These codepaths are a bit of a mess right now, because we're supporting several different use-cases that we're [in the middle of deprecating](https://github.com/sigstore/cosign/issues/4696).

Once we do, we'll always be using `sigstore-go` to verify, and then we can print this information out based on the actual verification results instead of trying to infer what the verification results were based on the verification settings. It's really tricky to reason through what it means when `co.SigVerifier` is or is not `nil` and `c.CertRef` is or is not `""`, for example.

So I think my preference here would be to wait a bit, finish deprecating the old codepaths, have all verification go through `sigstore-go` and use the verification results it returns to output what was verified.

Additionally, what we output from `verify` and `verify-blob` is quite different - this is another thing we could fix when the codepaths are simplified.

PR #4914, opened 2026-06-03
Update sigstore-go to v1.2.0

### LINKED ISSUE #3599: Attestations require uploading entire payload to rekor

codysoyland:
## Summary

Bumps sigstore-go from v1.1.4 to v1.2.0, which encodes DSSE envelopes as hashedrekord entries on Rekor v2.

- Bump sigstore-conformance action to v0.0.29 and drop the `message-digest-mismatch` xfail; cosign now correctly rejects bundles with mismatched artifact digests.
- Bump go-containerregistry to v0.21.6: close the `Bundle()` layer reader to release tokens for its new per-fetcher pull limiter (was deadlocking multi-bundle verifies), and switch kind-verify-attestation to `--allow-http-registry` since v0.21.6 narrowed the `.local` auto-insecure regex to `.localhost`.
- Adjust tests for transitive deps: timestamp-authority v2.1.2 requires `default-policy-oid`; tlog body parsing now requires a valid Rekor v1 or v2 body.

Closes #3599


Hayden-IO (commented):
Is our existing e2e test coverage sufficient or do we need another test?

codysoyland:
> Is our existing e2e test coverage sufficient or do we need another test?

It may be worth another test. I'll take a closer look.

.github/workflows/kind-verify-attestation.yaml
```diff
@@ -138,25 +138,25 @@ jobs:
 
     - name: Sign demoimage with cosign
       run: |
-        ./cosign sign --signing-config=${signingconfig} --trusted-root=${trustedroot} --yes --allow-insecure-registry ${demoimage} --identity-token ${OIDC_TOKEN}
+        ./cosign sign --signing-config=${signingconfig} --trusted-root=${trustedroot} --yes --allow-http-registry ${demoimage} --identity-token ${OIDC_TOKEN}
```
codysoyland, line 141: `--allow-http-registry` is the correct flag here. Previously it was automagically using http due to a `.local` domain. go-containerregistry v0.21.6 changed the regex that auto-detects a local registry as plain HTTP from .local to .localhost, so registry.local:5000 is no longer treated as insecure automatically.

pkg/cosign/verify_bundle_test.go
```diff
@@ -435,7 +435,19 @@ func (m *mockVerifierForBundle) VerifySignature(_, _ io.Reader, _ ...signature.V
 }
 
 func makeTlogEntry(t *testing.T, integratedTime int64) *tlog.Entry {
-	body := []byte(`{}`)
+	body := []byte(`{
```
codysoyland, line 438: Previously, nothing parsed this. The updated verification now decodes the entry body and requires a valid Rekor v1 or v2 structure, so an empty object fails.

test/e2e_tsa_test.go
```diff
@@ -405,6 +405,7 @@ func setUpTSAServerWithTLS(t *testing.T, td, timestampCACert, timestampServerKey
 	viper.Set("timestamp-signer", "memory")
 	viper.Set("timestamp-signer-hash", "sha256")
 	viper.Set("disable-ntp-monitoring", true)
+	viper.Set("default-policy-oid", "1.3.6.1.4.1.57264.2")
```
codysoyland, line 408: timestamp-authority v2.1.2 requires a TSA policy OID to be configured.

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -160,7 +160,7 @@ func TestVerifyBlob(t *testing.T) {
 	unexpiredCertPem, _ := cryptoutils.MarshalCertificateToPEM(unexpiredLeafCert)
 
 	expiredLeafCert, _ := test.GenerateLeafCertWithExpiration(identity, issuer,
-		time.Now().Add(-time.Hour), leafPriv, rootCert, rootPriv)
+		time.Now().Add(-10*time.Minute), leafPriv, rootCert, rootPriv)
```
codysoyland, line 163: These changes were made to keep the timestamp's genTime (ts.Time) within both cert validity windows (the TSA's, newly enforced in v2.1.2, and the Fulcio leaf's) while keeping the leaf expired.

pkg/oci/remote/signatures.go
```diff
@@ -76,6 +76,7 @@ func Bundle(ref name.Reference, opts ...Option) (*sgbundle.Bundle, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer layer0.Close()
```
codysoyland, line 79: go-containerregistry added a pull limiter and these unclosed writers caused it to deadlock.

codysoyland:
Added e2e tests for rekor2 in https://github.com/sigstore/cosign/pull/4916. I pushed them separately to keep this PR small, but I can merge them if preferred.

Also uncovered a data race: https://github.com/sigstore/cosign/pull/4917/changes

steiza (approved):
Nice work!

PR #4916, opened 2026-06-03
test(e2e): cover Rekor v2 hashedrekord behavior

codysoyland:
## Summary

Stacked on #4914. Adds e2e coverage for the Rekor v2 path that PR introduces (sigstore-go v1.2.0 encoding DSSE attestations as hashedrekord on Rekor v2).

- New `TestSignAttestVerifyRekorV2` round-trips sign + attest + verify against rekor-tiles (the only test that does so positively) and asserts each bundle's tlog entry is `hashedrekord/0.0.2`. This guards the v1.2.0 behavior change: attestations are now hashedrekord on Rekor v2 rather than the dsse type v1 emitted.
- Adds a `rekorV2URL` test constant.
- Fixes `TestSignRekorV2NoTSA`, which built its signing config with the v1 Rekor URL but `api-version=2`. It never actually reached rekor-tiles — it only happened to pass because it short-circuits on the missing-TSA check.

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -82,12 +82,14 @@ import (
 	"github.com/sigstore/cosign/v3/pkg/oci/mutate"
 	ociremote "github.com/sigstore/cosign/v3/pkg/oci/remote"
 	sigs "github.com/sigstore/cosign/v3/pkg/signature"
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	"github.com/sigstore/sigstore-go/pkg/root"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/payload"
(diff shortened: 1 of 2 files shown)


codysoyland:
This exposes a pre-existing data race: Fix here: https://github.com/sigstore/cosign/pull/4917

Hayden-IO:
@codysoyland Want to rebase, we'll get this in before cutting v3.1.0 to make sure there's no issues?

test/e2e_test.go
```diff
@@ -1284,6 +1286,155 @@ func TestSignRekorV2NoTSA(t *testing.T) {
 	}
 }
 
+// TestSignAttestVerifyRekorV2 exercises the full sign/attest/verify path
+// against rekor-tiles (Rekor v2). It asserts that both signatures and DSSE
+// attestations land as hashedrekord/0.0.2 tlog entries — which is what
+// sigstore-go v1.2.0+ produces on Rekor v2 (DSSE envelopes are encoded as
```
Hayden-IO: Sorry to nitpick comments but referencing a dependency version stands out. Can we clean up the comments in the test to add additional context when necessary?

codysoyland: No apology needed, thanks for reviewing!

Hayden-IO (approved):
LGTM, and I'm glad we have test coverage for the sign-with-tsa path as well now too

PR #4917, opened 2026-06-04
fix(verify): copy CheckOpts inside VerifyNewBundle to fix data race

codysoyland:
## Summary

`verifyImageAttestationsSigstoreBundle` fans out one goroutine per bundle, all sharing a single `*CheckOpts`. Inside each goroutine, `VerifyNewBundle -> rekorV2Bundle` writes `co.UseSignedTimestamps` when it sees a Rekor v2 entry, which races sibling goroutines reading `co` via `co.verificationOptions()`.

This fixes the race at the chokepoint every caller funnels through: `VerifyNewBundle` now copies `co` once and verifies against the copy, so the opportunistic write can't escape to a shared `*CheckOpts`. A shallow copy suffices — `rekorV2Bundle` only writes `UseSignedTimestamps` (a `bool`), and no other path under `VerifyNewBundle` mutates `co`'s pointer/map fields.

## Test

Adds `TestVerifyNewBundleConcurrentNoDataRace`, which fans out concurrent verifications of a Rekor v2 bundle against one shared `CheckOpts`. It passes normally and fails under `-race` if the per-call copy is removed.

### DIFF
--- a/pkg/cosign/verify_bundle.go
+++ b/pkg/cosign/verify_bundle.go
@@ -23,10 +23,15 @@ import (
 
 // VerifyNewBundle verifies a Sigstore bundle with the given parameters
 func VerifyNewBundle(_ context.Context, co *CheckOpts, artifactPolicyOption verify.ArtifactPolicyOption, bundle verify.SignedEntity) (*verify.VerificationResult, error) {
-	if err := rekorV2Bundle(bundle, co); err != nil {
+	// Copy co so rekorV2Bundle's UseSignedTimestamps write stays per-call and
+	// doesn't race a *CheckOpts shared across goroutines.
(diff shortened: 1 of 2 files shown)


steiza (approved):
Nice catch!

I don't love the fact that `rekorV2Bundle()` mutates `CheckOpts`, or that `VerifyNewBundle()` takes a pointer to `CheckOpts`, but I think this is a reasonable fix with the TODO you added.

codysoyland:
> I don't love the fact that `rekorV2Bundle()` mutates `CheckOpts`, or that `VerifyNewBundle()` takes a pointer to `CheckOpts`...

Same 😭

Hayden-IO:
Happy to chat about changing this at some point, since this was something i added originally

PR #4918, opened 2026-06-05
Remove unused signing code

aaronlew02:
#### Summary

This change removes signing code that remains unused as signing occurs exclusively via sigstore-go.

Tests are slightly refactored to eliminate reliance on unused signing code.


pkg/cosign/verify_test.go
```diff
@@ -2563,3 +2580,15 @@ func TestGetLocalBundles_InvalidPath(t *testing.T) {
 	assert.Nil(t, hash)
 	assert.Nil(t, bundles)
 }
+
+func getTimestampedSignature(sigBytes []byte, tsaClient *tsaMock.TSAClient) ([]byte, error) {
```
steiza, line 2584: Minor: should we move this into a helper so it can be shared between here and `test/e2e_test.go`?

🤷 I could go either way; it's a pretty small function.

steiza (approved):
Amazing! These giant reductions in code we have to maintain is exactly why we started this Cosign modernization effort ~2 years ago.

Unfortunately Go's dead code detection is pretty good, and so the resulting binary is pretty much the same size. 😅 But regardless, having less code to maintain and surface area for LLM-assisted vulnerability discovery is definitely a win.

pkg/cosign/verify_test.go
aaronlew02: I thought about combining the two `getTimestampedSignature` helpers but decided against it for several reasons:
1. These functions are helping test legacy (i.e., old-bundle) verification, which will be removed at some point. Keeping each `getTimestampedSignature` localized eases the deletion burden later.
2. Importing the same external helper increases coupling between e2e and unit tests.
3. The combined helper would need to accept the `client.TimestampAuthorityClient` interface to allow for mock inputs. This isn't much of a deterrent but adds complexity.

PR #4933, opened 2026-06-09
Fix build for Go version 1.26.3

Hayden-IO:
Needed to update the identity string as well. Also downgrade the Dockerfile version to match the release version, will bump all at once when there's a new golang-cross builder.


steiza (approved):
LGTM!

PR #4934, opened 2026-06-09
providers/filesystem: honor SIGSTORE_OIDC_TOKEN_FILE env var

dustinkirkland:
## Summary

The filesystem token provider currently reads from a hardcoded path (`/var/run/sigstore/cosign/oidc-token`). On container / Kubernetes workloads that's fine — the orchestrator bind-mounts the path at pod startup. On desktop or developer workstations, however, using this provider requires a one-time `sudo mkdir -p /var/run/sigstore/cosign && sudo chown \$USER` setup before anything writable can land there. That sudo step is the only privileged moment in an otherwise unprivileged "headless signing" workflow.

This PR lets `SIGSTORE_OIDC_TOKEN_FILE` override the default path with any user-writable location (e.g. `~/.sigstore/token`), so the provider can be used unprivileged on dev machines. Behavior is unchanged when the env var is unset — the historical hardcoded path remains the default.

## Why this matters

I'm prototyping a fully-headless gitsign signing flow against the public Fulcio for SSH-only dev workflows (where opening a graphical browser on the dev host is the original UX pain point — see https://github.com/sigstore/sigstore/blob/main/pkg/oauthflow/device.go). The flow:

```
sigstore-headless-login          # RFC 8628 device flow, writes JWT to disk
GITSIGN_TOKEN_PROVIDER=filesystem git commit ...
```

works end-to-end today with one wart: every user has to run `sudo mkdir` on every dev machine before they can write the token file. This PR removes that wart.

## Tests

```
$ go test ./pkg/providers/filesystem/...
ok  	github.com/sigstore/cosign/v3/pkg/providers/filesystem	0.002s
```

Three new tests cover the default path, env-var override, and Provide() reading from the env-var-pointed location. No existing tests changed.

## Backward compatibility

Strictly additive. The new env var is opt-in; every existing caller that doesn't set it sees the previous behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/providers/filesystem/filesystem.go
+++ b/pkg/providers/filesystem/filesystem.go
@@ -31,22 +31,37 @@ type filesystem struct{}
 var _ providers.Interface = (*filesystem)(nil)
 
 const (
-	// FilesystemTokenPath is the path to where we read an OIDC
-	// token from the filesystem.
+	// FilesystemTokenPath is the default path to where we read an OIDC
+	// token from the filesystem. Used when FilesystemTokenFileEnvVar is unset.
 	// nolint
 	FilesystemTokenPath = "/var/run/sigstore/cosign/oidc-token"
+
+	// FilesystemTokenFileEnvVar, when set, overrides FilesystemTokenPath with
+	// any user-writable location. This allows the filesystem provider to be
+	// used on desktop / dev workstations without the one-time sudo step
+	// required to create /var/run/sigstore/cosign.
+	FilesystemTokenFileEnvVar = "SIGSTORE_OIDC_TOKEN_FILE"
 )
+
+// tokenPath returns the path the filesystem provider reads the OIDC token
+// from: SIGSTORE_OIDC_TOKEN_FILE if set, else the historical default.
+func tokenPath() string {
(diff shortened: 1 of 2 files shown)


pkg/providers/filesystem/filesystem.go
```diff
@@ -31,22 +31,37 @@ type filesystem struct{}
...
+	// FilesystemTokenPath is the default path to where we read an OIDC
+	// token from the filesystem. Used when FilesystemTokenFileEnvVar is unset.
 	// nolint
 	FilesystemTokenPath = "/var/run/sigstore/cosign/oidc-token"
+
+	// FilesystemTokenFileEnvVar, when set, overrides FilesystemTokenPath with
+	// any user-writable location. This allows the filesystem provider to be
+	// used on desktop / dev workstations without the one-time sudo step
+	// required to create /var/run/sigstore/cosign.
+	FilesystemTokenFileEnvVar = "SIGSTORE_OIDC_TOKEN_FILE"
```
Hayden-IO, line 43: You could also pass an ID token via the `SIGSTORE_ID_TOKEN` env var. Would this be sufficient?

PR #4947, opened 2026-06-14
check bundle layer size before reading in remote.Bundle

alhudz:
Repro: point `cosign verify` at an image whose `.sig`-style bundle reference resolves to an image with one `application/vnd.dev.sigstore.bundle` layer that declares (or expands to) a large size. `GetBundles` -> `ociremote.Bundle` reads it before any validation.
Cause: `remote.Bundle` calls `layers[0].Uncompressed()` + `io.ReadAll` with no size cap, while the sibling registry readers (`pkg/oci/internal/signature/layer.go`, `pkg/oci/remote/remote.go`, `pkg/oci/static/file.go`) all gate on `payloadsize.CheckSize` first, so a hostile registry can make cosign allocate unbounded memory on a layer it has not yet trusted.
Fix: apply the same `payloadsize.CheckSize(uint64(layer.Size()))` check before reading, honouring `COSIGN_MAX_ATTACHMENT_SIZE` like the other layers do.

Added a regression test that fails on `main` (the read path is reached) and passes with the cap in place.

### DIFF
--- a/pkg/oci/remote/signatures_test.go
+++ b/pkg/oci/remote/signatures_test.go
@@ -17,7 +17,9 @@ package remote
 
 import (
 	"errors"
+	"io"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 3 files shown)


pkg/oci/remote/signatures_test.go
```diff
@@ -17,16 +17,61 @@ package remote
...
+}
+func (l *oversizeLayer) Compressed() (io.ReadCloser, error) {
+	return nil, errors.New("Compressed should not be called")
+}
+func (l *oversizeLayer) Uncompressed() (io.ReadCloser, error) {
+	return nil, errors.New("Uncompressed should not be called once size is capped")
+}
+
+func TestBundleRejectsOversizeLayer(t *testing.T) {
+	ri := remote.Image
```
steiza: I don't think we want to be patching other modules as part of a unit test - maybe this would be better off in `e2e_test.go` where we have a container registry running?

steiza (commented):
Nice catch! I suppose if we have size limits, we should be enforcing them consistently. We do know that some folks have very large bundles that include SBOMs though that are > 200 MB. Of course Rekor v1 has size limits as well.

Regardless, I think we want to test this change in `e2e_test.go` instead of in the unit test.

pkg/oci/remote/signatures_test.go
alhudz: Moved it to `e2e_test.go` as `TestBundleLayerSizeCap`, so it runs against the real registry from `reg(t)` and drops the fake `oversizeLayer` / `remoteImage` patching. It signs a new-format bundle, then lowers `COSIGN_MAX_ATTACHMENT_SIZE` below the layer size and checks `ociremote.Bundle` returns the limit error. Fails on `main` (the read path is reached), passes with the cap in place.

alhudz:
On the large SBOM bundles: this just reuses the same `COSIGN_MAX_ATTACHMENT_SIZE` knob the other registry readers already honour, so anyone pulling >200MB content sets that env var and it covers this path too. Default stays 128MiB, same as the siblings, so the only change is that an untrusted oversize layer gets rejected before the `io.ReadAll` instead of after.

alhudz:
gentle ping

cmurphy:
Something misfired in the verify-attestation check, can you rebase to have it run again?

alhudz:
Rebased onto current `main`, just the same two commits replayed with no other changes. The failed `verify-attestation` jobs were cancelled during the `Install cluster + sigstore` step rather than failing on a test, so a rerun should clear them. `TestBundleLayerSizeCap` still passes locally with `-tags e2e` against the fake registry.

The new runs are sitting at `action_required`, so they need a workflow approval from your side before they start.

cmurphy:
I believe https://github.com/sigstore/cosign/pull/5052 covers this case and I already merged it, so I will close this one. Feel free to reopen if I have missed something.

PR #4959, opened 2026-06-18
Require new bundle format for signing and verification

aaronlew02:
#### Summary

This change mandates the new bundle format for signing and verification.

For verification, this enables the exclusive use of `sigstore-go` and requires a trusted root. For signing, this disables individual service inputs (i.e., outside of a signing config) and detached outputs.

#### Release Note

* Signers requiring services  **MUST** use a signing config (either from TUF via `--use-signing-config` [`true` by default] or explicitly provided via `--signing-config`).
    * This applies to **ALL** keyless signers.
    * This applies to key-based signers requesting a certificate, RFC 3161 timestamp, and/or transparency log entry.
* Signers **NOT** requiring services **MUST** set `--use-signing-config=false`.
    * This applies to offline key-based signers.
* Verifiers validating service-provided verification material **MUST** use a trusted root (either fetched from TUF automatically or explicitly provided via `--trusted-root`).
    * This applies to **ALL** keyless verifiers.
    * This applies to key-based verifiers with a certificate, RFC 3161 timestamp, and/or transparency log entry.

* Several flags have been removed:
    * Service configuration flags obsoleted by the signing config and trusted root:
        * `--rekor-url` (for signing, attestation, and verification only), `--fulcio-url`, `--oidc-issuer`, `--timestamp-server-url`, `--timestamp-certificate-chain`, `--ca-roots`, `--ca-intermediates`, `--certificate-chain`, `--certificate`, `--tlog-upload`, `--offline`
    * Flags obsoleted by the new bundle format:
        * `--new-bundle-format`
        * `--rekor-entry-type`
        * Input overrides: `--record-creation-timestamp`, `--signature`/`--signature-path`, `--payload`, `--signature-digest-algorithm`
        * Detached outputs: `--sct`, `--rfc3161-timestamp`/`--rfc3161-timestamp-bundle`, `--b64`, `--output`/`--output-signature`/`--output-payload`/`--output-attestation`/`--output-certificate`
    * Flags obsoleted by the shift to OCI 1.1 Referrers:
        * `--replace`, `--attachment`, `--experimental-oci11`, `--sign-container-identity`
    * Protocol and verification flow adjustments:
        * `--insecure-skip-verify` (SCT verification is no longer performed during signing)
        * `--private-infrastructure` (Obsoleted in favor of `--insecure-ignore-tlog`)

* Service configuration environment variables are no longer used in production (but remain used in testing):
    * `SIGSTORE_ROOT_FILE`, `SIGSTORE_REKOR_PUBLIC_KEY`, `SIGSTORE_CT_LOG_PUBLIC_KEY_FILE`, `SIGSTORE_TSA_CERTIFICATE_FILE`

* **ALL** signers using a signing config containing at least one transparency log must agree to the immutable record privacy statement.


aaronlew02:
# Testing

The following commands can be used to test the changes in this PR.

## Setup

Before running the tests, ensure that you have generated a local test keypair (which is used for key-based tests):

```shell
# Generate a keypair (writes to cosign.key and cosign.pub)
cosign generate-key-pair
```


## 1\. TUF-provided Signing Config & Trusted Root

Verify that a TUF-provided signing config and trusted root are resolved correctly for keyless operations and key-based, certificate-backed operations:

### Keyless Operations

#### Image signing & verification
```shell
# Sign using TUF signing config
cosign sign --use-signing-config=true [IMAGE_URI]

# Verify using TUF trusted root
cosign verify --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
```

#### Blob signing & verification
```shell
# Sign a blob to generate a bundle
cosign sign-blob --use-signing-config=true --bundle blob.bundle [BLOB_FILE]

# Verify the blob using the bundle and TUF trusted root
cosign verify-blob --bundle blob.bundle --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]
```

#### Image attestation & verification
```shell
# Attest the image
cosign attest --use-signing-config=true --predicate [PREDICATE_FILE] [IMAGE_URI]

# Verify the attestation
cosign verify-attestation --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
```

#### Blob attestation & verification
```shell
# Attest a blob to generate a bundle
cosign attest-blob --use-signing-config=true --predicate [PREDICATE_FILE] --bundle attest.bundle [BLOB_FILE]

# Verify the blob attestation
cosign verify-blob-attestation --bundle attest.bundle --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]
```

### Key-based / Certificate-backed Operations

#### Key-based, certificate-backed signing & verification
```shell
# Sign using local key and request Fulcio certificate using TUF signing config
cosign sign --key cosign.key --issue-certificate --use-signing-config=true [IMAGE_URI]

# Verify using TUF trusted root and identity checks
cosign verify --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
```

#### Key-based signing with transparency log and RFC 3161 timestamp
```shell
# Sign a blob using key and request log entry/timestamp
cosign sign-blob --key cosign.key --use-signing-config=true --bundle blob.bundle [BLOB_FILE]

# Verify the blob bundle (verifying the signature, log proof, and timestamp against TUF trusted root)
cosign verify-blob --key cosign.pub --bundle blob.bundle [BLOB_FILE]
```

---

## 2\. Explicit Local Signing Config & Trusted Root

Verify that explicitly provided local config and trusted root files are loaded and utilized correctly:

### Keyless Operations

#### Image signing & verification
```shell
# Sign using a local signing config file
cosign sign --signing-config [PATH_TO_SIGNING_CONFIG] [IMAGE_URI]

# Verify using an explicit trusted root file
cosign verify --trusted-root [PATH_TO_TRUSTED_ROOT] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
```

#### Blob signing & verification
```shell
# Sign a blob using a local signing config file
cosign sign-blob --signing-config [PATH_TO_SIGNING_CONFIG] --bundle blob.bundle [BLOB_FILE]

# Verify the blob using an explicit trusted root file
cosign verify-blob --trusted-root [PATH_TO_TRUSTED_ROOT] --bundle blob.bundle --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]
```

#### Image attestation & verification
```shell
# Attest the image using a local signing config file
cosign attest --signing-config [PATH_TO_SIGNING_CONFIG] --predicate [PREDICATE_FILE] [IMAGE_URI]

# Verify the attestation using an explicit trusted root file
cosign verify-attestation --trusted-root [PATH_TO_TRUSTED_ROOT] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
```

#### Blob attestation & verification
```shell
# Attest a blob using a local signing config file
cosign attest-blob --signing-config [PATH_TO_SIGNING_CONFIG] --predicate [PREDICATE_FILE] --bundle attest.bundle [BLOB_FILE]

# Verify the blob attestation using an explicit trusted root file
cosign verify-blob-attestation --trusted-root [PATH_TO_TRUSTED_ROOT] --bundle attest.bundle --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]
```

### Key-based Operations

#### Key-based, certificate-backed signing & verification
```shell
# Sign using local key and request Fulcio certificate using local signing config
cosign sign --key cosign.key --issue-certificate --signing-config [PATH_TO_SIGNING_CONFIG] [IMAGE_URI]

# Verify using local trusted root file and identity checks
cosign verify --trusted-root [PATH_TO_TRUSTED_ROOT] --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [IMAGE_URI]
```

#### Key-based signing with transparency log and RFC 3161 timestamp
```shell
# Sign a blob using key and request log entry/timestamp via local signing config
cosign sign-blob --key cosign.key --signing-config [PATH_TO_SIGNING_CONFIG] --bundle blob.bundle [BLOB_FILE]

# Verify the blob bundle against local trusted root file
cosign verify-blob --key cosign.pub --trusted-root [PATH_TO_TRUSTED_ROOT] --bundle blob.bundle [BLOB_FILE]
```

### Service Omission Tests (Config-driven behavior)

Verify that omitting specific services in the signing config successfully skips their use during signing:

#### Omitting Rekor (Skips log upload)
```shell
# 1. Generate a signing config file without Rekor configured
cosign signing-config create --with-default-services --no-default-rekor --out signing-config-no-rekor.json

# 2. Sign a blob (succeeds without attempting log upload; generated bundle will not contain a log entry)
cosign sign-blob --signing-config signing-config-no-rekor.json --bundle blob-no-rekor.bundle [BLOB_FILE]

# 3. Verify the bundle (should FAIL by default due to missing log entry/SET proof)
cosign verify-blob --trusted-root [PATH_TO_TRUSTED_ROOT] --bundle blob-no-rekor.bundle --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]

# 4. Verify the bundle with tlog bypassed (should SUCCEED)
cosign verify-blob --trusted-root [PATH_TO_TRUSTED_ROOT] --bundle blob-no-rekor.bundle --insecure-ignore-tlog --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]
```

#### Omitting TSA (Skips RFC 3161 timestamping)
```shell
# 1. Generate a signing config file without TSA configured
cosign signing-config create --with-default-services --no-default-tsa --out signing-config-no-tsa.json

# 2. Sign a blob (succeeds without requesting timestamp; generated bundle will not contain a timestamp)
cosign sign-blob --signing-config signing-config-no-tsa.json --bundle blob-no-tsa.bundle [BLOB_FILE]

# 3. Verify the bundle (should SUCCEED; falls back to using the tlog integrated time to verify certificate expiry)
cosign verify-blob --trusted-root [PATH_TO_TRUSTED_ROOT] --bundle blob-no-tsa.bundle --certificate-identity [IDENTITY] --certificate-oidc-issuer [ISSUER] [BLOB_FILE]
```

---

## 3\. Simple Key-based Signing & Verification (No services)

Verify that simple key-based operations (which do not interact with Fulcio, Rekor, or a timestamp authority) work cleanly:

```shell
# Sign an image using a private key (no signing config or OCI upload)
cosign sign --key cosign.key [IMAGE_URI]

# Verify the image using the public key
cosign verify --key cosign.pub [IMAGE_URI]

# Sign a blob using a private key (generates a simple bundle without log/timestamp verification material)
cosign sign-blob --key cosign.key --bundle blob.bundle [BLOB_FILE]

# Verify the blob using the public key and bundle
cosign verify-blob --key cosign.pub --bundle blob.bundle [BLOB_FILE]
```

---

## 4\. CLI Flag Validations (Negative Testing)

Verify that the CLI rejects removed flags and enforces mandatory new arguments:

```shell
# 1. Rejects removed flags (should fail with unknown flag errors)
cosign sign --rekor-url [URL] [IMAGE_URI]
cosign sign --timestamp-server-url [URL] [IMAGE_URI]
cosign verify --rekor-url [URL] [IMAGE_URI]
cosign verify --signature [SIG_PATH] --payload [PAYLOAD_PATH] [IMAGE_URI]
cosign verify-blob --signature [SIG_PATH] [BLOB_FILE]

# 2. Enforces signing configuration for keyless and certificate-based signing/attestation (should fail with specific keyless or certificate-based error messages)
cosign sign --use-signing-config=false [IMAGE_URI]
cosign sign --key cosign.key --issue-certificate --use-signing-config=false [IMAGE_URI]
cosign sign-blob --use-signing-config=false --bundle blob.bundle [BLOB_FILE]
cosign attest --use-signing-config=false [IMAGE_URI]
cosign attest-blob --use-signing-config=false --bundle attest.bundle [BLOB_FILE]

# 3. Enforces bundle for blob signing and attestation (should fail with error requiring bundle)
cosign sign-blob [BLOB_FILE]
cosign attest-blob [BLOB_FILE]
```

cmd/cosign/cli/attest/attest_blob.go
```diff
@@ -147,89 +133,22 @@ func (c *AttestBlobCommand) Exec(ctx context.Context, artifactPath string) error
...
-		var err error
-		c.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(c.KeyOpts, c.TlogUpload)
-		if err != nil {
-			return fmt.Errorf("creating signing config: %w", err)
-		}
+		c.SigningConfig = signcommon.NewEmptySigningConfig()
 	}
 
-	bundleBytes, _, _, err := signcommon.NewAttestationBundle(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
+	_, err = signcommon.ShouldUploadToTlog(ctx, c.KeyOpts, nil, len(c.SigningConfig.RekorLogURLs()) > 0)
```
Hayden-IO, line 139: This is throwing away the result of `ShouldUploadToTlog`, so the attestation would be unconditionally uploaded to the transparency log.

aaronlew02: Whether the attestation would be uploaded to the transparency log depends on:
* The value of `len(c.SigningConfig.RekorLogURLs()) > 0`, which if `true` would enable the upload confirmation prompt.
* Whether the user accepts the prompt. If they don't, execution stops with an error. If they do, the signing config goes to `sigstore-go`, which would upload because a log is present.

Hayden-IO: Ah, so this is just about presenting the prompt? I see we do the same in sign/attest but override what's in the signing config, what's the reason for that?

aaronlew02: That's to support users who want to sign a private container image without uploading to the transparency log.

If a user uses a signing config with a Rekor URL, but a private repository is detected:
1. The user will be prompted to confirm that they wish to upload to Rekor despite the repository being private. If the user disagrees, `shouldUpload` returns false, and execution proceeds without a Rekor upload. If the user agrees...
2. The user will receive the Immutable Records confirmation prompt. If the user disagrees, execution stops with an error. If the user agrees, `shouldUpload` returns true and the signing config goes to `sigstore-go`, which would upload because a log is present.

This matches current functionality.

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -18,82 +18,53 @@ package verify
...
 	CertGithubWorkflowSHA        string
 	CertGithubWorkflowName       string
 	CertGithubWorkflowRepository string
 	CertGithubWorkflowRef        string
 	IgnoreSCT                    bool
-	SCTRef                       string
 	Offline                      bool
 	UseSignedTimestamps          bool
 	IgnoreTlog                   bool
-	HashAlgorithm                crypto.Hash
```
Hayden-IO, line 71: To confirm, hash algorithm will now only be determined by the default mapping between signing scheme and hash alg?

cmd/cosign/cli/verify/verify_blob.go
```diff
@@ -113,210 +84,51 @@ func (c *VerifyBlobCmd) Exec(ctx context.Context, blobRef string) error {
...
-			return err
-		}
-		opts = append(opts, static.WithCertChain(certPEM, chainPEM))
+		artifactPolicyOption = sgverify.WithArtifactDigest(alg, digest)
+	} else {
+		artifactPolicyOption = sgverify.WithArtifact(bytes.NewReader(blobBytes))
 	}
 
-	sig, err := base64signature(c.SigRef, c.BundlePath)
+	_, err = cosign.VerifyNewBundle(ctx, co, artifactPolicyOption, bundle)
```
Hayden-IO, line 123: For a later PR - we should implement the change mentioned in https://github.com/sigstore/cosign/blob/98ff280b1942d3d49dbe8b34f41ad6e8154f4d88/pkg/cosign/verify_bundle.go#L25-L29. We should move the function into an internal package as well when we do. I also think we could drop the first return value since no caller is using it.

cmd/cosign/cli/verify/verify_blob.go
Hayden-IO: And, how does this interact with when you parse out the algorithm from the payloadDigest?

cmd/cosign/cli/sign/sign_blob.go
```diff
@@ -53,26 +49,19 @@ func getPayload(ctx context.Context, payloadPath string, hashFunction crypto.Has
...
+func SignBlobCmd(ctx context.Context, ro *options.RootOptions, ko options.KeyOpts, payloadPath, certPath, certChainPath string) error {
 	var payload internal.HashReader
 
 	ctx, cancel := context.WithTimeout(ctx, ro.Timeout)
 	defer cancel()
 
-	var shouldUpload bool
-	var err error
-
 	if ko.SigningConfig == nil {
```
Hayden-IO, line 58: In what case will this be nil? Currently, you can either provide a signing config, or specify --use-signing-config to fetch it from TUF. If we're removing that latter flag, then the signing config would always populated, either by TUF or with one explicitly provided (which could be empty).

(this might be answered in a later commit, feel free to point to that if so)

cmd/cosign/cli/sign/sign_blob_test.go
```diff
@@ -119,14 +108,17 @@ func TestSignBlobCmd(t *testing.T) {
...
 	keyOpts.SigningConfig = sc
-	sigBytes, err := SignBlobCmd(t.Context(), rootOpts, keyOpts, blobPath, "", "", true, "", "", true)
-	if err != nil {
+	if err := SignBlobCmd(t.Context(), rootOpts, keyOpts, blobPath, "", ""); err != nil {
 		t.Fatalf("unexpected error %v", err)
 	}
-	decodedSig, err := base64.StdEncoding.DecodeString(string(sigBytes))
-	if err != nil {
-		t.Fatalf("failed to decode base64 signature: %v", err)
+	var b1 struct {
```
Hayden-IO, line 114: nit, a better var name would help readability, even if this is just an inlined struct

cmd/cosign/cli/sign/sign_blob_test.go
```diff
@@ -119,14 +108,17 @@ func TestSignBlobCmd(t *testing.T) {
...
+		MessageSignature struct {
+			Signature string `json:"signature"`
+		} `json:"messageSignature"`
 	}
+	bytes1, _ := os.ReadFile(bundlePath)
```
Hayden-IO, line 119: even though it's unnecessary, i'd check the second return val and fail fatally if not nil

cmd/cosign/cli/signblob.go
```diff
@@ -26,44 +25,54 @@ import (
...
 
 func SignBlob() *cobra.Command {
 	o := &options.SignBlobOptions{}
-	viper.RegisterAlias("output", "output-signature")
 
 	cmd := &cobra.Command{
 		Use:   "sign-blob",
-		Short: "Sign the supplied blob, outputting the base64-encoded signature to stdout",
-		Example: `  cosign sign-blob --key <key path>|<kms uri> <blob>
+		Short: "Sign the supplied blob, outputting the bundle to a file",
```
Hayden-IO, line 35: bundle -> "signature bundle"

cmd/cosign/cli/signblob.go
```diff
@@ -26,44 +25,54 @@ import (
...
+		Example: `  cosign sign-blob --key <key path>|<kms uri> --bundle <bundle.json> <blob>
 
   # sign a blob with a local key pair file
-  cosign sign-blob --key cosign.key <FILE>
+  cosign sign-blob --key cosign.key --bundle <bundle.json> <FILE>
```
Hayden-IO, line 39: since we're here, can you also add an example of calling sign-blob without a key? i think this was a side effect of keys being the default back in v1

Hayden-IO: also in the example, we should denote --key is optional

cmd/cosign/cli/signblob.go
```diff
@@ -26,44 +25,54 @@ import (
...
 				return &options.KeyParseError{}
 			}
 
-			if o.NewBundleFormat && o.BundlePath == "" {
-				return fmt.Errorf("must specify --bundle with --new-bundle-format")
+			if o.BundlePath == "" {
+				return fmt.Errorf("please specify --bundle")
+			}
+
+			var signType string
```
Hayden-IO, line 66: what's the purpose of `signType`?

cmd/cosign/cli/signblob.go
```diff
@@ -93,43 +102,29 @@ func SignBlob() *cobra.Command {
...
 				IDToken:                        o.Fulcio.IdentityToken,
 				FulcioAuthFlow:                 o.Fulcio.AuthFlow,
-				InsecureSkipFulcioVerify:       o.Fulcio.InsecureSkipFulcioVerify,
-				RekorURL:                       o.Rekor.URL,
-				OIDCIssuer:                     o.OIDC.Issuer,
 				OIDCClientID:                   o.OIDC.ClientID,
 				OIDCClientSecret:               oidcClientSecret,
 				OIDCRedirectURL:                o.OIDC.RedirectURL,
 				OIDCDisableProviders:           o.OIDC.DisableAmbientProviders,
+				OIDCProvider:                   o.OIDC.Provider,
```
Hayden-IO, line 111: was this missing before? if so, can we split out that change?

cmd/cosign/cli/signblob.go
```diff
@@ -93,43 +102,29 @@ func SignBlob() *cobra.Command {
...
-				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL, o.TrustedRootPath, o.TlogUpload,
-				o.NewBundleFormat, o.BundlePath, o.Key, o.IssueCertificate,
-				o.Output, "", o.OutputCertificate, "", o.OutputSignature, o.RFC3161TimestampPath); err != nil {
+				o.TrustedRootPath, o.Key); err != nil {
 				return err
 			}
 
 			for _, blob := range args {
-				// TODO: remove when the output flag has been deprecated
-				if o.Output != "" {
```
Hayden-IO, line 127: are we removing this flag in another commit/PR?

Hayden-IO (commented):
lgtm

cmd/cosign/cli/attest_blob.go
```diff
@@ -71,48 +83,35 @@ func AttestBlob() *cobra.Command {
...
-				Timeout:           ro.Timeout,
-				RekorEntryType:    o.RekorEntryType,
+				KeyOpts:       ko,
+				CertPath:      o.Cert,
+				CertChainPath: o.CertChain,
+				ArtifactHash:  o.Hash,
+				PredicateType: o.Predicate.Type,
+				PredicatePath: o.Predicate.Path,
+				StatementPath: o.Predicate.Statement,
+				Timeout:       ro.Timeout,
```
Hayden-IO, line 114: For another change, should we pass through timeout for sign/sign-blob as well?

Hayden-IO (commented):
The comments left on the sign-blob commit are relevent here as well, but otherwise lgtm

cmd/cosign/cli/verify/verify_attestation.go
```diff
@@ -124,65 +109,29 @@ func (c *VerifyAttestationCommand) Exec(ctx context.Context, images []string) (e
...
-	err = SetLegacyClientsAndKeys(ctx, c.IgnoreTlog, shouldVerifySCT(c.IgnoreSCT, c.KeyRef, c.Sk), keylessVerification(c.KeyRef, c.Sk), c.RekorURL, c.TSACertChainPath, c.CertChain, c.CARoots, c.CAIntermediates, co)
-	if err != nil {
-		return fmt.Errorf("setting up clients and keys: %w", err)
-	}
-
-	// User provides a key or certificate. Otherwise, verification requires a Fulcio certificate
-	// provided in an attached bundle or OCI annotation. LoadVerifierFromKeyOrCert must be called
-	// after initializing trust material in order to verify certificate chain.
+	// User provides a key. Otherwise, verification requires a Fulcio certificate
+	// provided in an attached bundle or OCI annotation.
```
Hayden-IO, line 127: @steiza I didn't think about this! What should we be doing about verification where the signature is stored in an annotation? Do we deprecate this with v4, or wait?

I'm OK with waiting, this is unrelated to the sigstore-go migration.

Hayden-IO (commented):
Overall, lgtm

cmd/cosign/cli/attest/attest.go
```diff
@@ -135,115 +118,34 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
...
-
-	// We don't actually need to access the remote entity to attach things to it
-	// so we use a placeholder here.
-	se := ociremote.SignedUnknown(digest, ociremoteOpts...)
-
-	ddVerifier, err := signature.LoadVerifier(pubKey, signcommon.ProtoHashAlgoToHash(hashAlgProto))
-	if err != nil {
-		return fmt.Errorf("loading verifier: %w", err)
-	}
-	dd := cremote.NewDupeDetector(ddVerifier)
```
Hayden-IO, line 230: This should be outdated at this point, but I just want to explicitly confirm that there's no reason to keep this around. Duplications aren't relevant because we're not attaching a signature multiple times to an annotation. I'm not certain why record creation timestamp was added though, that's the one that gives me pause.

cc @steiza

cmd/cosign/cli/attest.go
```diff
@@ -31,7 +31,7 @@ func Attest() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "attest",
 		Short: "Attest the supplied container image",
-		Example: `  cosign attest --key <key path>|<kms uri> [--predicate <path>] [--a key=value] [--no-upload=true|false] [--record-creation-timestamp=true|false] [--f] [--r] <image uri>
+		Example: `  cosign attest --key <key path>|<kms uri> [--predicate <path>] [--no-upload=true|false] [--yes] <image uri>
```
Hayden-IO, line 34: --key should be optional

cmd/cosign/cli/attest.go
```diff
@@ -61,16 +61,24 @@ func Attest() *cobra.Command {
...
 
 		Args:             cobra.MinimumNArgs(1),
 		PersistentPreRun: options.BindViper,
 		PreRunE: func(_ *cobra.Command, _ []string) error {
-			if o.NewBundleFormat && o.NoUpload && o.BundlePath == "" {
-				return fmt.Errorf("must enable upload to the OCI registry or specify a local --bundle path with --new-bundle-format")
+			if o.NoUpload && o.BundlePath == "" {
+				return fmt.Errorf("must enable upload to the OCI registry or specify a local --bundle path")
+			}
+			var attestType string
```
Hayden-IO, line 72: Same here, what's `attestType` for?

cmd/cosign/cli/verify/verify.go
```diff
@@ -236,27 +152,15 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-				if err == nil {
-					verified = verifiedOutput
-				}
-			} else {
-				ref, err = sign.GetAttachedImageRef(ref, c.Attachment, ociremoteOpts...)
-				if err != nil {
-					return fmt.Errorf("resolving attachment type %s for image %s: %w", c.Attachment, img, err)
-				}
+			// OCI bundle always contains attestation
+			verified, bundleVerified, err = cosign.VerifyImageAttestations(ctx, ref, co, c.NameOptions...)
```
Hayden-IO, line 156: Outside of the scope for this PR, but verify and verify-attestation are now looking very similar, so we might be able to merge them as part of v4.

cmd/cosign/cli/verify.go
```diff
@@ -80,54 +80,31 @@ against the transparency log.`,
 		Args:             cobra.MinimumNArgs(1),
 		PersistentPreRun: options.BindViper,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if o.CommonVerifyOptions.PrivateInfrastructure {
```
Hayden-IO, line 83: if we're keeping IgnoreTlog, can we keep PrivateInfrastructure? This was just meant as an alias.

Hayden-IO (commented):
overall lgtm, other comments from other threads apply here

cmd/cosign/cli/signcommon/common.go
```diff
@@ -598,64 +560,18 @@ func NewLegacyBundleFromProtoBundleComponents(bc *BundleComponents) ([]byte, err
...
-		tsaConfig = root.ServiceConfiguration{
-			Selector: prototrustroot.ServiceSelector_ANY,
-			Count:    1,
-		}
-	}
-
-	return root.NewSigningConfig(
+// NewEmptySigningConfig returns a signing config with no services configured.
+func NewEmptySigningConfig() *root.SigningConfig {
+	sc, _ := root.NewSigningConfig(
```
Hayden-IO, line 565: nit, i'd add a comment for why it's ok to not check the returned error value

Hayden-IO (commented):
This is the first of the commits where I'd suggest we need to split this out into its own PR, because this commit does a lot and it's not as obvious that there aren't subtle changes, in comparison to the other commits where for the most part we're just dropping the `!bundle{}` block.

Are there any functions that could be added in order to support other changes in this PR, and then in later PRs, we do smaller refactors removing dead code?

pkg/cosign/verify.go
```diff
@@ -1865,7 +992,7 @@ func getLocalBundleDescriptors(path string) ([]bundleDescriptor, *v1.Hash, error
 	// Find the target image digest from the index manifest
 	var targetDigest v1.Hash
 	for _, m := range manifest.Manifests {
-		if val, ok := m.Annotations["kind"]; ok && val == "dev.cosignproject.cosign/image" {
+		if val, ok := m.Annotations["kind"]; ok && (val == "dev.cosignproject.cosign/image" || val == "dev.cosignproject.cosign/imageIndex") {
```
Hayden-IO, line 995: Can we move small changes like this out into their own PRs with their own tests?

pkg/cosign/verify.go
```diff
@@ -138,9 +124,6 @@ type CheckOpts struct {
 	IgnoreSCT bool
 	// Detached SCT. Optional, as the SCT is usually embedded in the certificate.
 	SCT []byte
```
Hayden-IO, line 126: SCT fields no longer needed?

pkg/cosign/verify_test.go
```diff
@@ -96,1565 +72,16 @@ func (m *mockVerifier) VerifySignature(signature, message io.Reader, opts ...sig
...
-
-	// Grab the CTLog public keys again so we get them from env.
-	co.CTLogPubKeys, err = GetCTLogPubs(context.Background())
-	if err != nil {
-		t.Fatalf("Failed to get CTLog public keys from TUF: %v", err)
-	}
-	_, err = ValidateAndUnpackCert(chain[0], co)
-	if err != nil {
-		t.Errorf("ValidateAndUnpackCert expected no error, got err = %v", err)
+func appendSlices(slices [][]byte) []byte {
```
Hayden-IO, line 75: Can also use `bytes.Join` which should do the same thing

Hayden-IO (commented):
Here is well, ideally this commit would only drop dead code, and so we can take care of that in its own PR.

pkg/cosign/verify_test.go
```diff
@@ -96,1565 +72,16 @@ func (m *mockVerifier) VerifySignature(signature, message io.Reader, opts ...sig
...
-		IgnoreSCT:  true,
-	}
-
-	_, err := ValidateAndUnpackCert(leafCert, co)
-	require.Contains(t, err.Error(), "none of the expected identities matched what was in the certificate")
-	err = CheckCertificatePolicy(leafCert, co)
-	require.Contains(t, err.Error(), "none of the expected identities matched what was in the certificate")
-}
-
-func TestValidateAndUnpackCertInvalidGithubWorkflowTrigger(t *testing.T) {
```
Hayden-IO, line 1247: I believe we are losing some test coverage with these being removed. `verificationOptions()` (pkg/cosign/verify.go:213-219) still builds a verification policy with these options. Can we make sure that we have matching test coverage, either in a e2e test or with bundle verification?

pkg/cosign/verify_test.go
```diff
@@ -96,1565 +72,16 @@ func (m *mockVerifier) VerifySignature(signature, message io.Reader, opts ...sig
...
-	co := &CheckOpts{
-		TrustedMaterial: trustedRoot,
-		IgnoreSCT:       true,
-		Identities:      []Identity{{Subject: subject, Issuer: oidcIssuer}},
-	}
-	_, err = ValidateAndUnpackCert(leafCert, co)
-	assert.NoError(t, err)
-}
-
-func TestValidateAndUnpackCertWithSCT(t *testing.T) {
```
Hayden-IO, line 1609: Do we have any tests when IgnoreSCT is set to false, and that a certificate without an embedded SCT is rejected? This would cover pkg/cosign/verify.go:242-245.

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -1000,807 +332,7 @@ func makeLocalNewBundle(t *testing.T, sig []byte, digest [32]byte) string {
 	return bundlePath
 }
 
-func TestVerifyBlobCmdWithBundle(t *testing.T) {
```
Hayden-IO, line 1003: I think we have coverage for the success case, but this drops coverage for failing cases. Can we either update these tests to use sigstore-go's mocks for verification, or confirm we have adequate test coverage elsewhere?

cmd/cosign/cli/verify/verify_blob_test.go
```diff
@@ -1000,807 +332,7 @@ func makeLocalNewBundle(t *testing.T, sig []byte, digest [32]byte) string {
...
-			CheckClaims:   false, // Intentionally false. This checks the subject claim. This is tested in verify_blob_attestation_test.go
-			PredicateType: "customFoo",
-		}
-		if err := cmd.Exec(context.Background(), blobPath); err != nil {
-			t.Fatal(err)
-		}
-	})
-}
-
-func TestVerifyBlobCmdInvalidRootCA(t *testing.T) {
```
Hayden-IO, line 1520: Similarly here, there's a number of failure edge cases we should confirm are tested.

cmd/cosign/cli/verify/common.go
```diff
@@ -183,18 +76,9 @@ func SetTrustedMaterial(ctx context.Context, trustedRootPath, certChain, caRoots
 	if verifyOnlyWithKey {
```
Hayden-IO, line 76: Do we have any tests for this case?

pkg/cosign/verify_test.go
```diff
@@ -96,1565 +72,16 @@ func (m *mockVerifier) VerifySignature(signature, message io.Reader, opts ...sig
...
-	if err != nil {
-		t.Errorf("ValidateAndUnpackCert expected no error, got err = %v", err)
-	}
-	err = CheckCertificatePolicy(leafCert, co)
-	if err != nil {
-		t.Errorf("CheckCertificatePolicy expected no error, got err = %v", err)
-	}
-}
-
-func TestValidateAndUnpackCertSuccessWithUriSan(t *testing.T) {
```
Hayden-IO, line 1115: While they aren't used currently by any publicly trusted identity provider, I believe sigstore-go does indirectly support certificates with the URI or OtherName field set, because of https://github.com/sigstore/sigstore/blob/f8fb68a433676bdb6c8b44899ded2f747fcfd6b3/pkg/cryptoutils/sans.go#L133.

Can we confirm we have test coverage somewhere - in sigstore-go or here - that handles this?

test/e2e_test.go
```diff
@@ -3408,7 +3101,8 @@ func TestAttachWithRFC3161Timestamp(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	must(verifyKeylessTSA(imgName, file.Name(), pemrootRef, true, true), t)
+	// TODO: Re-enable verification if/once attach.SignatureCmd uploads signatures in the new bundle format.
```
Hayden-IO, line 3104: What does attach do now after this change?

cmd/cosign/cli/verify/verify.go
```diff
@@ -236,27 +152,15 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
 			}
 
-			if co.NewBundleFormat {
-				// OCI bundle always contains attestation
-				verified, bundleVerified, err = cosign.VerifyImageAttestations(ctx, ref, co, c.NameOptions...)
-				if err != nil {
-					return err
-				}
-
-				verifiedOutput, err := transformOutput(verified, ref.Name())
```
Hayden-IO, line 246: It looks like this was dropped, is that intentional?

cmd/cosign/cli/signblob.go
```diff
@@ -93,43 +102,29 @@ func SignBlob() *cobra.Command {
...
 			}
 			if err := signcommon.LoadTrustedMaterialAndSigningConfig(cmd.Context(), &ko, o.UseSigningConfig, o.SigningConfigPath,
-				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL, o.TrustedRootPath, o.TlogUpload,
-				o.NewBundleFormat, o.BundlePath, o.Key, o.IssueCertificate,
-				o.Output, "", o.OutputCertificate, "", o.OutputSignature, o.RFC3161TimestampPath); err != nil {
+				o.TrustedRootPath, o.Key); err != nil {
 				return err
 			}
 
 			for _, blob := range args {
```
Hayden-IO, line 126: When multiple blobs are provided, this would overwrite the same bundle file with each iteration. This looks like a bug at HEAD as well.

Either, we change the command to not allow signing multiple blobs as once (i didn't realize you could, I think it'd be a fine change), or we need to prepend an artifact identifier to the name of the bundle (not as much of a fan of this, since we'd then need to document that the bundle name isn't actually the file name, it's the suffix).

cmd/cosign/cli/verify/verify.go
```diff
@@ -134,76 +101,32 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
...
-		return err
-	}
-
 	if c.CheckClaims {
-		if co.NewBundleFormat {
-			co.ClaimVerifier = cosign.IntotoSubjectClaimVerifier
-		} else {
-			co.ClaimVerifier = cosign.SimpleClaimVerifier
-		}
+		co.ClaimVerifier = cosign.IntotoSubjectClaimVerifier
```
Hayden-IO, line 118: Isn't ClaimVerifier checked in the now deleted Cosign verification library, https://github.com/sigstore/cosign/pull/4959/changes#diff-8a85c8e688d61e16b8af8e09832ed2bef89c1163b0e9601a8363c782c387c006L932-L936? Does sigstore-go handle this and this is now dead code? Or is there missing functionality?

cmd/cosign/cli/sign.go
```diff
@@ -128,21 +126,15 @@ race conditions or (worse) malicious tampering.
 				TSAClientCert:                  o.TSAClientCert,
 				TSAClientKey:                   o.TSAClientKey,
 				TSAServerName:                  o.TSAServerName,
-				TSAServerURL:                   o.TSAServerURL,
 				IssueCertificateForExistingKey: o.IssueCertificate,
```
Hayden-IO, line 129: `BundlePath` is missing here, but optionally used within `SignCmd()`.

Hayden-IO (changes requested):

cmd/cosign/cli/signcommon/common.go
Hayden-IO, line 442: This is an outdated error, we won't support fetching the individual target files.

cmd/cosign/cli/verify/verify.go
aaronlew02: This remains, but the code has shifted (see [here](https://github.com/sigstore/cosign/pull/4959/changes/BASE..6c0cc7249bae9a72c445ebd0e9b6135592b83c54#diff-fc9a26a4caa1d8684a4bdd9102f187f44e1f7fce7c06a19f4185035f13317cc8R161)).

cmd/cosign/cli/attest.go
aaronlew02: To specify either keyless or certificate-based attesting as the source of the error [here](https://github.com/sigstore/cosign/pull/4959/changes/BASE..baa6db068d7dfb32d9ed3d95192fc941343df907#diff-1df8237e8d19ee23ad75e394c1299aa004765daa97ae7c4b249061b1156739eaR80) (small detail).

cmd/cosign/cli/signblob.go
aaronlew02: Yes. I've submitted #5079 to address this.

cmd/cosign/cli/signblob.go
aaronlew02: Yes, `--output` is removed in [`de506c6`](https://github.com/sigstore/cosign/pull/4959/commits/de506c6e4b473ae3ad891c4ce1cdcee3fac2725c).

pkg/cosign/verify.go
aaronlew02: This change has been extracted into #5080.

cmd/cosign/cli/signblob.go
aaronlew02: I have addressed this in #5081.

cmd/cosign/cli/signblob.go
aaronlew02: To specify either keyless or certificate-based attesting as the source of the error [here](https://github.com/sigstore/cosign/pull/4959/changes/BASE..d6d0a2f8b39c8a11b5dd2ce3f789ac33bf5b979f#diff-3d58997fa494d583128d8c6c3c292d2752eda635c2ba20f0b018b7d378132d93R74) (small detail).

cmd/cosign/cli/attest_blob.go
aaronlew02: In `sign` and `sign-blob`, `ro.Timeout` is passed directly to and used by `SignCmd` [here](https://github.com/sigstore/cosign/pull/4959/changes/4537ba628e9dc772df2d601a0a6185e5e45f8740#diff-72911b6572099462e5f2e8ff920f4f9228587ab100820a26d3e90377c36bd73fR58) and `SignBlobCmd` [here](https://github.com/sigstore/cosign/pull/4959/changes/d6d0a2f8b39c8a11b5dd2ce3f789ac33bf5b979f#diff-a0c16ba1eed287205caf89e0f6d43a4e032730a46be333d03fd0d740e7bdd81fR55), respectively.

cmurphy (commented):
I'm still working on reviewing this in-depth, but I have some high-level thoughts to start with. Thanks for working so diligently on this.

I really appreciate you breaking this PR into multiple commits, it really helps make this more reviewable.

The other goal I like to have with chunking PRs into separate commits is for debuggability and rollback safety, and the way these are separated doesn't quite hit that. If I revert to any commit before "Update shared flags and helper logic", cosign doesn't build for me. That means tools like "git bisect" won't work if we need to trace a bug back to a particular commit, and if we find any bug in this we can't just roll back to the prior commit, we have to revert the whole pull request (if it's not something we can just fix right away). If there's any way to break up that "Update shared flags and helper logic" commit so that the prior commits work without it, or reorder it to have it first in line so that every individual commit builds, that would be ideal. But I'll understand if you say the refactor is too complex or it would be an unreasonable amount of work to mess with git just to keep the history clean.

Along with splitting up the commits, it also helps for debuggability if each commit has a semi-detailed commit message. For me, I like to know how each commit fits into the overarching pull request goal, and have any changes that may be surprising or weird or not be fully aligned with the goal be well explained so I don't have to ask a million questions during the review or later when I'm looking through the git history to debug. Don't assume that the people reviewing the code already know everything - it really helps to spell it out. So a commit with subject "Update core logic and delete legacy verification code" isn't quite verbose enough me, it's more helpful to have a hint of what you mean by "updating core logic" without me having to dive into the whole 1000 line commit.

I see that some pieces of this were already broken out into separate PRs, thanks for doing that! As I review deeper I'll try to point out any other pieces that might be good candidates to split out if you haven't already found them. From the PR description, I see "ALL signers using a signing config containing at least one transparency log must agree to the immutable record privacy statement." - I wonder if that's a potential candidate for its own PR?

I see Hayden left a few comments a few weeks ago, it'd be nice to see those addressed.

A more general question for you and for @Hayden-IO and @steiza - this PR removes support for all old bundle and disparate signature materials for both signing and verifying. Is that the approach that has been widely agreed on for v4? There are going to be millions of artifacts that exist with the old formats that will still need to be verified. Would it be a smoother approach to delete support for *signing* with the legacy formats, but keep support for *verifying* legacy formats until v5?

And in that vein, I see the command docs were updated for the deleted flags, but since this is such a huge change in functionality, I think it would be good to have an entry in the README explaining this change. It would be good to have an update to docs.sigstore.dev ready to go too, at least before this is released.

aaronlew02:
@cmurphy Thank you so much for reviewing this very large change. I appreciate the advice and would like this to be as reviewable as possible.

I have a series of PRs in draft, which are still under construction, that attempt to split up this PR into buildable chunks while incorporating several merged and unmerged auxiliary changes as well address comments from this PR.

> this PR removes support for all old bundle and disparate signature materials for both signing and verifying. Is that the approach that has been widely agreed on for v4?

My understanding is that v4 is intended to use `sigstore-go` for all signing and verification (which is only possible with the new bundle format).

> Would it be a smoother approach to delete support for signing with the legacy formats, but keep support for verifying legacy formats until v5?

That makes sense -- I'm open to that. It would reduce the number of users broken by an update. Alternatively, v3 could be maintained for legacy verification until v5.

> since this is such a huge change in functionality, I think it would be good to have an entry in the README explaining this change. It would be good to have an update to docs.sigstore.dev ready to go too, at least before this is released.

Agreed. I'd be happy to update those.

steiza:
> I'm still working on reviewing this in-depth, but I have some high-level thoughts to start with. Thanks for working so diligently on this.
> I really appreciate you breaking this PR into multiple commits, it really helps make this more reviewable.

I second all of this!

> A more general question for you and for @Hayden-IO and @steiza - this PR removes support for all old bundle and disparate signature materials for both signing and verifying. Is that the approach that has been widely agreed on for v4? There are going to be millions of artifacts that exist with the old formats that will still need to be verified. Would it be a smoother approach to delete support for signing with the legacy formats, but keep support for verifying legacy formats until v5?

My understanding (although this is of course up for discussion among Cosign maintainers):

v3 (where we're at today)
- Defaults to protobuf bundles
- Always sign with sigstore-go (breaking apart protobuf bundle content if needed for "old" bundle support)
- Verifies with sigstore-go for protobuf bundles and legacy path for old bundles

v4 (what this PR is working towards)
- Only signs and verifies protobuf bundles
- Consequently only uses sigstore-go for signing and verifying
- Can still take old bundle material and convert it to protobuf with `cosign bundle create`

v5 (the future)
- :shrug: TBD!

> And in that vein, I see the command docs were updated for the deleted flags, but since this is such a huge change in functionality, I think it would be good to have an entry in the README explaining this change. It would be good to have an update to docs.sigstore.dev ready to go too, at least before this is released.

Agreed, and I think we've mostly done that, modulo the README update. The docs updates should mostly be done, as  https://github.com/sigstore/docs/issues/418 moved us off of deprecated Cosign v3 behavior, which should continue to be supported in Cosign v4.

pkg/cosign/verify_test.go
aaronlew02: This PR simply shifts the pre-existing `appendSlices`, but I've replaced it with `bytes.Join` in #5093.

pkg/cosign/verify.go
aaronlew02: `IgnoreSCT` is still needed for `--insecure-ignore-sct`, but `SCT` and `CTLogPubKeys` will be deleted in a follow-up PR for cleanup.

cmd/cosign/cli/signcommon/common.go
aaronlew02: I've added the following comment in #5083:

```go
// root.NewSigningConfig only errors when an unsupported media type is provided.
// Since root.SigningConfigMediaType02 is a supported constant, this call cannot fail.
```

cmd/cosign/cli/verify/common.go
aaronlew02: Yes, this case is covered by the e2e tests `TestSignVerifyBundleOffline`, `TestSignBlobVerifyBlob`, and `TestSignBlobNonSHA256`, as well as the unit test `TestVerifyBlobWithUnsetHashAlgorithmMatchesKeyDefault`.

cmd/cosign/cli/verify/verify.go
aaronlew02: `co.ClaimVerifier` is still used in `verifyImageAttestationsSigstoreBundles`, which is called by `VerifyImageAttestations`. It enables image signature annotation verification, which is not handled by `sigstore-go`.

However, `co.ClaimVerifier` *is* unused in `verify_blob_attestation.go`, so I've removed it in #5089.

cmd/cosign/cli/signcommon/common.go
aaronlew02: I have addressed this in #5125.

cmd/cosign/cli/sign.go
aaronlew02: I have addressed this in #5132.

cmd/cosign/cli/verify.go
aaronlew02: I've kept `PrivateInfrastructure` in #5093.

cmd/cosign/cli/attest.go
aaronlew02: I have addressed this in #5108.

cmd/cosign/cli/signblob.go
aaronlew02: I have addressed this in #5083.

cmd/cosign/cli/sign/sign_blob_test.go
aaronlew02: I have addressed this in #5083.

pkg/cosign/verify_test.go
aaronlew02: `sigstore-go` reads URI, email, and `OtherName` SANs in [`SummarizeCertificate`](https://github.com/sigstore/sigstore-go/blob/a061819fe7f864aa13c79377dbd01488b572b442/pkg/fulcio/certificate/summarize.go#L42), and has tests for each:
* [`TestSummarizeCertificateWithActionsBundle`](https://github.com/sigstore/sigstore-go/blob/a061819fe7f864aa13c79377dbd01488b572b442/pkg/fulcio/certificate/summarize_test.go#L25) for URI.
* [`TestSummarizeCertificateWithOauthBundle`](https://github.com/sigstore/sigstore-go/blob/a061819fe7f864aa13c79377dbd01488b572b442/pkg/fulcio/certificate/summarize_test.go#L74) for email.
* [`TestSummarizeCertificateWithOtherNameSAN`](https://github.com/sigstore/sigstore-go/blob/a061819fe7f864aa13c79377dbd01488b572b442/pkg/fulcio/certificate/summarize_test.go#L104) for `OtherName`.

cmurphy:
> My understanding (although this is of course up for discussion among Cosign maintainers):

This does sound familiar, now that I think about it. I didn't think about it then but I worry a bit now that this is the kind of change that could be very frustrating to users (but that they won't tell us about until after we've done it). Let's chat about it in today's meeting?

pkg/cosign/verify_test.go
aaronlew02: I have addressed this in #5137.

pkg/cosign/verify_test.go
aaronlew02: I have addressed this in #5137.

PR #4967, opened 2026-06-29
fix: don't warn about tag usage for cosign-generated references

### LINKED ISSUE #3995: Signing an attestation generates an unnecessary warning
**Description**

We are using cosign to do a number of signing/attesting/verifying tasks within our secure supply chain pipelines and all work very well.  Also, we use our own PKI to sign artifacts with cosign.  However, included in one of the jobs is an attestation piece (all variables have been verified):
```
DIGEST=$(skopeo inspect --format "{{ .Digest }}" "docker://${image}")
cosign attest --tlog-upload=false --yes --type cyclonedx --predicate ./artifacts/${artifactTag}-cyclonedx-sbom.json --key ${ARTIFACT_SIGNING_KEY} ${image}@${DIGEST}
cosign sign --key ${ARTIFACT_SIGNING_KEY} $(cosign triangulate ${image}@${DIGEST} --type attestation)
```

All of this works exactly as intended.  However, the final command `cosign sign blah blah` results in the following warning:
```
WARNING: Image reference <harbor-registry>/<repo>/langgraph-cli:sha256-abc123.att uses a tag, not a digest, to identify the image to sign.
    This can lead you to sign a different image than the intended one. Please use a
    digest (example.com/ubuntu@sha256:abc123...) rather than tag
    (example.com/ubuntu:latest) for the input to cosign. The ability to refer to
    images by tag will be removed in a future release.
```
Again, this doesn't cause a functional error in the job but I am concerned that future releases _could_ affect it.  The warning message is misleading because it isn't an image reference at all, it is a reference to an attestation _of an image_ and that is what we are signing in this command. This command is signing the attestation itself, not the image (that occurs in a separate job and works fine).  

Attestation artifacts will never have a format of `example.com/ubuntu@sha256:abc123`   
Rather they'll have a reference like this: `example.com/ubuntu:sha256-abc123.att`

**Version**

cosign v2.4.1

ricardbejarano:
## Summary

Fixes #3995.

When signing a cosign-generated discovery tag — e.g. the `sha256-<digest>.att` reference returned by `cosign triangulate ... --type attestation` — `ParseOCIReference` printed:

```
WARNING: Image reference <repo>:sha256-abc123.att uses a tag, not a digest, to identify the image to sign.
    ...
```

That warning is misleading here: these tags are *derived from an image digest*, so they already pin to specific content, and the user is deliberately signing the attestation/signature/SBOM artifact rather than a mutable image tag.

## Changes

- `ParseOCIReference` now skips the tag warning when the reference is one of cosign's own digest-derived discovery tags (`sha256-<digest>.sig`, `.att`, `.sbom`). Ordinary mutable tags still warn as before, and digest references continue to be silent.
- The suffix set reuses the existing `ociremote.{Signature,Attestation,SBOM}TagSuffix` constants so it stays in sync with cosign's tag scheme.
- Extended `Test_ParseOCIReference` with cases for each generated suffix (no warning) and a look-alike tag `latest.att` (still warns).

## Notes

CHANGELOG.md is release-generated, so it is intentionally left untouched. No Go toolchain was available locally; relied on CI for `go test`/lint.

### DIFF
--- a/cmd/cosign/cli/signcommon/common.go
+++ b/cmd/cosign/cli/signcommon/common.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 
 	"net/http"
 	"time"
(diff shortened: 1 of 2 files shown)


cmurphy:
I think the warning still applies here, since it is still a tag and so it still is mutable, right?

steiza:
@cmurphy is correct - even if the name of the tag matches the digest it should resolve to, that doesn't mean that folks are checking to ensure that the tag hasn't been updated to point to another digest.

PR #4970, opened 2026-06-30
fix(ci): Pass --insecure-registry to ko publish

aaronlew02:
#### Summary

This change passes the `--insecure-registry` flag to `ko publish` in the `kind-verify-attestation` workflow, as `go-containerregistry` no longer automatically treats `*.local` as an insecure HTTP registry hostname (google/go-containerregistry#2281).

### DIFF
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -94,7 +94,7 @@ jobs:
           fmt.Println("hello world")
         }
         EOF
-        demoimage=`ko publish -B example.com/demo`
+        demoimage=$(ko publish --insecure-registry -B example.com/demo)
         echo "demoimage=$demoimage" >> $GITHUB_ENV
         echo Created image $demoimage
         popd


.github/workflows/kind-verify-attestation.yaml
loosebazooka, line 87: should this be running with `set -eE` to exit on failures and subshell failures?

loosebazooka (approved):

.github/workflows/kind-verify-attestation.yaml
aaronlew02: > By default, fail-fast behavior is enforced using `set -e` for both `sh` and `bash`. When `shell: bash` is specified, `-o pipefail` is also applied to enforce early exit from pipelines that generate a non-zero exit status.

([source](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax))

`shell: bash` is a specified default (L24).

`set -E` doesn't seem to apply here since there are no error traps (`trap ... ERR`).

bobcallaway (approved):

PR #4971, opened 2026-06-30
fix(download): Add predicate type not found error for new bundle format

aaronlew02:
#### Summary

This change adds error handling for when no attestations are found for the given predicate type, when using the new bundle format, to `download attestation`.


aaronlew02:
This has been closed in favor of #4976.

Issue #4573, opened 2025-12-10
cosign 3.0.3: cosign attestation download skips new-bundle-format style annotations, if any sort of bundle is found via referrers api


tamcore:
**Description**

With cosign v3.0.3, new-bundle-format style attestations are correctly detected on an image. It correctly checks the referrers API for artifacts for references.

But it already considers itself to be done, if the referrers api even returns just a simple cosign signature, without an attestation.

In that case, expected behaviour would be, for it to move on and check for an new-bundle-format=false style `sha256-xyz.att` attestation as well.

If I comment out the line

https://github.com/sigstore/cosign/blob/3f32cea203c59a93323a6bebfebff03417520143/cmd/cosign/cli/download/attestation.go#L75

it works just fine and it detects both old and new-style attestations just fine.

**Version**

3.0.3

Hayden-IO:
Thanks for filing this, the proposed fix looks correct to me. @steiza can you confirm?

steiza:
Yeah, previously I've thought that people were either dealing with objects that were signed with protobuf bundles or using old bundle formats / the old OCI storage mechanism.

For `verify` type subcommands we probably still want to try protobuf bundles first, and return success early if they are found and meet the verification criteria.

But I can see how for inspection subcommands (`download`, `tree`, `save`, `load`, ...) you would want it to handle both cases and not just stop after the protobuf bundle is handled.

All that is a long way of saying, yes, I think this change is fine for `download`.

steiza:
This was fixed with #4996.

PR #4976, opened 2026-07-01
download: also emit old-format attestations when a bundle exists (#4573)

arpitjain099:
I work on supply-chain security tooling and ran into this while inspecting an image's attestations.

`cosign attestation download` returns zero old-format `.att` attestations whenever the image also carries a new-bundle-format signature. `GetBundles` returns every bundle referrer (including signature-only ones), so the new-format branch is taken even when there are no matching attestations, and the command then returned before it ever fetched the old-format `sha256-<digest>.att` entries. This is #4573, and @steiza confirmed on that issue that download (unlike verify) should handle both formats.

The fix drops that early return so download writes any new-format bundles and then also fetches the old-format attestations, emitting both. A missing old-format entity is only treated as an error when nothing was written yet, so existing behavior is preserved for old-format-only images. verify paths are untouched, and I've scoped this to inspection-only download semantics so it doesn't conflict with the broader new-bundle work in the draft #4959.

The download package had no tests, so I added the first ones: the main case fails on the old code (old-format attestation dropped) and passes with the fix, plus two cases covering old-format-only errors and tolerating a missing old-format entity. `go test`, `go vet`, and `gofmt` are clean on the package.

### DIFF
--- a/cmd/cosign/cli/download/attestation.go
+++ b/cmd/cosign/cli/download/attestation.go
@@ -28,6 +28,13 @@ import (
 	ociremote "github.com/sigstore/cosign/v3/pkg/oci/remote"
 )
 
+// Indirected for testing. These wrap the two registry-touching stages so the
+// command flow (which format(s) to emit) can be exercised without a registry.
+var (
+	writeNewBundles      = writeNewBundlesImpl
+	writeOldAttestations = writeOldAttestationsImpl
+)
+
 func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOptions options.AttestationDownloadOptions, imageRef string, out io.Writer) error {
 	ref, err := name.ParseReference(imageRef, regOpts.NameOptions()...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/download/attestation.go
```diff
@@ -46,35 +53,70 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOpt
...
-				}
-				statement, err := envelope.Statement()
-				if err != nil || statement == nil {
-					continue
-				}
-				if statement.PredicateType != predicateType {
-					continue
-				}
+	if err != nil || len(newBundles) == 0 {
+		return 0, nil
```
steiza, line 88: We probably don't want to swallow the error here?
```suggestion
		return 0, err
```

steiza (commented):
I do appreciate that this adds tests, but I wonder if we should have just removed the https://github.com/sigstore/cosign/blob/3f32cea203c59a93323a6bebfebff03417520143/cmd/cosign/cli/download/attestation.go#L75 line 😅 - the tests add quite a bit of complexity.

At the very least, let's not swallow the error from `cosign.GetBundles()`.

steiza:
We landed #4996 instead, but thank you for the push to get this fixed!

arpitjain099:
> We landed #4996 instead, but thank you for the push to get this fixed!

ah I was slow to respond but thank you for merging the change! Happy to see the changes made it through!

PR #4979, opened 2026-07-02
Fix panic in dockerfile verify on malformed FROM lines

mailnike:
## Summary

`cosign dockerfile verify` panics (`index out of range`) instead of returning an error on two malformed but parseable `FROM` lines. `getImageFromLine` is the culprit.

### Repro 1: unresolvable build arg
```dockerfile
ARG BASE_IMAGE
FROM ${BASE_IMAGE}
```
When `BASE_IMAGE` is not in the environment, `os.Expand` replaces `${BASE_IMAGE}` with an empty string, so after trimming `FROM` the line is blank, `strings.Fields` returns an empty slice, and `fields[len(fields)-1]` indexes `fields[-1]` -> `panic: runtime error: index out of range [-1]`.

### Repro 2: trailing `AS`
```dockerfile
FROM gcr.io/test/image AS
```
The `AS` is the last field, so `fields[i+1]` reads past the slice -> `panic: index out of range [2] with length 2`.

## Change

`getImageFromLine` now returns `(string, error)`:
- errors with `no image found in line %q (is a build arg missing from the environment?)` when no image token remains,
- errors with `expected stage name after "AS" in line %q` when `AS` has no following stage name,

and the caller propagates the error. Fail-closed is intentional: silently skipping an unresolvable base image would weaken a verification command.

## Testing

Adds `TestGetImagesFromDockerfileInvalidFromLines` (both cases). Verified both directions: the test panics/fails on the unpatched code and passes with the fix. `gofmt`, `go vet ./cmd/cosign/cli/dockerfile/...`, and the full package test suite pass; `go build ./cmd/cosign/...` is clean (single call site).

No open issue tracks this crash (the related #434 / #435 are closed and were about ARG *support*, a separate feature).

### DIFF
--- a/cmd/cosign/cli/dockerfile/verify.go
+++ b/cmd/cosign/cli/dockerfile/verify.go
@@ -93,7 +93,10 @@ func (fc *finderCache) getImagesFromDockerfile(ctx context.Context, dockerfile i
 		lineUpper := strings.ToUpper(line)
 		switch {
 		case strings.HasPrefix(lineUpper, "FROM"):
-			image := fc.getImageFromLine(line)
+			image, err := fc.getImageFromLine(line)
+			if err != nil {
+				return nil, err
+			}
 			switch {
 			case image == "scratch":
 				ui.Infof(ctx, "- scratch image ignored")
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
thanks!

PR #4981, opened 2026-07-03
fix: pass NewBundleFormat to KeyOpts in sign command

### LINKED ISSUE #4980: NewBundleFormat not passed to KeyOpts in sign command, breaking TSA usage
When using `cosign sign` with `--use-signing-config=false` and `--timestamp-server-url`, the
`NewBundleFormat` flag is never copied from `SignOptions` to `KeyOpts` in `cmd/cosign/cli/sign.go`.

This causes `GetRFC3161Timestamp` in `signcommon/common.go` to always see `NewBundleFormat=false`,
resulting in:

    Error: expected either new bundle or an rfc3161-timestamp path when using a TSA server

The `attest` command is unaffected — it already sets `NewBundleFormat` in `KeyOpts`.

JasonPowr:
`NewBundleFormat` is parsed from the CLI flag but never copied to `KeyOpts` in `sign.go`,
causing TSA signing to fail with:

    "expected either new bundle or an rfc3161-timestamp path when using a TSA server"

This only affects the manual flags path (`--use-signing-config=false` with explicit
`--timestamp-server-url`). The default signing config path is unaffected as it sources
TSA and bundle format settings from the trusted root / signing config, bypassing `KeyOpts`.

Fix: add the missing `NewBundleFormat: o.NewBundleFormat` to the `KeyOpts` struct initialization.

Fixes: https://github.com/sigstore/cosign/issues/4980

### DIFF
--- a/cmd/cosign/cli/sign.go
+++ b/cmd/cosign/cli/sign.go
@@ -130,6 +130,7 @@ race conditions or (worse) malicious tampering.
 				TSAServerName:                  o.TSAServerName,
 				TSAServerURL:                   o.TSAServerURL,
 				IssueCertificateForExistingKey: o.IssueCertificate,
+				NewBundleFormat:                o.NewBundleFormat,
 			}
 			if err := signcommon.LoadTrustedMaterialAndSigningConfig(cmd.Context(), &ko, o.UseSigningConfig, o.SigningConfigPath,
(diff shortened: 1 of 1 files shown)


Hayden-IO (approved):
Thanks!

PR #4991, opened 2026-07-06
Deprecate --payload for sign and verify commands

aaronlew02:
#### Summary

This change deprecates the `--payload` flag, which will be deprecated  for the `sign` and `verify` commands, as future versions will use the new bundle format exclusively.

### DIFF
--- a/cmd/cosign/cli/options/sign.go
+++ b/cmd/cosign/cli/options/sign.go
@@ -109,6 +109,7 @@ func (o *SignOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.PayloadPath, "payload", "",
 		"path to a payload file to use rather than generating one")
 	// _ = cmd.MarkFlagFilename("payload") // no typical extensions
+	_ = cmd.Flags().MarkDeprecated("payload", "payload will always be generated automatically in future versions")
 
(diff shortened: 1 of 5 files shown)


Hayden-IO:
Do we have a sense of how `payload` was used in signing? I agree with this being deprecated, but just wanted to check if we're aware of anyone relying on this currently.

loosebazooka:
> Do we have a sense of how payload was used in signing? I agree with this being deprecated, but just wanted to check if we're aware of anyone relying on this currently.


It does feel like no one should be using this, and should be using `attest` if they want to have a custom payload

steiza (approved):
I'm in favor of this deprecation.

If you want to sign a "payload" instead of a hash of a file on disk, you should use `cosign attest --predicate`.

Maybe in the distant future we want to collapse `sign` and `attest` based on the presence of a `--payload` or `--predicate`/`--statement`? But we should finish some of our current cleanup efforts before planning out the next round 😅

aaronlew02:
> Maybe in the distant future we want to collapse sign and attest based on the presence of a --payload or --predicate/--statement?

@Hayden-IO has brought this up as well. This sounds like a good idea, particularly given how similarly the two commands function with the new bundle format.

Hayden-IO (approved):

PR #4994, opened 2026-07-08
fix: avoid panic in dockerfile verify when FROM expands to empty

ogulcanaydogan:
`cosign dockerfile verify` panics with an index-out-of-range on a Dockerfile that uses a required build arg with no default:

```dockerfile
ARG BASE_IMAGE
FROM ${BASE_IMAGE}
```

`getImageFromLine` expands the templated vars and then returns `fields[len(fields)-1]`. Static analysis can't know the `--build-arg` value, so `${BASE_IMAGE}` expands to an empty string, `fields` ends up empty, and the slice index goes out of range instead of failing cleanly.

This returns an empty image reference in that case and lets the caller skip it, next to the existing `scratch` and stage-reference skips. Added a test case (`arg-with-no-default-value`) to `TestGetImagesFromDockerfile` that reproduces the panic without the fix.

### DIFF
--- a/cmd/cosign/cli/dockerfile/verify.go
+++ b/cmd/cosign/cli/dockerfile/verify.go
@@ -95,6 +95,8 @@ func (fc *finderCache) getImagesFromDockerfile(ctx context.Context, dockerfile i
 		case strings.HasPrefix(lineUpper, "FROM"):
 			image := fc.getImageFromLine(line)
 			switch {
+			case image == "":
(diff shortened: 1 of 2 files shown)


steiza (approved):
Looks good to me; thanks!

steiza:
Ah it looks like https://github.com/sigstore/cosign/pull/4979 came in a bit before you - if I can fix the DCO bot on that PR I will land that one instead.

steiza:
We landed #4979 instead.

PR #4996, opened 2026-07-08
Allow attestation download to handle both bundle types

steiza:
As suggested on https://github.com/sigstore/cosign/issues/4573

#### Summary

As suggested on https://github.com/sigstore/cosign/issues/4573

#### Release Note

* `cosign download attestation` includes both protobuf bundles as well as the old `.att` bundles

#### Documentation

N/A

### DIFF
--- a/cmd/cosign/cli/download/attestation.go
+++ b/cmd/cosign/cli/download/attestation.go
@@ -72,7 +72,6 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOpt
 				return err
 			}
 		}
-		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


Hayden-IO (dismissed):

steiza:
This is trickier than I thought, as the behavior of `cosign download attestation` is not well defined when attestations are not found.

For new bundles, if nothing is found we simply don't print out anything. For old bundles, if nothing was found we returned an error.

The only place we use `cosign.FetchAttestations()` is in `cosign download attestation` and in the end-to-end tests (where we always fail if an error is returned). So I believe this change in behavior is safe.

Hayden-IO (approved):

PR #5001, opened 2026-07-10
fix(download): Validate predicate type for new bundle format

aaronlew02:
#### Summary

This change expands predicate type validation in `download attestation` to correctly handle attestations in the new bundle format, ensuring an error is thrown if no matching predicate types are found across either format.

### DIFF
--- a/cmd/cosign/cli/download/attestation.go
+++ b/cmd/cosign/cli/download/attestation.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -46,6 +47,8 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOpt
 		}
 	}
 
+	var foundMatches bool
+
 	// Try bundles first
 	newBundles, _, err := cosign.GetBundles(ctx, ref, ociremoteOpts)
 	if err == nil && len(newBundles) > 0 {
@@ -71,6 +74,7 @@ func AttestationCmd(ctx context.Context, regOpts options.RegistryOptions, attOpt
 			if err != nil {
 				return err
 			}
+			foundMatches = true
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


cmurphy (commented):
Could you include a test to show that this fixes the issue you found?

aaronlew02:
@cmurphy Sorry, this PR lacks a bit of context.

This PR is auxiliary to #4959, which currently pulls in the changes in this PR to make the e2e test `TestAttestationDownloadWithBadPredicateType` pass with the new bundle format (see test failure [here](https://github.com/sigstore/cosign/actions/runs/29100836085/job/86388978000)).

This PR is also a follow-up to #4996.

aaronlew02:
I will be adding an e2e test so that this fix can be shipped in a v3 release.

aaronlew02:
I've added an e2e test.

steiza:
Yeah, like I mentioned on https://github.com/sigstore/cosign/pull/4996#issuecomment-4925793932, the behavior of `cosign download attestation` was not well defined when an attestation was not found.

With #4996 we made it so if no attestations are found (and you didn't specify a predicate type) we return an empty list, and this changes makes it so that if no attestations are found and you **did** specify a predicate type we return an error.

What should the behavior be? I honestly don't know. I sort of feel like errors should be reserved for things like you couldn't contact the container registry. To my knowledge the previous behaviors weren't documented.

Should we return an empty list here instead of erroring? I'm open to feedback here.

aaronlew02:
@steiza That's a good question. It makes sense to return an empty list instead of an error, even with a desired predicate type, since the search completes successfully.

I prefer keeping this PR scoped to its specific bug fix but would be happy to address this in another PR.

cmurphy:
@steiza given the existing test that @aaronlew02 pointed out it looks like returning an error has been the standard pre-new-bundle. I don't have a strong opinion of what is "right" but I'm leaning toward using the error just for consistency.

steiza (approved):
SGTM!

cmurphy (approved):

PR #4997, opened 2026-07-09
fix: include artifactType in OCI 1.1 signature referrer manifest

anithapriyanatarajan:
#### Summary

Closes #4995

`WriteSignaturesExperimentalOCI` was constructing the OCI 1.1 referrer manifest by marshalling a plain `v1.Manifest` struct, which does not include the top-level `artifactType` field required by the OCI 1.1 Image Manifest spec. As a result, signature referrer manifests uploaded via the experimental OCI path were missing `artifactType`, making them
non-compliant and difficult to discover by tools that filter referrers by type.

**Fix:** 
Replace the `json.Marshal(&m)` + `taggableManifest` approach with the existing `referrerManifest` wrapper (already used by `WriteReferrer` and `WriteAttestationsReferrer`), which serializes thetop-level `artifactType` field correctly. The value is `application/vnd.dev.cosign.artifact.sig.v1+json`, consistent with what `ociexperimental.ArtifactType("sig")` already computed in the function.

#### Release Note

Fixed `WriteSignaturesExperimentalOCI` to include the top-level `artifactType` field (`application/vnd.dev.cosign.artifact.sig.v1+json`) in OCI 1.1 referrer manifests produced for signatures, aligning with the
OCI 1.1 Image Manifest specification. Previously uploaded signatures are unaffected; new uploads will produce a different manifest digest due to the corrected content.

#### Documentation

No documentation update required. This is a bug fix to an experimental OCI 1.1 code path with no user-facing CLI or API changes.

PR #5002, opened 2026-07-13
fix: include artifactType in OCI 1.1 signature referrer manifest (cherry-pick PR-4997 to release-2.6)

### LINKED ISSUE #4995: WriteSignaturesExperimentalOCI does not set top-level artifactType on signature referrer manifests
**Description**

When Chains (or any cosign library consumer) stores image signatures using the OCI 1.1 Referrers API via [WriteSignaturesExperimentalOCI](https://github.com/sigstore/cosign/blob/main/pkg/oci/remote/write.go#L216), the resulting referrer manifest does not have a top-level artifactType field set. Only config.mediaType is set.

This is inconsistent with how attestation referrers are written — [WriteAttestationNewBundleFormat](https://github.com/sigstore/cosign/blob/main/pkg/oci/remote/write.go#L405) (via WriteReferrer) correctly sets artifactType on the manifest. Signatures should behave the same way.

**Impact**

The OCI 1.1 spec defines artifactType as the canonical field for identifying referrer content type. Tools that filter referrers by artifactType — including cosign tree and cosign download signature — cannot find signature referrers produced by this function:

- cosign tree silently omits the signature from its output; only attestations appear
- cosign download signature returns the attestation instead of the signature (related: #4573)
- oras discover shows the signature as type `<unknown>` because no artifactType is set

Image

- cosign verify is unaffected because it discovers signatures by config.mediaType, not artifactType.

Sample crane manifest **MISSING** artifactType for signature

Image

Crane manifest with correct artifactType for attestation

Image

**Expected behaviour**

Signature referrer manifests should have `artifactType` set to `application/vnd.dev.cosign.artifact.sig.v1+json`, consistent with how attestation referrer manifests already have their artifactType set. This would make signatures discoverable by any standards-compliant referrers client without requiring knowledge of the cosign-specific `config.mediaType` convention.

**Version**

Consuming [library](https://github.com/tektoncd/chains/blob/main/go.mod#L24) - github.com/sigstore/cosign/v2 v2.6.3 

**Related**

#4335 — Complete OCI 1.1 Referrers API Support Across All Cosign Commands
#4573 — cosign download attestation returns signature bundle when referrer found

anithapriyanatarajan:
#### Summary

Cherry-pick of #4997 onto `release-2.6`.

Closes #4995

`WriteSignaturesExperimentalOCI` was constructing the OCI 1.1 referrer manifest by marshalling a plain `v1.Manifest` struct, which does not include the top-level `artifactType` field required by the OCI 1.1 Image Manifest spec. As a result, signature referrer manifests uploaded via the experimental OCI path were missing `artifactType`, making them non-compliant and difficult to discover by tools that filter referrers by type.

**Fix:** Replace the `json.Marshal(&m)` + `taggableManifest` approach with the existing `referrerManifest` wrapper (already used by `WriteReferrer` and `WriteAttestationsReferrer`), which serializes the top-level `artifactType` field correctly. The value is `application/vnd.dev.cosign.artifact.sig.v1+json`, consistent with what `ociexperimental.ArtifactType("sig")` already computed in the function.

**Adaptation from original:** `rm.targetRef(d.Repository)` is called without options, as the `release-2.6` version of `targetRef` does not accept variadic options (added later on `main`). The fix is otherwise identical.

#### Release Note

Fixed `WriteSignaturesExperimentalOCI` to include the top-level `artifactType` field (`application/vnd.dev.cosign.artifact.sig.v1+json`) in OCI 1.1 referrer manifests produced for signatures, aligning with the OCI 1.1 Image Manifest specification.

#### Documentation

No documentation update required.

### DIFF
--- a/.github/workflows/kind-verify-attestation.yaml
+++ b/.github/workflows/kind-verify-attestation.yaml
@@ -94,7 +94,7 @@ jobs:
           fmt.Println("hello world")
         }
         EOF
-        demoimage=`ko publish -B example.com/demo`
+        demoimage=$(ko publish --insecure-registry -B example.com/demo)
         echo "demoimage=$demoimage" >> $GITHUB_ENV
         echo Created image $demoimage
         popd
--- a/pkg/oci/remote/write.go
+++ b/pkg/oci/remote/write.go
@@ -193,22 +193,15 @@ func WriteSignaturesExperimentalOCI(d name.Digest, se oci.SignedEntity, opts ...
 	artifactType := ociexperimental.ArtifactType("sig")
 	m.Config.MediaType = types.MediaType(artifactType)
 	m.Subject = desc
-	b, err = json.Marshal(&m)
-	if err != nil {
-		return err
-	}
-	digest, _, err := v1.SHA256(bytes.NewReader(b))
-	if err != nil {
-		return err
-	}
-	targetRef, err := name.ParseReference(fmt.Sprintf("%s/%s@%s", d.RegistryStr(), d.RepositoryStr(), digest.String()))
+	rm := referrerManifest{m, artifactType}
+	targetRef, err := rm.targetRef(d.Repository)
 	if err != nil {
 		return err
 	}
 	// TODO: use ui.Infof
 	fmt.Fprintf(os.Stderr, "Uploading signature for [%s] to [%s] with config.mediaType [%s] layers[0].mediaType [%s].\n",
 		d.String(), targetRef.String(), artifactType, ctypes.SimpleSigningMediaType)
-	return remotePut(targetRef, &taggableManifest{raw: b, mediaType: m.MediaType}, o.ROpt...)
+	return remotePut(targetRef, rm, o.ROpt...)
 }
 
 type taggableManifest struct {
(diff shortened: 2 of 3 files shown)


Hayden-IO:
This is the challenge with these backports, figuring out what's needed for CI that runs against scaffolding@main. You'll need to cherry pick https://github.com/sigstore/cosign/pull/4970 in as well.

anithapriyanatarajan:
> You'll need to cherry pick #4970 in as well.

Thank you for pointing out. Cherry-picked #4970 and the CI is passing now.

Hayden-IO (approved):
thanks!

steiza (approved):
LGTM!

Did you also want https://github.com/sigstore/cosign/pull/4996? Maybe in a separate pull request against `release-2.6`? Just trying to figure out if we should land this as-is (which is fine by me).

anithapriyanatarajan:
>Did you also want https://github.com/sigstore/cosign/pull/4996? Maybe in a separate pull request against release-2.6? Just trying to figure out if we should land this as-is (which is fine by me).

Was mid way doing that. Based on initial assessment `cosign.GetBundles` signature is different in release-2.6. Will submit the cherry-pick with required edits and tag you. Thank you.

anithapriyanatarajan:
@Hayden-IO @steiza - could you help merge this, only then CI will pass for https://github.com/sigstore/cosign/pull/5017. Thank you

PR #5012, opened 2026-07-13
fix(verify): skip identity validation for security keys

aaronlew02:
#### Summary

This change skips identity validation for security keys.

Identity validation is supposed to be skipped for all key-based signing; however, previously, only providing a public key reference triggered the skip.


Hayden-IO (approved):

Hayden-IO:
Thanks! Surprised this was never noticed before!

PR #5016, opened 2026-07-15
Skip nil subject entries in IntotoSubjectClaimVerifier

arpitjain099:
## Summary

`IntotoSubjectClaimVerifier` panics with a nil pointer dereference when an in-toto statement's subject array contains a null entry. `pkg/cosign/verifiers.go:80` does `dgst, ok := subj.Digest["sha256"]` without checking that `subj` is non-nil.

Reproducer statement (base64'd into a DSSE envelope with payloadType `application/vnd.in-toto+json`):

```json
{"_type":"https://in-toto.io/Statement/v0.1","predicateType":"https://spdx.dev/Document","subject":[null],"predicate":"x"}
```

Before the fix this is a SIGSEGV:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
github.com/sigstore/cosign/v3/pkg/cosign.IntotoSubjectClaimVerifier(...)
	pkg/cosign/verifiers.go:80
```

## Why the nil shows up

`Statement.UnmarshalJSON` (`pkg/cosign/attestation/attestation.go`) tries `protojson.Unmarshal` first, and falls back to `encoding/json` when the predicate is a string rather than an object. That fallback is the interesting part: `encoding/json` happily decodes a JSON `null` inside the subject array into a `nil *ResourceDescriptor`, and the claim verifier then dereferences it. The string predicate matters because it's what forces the legacy path, so an SPDX-style attestation is a natural carrier.

I kept the fix in the verifier since that's where the deref is. Rejecting null subjects in `UnmarshalJSON` would also work and might be worth doing separately, but it's a wider behavior change and I didn't want to bundle it in here.

## Scope, and what this is not

I want to be precise about severity, since this touches verification code. `ClaimVerifier` runs *after* signature verification, so reaching it needs a validly signed attestation. In keyless that means an attacker signs their own crafted attestation with their own Fulcio cert and serves it from an image they control; a verifier running `cosign verify-attestation --check-claims` (or `verify --check-claims`) against that image then crashes.

So this is a client-side crash on hostile input, not a signature or verification bypass. Nothing gets accepted that shouldn't be. The claim check still fails for the crafted input after the fix, it just fails properly: `no matching subject digest found`.

## Change

Skip nil entries in the subject loop. Five lines plus a comment.

## Testing

Added a `subject:[null]` case to the existing `Test_IntotoSubjectClaimVerifier` table, which goes through `static.NewSignature` into the real `IntotoSubjectClaimVerifier` entry point. It panics on main and passes with the fix. Existing cases in that table are untouched and still pass, and `go test ./pkg/cosign/...` plus `go build ./...` are clean.


pkg/cosign/verifiers.go
```diff
@@ -77,6 +77,11 @@ func IntotoSubjectClaimVerifier(sig oci.Signature, imageDigest v1.Hash, annotati
 		return err
 	}
 	for _, subj := range st.Subject {
+		// A null entry in the subject array decodes to a nil descriptor on the
+		// legacy encoding/json path, so skip it rather than dereferencing it.
+		if subj == nil {
+			continue
```
Hayden-IO: This would be an invalid statement as per the [spec](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md), so I'd throw an error rather than continue.

pkg/cosign/verifiers_test.go
```diff
@@ -63,6 +63,11 @@ const (
...
+	// A statement whose subject array holds a null entry. The string predicate
```
Hayden-IO: This comment can be cleaned up or removed, it's too in the weeds and isn't providing additional context.

pkg/cosign/verifiers.go
```diff
@@ -77,6 +77,11 @@ func IntotoSubjectClaimVerifier(sig oci.Signature, imageDigest v1.Hash, annotati
 		return err
 	}
 	for _, subj := range st.Subject {
+		// A null entry in the subject array decodes to a nil descriptor on the
```
Hayden-IO: Comment is unnecessary, code is self-documenting.

pkg/cosign/verifiers.go
arpitjain099: Good call, an all-null or null-containing subject is invalid per the spec. Switched it to return an error instead of skipping. The test case for it already expected a failure, so it still covers the path.

pkg/cosign/verifiers_test.go
arpitjain099: Trimmed it down to one line.

pkg/cosign/verifiers.go
arpitjain099: Removed it.

Hayden-IO (approved):
Thanks!

steiza (approved):
I love how the unification of verification paths has made it so we only need to fix this in one place.

There is separate verification for the "old" Cosign bundles, but they are deprecated and we will soon be removing them from the codebase.

PR #5019, opened 2026-07-17
Auto-detect default digest algorithm for public keys

aaronlew02:
#### Summary

This change updates public key verifier loaders to automatically detect the digest algorithm for a given public key.

This renders the deprecated `--signature-digest-algorithm` (to be removed in #4959) redundant.


codysoyland (approved):

Hayden-IO (approved):
LGTM. This does break a verifier providing their own ECDSA-P384 or P521 key who wasn't using the digest flag, so we should have this go out in the next major revision.

PR #5022, opened 2026-07-20
fix(pkcs11key): return an error instead of panicking when no key pair matches

### LINKED ISSUE #4540: Cosign signs successfully on Windows but fails on Ubuntu 24 when using PKCS#11/HSM
**Description**

I encounter this error when trying to use Cosign to sign with the HSM. I’ve already checked the slot and objects with `pkcs11-tool`, and everything looks fine.
What’s strange is that with the same Cosign version (v3.0.2) I can sign successfully on Windows, but I get the error below when trying on Ubuntu (Ubuntu 24).
Could you please help me check this?

`panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x260f596]
goroutine 1 [running]:
github.com/sigstore/cosign/v3/pkg/cosign/pkcs11key.(*Key).PublicKey(0x12835b01d807aa98?, {0xc000002380?, 0x80deb1fe72be5d74?, 0xc19bf1749bdc06a7?})
    /opt/cosign/cosign/pkg/cosign/pkcs11key/pkcs11key.go:199 +0x16
github.com/sigstore/cosign/v3/internal/key.(*SignerVerifierKeypair).GetPublicKeyPem(0x300000002?)
    /opt/cosign/cosign/internal/key/svkeypair.go:115 +0x22
github.com/sigstore/cosign/v3/pkg/cosign/bundle.SignData({0x49d3970, 0xc000413ce0}, {0x49a7678, 0xc000b3e000?}, {0x49e4780, 0xc000b9e580}, {0x0?, 0x0?}, 0xc000775290, {0x49dad80, ...})
    /opt/cosign/cosign/pkg/cosign/bundle/sign.go:57 +0x476
github.com/sigstore/cosign/v3/cmd/cosign/cli/sign.SignBlobCmd(_, {0x0, {0x0, 0x0}, {0x7ffe1511735d, 0x45}, {0x4048341, 0x1b}, {0x404394e, 0x1a}, ...}, ...)
    /opt/cosign/cosign/cmd/cosign/cli/sign/sign_blob.go:98 +0x6cc
github.com/sigstore/cosign/v3/cmd/cosign/cli.SignBlob.func2(0xc0008a8908?, {0xc000816ba0, 0x1, 0x3fff29a?})
    /opt/cosign/cosign/cmd/cosign/cli/signblob.go:168 +0x7be
github.com/spf13/cobra.(*Command).execute(0xc0008a8908, {0xc000816b40, 0x6, 0x6})
    /root/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1015 +0xaaa
github.com/spf13/cobra.(*Command).ExecuteC(0xc00084ec08)
    /root/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1148 +0x46f
github.com/spf13/cobra.(*Command).Execute(0x0?)
    /root/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1071 +0x13
main.main()
    /opt/cosign/cosign/cmd/cosign/main.go:64 +0x44f`

pujitha24:
Motivation:
crypto11.FindKeyPair (github.com/ThalesIgnite/crypto11@v1.2.5, keys.go:240-244)
returns (nil, nil) — no error — when no key pair on the PKCS#11 token matches
the configured key id/label (e.g. wrong slot, wrong object name, a HSM/driver
that behaves differently across platforms). GetKeyWithURIConfig did not check
for this, so it happily returned a *Key with a nil signer. Any caller that
then invokes PublicKey(), Sign(), SignMessage(), or VerifySignature() directly
on that Key hits a nil pointer dereference: exactly the panic reported in
#4540 at pkcs11key.go:199 (`k.signer.Public()`), by two independent users
signing against real PKCS#11 tokens/HSMs.

The package already guards against a nil signer in Verifier()/SignerVerifier()
(present since the original PKCS11 support commit in 2021), and cosign's two
callers of GetKeyWithURIConfig (pkg/signature/keys.go) do route through those
methods before returning a usable signer/verifier. This change closes the gap
at the source — GetKeyWithURIConfig itself — so a Key can never be handed out
with a nil signer in the first place. This is a stricter, single-point
guarantee than the two narrower existing checks, and protects any current or
future caller of the exported pkcs11key API that doesn't happen to go through
Verifier()/SignerVerifier() first.

This is a correctness/UX fix, not a security issue: no auth bypass or data
corruption is involved, and cosign is a one-shot CLI so a panic and a returned
error both terminate the process with a non-zero exit code either way. The
concrete improvement is a clear, actionable error message ("signer not set:
no key pair found for id=... label=... in slot/token") instead of a Go panic
and stack trace.

Approach:
Add a nil check right after the FindKeyPair calls in GetKeyWithURIConfig,
returning a descriptive error that wraps the existing SignerNotSet sentinel
(reusing the same idiom already used in Verifier()/SignerVerifier()). The
sentinel errors and the new error-construction helper are extracted into a
new untagged pkg/cosign/pkcs11key/errors.go so they, and a unit test for the
helper, can run without the CGO/hardware-dependent pkcs11key build tag.

Validation:
- go build ./...
- go test ./pkg/cosign/pkcs11key/... -v            (untagged, runs in default CI)
- CGO_ENABLED=1 go build -tags=pkcs11key,pivkey ./...
- CGO_ENABLED=1 go test -tags=pkcs11key ./pkg/cosign/pkcs11key/... -v
- go vet ./pkg/cosign/pkcs11key/... (both with and without -tags=pkcs11key)
- gofmt -l pkg/cosign/pkcs11key/
All of the above pass. Full end-to-end validation against a real/simulated
PKCS#11 token (SoftHSMv2) was not possible in this sandbox, consistent with
this repo's own e2e test (test/pkcs11_test.go) being gated behind an
additional "softhsm" build tag requiring SoftHSMv2 to be installed locally.

Fixes #4540


pkg/cosign/pkcs11key/pkcs11key.go
```diff
@@ -173,6 +167,12 @@ func GetKeyWithURIConfig(config *Pkcs11UriConfig, askForPinIfNeeded bool) (*Key,
 	if err != nil {
 		return nil, err
 	}
+	// FindKeyPair returns a nil signer with a nil error when no matching
+	// key pair is found, so we must check explicitly to avoid handing back
+	// a Key with a nil signer, which would later panic on use.
+	if signer == nil {
```
Hayden-IO, line 179: I appreciate adding a test, but since it's just adding an error message, can we inline the error message here? Shouldn't need a test.

pujitha24:
Good call, done. I inlined the error message directly at the `signer == nil` check in `GetKeyWithURIConfig` (removed the `errSignerNotFound` helper), and removed `pkg/cosign/pkcs11key/errors.go` and its unit test, moving the pre-existing sentinel error vars (`ContextNotInitialized`, `SignerNotSet`, `CertNotSet`) back into `pkcs11key.go`. Verified `go build ./...`, the tagged `pkcs11key` build/vet, and `gofmt` all still pass.

Hayden-IO (approved):

pujitha24:
Thanks for the review, @Hayden-IO! The error message is inlined at the `signer == nil` check as suggested, the thread's resolved, and all checks are green. Happy to make any further tweaks — otherwise this should be good to go whenever you have a moment. 🙏

PR #5023, opened 2026-07-21
fix(verify): detect local sigstore bundles for saved image indexes

### LINKED ISSUE #4937: Verifying OCI image index offline
**Description**

I've been trying to sign & verify offline with key. I'm using `sign` / `save` / `verify`.
The verify step fails with:

```
Error: no signatures associated with the image saved in img
error during command execution: no signatures associated with the image saved in img
```

This is the content of the index:

```
{
   "schemaVersion": 2,
   "manifests": [
      {
         "size": 856,
         "digest": "sha256:8f76951b91383d78e515159899508a92ffdd4d56c814a7c380ac13a919d089e0",
         "annotations": {
            "kind": "dev.cosignproject.cosign/imageIndex"
         }
      }
   ]
}
```

I'm wondering if the issue doesn't come from https://github.com/sigstore/cosign/pull/4626 which only look for `dev.cosignproject.cosign/image` or am I doing something wrong?

What I tried that work:
- `verify` directly from registry
- `sign` with `--recursive`, recover the hash of the platform leaf, `save` with that platform leaf and then `verify`

Version: 3.1.0

pujitha24:
Motivation:
Offline verification of a locally-saved multi-arch image index failed
with "no signatures associated with the image saved in <path>", even
though the same reference verified successfully against the registry
(sigstore/cosign#4937). `cosign save` on an image index writes the root
OCI layout entry annotated `dev.cosignproject.cosign/imageIndex`
(pkg/oci/layout/write.go), but `getLocalBundleDescriptors`
(pkg/cosign/verify.go), which backs `HasLocalBundles` /
`HasLocalAttestationBundles` / `GetLocalBundles`, only recognized the
`dev.cosignproject.cosign/image` annotation used for single images.
As a result, `cosign verify --local-image` on a saved image index never
found the target digest, `HasLocalBundles` reported false, and
verification silently fell back to the legacy v2 signature path, which
has no signature data to find for images signed with the new bundle
format, producing the misleading error above.

Approach:
Extend the annotation check in `getLocalBundleDescriptors` to also
accept `dev.cosignproject.cosign/imageIndex`, matching the annotation
`pkg/oci/layout.WriteSignedImageIndex` already writes for saved image
indexes. The two annotations are mutually exclusive on a given saved
layout's root entry (an entry is either a single image or an image
index, never both), so this cannot introduce ambiguity or select the
wrong digest. This is a detection fix only: it does not change how
found bundles are verified, so it does not weaken verification.

Validation:
- go build ./...
- go test ./pkg/cosign/...  (all pass)
- Added TestHasLocalSigstoreBundles_ImageIndex, which builds a saved
  image-index OCI layout with an OCI 1.1 referrer manifest carrying a
  sigstore bundle layer, and asserts hasLocalSigstoreBundles detects
  it. Verified (via git stash) that this test fails on the pre-fix
  code with the same underlying miss this issue reports, and passes
  with the fix.

Fixes #4937

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1843,10 +1843,11 @@ func getLocalBundleDescriptors(path string) ([]bundleDescriptor, *v1.Hash, error
 		return nil, nil, fmt.Errorf("getting index manifest: %w", err)
 	}
 
-	// Find the target image digest from the index manifest
+	// Find the target image digest from the index manifest. The saved entity
+	// may be a single image or a multi-arch image index, annotated accordingly.
 	var targetDigest v1.Hash
 	for _, m := range manifest.Manifests {
-		if val, ok := m.Annotations["kind"]; ok && val == "dev.cosignproject.cosign/image" {
+		if val, ok := m.Annotations["kind"]; ok && (val == "dev.cosignproject.cosign/image" || val == "dev.cosignproject.cosign/imageIndex") {
 			targetDigest = m.Digest
 			break
 		}
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -1843,8 +1843,9 @@ func getLocalBundleDescriptors(path string) ([]bundleDescriptor, *v1.Hash, error
 		return nil, nil, fmt.Errorf("getting index manifest: %w", err)
 	}
 
-	// Find the target image digest from the index manifest. The saved entity
-	// may be a single image or a multi-arch image index, annotated accordingly.
+	// Find the target digest from the index manifest. The saved entity may be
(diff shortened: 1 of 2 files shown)


pujitha24:
This has been sitting rebased and green for a couple weeks now — happy to make any changes if something would help move review along.

pujitha24:
Closing this — #5080 landed the same fix (extending `getLocalBundleDescriptors` to accept the `dev.cosignproject.cosign/imageIndex` annotation) plus an equivalent regression test, and it's already merged into main. No need to carry this one forward.

PR #5029, opened 2026-07-28
fix(sign): pass digest hash algorithm to underlying signer in SignData

### LINKED ISSUE #5018: HashiCorp Vault ECDSA P-521 key fails to sign images after upgrading from Cosign v2.4.1 to v3.1.1
## Description

After upgrading Cosign from v2.4.1 to v3.1.1, signing an OCI image with an existing ECDSA P-521 key stored in HashiCorp Vault Transit KMS fails with:

unexpected length of digest for hash function specified

The same Vault key and signing setup worked successfully with Cosign v2.4.1.

### Environment
Cosign version before upgrade: v2.4.1
Cosign version after upgrade: v3.1.1
KMS provider: HashiCorp Vault Transit Engine
Vault key type: ecdsa-p521
Key reference: hashivault://cosign
Signing mode: signing configuration supplied with --signing-config
Target: OCI image referenced by digest

signing_config.json
```
{
  "mediaType": "application/vnd.dev.sigstore.signingconfig.v0.2+json",
  "rekorTlogConfig": {
    "selector": "ANY"
  },
  "tsaConfig": {
    "selector": "ANY"
  }
}
```

### Steps to reproduce

1. Create or use an existing ECDSA P-521 key in HashiCorp Vault Transit: type: ecdsa-p521
2. Configure Cosign to access the Vault KMS key.
3. With Cosign v2.4.1, sign an image using the key. Signing succeeds.
4. Upgrade Cosign to v3.0.6.
5. Run:
```
cosign sign \
  --signing-config signing_config.json \
  --key hashivault://cosign \
  "$IMAGE"
```

### Actual behavior

Signing fails with the following error:

Error: signing [###/podman-signer@sha256:7e7fc7d716f85056e3685d48f4200d0b4540fb54f6d6e1cae567f5e36c76b164]: signing digest: signing bundle: error signing bundle: unexpected length of digest for hash function specified
error during command execution: signing [###/podman-signer@sha256:7e7fc7d716f85056e3685d48f4200d0b4540fb54f6d6e1cae567f5e36c76b164]: signing digest: signing bundle: error signing bundle: unexpected length of digest for hash function specified

### Expected behavior

Cosign v3.1.1 should continue to support signing with an existing HashiCorp Vault ECDSA P-521 key, as Cosign v2.4.1 did.

Cosign should provide a supported way to configure the digest algorithm used by the KMS signer. Or make the bundle-signing code to produce a SHA-512 digest when signing images since configured signer for images is hardcoded to SHA-256 digest.
https://github.com/sigstore/cosign/blob/089731c31f750153e9d91ec248c990e3b1991c2b/pkg/signature/keys.go#L157

### Additional context

This appears to be a regression related to the newer bundle-signing path.

The error suggests that different parts of the signing flow disagree about the digest algorithm:

The ECDSA P-521 public key causes the bundle-signing code to produce a SHA-512 digest.
The HashiCorp Vault KMS signer may have been initialized to expect a SHA-256 digest.
The resulting 64-byte digest is then rejected by a signer configured for a 32-byte SHA-256 digest.

Could Cosign expose a CLI or signing-configuration option that allows users to select the hash algorithm for an existing KMS key?

## Version
 v3.1.1

pujitha24:
Fixes #5018

Motivation

Signing with a HashiCorp Vault Transit KMS ECDSA P-521 key fails with
"signing digest: signing bundle: error signing bundle: unexpected
length of digest for hash function specified". This is a regression
in the bundle-signing path introduced when SignerVerifierKeypair was
added: it correctly computes the digest using the key's own default
hash algorithm (SHA-512 for P-521), but never told the underlying
signer.SignMessage which hash algorithm that digest was produced with.

KMS-backed SignerVerifiers created via
pkg/signature/keys.go:SignerVerifierFromKeyRef are constructed with a
hardcoded crypto.SHA256 default (unrelated to the actual key's curve),
so sigstore-go's ComputeDigestForSigning fell back to that default,
saw a 64-byte digest against an expected 32-byte SHA-256 digest, and
rejected it with the mismatch error above.

Approach

internal/key/svkeypair.go: SignerVerifierKeypair.SignData now also
passes signatureoptions.WithCryptoSignerOpts(hashType) alongside
WithDigest(digest), so the signer is told exactly which hash algorithm
produced the digest, regardless of what hash it was constructed with
by default. This affects every SignerVerifier-backed signing path (any
KMS provider, not just Vault) whenever the key's natural hash isn't
SHA-256 (e.g. P-384/P-521 ECDSA, larger RSA keys). For keys whose
natural hash is already SHA-256 (e.g. default P-256), this is a no-op:
the newly-passed option matches the signer's existing default.

Validation

go build ./...
go test ./internal/key/... ./pkg/signature/... ./pkg/cosign/bundle/... -count=1
  ok  	github.com/sigstore/cosign/v3/internal/key
  ok  	github.com/sigstore/cosign/v3/pkg/signature
  ok  	github.com/sigstore/cosign/v3/pkg/cosign/bundle

Added TestKMSKeypair_ECDSA_P521_SignData, which reproduces the digest
length/hash mismatch reported in the issue (verified it fails without
this fix with "expected underlying signer to be given crypto.SHA512
opts matching the digest, got <nil>", and passes with it). This was
verified with a unit test against a mock SignerVerifier reproducing
the KMS signer's option-handling behavior; it was not validated
against a live HashiCorp Vault instance.

### DIFF
--- a/internal/key/svkeypair.go
+++ b/internal/key/svkeypair.go
@@ -133,7 +133,7 @@ func (k *SignerVerifierKeypair) SignData(ctx context.Context, data []byte) ([]by
 		h := hashType.New()
 		h.Write(data)
 		digest = h.Sum(nil)
-		sOpts = append(sOpts, signatureoptions.WithDigest(digest))
+		sOpts = append(sOpts, signatureoptions.WithDigest(digest), signatureoptions.WithCryptoSignerOpts(hashType))
 	}
 
 	sig, err := k.sv.SignMessage(bytes.NewReader(data), sOpts...)
--- a/internal/key/svkeypair_test.go
+++ b/internal/key/svkeypair_test.go
@@ -38,9 +38,10 @@ import (
 
 // mockSignerVerifier is a mock implementation of signature.SignerVerifier for testing.
 type mockSignerVerifier struct {
-	pubKey    crypto.PublicKey
-	pubKeyErr error
-	signErr   error
+	pubKey     crypto.PublicKey
+	pubKeyErr  error
+	signErr    error
+	signerOpts crypto.SignerOpts
 }
 
 func (m *mockSignerVerifier) PublicKey(_ ...signature.PublicKeyOption) (crypto.PublicKey, error) {
@@ -50,10 +51,15 @@ func (m *mockSignerVerifier) PublicKey(_ ...signature.PublicKeyOption) (crypto.P
 	return m.pubKey, nil
 }
 
-func (m *mockSignerVerifier) SignMessage(_ io.Reader, _ ...signature.SignOption) ([]byte, error) {
+func (m *mockSignerVerifier) SignMessage(_ io.Reader, opts ...signature.SignOption) ([]byte, error) {
 	if m.signErr != nil {
 		return nil, m.signErr
 	}
+	var signerOpts crypto.SignerOpts
+	for _, opt := range opts {
+		opt.ApplyCryptoSignerOpts(&signerOpts)
+	}
+	m.signerOpts = signerOpts
 	return []byte("mock-signature"), nil
 }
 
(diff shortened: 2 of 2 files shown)


pujitha24:
Just checking in on this one — it's still rebased and green, happy to make any changes if something would help move review along.

PR #5032, opened 2026-07-31
fix: prevent shell completions for various options not taking filenames

scop:
#### Summary

SSIA.

This is non-exhaustive, there's surely more to go through. But it's an improvement.

#### Release Note

Prevent shell completion for various options that do not take filenames as arguments.

#### Documentation

No


Hayden-IO (approved):
Thanks!

PR #5036, opened 2026-08-02
fix(blob): compare file checksums case-insensitively

semx:
## Summary

`cosign initialize --root-checksum` rejects a correct checksum when it is written in upper case.

`hex.EncodeToString` always emits lower case, and `pkg/blob/load.go:107` compares it to the user-supplied value with `!=`. Hex is case-insensitive, and plenty of sources publish digests in upper case, so a user holding a perfectly good checksum is told it is wrong.

Reproducer against the real sigstore root:

```console
$ curl -sO https://tuf-repo-cdn.sigstore.dev/12.root.json
$ shasum -a 256 12.root.json | awk '{print toupper($1)}'
5FE1E509A47277183A004745942C0A116379297DA9F1CCA529774320F1A5DFAB

$ cosign initialize --root 12.root.json \
    --root-checksum 5FE1E509A47277183A004745942C0A116379297DA9F1CCA529774320F1A5DFAB
Error: incorrect checksum for file 12.root.json: expected 5FE1E509A47277183A004745942C0A116379297DA9F1CCA529774320F1A5DFAB but got 5fe1e509a47277183a004745942c0a116379297da9f1cca529774320f1a5dfab
```

The error message gives it away: expected and got are the same digest.

## Why this is more than cosmetic

`LoadFileOrURLWithChecksum` has exactly one caller, `cmd/cosign/cli/initialize/init.go:59`, and that is the TUF root of trust. So the person who hits this is doing the right thing: pinning a root they fetched out of band. What the tool tells them is that the root does not match. The nearest way out is `--force-skip-checksum-validation`, which is exactly the habit this flag exists to prevent.

## Change

`strings.EqualFold` instead of `!=`. Wrong checksums stay wrong: a `DEADBEEF`-prefixed digest is still rejected, and the prefixed forms (`sha256:`, `sha512:`) behave the same as before apart from case.

## Testing

Added upper-case sha256 and sha512 cases to the existing `TestLoadURLWithChecksum`. They fail on main and pass with the fix; the existing cases are untouched. `go vet ./pkg/blob/`, `gofmt` and `go build ./...` are clean.

Also ran both binaries end to end against sigstore's `12.root.json`:

| build | checksum | exit |
|---|---|---|
| main | lower case | 0 |
| main | UPPER CASE | 1 |
| fix | UPPER CASE | 0, initialization completes |
| fix | `sha256:` + UPPER CASE | 0 |
| fix | `DEADBEEF...` | 1, still rejected |

## Release Note

`cosign initialize --root-checksum` now accepts upper-case and mixed-case hex digests.

## Documentation

NA

### DIFF
--- a/pkg/blob/load.go
+++ b/pkg/blob/load.go
@@ -15,6 +15,7 @@
 package blob
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"crypto/sha512"
 	"encoding/hex"
@@ -103,11 +104,13 @@ func LoadFileOrURLWithChecksum(fileRef string, checksum string) ([]byte, error) 
 	}
 
 	checksumAlgo.Write(fileContent)
-	computedChecksum := hex.EncodeToString(checksumAlgo.Sum(nil))
-	// Hex is case-insensitive and checksums are commonly published in upper case,
-	// so compare without regard to case rather than rejecting a correct digest.
-	if !strings.EqualFold(computedChecksum, checksumValue) {
-		return nil, fmt.Errorf("incorrect checksum for file %s: expected %s but got %s", fileRef, checksumValue, computedChecksum)
+	computedChecksum := checksumAlgo.Sum(nil)
+	// Compare the digests rather than their hex spellings: hex is case-insensitive
+	// and checksums are frequently published in upper case, which a plain string
+	// comparison would reject. A value that is not valid hex cannot match either.
+	expectedChecksum, err := hex.DecodeString(checksumValue)
+	if err != nil || !bytes.Equal(computedChecksum, expectedChecksum) {
+		return nil, fmt.Errorf("incorrect checksum for file %s: expected %s but got %s", fileRef, checksumValue, hex.EncodeToString(computedChecksum))
 	}
 
 	return fileContent, nil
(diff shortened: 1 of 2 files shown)


pkg/blob/load.go
```diff
@@ -104,7 +104,9 @@ func LoadFileOrURLWithChecksum(fileRef string, checksum string) ([]byte, error)
 
 	checksumAlgo.Write(fileContent)
 	computedChecksum := hex.EncodeToString(checksumAlgo.Sum(nil))
-	if computedChecksum != checksumValue {
+	// Hex is case-insensitive and checksums are commonly published in upper case,
+	// so compare without regard to case rather than rejecting a correct digest.
+	if !strings.EqualFold(computedChecksum, checksumValue) {
```
Hayden-IO: Could we instead compare the bytes of the decoded `checksumValue` with `checksumAlgo.Sum()`? No concerns of case insensitivity then.

semx: Good call, done in the follow-up commit. Decoding `checksumValue` and comparing it to `checksumAlgo.Sum(nil)` with `bytes.Equal` drops the case question entirely instead of working around it.

One detail worth noting: I kept both the decode error and the mismatch on the same branch, so a value that is not valid hex still reports `incorrect checksum` rather than a new error type. That keeps the existing "certainly not a correct checksum value" case in `TestLoadURLWithChecksum` meaningful — it cannot decode to the digest, so it is still wrong for the same reason.

Verified end to end against sigstore's own `12.root.json`:

| checksum | exit |
|---|---|
| lower case | 0 |
| UPPER CASE | 0 |
| `sha256:` + UPPER CASE | 0 |
| one byte altered | 1 |
| truncated | 1 |
| not hex at all | 1 |

Hayden-IO (approved):

PR #5042, opened 2026-08-06
fix: only show public-instance privacy statement for the default Rekor URL

### LINKED ISSUE #4400: Problems with local Rekor server & `cosign attest`
**Question**

I tried attesting an image with an SPDX SBOM with the command `cosign attest --key cosign.key --type spdxjson --predicate mysbom.spdx.json sampleregistry.com/myorg/myimage@sha256:a`. When I specify `--tlog-upload=false`, everything runs just fine. But when I run Rekor server locally with the [provided compose file](https://github.com/sigstore/rekor/blob/main/docker-compose.yml), download the Rekor pub key from `http://localhost:3000/api/v1/log/publicKey` into a file to which I point the env variable `SIGSTORE_REKOR_PUBLIC_KEY` and specify `--rekor-url=http://localhost:3000`, I receive the following message:

```
        The sigstore service, hosted by sigstore a Series of LF Projects, LLC, is provided pursuant to the Hosted Project Tools Terms of Use, available at https://lfprojects.org/policies/hosted-project-tools-terms-of-use/.
        Note that if your submission includes personal data associated with this signed artifact, it will be part of an immutable record.
        This may include the email address associated with the account with which you authenticate your contractual Agreement.
        This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later, and is subject to the Immutable Record notice at https://lfprojects.org/policies/hosted-project-tools-immutable-records/.

By typing 'y', you attest that (1) you are not submitting the personal data of any other person; and (2) you understand and agree to the statement and the Agreement terms at the URLs listed above.
Are you sure you would like to continue? [y/N] 
```

Which obviously seems wrong as I wanted to use the local Rekor server and not the public one. Is there something I am obviously doing wrong or missing?

pujitha24:
Motivation:
When uploading to a transparency log, cosign always printed the
"hosted by sigstore a Series of LF Projects, LLC" privacy statement
and confirmation prompt, even when the user pointed --rekor-url at
their own local or self-hosted Rekor instance. The upload itself
correctly went to the custom instance, but the displayed text
described the public sigstore.dev data-retention/Terms of Use policy,
which is misleading and confusing for anyone running their own Rekor
server (e.g. via the rekor docker-compose setup).

This is a UX/cosmetic issue only: no upload destination or signing
behavior changes, only the confirmation text shown to the user.

Approach:
In ShouldUploadToTlog (cmd/cosign/cli/signcommon/common.go), only show
the public-instance privacy statement when the configured Rekor URL is
empty (unset) or equal to options.DefaultRekorURL
(https://rekor.sigstore.dev). A custom --rekor-url no longer triggers
the public-instance notice.

Validation:
- go build ./...
- go test $(go list ./... | grep -v third_party/) (equivalent to `make test`) — all packages pass
- make lint — 0 issues
- Added TestShouldUploadToTlog_CustomRekorURLSkipsPublicInstanceStatement
  in cmd/cosign/cli/signcommon/common_test.go, which calls
  ShouldUploadToTlog with a custom RekorURL and asserts the public-instance
  statement is not printed. Manually reverted the guard and reran the test
  to confirm it fails without the fix and passes with it restored.

Report: https://github.com/sigstore/cosign/issues/4400

Fixes #4400

### DIFF
--- a/cmd/cosign/cli/signcommon/common.go
+++ b/cmd/cosign/cli/signcommon/common.go
@@ -119,7 +119,9 @@ func GetKeypairAndToken(ctx context.Context, ko options.KeyOpts, cert, certChain
 func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, tlogUpload bool) (bool, error) {
 	upload := shouldUploadToTlog(ctx, ko, ref, tlogUpload)
 	var statementErr error
-	if upload {
+	// Only warn about the public good instance's data retention policy when
+	// actually uploading to it; a custom --rekor-url points elsewhere.
+	if upload && (ko.RekorURL == "" || ko.RekorURL == options.DefaultRekorURL) {
 		privacy.StatementOnce.Do(func() {
 			ui.Infof(ctx, privacy.Statement)
 			ui.Infof(ctx, privacy.StatementConfirmation)
--- a/cmd/cosign/cli/signcommon/common.go
+++ b/cmd/cosign/cli/signcommon/common.go
@@ -25,7 +25,9 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
+	"strings"
 
 	"net/http"
 	"time"
@@ -121,7 +123,7 @@ func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Refere
 	var statementErr error
 	// Only warn about the public good instance's data retention policy when
 	// actually uploading to it; a custom --rekor-url points elsewhere.
-	if upload && (ko.RekorURL == "" || ko.RekorURL == options.DefaultRekorURL) {
+	if upload && isPublicGoodRekorURL(ko.RekorURL) {
 		privacy.StatementOnce.Do(func() {
 			ui.Infof(ctx, privacy.Statement)
 			ui.Infof(ctx, privacy.StatementConfirmation)
@@ -133,6 +135,33 @@ func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Refere
 		})
 	}
 	return upload, statementErr
+}
+
+// publicGoodRekorHostSuffixes are the hostname suffixes of Rekor instances operated
+// as part of the sigstore public good instance (production and staging). A literal
+// comparison against options.DefaultRekorURL is not sufficient because the public
+// good instance is served from multiple region- and year-specific hostnames.
+var publicGoodRekorHostSuffixes = []string{".sigstore.dev", ".sigstage.dev"}
+
+// isPublicGoodRekorURL reports whether rekorURL points at the sigstore public good
+// instance (production or staging), which is the only case where the data-retention
+// privacy statement applies. An empty rekorURL is treated as the default public
+// instance.
+func isPublicGoodRekorURL(rekorURL string) bool {
+	if rekorURL == "" {
+		return true
+	}
+	parsed, err := url.Parse(rekorURL)
+	if err != nil || parsed.Hostname() == "" {
+		return false
+	}
+	host := strings.ToLower(parsed.Hostname())
+	for _, suffix := range publicGoodRekorHostSuffixes {
+		if host == suffix[1:] || strings.HasSuffix(host, suffix) {
+			return true
+		}
+	}
+	return false
 }
 
 func shouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, tlogUpload bool) bool {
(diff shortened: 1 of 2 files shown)


cmd/cosign/cli/signcommon/common.go
```diff
@@ -119,7 +119,9 @@ func GetKeypairAndToken(ctx context.Context, ko options.KeyOpts, cert, certChain
 func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Reference, tlogUpload bool) (bool, error) {
 	upload := shouldUploadToTlog(ctx, ko, ref, tlogUpload)
 	var statementErr error
-	if upload {
+	// Only warn about the public good instance's data retention policy when
+	// actually uploading to it; a custom --rekor-url points elsewhere.
+	if upload && (ko.RekorURL == "" || ko.RekorURL == options.DefaultRekorURL) {
```
bobcallaway: im not sure this is sufficient; it would be better to match against logs listed in the TUF root that maps back to the pgi (urls will change over time); @Hayden-IO wdyt?

pujitha24:
Good point — comparing against the literal default URL string is fragile since Rekor URLs can change over time, so this wouldn't reliably track "is this actually the public-good instance" going forward. Matching against the transparency logs listed in the TUF trusted root sounds like the more durable approach, but I don't want to guess at that implementation without being sure I'm using the right trust-root API correctly. I'll hold off on further changes here pending @Hayden-IO's input on the preferred approach — happy to implement once there's agreement on direction.

cmd/cosign/cli/signcommon/common.go
Hayden-IO: Yea, we can't use the static Rekor URL, we already have another set of region-specific and year-specific URLs for newer Rekor instances. The other problem is that the TUF repo isn't sufficient as a signal, because a user could also provide their own configuration file that uses the public instance.

I'd be OK with baking in *.sigstage.dev and *.sigstore.dev though.

pujitha24:
Thanks @Hayden-IO — that's a good, concrete signal to key off of. I've updated the check to match the Rekor URL's hostname against `*.sigstore.dev` and `*.sigstage.dev` (including the bare domains) instead of the literal default-URL string, so region/year-specific public-good instances are still recognized. Added test cases covering a region-specific URL (`rekor.us-central1.sigstore.dev`), staging, a custom self-hosted URL, and a couple of lookalike-domain cases (e.g. `sigstore.dev.evil.example.com`) to make sure the suffix match respects domain boundaries. `go build`/`go test`/`go vet` pass for the affected package.

cmd/cosign/cli/signcommon/common.go
```diff
@@ -133,6 +137,33 @@ func ShouldUploadToTlog(ctx context.Context, ko options.KeyOpts, ref name.Refere
...
+// good instance is served from multiple region- and year-specific hostnames.
+var publicGoodRekorHostSuffixes = []string{".sigstore.dev", ".sigstage.dev"}
+
+// isPublicGoodRekorURL reports whether rekorURL points at the sigstore public good
+// instance (production or staging), which is the only case where the data-retention
+// privacy statement applies. An empty rekorURL is treated as the default public
+// instance.
+func isPublicGoodRekorURL(rekorURL string) bool {
+	if rekorURL == "" {
+		return true
```
Hayden-IO: should this be false, or just remove and let `url.Parse` handle the empty string? I assume this is irrelevant though because if `upload == true`, there must be a valid log URL.

cmd/cosign/cli/signcommon/common_test.go
```diff
@@ -206,6 +207,60 @@ func Test_ParseOCIReference(t *testing.T) {
 	}
 }
 
+func TestShouldUploadToTlog_CustomRekorURLSkipsPublicInstanceStatement(t *testing.T) {
```
Hayden-IO: can these tests be grouped together using table driven testing?

cmd/cosign/cli/signcommon/common.go
Hayden-IO: Should we change this to `false`?

Hayden-IO (approved):

pujitha24:
@Hayden-IO good catch — done. `isPublicGoodRekorURL` now returns `false` for an empty `RekorURL` instead of `true`, since an empty URL means `NewSigningConfigFromKeyOpts` doesn't configure a Rekor service to upload to at all, so the public-good privacy statement shouldn't be shown in that case. Updated the corresponding test case (`"empty is not public good"`) to match. `go test ./cmd/cosign/cli/signcommon/...` passes.

PR #5046, opened 2026-08-07
Keep an attached certificate chain that carries no root

### LINKED ISSUE #3976: Intermediate cert and root CA must come via same source for verify to succeed
**Description**

Intermediate cert and root CA must come in via same source for verify to succeed.

**Expected behavior**

As long as chain of trust is complete, signing cert + intermediate cert + root CA should be able to be attached in registry or supplied to cosign verify via command line arguments. Specifically, it should be possible to have signing cert + intermediate cert attached to image in registry, while supplying only root CA for verification.

**Actual behavior**

1. intermediate cert + root CA are both attached as certchain using `cosign attach --cert-chain ...`, then verifiation versus root CA only works (but attached root cert is duplicated in two places)
2. if intermediate only is attached as cert chain, then verification with root CA only FAILS. In this case intermediate has to be supplied with root CA to verify (intermediate is duplicated in two places).

**Reproduction**

f I remove the root CA from the cert chain file [here](https://github.com/Nordix/metal3-clusterapi-docs/blob/226bedb94fa36220dd1176aaf2bd277a0842c639/security/container-signing-poc/cosign/gencrt.sh#L140), then cosign verify [here](https://github.com/Nordix/metal3-clusterapi-docs/blob/226bedb94fa36220dd1176aaf2bd277a0842c639/security/container-signing-poc/cosign/Makefile#L65) fails.

Relevant parts:

```text
# fails
cat "${OUTPUT_DIR}"/sub-ca.crt  > "${OUTPUT_DIR}"/certificate_chain.pem
# works
cat "${OUTPUT_DIR}"/sub-ca.crt "${OUTPUT_DIR}"/ca.crt > "${OUTPUT_DIR}"/certificate_chain.pem

cosign attach signature \
	--payload examples/payload.json \
	--signature examples/payloadbase64.sig \
	--certificate examples/leaf.crt \
	--certificate-chain examples/certificate_chain.pem \
	127.0.0.1:5003/alpine@sha256:33735bd63cf84d7e388d9f6d297d348c523c044410f553bd878c6d7829612735

cosign verify \
	--ca-roots=examples/ca.crt \
	--certificate-identity-regexp '.*' \
	--certificate-oidc-issuer-regexp '.*' \
	--private-infrastructure \
	--insecure-ignore-sct \
127.0.0.1:5003/alpine@sha256:33735bd63cf84d7e388d9f6d297d348c523c044410f553bd878c6d7829612735

Error: no matching signatures: cert verification failed: x509: certificate signed by unknown authority. Check your TUF root (see cosign initialize) or set a custom root with env var SIGSTORE_ROOT_FILE
main.go:69: error during command execution: no matching signatures: cert verification failed: x509: certificate signed by unknown authority. Check your TUF root (see cosign initialize) or set a custom root with env var SIGSTORE_ROOT_FILE
```

**Version**

```text
❯ cosign version

GitVersion:    v2.4.1
GitCommit:     unknown
GitTreeState:  unknown
BuildDate:     unknown
GoVersion:     go1.23.1
Compiler:      gc
Platform:      linux/amd64
```

**References**

1. [Slack discussion](https://sigstore.slack.com/archives/C01PZKDL4DP/p1733150465358289)

Eljees:
Fixes #3976.

When a signature carries an attached certificate chain, the intermediate pool is built like this:

```go
if len(chain) > 1 {
    if co.IntermediateCerts == nil {
        pool = x509.NewCertPool()
        for _, cert := range chain[:len(chain)-1] {
            pool.AddCert(cert)
        }
    }
}
```

Both parts assume the chain ends with a root. That holds when the signer attaches the whole chain, but not when the root CA is supplied separately — then every certificate in the chain is an intermediate. With a single intermediate attached, `len(chain) > 1` is false, the pool is never built, and the attached intermediate is dropped entirely.

That is the asymmetry in the report. Running the current logic verbatim against generated certificates:

```
chain=[intermediate]        -> pool=nil        the attached intermediate is discarded
chain=[intermediate, root]  -> pool with 1 cert
```

Which matches both observations: attaching intermediate + root works (with the root duplicated), attaching the intermediate alone fails unless it is also passed on the command line.

The pool building moves into `intermediatePoolFromChain`, which drops the last element only when it is actually self-signed, and keeps it otherwise. Existing behaviour for chains that do end with a root is unchanged.

Covered by a table test over the five shapes: empty, intermediate only, intermediate + root, two intermediates, root only. `pkg/cosign` is green with the change and on a clean tree.

`go build ./...` could not finish here — the module proxy timed out fetching `open-policy-agent/opa`, unrelated to this change; `go vet ./pkg/cosign/` and `gofmt` are clean.

AI-assisted (LLM used for drafting); the reproduction and runs above are mine.

### DIFF
--- a/pkg/cosign/verify.go
+++ b/pkg/cosign/verify.go
@@ -574,6 +574,30 @@ func validateCertExtensions(ce CertExtensions, co *CheckOpts) error {
 	return nil
 }
 
+// intermediatePoolFromChain builds the intermediate pool from a certificate chain that
+// was attached alongside a signature. The chain ends with a root only when the signer
+// attached the whole chain; when the root is supplied separately (co.RootCerts), every
+// certificate in the chain is an intermediate and must be kept, including the last one.
+// Returns nil when the chain contributes no intermediates.
+func intermediatePoolFromChain(chain []*x509.Certificate) *x509.CertPool {
+	intermediates := chain
+	if len(chain) > 0 {
+		last := chain[len(chain)-1]
+		if bytes.Equal(last.RawIssuer, last.RawSubject) {
+			// Self-signed: this is a root, not an intermediate.
+			intermediates = chain[:len(chain)-1]
+		}
+	}
+	if len(intermediates) == 0 {
+		return nil
+	}
+	pool := x509.NewCertPool()
+	for _, cert := range intermediates {
+		pool.AddCert(cert)
+	}
+	return pool
+}
+
 // ValidateAndUnpackCertWithChain creates a Verifier from a certificate. Verifies that the certificate
 // chains up to the provided root. Chain should start with the parent of the certificate and end with the root.
 // Optionally verifies the subject and issuer of the certificate.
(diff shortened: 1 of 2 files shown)


Hayden-IO:
Thank you for looking into this, but this code path will be removed as of https://github.com/sigstore/cosign/pull/4959/, so no need to fix these edge cases at this point.

PR #5048, opened 2026-08-07
fix: include annotations in implied payload for offline verify

### LINKED ISSUE #4024: Error in offline validation when using additional annotations
**Description**

When using annotations, I am able to validate them by using the signature being stored in the registry but not when using the offline validation capability. 

For being able to recreate the tasks I did:

cosign sign --upload=false --key awskms:///arn:aws:kms:eu-west-3:xxxxxx:alias/cosign-key  -a appname=myapp   --output-signature=mysig1.sig xxx/yyy:latest

cosign verify -a appname=myapp   --key cosign.pub --signature mysig1.sig  xxx/yyy:latest

I'm getting the following error message:
WARNING: using obsolete implied signature payload data (with digested reference index.docker.io/xxx/bla@sha256:db071ebcec3e74bfb9a6e0358a233f7b4cc38585d3201239b9239d2e287d7e9a); specify it explicitly with --payload instead
Error: no matching signatures: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest  &{Code:400 Message:verifying signature: crypto/rsa: verification error}
main.go:69: error during command execution: no matching signatures: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest  &{Code:400 Message:verifying signature: crypto/rsa: verification error}

When doing the validation online without having specified --upload=false or not specifying any annotation and doing then the offline validation by using --signature=mysig1.sig the validation is successful. 

I am not sure if this is actually a bug but when reading the documentation my expectation is that cosign verify -a appname=myapp   --key cosign.pub --signature mysig1.sig  xxx/yyy:latest should work in general.

**Version**

GitVersion:    2.4.1
GitCommit:     9a4cfe1aae777984c07ce373d97a65428bbff734
GitTreeState:  "clean"
BuildDate:     2024-10-03T17:01:50Z
GoVersion:     go1.23.2
Compiler:      gc
Platform:      darwin/arm64

pujitha24:
Motivation:
When signing with `cosign sign --upload=false -a key=value
--output-signature=sig.file`, cosign builds the signed payload from
payload.Cosign{Image, ClaimedIdentity, Annotations} including the
annotations (cmd/cosign/cli/sign/sign.go). When later verifying
offline with `cosign verify -a key=value --signature sig.file` and no
explicit `--payload`, cosign reconstructs an "implied" payload via
ObsoletePayload to compare against the signature, but that
reconstruction dropped annotations entirely. The reconstructed payload
therefore never matched the payload that was actually signed, so
verification always failed with "crypto/rsa: verification error" (or
equivalent) any time annotations were used with offline signing.

Approach:
Thread the caller-supplied annotations through ObsoletePayload so the
implied payload it builds matches what was used at signing time. The
verify call site (pkg/cosign/verify.go) now passes co.Annotations,
which is populated from the `-a` flag. The attach-signature call site
(cmd/cosign/cli/attach/sig.go) has no annotations flag, so it passes
nil, preserving its existing behavior exactly.

This is a pure payload-construction/serialization fix with no
dependency on registry, network, or transparency-log state: the
default (no-annotations) case is byte-for-byte unchanged.

Validation:
- go build ./...
- go test ./pkg/cosign/... ./cmd/cosign/cli/attach/... ./cmd/cosign/cli/verify/...
- go test $(go list ./... | grep -v third_party/)
  All pass. Added TestObsoletePayloadWithAnnotations in
  pkg/cosign/obsolete_test.go, which fails to compile against the
  pre-fix ObsoletePayload signature and asserts the annotation now
  appears in the marshaled payload's "optional" field once the fix is
  applied; confirmed by temporarily reverting the source changes and
  observing the test target fail to build.

Report: https://github.com/sigstore/cosign/issues/4024

Fixes #4024

### DIFF
--- a/cmd/cosign/cli/attach/sig.go
+++ b/cmd/cosign/cli/attach/sig.go
@@ -55,7 +55,7 @@ func SignatureCmd(ctx context.Context, regOpts options.RegistryOptions, sigRef, 
 
 	var payload []byte
 	if payloadRef == "" {
-		payload, err = cosign.ObsoletePayload(ctx, digest)
+		payload, err = cosign.ObsoletePayload(ctx, digest, nil)
 	} else {
 		payload, err = os.ReadFile(filepath.Clean(payloadRef))
 	}
--- a/pkg/cosign/obsolete.go
+++ b/pkg/cosign/obsolete.go
@@ -26,8 +26,8 @@ import (
 // ObsoletePayload returns the implied payload that some commands expect to match
 // the signature if no payload is provided by the user.
 // DO NOT ADD ANY NEW CALLERS OF THIS.
-func ObsoletePayload(ctx context.Context, digestedImage name.Digest) ([]byte, error) {
-	blob, err := (&payload.Cosign{Image: digestedImage}).MarshalJSON()
+func ObsoletePayload(ctx context.Context, digestedImage name.Digest, annotations map[string]interface{}) ([]byte, error) {
+	blob, err := (&payload.Cosign{Image: digestedImage, Annotations: annotations}).MarshalJSON()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


pujitha24:
This has been rebased and green for a week now — happy to make any changes that would make review easier, just let me know.

cmurphy (approved):

PR #5051, opened 2026-08-10
verify: fetch bundles once and reuse them for verification

codysoyland:
The bundle-format auto-detect in `cosign verify` and `cosign verify-attestation` calls `GetBundles` just to learn whether any bundles exist — downloading the referrers index plus every bundle manifest and blob — and throws the result away; verification then calls `GetBundles` again. Every bundle on the image is downloaded twice.

This keeps the bundles the auto-detect fetches and verifies them directly through a new `cosign.VerifyImageAttestationsWithBundles(ctx, bundles, hash, co)` — the tail of the existing bundle verification, split at the `GetBundles` seam. It takes already-fetched bundles and the digest they were fetched for, and performs no registry requests (pinned by an instrumented-registry test). The fallback decision is unchanged: any `GetBundles` error or an empty result selects legacy verification exactly as before. Later images in a multi-image invocation use the fetching path as before, and the `download` commands are untouched.

Measured request counts (all verifications exit 0):

| scenario | before | after |
|---|---|---|
| 5 bundle referrers, live production registry | 36 | 19 |
| 1 bundle referrer (zot v2.1.20; also go-containerregistry's test registry) | 8 | 5 |

Bundle fetches go from 2N to N — everything is downloaded exactly once and reused in-process.

Assisted by Claude Fable 5

### DIFF
--- a/cmd/cosign/cli/verify/verify.go
+++ b/cmd/cosign/cli/verify/verify.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/options"
 	"github.com/sigstore/cosign/v3/cmd/cosign/cli/sign"
@@ -145,6 +146,11 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 	}
 	vOfflineKey := verifyOfflineWithKey(c.KeyRef, c.CertRef, c.Sk, co)
 
+	// Bundles fetched while auto-detecting the bundle format, reused to
+	// verify the first image without fetching them again.
+	var bundles []*sgbundle.Bundle
+	var bundlesHash *v1.Hash
+
 	// Auto-detect bundle format for local images
 	if c.LocalImage {
 		hasBundles, err := cosign.HasLocalBundles(images[0])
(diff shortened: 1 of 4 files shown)


steiza (dismissed):
Oh man, part of me wants to say let's just get rid of old bundles, but this would be useful for a future Cosign v3 release.

We landed some changes that created merge conflicts though, so can you rebase?

codysoyland:
> Oh man, part of me wants to say let's just get rid of old bundles, but this would be useful for a future Cosign v3 release.

Oh yeah, since we don't need to do the auto-detect when we drop support for legacy signatures -- that makes sense. Selfishly, as an operator of the registry with the largest collection of sigstore artifacts, I'd like to have this patch to reduce pressure on it. 😅 

The branch is now rebased.

cmd/cosign/cli/verify/verify.go
```diff
@@ -236,7 +245,12 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 
 			if co.NewBundleFormat {
 				// OCI bundle always contains attestation
-				verified, bundleVerified, err = cosign.VerifyImageAttestations(ctx, ref, co, c.NameOptions...)
+				if i == 0 && len(bundles) > 0 {
+					// Reuse the bundles fetched by the format auto-detect.
+					verified, bundleVerified, err = cosign.VerifyImageAttestationsWithBundles(ctx, bundles, bundlesHash, co)
```
steiza, line 250: I don't think this is quite right if you give `cosign verify` a list of images with which to verify.

Above (around line 160) we check `images[0]` to see if there are "new" (protobuf) bundles or not, and then cache the returned bundles as `bundles`.

If I understand correctly, here we're verifying those resulting `bundles` (from `images[0]`) over and over again, regardless of which item in `images` we are in.

tiagovilasboas:
### Staff AppSec review

**Verdict: Approve with comments** (conversation comment — formal review UI often disabled.)

Nice reduction of double-fetch on the bundle auto-detect path — pinning verification to the `GetBundles` digest via `VerifyImageAttestationsWithBundles` is the right seam (and the instrumented-registry test that asserts zero further registry GETs is exactly the regression we want).

Review focuses:

1. **Digest binding / tag TOCTOU.** Probe uses `images[0]` then reuses `(bundles, bundlesHash)` only for `i == 0`. Please document (or assert) that verification is against the digest observed at probe time if the tag moves between probe and the later `ParseReference` in the loop — output/UX should not imply the live tag tip was what got verified.

2. **API contract.** `hash == nil` → hard error is good. Can a non-nil hash ever disagree with the bundle’s subject and still “succeed” partially? A unit case with mismatched hash vs bundle subject would lock the fail-closed behavior.

3. **Parity.** `verify_attestation.go` gates reuse with `co.NewBundleFormat && len(bundles) > 0`; `verify.go` nests under `co.NewBundleFormat` then `len(bundles) > 0`. Same outcome today — worth aligning the conditions so a future flag refactor doesn’t diverge.

4. **Coverage.** Patch coverage on the CLI branches is still thin (~22%). Even one Exec-level test that hits the reuse branch (not only the pkg helper) would help steiza’s merge confidence.

Optional: changelog / release note that operators of large registries should see fewer referrer downloads on `verify` / `verify-attestation` auto-detect. Not blocking.

PR #5054, opened 2026-08-13
fix: use correct Rekor URLs to activate privacy warning

aaronlew02:
#### Summary

This change updates the Rekor public good instance privacy warning for `sign` and `sign-blob` to appear only when the Rekor public good instance is to be used.

Previously, `ShouldUploadToTlog` only checked `KeyOpts.RekorURL`, despite the fact that  `KeyOpts.RekorURL` defaults to the public good Rekor URL even when a signing config is provided containing a different transparency log URL. Now, `ShouldUploadToTlog` instead checks the signing config created for users that do not provide one manually. 

To make this work, the following refactors occur:
1. The constructed signing config no longer omits `ko.RekorURL` when `--tlog-upload=false`. `NewSigningConfigFromKeyOpts` no longer has a `tlogUpload` input, and the relevant logic has been extracted for after `ShouldUploadToTlog` is called.
2. That signing config is now constructed before (instead of after) `ShouldUploadToTlog` is called.

### DIFF
--- a/cmd/cosign/cli/sign/sign.go
+++ b/cmd/cosign/cli/sign/sign.go
@@ -202,9 +202,12 @@ func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpt
 		if err != nil {
 			return fmt.Errorf("should upload to tlog: %w", err)
 		}
-		ko.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(ko, shouldUpload)
+		ko.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(ko)
 		if err != nil {
 			return fmt.Errorf("creating signing config: %w", err)
+		}
+		if !shouldUpload {
+			ko.SigningConfig = ko.SigningConfig.WithRekorLogURLs()
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


Hayden-IO:
https://github.com/sigstore/cosign/pull/5055 for the linter fix

cmd/cosign/cli/sign/sign.go
```diff
@@ -198,13 +198,16 @@ func signDigestBundle(ctx context.Context, digest name.Digest, ko options.KeyOpt
 	}
 
 	if ko.SigningConfig == nil {
+		ko.SigningConfig, err = signcommon.NewSigningConfigFromKeyOpts(ko)
+		if err != nil {
+			return fmt.Errorf("creating signing config: %w", err)
+		}
 		shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, ko, digest, signOpts.TlogUpload)
```
Hayden-IO, line 205: Because this is what displays the privacy statement, should this be unconditionally called outside of this block?

aaronlew02: Since even before #4618, the privacy statement only appears when a signing config is not provided.

We can change this to support manually-provided signing configs, but that would be a potentially breaking change in UX.

aaronlew02: In #4959 (planned for v4), this is called unconditionally.

Hayden-IO: Ah I'm getting my wires crossed between these changes.

We probably should have had the privacy statement added with the signing config path at the start of v3, but I agree, it's a breaking change and we might as well just wait to unify this in v4.

Hayden-IO:
can you rebase?

Hayden-IO (approved):

bobcallaway:
@Hayden-IO quick question: I thought you wanted a URL-based match when we were discussing in the other PR?

Hayden-IO:
That shouldn't change, matching should still be based on URLs. The previous PR didn't account for the divergent codepaths between the Cosign signing API and the sigstore-go signing API, so it was only checking the `--rekor-url` flag value and skipping the URLs provided in the signing config. Now, the signing config is the source of truth, with it being populated from per-service flag values.

Hayden-IO (approved):

PR #5058, opened 2026-08-14
fix: strip trailing whitespace from attached signature files/stdin

### LINKED ISSUE #4207: Cosign Verification Failure After Image Transfer
I am experiencing a Cosign verification failure after transferring a signed container image between two OCI registries hosted on different servers. While the image digest remains unchanged, the signature payload appears to be modified upon attachment, causing verification errors.

Expected behaviour
-------------------------------------
We expect the Cosign signature to remain valid after transferring the signed image to a different OCI registry. The digest should stay unchanged, and verification should succeed in target mechine

The commands used in my scenario:

`Machine 1: Docker registry
-------------------------------
```
cosign sign --key cosign.key localhost:5000/my-ubuntu-4:1.0
skopeo copy --preserve-digests --src-tls-verify=false docker://localhost:5000/my-ubuntu-4:1.0 dir:/home/sarath/my-ubuntu-skopeo/
cd /home/sarath/my-ubuntu-skopeo/
cosign download signature localhost:5000/my-ubuntu-4:1.0 | jq -r '.Base64Signature' > signature.sig
cosign download signature localhost:5000/my-ubuntu-4:1.0 > payload.json
tar -cvf my-ubuntu-skopeo.tar -C /home/sarath/my-ubuntu-skopeo/ .

```
Machine 2: Containerd registry
---------------------------------------
```
tar -xvf my-ubuntu-skopeo.tar
skopeo copy --preserve-digests --dest-tls-verify=false dir:/home/sarath/dockerTest/my-ubuntu-skopeo docker://localhost:5000/my-ubuntu-4:1.0
cosign attach signature --payload payload.json --signature signature.sig localhost:5000/my-ubuntu-4:1.0

we encountered with an error:

cosign verify --key cosign.pub localhost:5000/my-ubuntu-4:1.0
Error: no matching signatures: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest {"code":400,"message":"verifying signature: invalid signature when validating ASN.1 encoded signature"}
error during command execution: no matching signatures: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest {"code":400,"message":"verifying signature: invalid signature when validating ASN.1 encoded signature"}
```

We used Skopeo to preserve the image digest, ensuring that the digest remains the same on both the source and target machines. Additionally, the signature is attached to the tag,but still facing issue 

`  curl -X GET 'http://localhost:5000/v2/my-ubuntu-4/tags/list'`

Image

```
Version:
GitVersion:    v2.5.0
GitCommit:     38bb98697005cdc5c092f031594c0e45d039f4a0
GitTreeState:  clean
BuildDate:     2025-04-07T20:33:49Z
GoVersion:     go1.24.1
Compiler:      gc
Platform:      linux/amd64
```

pujitha24:
Motivation:
The linked report below describes a failure after detaching a
signature with `cosign download signature ... | jq -r
'.Base64Signature' > signature.sig` and later re-attaching it on
another registry with `cosign attach signature --signature
signature.sig ...`. The reporter's own --verbose debug output shows
the OCI manifest annotation dev.cosignproject.cosign/signature on the
re-attached image literally contains the base64 signature value with
a trailing newline embedded in it, unlike the original signing
machine's copy. That newline comes from shell redirection after
`jq -r`; cosign's attach signature command stores whatever bytes it
reads from the --signature file or stdin verbatim, with no trimming.

Storing extraneous whitespace inside a signature annotation is a
real, demonstrated defect regardless of whether it fully explains the
"invalid signature when validating ASN.1 encoded signature" error the
reporter hit: Go's own base64 decoder tolerates embedded \r/\n and
decodes to the same bytes, so this alone would not necessarily
reproduce that exact failure, and the report's repro also mixed up
`cosign download signature`'s JSON envelope for --payload, a separate
issue. This change fixes the annotation-pollution defect on its own
merits; it is not a claim that it resolves the full reported scenario
end-to-end.

Approach:
Trim leading and trailing whitespace from bytes read from a signature
file or stdin in signatureBytes() (cmd/cosign/cli/attach/sig.go)
before they are stored as the signature annotation. Base64 has no
whitespace in its alphabet, so trimming cannot remove meaningful
signature bytes. The literal-argument path is untouched.

Validation:
- go test ./cmd/cosign/cli/attach/... — added
  TestSignatureBytesTrimsTrailingWhitespace, which fails before this
  change (returned bytes include the trailing "\n") and passes after;
  verified this directly by stashing the fix and re-running the test.
  Also added TestSignatureBytesFileWithoutTrailingWhitespaceUnaffected
  confirming clean input is unaffected.
- go build ./... and go vet ./cmd/cosign/cli/attach/... pass.
- go test ./cmd/cosign/... passes (all subpackages).
- golangci-lint run ./cmd/cosign/cli/attach/... (v2.12.2, matching
  this repo's Dockerfile.golangci-lint) reports 0 issues.

Report: https://github.com/sigstore/cosign/issues/4207

Fixes #4207

### DIFF
--- a/cmd/cosign/cli/attach/sig_test.go
+++ b/cmd/cosign/cli/attach/sig_test.go
@@ -1,0 +1,60 @@
+//
+// Copyright 2025 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attach
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestSignatureBytesTrimsTrailingWhitespace(t *testing.T) {
+	const want = "MEQCIGnMnU8FiEcNVIgeR6HxNp4Gaqg5q0XLSNhBO5pp+QaNAiB40iDMGGQdthm8U9qQeCAzp6ecLdKCG3R/yq5S2uMa6g=="
+
+	dir := t.TempDir()
+	sigFile := filepath.Join(dir, "signature.sig")
+	// Simulate a signature file produced with `jq -r '.Base64Signature' > signature.sig`,
+	// which appends a trailing newline.
+	if err := os.WriteFile(sigFile, []byte(want+"\n"), 0600); err != nil {
+		t.Fatal(err)
+	}
+
+	got, err := signatureBytes(sigFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(got) != want {
+		t.Errorf("signatureBytes() = %q, want %q", string(got), want)
+	}
+}
+
+func TestSignatureBytesFileWithoutTrailingWhitespaceUnaffected(t *testing.T) {
+	const want = "MEQCIGnMnU8FiEcNVIgeR6HxNp4Gaqg5q0XLSNhBO5pp+QaNAiB40iDMGGQdthm8U9qQeCAzp6ecLdKCG3R/yq5S2uMa6g=="
+
+	dir := t.TempDir()
+	sigFile := filepath.Join(dir, "signature.sig")
+	if err := os.WriteFile(sigFile, []byte(want), 0600); err != nil {
+		t.Fatal(err)
+	}
+
+	got, err := signatureBytes(sigFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(got) != want {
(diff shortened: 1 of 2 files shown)


pujitha24:
The failing checks here are not caused by this diff — flagging in case that is what is holding it up.

The e2e-registry job fails at "Install cluster + sigstore" with `curl: (35) Recv failure: Connection reset by peer` while downloading `setup-kind.sh` from the sigstore/scaffolding releases — a network failure before any cosign code executes.

Happy to rebase to re-trigger CI if that would help, or to dig further if you think I have this wrong.

cmd/cosign/cli/attach/sig_test.go
```diff
@@ -0,0 +1,60 @@
+//
+// Copyright 2025 The Sigstore Authors.
```
cmurphy: Could you update the copyright to 2026?

cmurphy (commented):
Very clever catch. Could you fix the copyright year? That should also re-trigger the e2e tests.

pujitha24:
Fixed, updated to 2026 in the new test file. Pushing this now, hopefully it kicks off a fresh e2e run.

cmurphy (approved):

PR #5059, opened 2026-08-14
Add offline signing flag

### LINKED ISSUE #5013: Offline signing flag
**Description**

v4 will require that users who need services for signing must specify those services via a signing config. As of #4959, this can be done either by providing a signing config explicitly (via `--signing-config`) or relying on one fetched automatically by TUF (`--use-signing-config`). Users signing offline must set `--use-signing-config=false`. 

Since `--use-signing-config=false` was originally intended to allow for specification of services via URL flags instead of a signing config, the flag can become semantically confusing for users.

This issue proposes removing `--use-signing-config` in favor of:
* Signers requiring TUF services simply omitting `--signing-config`
* Signers **NOT** requiring TUF services omitting `--signing-config` and using a new flag (e.g., `--offline`)

aaronlew02:
Closes #5013

#### Summary

This change adds an `--offline` flag to signing and attestation commands to sign without connectivity to any online services.

This change removes the deprecated `--use-signing-config` flag.

#### Release Note

* A signing config and trusted root are fetched from TUF by default.
* Either file can instead be provided manually with `--signing-config` or `--trusted-root`.
* `--offline` will:
    * Prevent either file from being fetched by TUF. 
    * Disallow `--signing-config`, `--trusted-root`, `--issue-certificate`, and custom service URLs.
    * Require a private key (`--key`).


cmd/cosign/cli/signcommon/common.go
```diff
@@ -468,49 +467,55 @@ func ParseSignatureAlgorithmFlag(signingAlgorithm string) (pb_go_v1.PublicKeyDet
...
+		ko.SigningConfig = NewEmptySigningConfig()
+		return nil
 	}
-	if (useSigningConfig || signingConfigPath != "") && !tlogUpload {
-		return fmt.Errorf("--tlog-upload=false is not supported with --signing-config or --use-signing-config. Provide a signing config with --signing-config without a transparency log service, which can be created with `cosign signing-config create` or `curl https://raw.githubusercontent.com/sigstore/root-signing/refs/heads/main/targets/signing_config.v0.2.json | jq 'del(.rekorTlogUrls)'` for the public instance")
+	if serviceURLsSpecified {
+		return fmt.Errorf("cannot specify service URLs when using a signing config")
+	}
+	if !tlogUpload {
+		return fmt.Errorf("--tlog-upload=false is not supported with a signing config. Provide a signing config with --signing-config without a transparency log service, which can be created with `cosign signing-config create` or `curl https://raw.githubusercontent.com/sigstore/root-signing/refs/heads/main/targets/signing_config.v0.2.json | jq 'del(.rekorTlogUrls)'` for the public instance")
```
Hayden-IO: Can this error message be updated to mention the offline flag as an alternative now? Or as the only alternative, rather than mentioning an empty signing config? I think we can remove the mention of deleting just the Rekor URLs because it still logs the certificate issuance, and I think those who would want to skip the tlog upload would also want to skip Fulcio/CT logging too.

cmd/cosign/cli/signcommon/common.go
```diff
@@ -468,49 +467,55 @@ func ParseSignatureAlgorithmFlag(signingAlgorithm string) (pb_go_v1.PublicKeyDet
...
+	if !newBundleFormat && bundlePath == "" {
+		return fmt.Errorf("must provide --new-bundle-format or --bundle where applicable with a signing config")
+	}
+	if trustedRootPath != "" {
+		ko.TrustedMaterial, err = root.NewTrustedRootFromPath(trustedRootPath)
+		if err != nil {
+			return fmt.Errorf("loading trusted root: %w", err)
+		}
+	} else {
+		ko.TrustedMaterial, err = cosign.TrustedRoot()
```
Hayden-IO: If this is meant to be entirely offline, Cosign shouldn't fetch metadata from the TUF repo, and it shouldn't need to since it won't rely on any infrastructure for signing or logging. Same with signing config below.

cmd/cosign/cli/options/attest.go
```diff
@@ -129,15 +129,15 @@ func (o *AttestOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.NewBundleFormat, "new-bundle-format", true, "attach a Sigstore bundle using OCI referrers API")
 	_ = cmd.Flags().MarkDeprecated("new-bundle-format", "this will be the only supported format in future versions")
 
-	cmd.Flags().BoolVar(&o.UseSigningConfig, "use-signing-config", true,
-		"whether to use a TUF-provided signing config for the service URLs")
-	_ = cmd.Flags().MarkDeprecated("use-signing-config", "an offline signing flag will be added in the future; TUF will continue to provide a signing config by default if one is not provided manually")
+	cmd.Flags().BoolVar(&o.Offline, "offline", false,
+		"only allow offline signing with a local key without contacting network services")
```
Hayden-IO: Can the flag's documentation call out that the key usage isn't logged and therefore isn't auditable? I want to convey that the signer is acknowledging that using this flag means they are not mitigating some of the threats outlined in https://docs.sigstore.dev/about/threat-model/.

cmd/cosign/cli/signcommon/common.go
Hayden-IO, line 539: It'd be cool if we could also have an e2e test that demonstrates signing is offline. Something like, clear the TUF repo directory, call sign, verify the directory is still empty?

cmd/cosign/cli/signcommon/common.go
```diff
@@ -468,49 +467,55 @@ func ParseSignatureAlgorithmFlag(signingAlgorithm string) (pb_go_v1.PublicKeyDet
...
+		if keyRef == "" {
+			return fmt.Errorf("offline signing requires a private key")
+		}
+		if issueCertificate {
+			return fmt.Errorf("cannot issue certificate when offline")
+		}
+		if serviceURLsSpecified {
+			return fmt.Errorf("cannot specify service URLs when signing offline")
+		}
+		ko.SigningConfig = NewEmptySigningConfig()
```
Hayden-IO: nit, can you add a comment documenting this line? I think it's easy to miss that this is what prevents any network calls.

Hayden-IO (commented):
Looks good, just a few small suggestions! Can you also update the README to document this? While we need to do a lot of updates for the main README, I think this is worth calling out.

cmd/cosign/cli/signcommon/common.go
aaronlew02: > Can this error message be updated to mention the offline flag as an alternative now?

Yes!

> Or as the only alternative, rather than mentioning an empty signing config?

Done ([`13f7fa1`](https://github.com/sigstore/cosign/pull/5059/commits/13f7fa1b5132e747d2e40d956ce102b4b5351b33)).

> I think we can remove the mention of deleting just the Rekor URLs because it still logs the certificate issuance, and I think those who would want to skip the tlog upload would also want to skip Fulcio/CT logging too.

Great point.

cmd/cosign/cli/signcommon/common.go
aaronlew02: > It'd be cool if we could also have an e2e test that demonstrates signing is offline.

I had updated `TestSignVerifyBundleOffline` to set `Offline` to `true` instead of `TlogUpload` to `false`.

> Something like, clear the TUF repo directory, call sign, verify the directory is still empty?

I have further updated `TestSignVerifyBundleOffline` to do this.

cmd/cosign/cli/options/attest.go
aaronlew02: Done ([`8868640`](https://github.com/sigstore/cosign/pull/5059/commits/8868640bc7748a521c73b0ed325cad5900226b28)).

cmd/cosign/cli/signcommon/common.go
aaronlew02: Done ([`68a26ba`](https://github.com/sigstore/cosign/pull/5059/commits/68a26ba84738ffffc3b5a3b23538b723b0b7c119)).

aaronlew02:
I've updated the `README` to include instructions for offline signing.

Hayden-IO (dismissed):
Looks great!

cmd/cosign/cli/signcommon/common.go
aaronlew02: Both of these changes are actually meaningless since that test doesn't execute `LoadTrustedMaterialAndSigningConfig` .

Instead, I've added a unit test for `LoadTrustedMaterialAndSigningConfig` that demonstrates that:
* Offline signing bypasses TUF, sets an empty signing config, and leaves the TUF directory empty.
* Online signing attempts to contact TUF and fails when the TUF mirror is invalid (negative control).

steiza (changes requested):
Sorry, I am way late to this pull request 😞 

But I want to be really careful about new flags we add - both in terms of limiting their number and also to be clear about what the flags do.

`--offline` does two things, I think:
- It turns off verification after signing, so that we don't try to fetch an updated trusted root via TUF
- It ensures that things signed with a key don't create a transparency log in Rekor

I think these are good use-cases to support, but I'm not sure we want to group things in this way.

For instance: if I'm signing with a key and don't want a transparency log entry but do want a signed timestamp (which I think everyone should want when signing with a key!) `--offline` doesn't make much sense because I'm probably reaching out over the network to the TSA.

I acknowledge I'm way late here and might be missing some context, so feedback welcome!

aaronlew02:
@steiza Thanks for taking a look! 

The goal of `--offline` is to sign without any online service connectivity, so the results you mention are consequences of this. However, users who want to use one type of online service but not another (e.g., a TSA but not a transparency log) should not use `--offline` and instead provide a signing config manually that contains only the desired service.

I've updated the error message for setting `--tlog-upload=false` with a signing config to clarify this.

cmurphy:
Sorry for coming to this so late as well. I remember some time ago having a discussion about deprecating `--tlog-upload`, with the consensus that to avoid uploading to rekor one could simply omit rekor from the signing config. Could this offline mode be accomplished by just having an entirely empty signing config?

I believe there is already an `--offline` flag for verification, which I know has been problematic because it's not truly "offline". I worry slightly that introducing another `--offline` flag for a different command will 1) get confused with the deprecated `--offline` flag that will be removed for verification, and 2) as Zach alluded to, "offline" (or "air-gapped") can mean slightly different things to different people, with different services enabled within an air-gapped environment, so prescribing one strict meaning for "offline" - even if it's much more correct than our older use of "offline" - might be missing core use cases.

aaronlew02:
@cmurphy 

> Sorry for coming to this so late as well.

No worries. Thanks for taking a look!

> Could this offline mode be accomplished by just having an entirely empty signing config?

Yes. However, in this PR, a user providing an empty signing config will still receive a trusted root from TUF, and while verification should succeed, this will not be "offline." To fully disconnect from TUF, the user would have to provide an empty trusted root as well. The `--offline` flag reduces the burden of being concerned with either file.

As noted in #5013, the new `--offline` flag is intended to resolve semantic confusion arising from #4959 changing the purpose of `--use-signing-config=false` with the removal of service URL flags.

Instead of adding an `--offline` flag, we could switch the [current](https://github.com/sigstore/cosign/blob/e9a91ea113bfc8102a4385c4ac477285e4b74cee/cmd/cosign/cli/signcommon/common.go#L536) order of events in `LoadTrustedMaterialAndSigningConfig` so that a signing config is always initialized first (by default from TUF, otherwise from a user-provided file -- including an empty one), and then skip fetching a trusted root from TUF (unless the user provides one) if the initialized signing config is empty.

cmurphy:
> Instead of adding an --offline flag, we could switch the current order of events in LoadTrustedMaterialAndSigningConfig so that a signing config is always initialized first (by default from TUF, otherwise from a user-provided file -- including an empty one), and then skip fetching a trusted root from TUF (unless the user provides one) if the initialized signing config is empty.

I like this suggestion. I think that allows for more flexibility to accommodation any environment.

aaronlew02:
@cmurphy #5125 rearranges `LoadTrustedMaterialAndSigningConfig` (keeping `--use-signing-config` for now) if we prefer to move forward with that.

aaronlew02:
This approach has been abandoned in favor of #5125.

PR #5062, opened 2026-08-17
fix(providers): clearer errors when GitHub OIDC token request fails

### LINKED ISSUE #4438: better error message when OIDC fails
Sigstore signing depends on many moving parts and sometimes one of them fails: this is acceptable. Currently projects that embed cosign can look a little bad though when this happens since the error messages don't make it clear what is failing.

A case I've seen multiple times is signing failing on GitHub because the GitHub OIDC provider is malfunctioning -- knowing that this is an issue on the GitHub service would be useful but currently it says: `error signing scorecard results: getting key from Fulcio: fetching ambient OIDC credentials: invalid character 'u' looking for beginning of value` (this example is  from a `ossf/scorecard-action` workflow failure)

This fails in `auth.ReadIDToken()` My assumption is that the last part of the error comes from json parser: We could improve this by replacing json errors with something like `Invalid identity token response from provider %s`.

ekalinin:
Fixes #4438

## Summary

The GitHub Actions provider decoded the response body without ever checking the status code. When the GitHub OIDC endpoint misbehaves and returns a non-JSON body (for example `upstream connect error or disconnect/reset before headers`, mentioned in the issue), the decoder trips on the first byte and the user sees:

```
error signing scorecard results: getting key from Fulcio: fetching ambient OIDC credentials: invalid character 'u' looking for beginning of value
```

which says nothing about where the failure actually is.

## Changes

- Check `resp.StatusCode` before decoding, and report the status together with the response body. The body is read through an `io.LimitReader` (256 bytes) so a large HTML or proxy error page does not flood the output, and it is formatted with `%q` so an empty body and embedded newlines stay readable on one line.
- Wrap decoder failures with the provider name, as suggested in the issue: `invalid identity token response from provider github-actions`.
- Add a `providerName` constant and use it both in `providers.Register` and in the error messages, so the two cannot drift apart.

The result:

```
fetching ambient OIDC credentials: identity token request to provider github-actions failed with status 502 Bad Gateway: "upstream connect error or disconnect/reset before headers"
```

Retrying on 5xx is deliberately left out: that is a behaviour change rather than a message change, and it belongs in its own PR.

## Tests

Table-driven `TestProvide_ErrorResponse` on top of the existing `httptest` harness, covering a 502 with a non-JSON body, truncation of a long body, an empty body, and malformed JSON on a 200. Each case was confirmed to fail against the unpatched provider.

## Release Note

```release-note
Improve error messages when the GitHub Actions OIDC provider returns a failing status or a malformed identity token response.
```

### DIFF
--- a/pkg/providers/github/github.go
+++ b/pkg/providers/github/github.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strings"
@@ -33,10 +34,15 @@ const (
 	RequestTokenEnvKey = env.VariableGitHubRequestToken
 	// Deprecated: use `env.VariableGitHubRequestURL` instead
 	RequestURLEnvKey = env.VariableGitHubRequestURL
+
+	// providerName is the name this provider is registered under.
+	providerName = "github-actions"
+	// maxErrorBodyBytes limits how much of a failed response body is quoted back in the error.
+	maxErrorBodyBytes = 256
 )
 
 func init() {
-	providers.Register("github-actions", &githubActions{})
+	providers.Register(providerName, &githubActions{})
 }
 
 type githubActions struct{}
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
thank you!

PR #5064, opened 2026-08-17
fix: verify-blob should not force SHA256 for non-default key hash algorithms

### LINKED ISSUE #5063: `verify-blob` requires deprecated `--signature-digest-algorithm` for non-SHA-256 bundled signatures
**Description**
`cosign verify-blob` warns that `--signature-digest-algorithm` is deprecated because `--bundle` includes the digest algorithm. However, verification of a SHA-512 blob signature fails unless the deprecated flag is explicitly supplied.

The following commands can be used to reproduce the issue.

Generate the key and signature:
```
$ openssl ecparam -name secp521r1 -genkey -noout -out p521.pem
$ cosign import-key-pair --key p521.pem
$ cosign sign-blob --key import-cosign.key --bundle foo.txt.sig --signing-config ./signing-config.json --yes foo.txt
```

Verifying without `--signature-digest-algorithm` fails:
```
$ cosign verify-blob --key import-cosign.pub --bundle foo.txt.sig --insecure-ignore-tlog=true foo.txt
WARNING: Skipping tlog verification is an insecure practice that lacks transparency and auditability verification for the blob.
Error: failed to verify signature: could not verify message: invalid signature when validating ASN.1 encoded signature
error during command execution: failed to verify signature: could not verify message: invalid signature when validating ASN.1 encoded signature
```

Verifying with `--signature-digest-algorithm` succeeds:
```
$ cosign verify-blob --key import-cosign.pub --bundle foo.txt.sig --insecure-ignore-tlog=true --signature-digest-algorithm sha512 foo.txt
Flag --signature-digest-algorithm has been deprecated, please use --bundle, which already includes the digest algorithm
WARNING: Skipping tlog verification is an insecure practice that lacks transparency and auditability verification for the blob.
Verified OK
```

**Version**
```
$ cosign version
cosign: A tool for Container Signing, Verification and Storage in an OCI registry

GitVersion:    v3.1.3
GitCommit:     11926fa5bbbbde47e88fc006b625a17769b743b2
GitTreeState:  clean
BuildDate:     2026-08-05T23:43:27Z
GoVersion:     go1.26.4
Compiler:      gc
Platform:      linux/amd64
```

pujitha24:
Motivation:
cosign verify-blob fails to verify legacy-bundle blob signatures made
with keys whose conventional digest algorithm is not SHA256 (e.g. an
ECDSA P-521 key, which pairs with SHA-512), unless the user explicitly
passes the deprecated --signature-digest-algorithm flag. Without it,
verification fails with "invalid signature when validating ASN.1
encoded signature".

Approach:
The --signature-digest-algorithm flag had a hardcoded default value of
"sha256", so its value was indistinguishable from an explicit
--signature-digest-algorithm sha256. SignatureDigestOptions.HashAlgorithm()
compounded this by also returning SHA256 for an empty/unset value.
VerifyBlobCmd.Exec then forced any zero HashAlgorithm to SHA256 before
loading the verifier, discarding the zero-value sentinel that
LoadPublicKeyRaw (pkg/signature/keys.go) relies on to fall back to
signature.LoadDefaultVerifier, which selects the correct digest
algorithm for the given key type (SHA256 for P-256/RSA, SHA384 for
P-384, SHA512 for P-521).

This change makes the flag default to an empty string, has
HashAlgorithm() return 0 (unset) when the flag isn't provided, and
removes the forced-SHA256 default in VerifyBlobCmd.Exec, letting the
key-loading path pick the digest algorithm that matches the key. The
same forced-default pattern still exists, unchanged, in the image
verify, verify-attestation, and verify-blob-attestation commands,
which are out of scope for this fix.

Validation:
Added TestVerifyBlobWithUnsetHashAlgorithmMatchesKeyDefault
(cmd/cosign/cli/verify/verify_blob_test.go), which signs a blob with a
generated P-521 key using SHA-512 and verifies it via VerifyBlobCmd.Exec
with HashAlgorithm left unset. Confirmed this test fails with the
exact reported error before this change (verified by reverting only
the implementation files) and passes after. Also added
TestSignatureDigestOptionsHashAlgorithm
(cmd/cosign/cli/options/signature_digest_test.go) covering the
unset/explicit/invalid cases for HashAlgorithm().

Ran:
  go build ./...
  go vet ./cmd/... ./pkg/...
  go test ./cmd/... ./pkg/...
  golangci-lint run ./cmd/cosign/cli/options/... ./cmd/cosign/cli/verify/...
All passed with no regressions.

Report: https://github.com/sigstore/cosign/issues/5063

Fixes #5063

### DIFF
--- a/cmd/cosign/cli/options/signature_digest.go
+++ b/cmd/cosign/cli/options/signature_digest.go
@@ -57,19 +57,20 @@ var _ Interface = (*SignatureDigestOptions)(nil)
 func (o *SignatureDigestOptions) AddFlags(cmd *cobra.Command) {
 	validSignatureDigestAlgorithms := strings.Join(supportedSignatureAlgorithmNames(), "|")
 
-	cmd.Flags().StringVar(&o.AlgorithmName, "signature-digest-algorithm", "sha256",
+	cmd.Flags().StringVar(&o.AlgorithmName, "signature-digest-algorithm", "",
 		fmt.Sprintf("digest algorithm to use when processing a signature (%s)", validSignatureDigestAlgorithms))
 	_ = cmd.Flags().MarkDeprecated("signature-digest-algorithm", "please use --bundle, which already includes the digest algorithm")
 }
 
 // HashAlgorithm converts the algorithm's name - provided as a string - into a crypto.Hash algorithm.
-// Returns an error if the algorithm name doesn't match a supported algorithm, and defaults to SHA256
-// in the event that the given algorithm is invalid.
+// Returns an error if the algorithm name doesn't match a supported algorithm. Returns 0 if the
+// algorithm hasn't been explicitly set, so that callers can fall back to a key-specific default
+// instead of always assuming SHA256.
 func (o *SignatureDigestOptions) HashAlgorithm() (crypto.Hash, error) {
 	normalizedAlgo := strings.ToLower(strings.TrimSpace(o.AlgorithmName))
 
 	if normalizedAlgo == "" {
-		return crypto.SHA256, nil
+		return 0, nil
 	}
 
 	algo, exists := supportedSignatureAlgorithms[normalizedAlgo]
--- a/cmd/cosign/cli/options/signature_digest_test.go
+++ b/cmd/cosign/cli/options/signature_digest_test.go
@@ -1,0 +1,55 @@
+//
+// Copyright 2026 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package options
+
+import (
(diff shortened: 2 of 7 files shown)


Hayden-IO (commented):
Thanks so much for taking care of this! Can we also update the verify, verify-attestation, and verify-attestation-blob paths?

https://github.com/sigstore/cosign/blob/e9af93ff293490fa2a7cb8088e712e21f11a5ded/cmd/cosign/cli/verify/verify.go#L99-L102

https://github.com/sigstore/cosign/blob/e9af93ff293490fa2a7cb8088e712e21f11a5ded/cmd/cosign/cli/verify/verify_attestation.go#L86-L89

https://github.com/sigstore/cosign/blob/e9af93ff293490fa2a7cb8088e712e21f11a5ded/cmd/cosign/cli/verify/verify_blob_attestation.go#L89-L92

Hayden-IO (approved):

Hayden-IO:
Once we work through a few more of the open PRs and issues, I'll plan to cut one more 3.x release before we get in #4959, which is a breaking change.

Hayden-IO:
can you run make docgen?

pujitha24:
Ran `make docgen` locally against this branch and it produced no diff — the doc/ files are already in sync with the current flag set, so there's nothing to commit there. Looking at the failed Verify Docgen run (on 938c6e9a), the actual cause is "The hosted runner lost communication with the server," which is a GitHub Actions infra hiccup, not a doc mismatch. Re-running that check should clear it.

Hayden-IO:
whoops, sorry, used to seeing that error for other reasons!

PR #5070, opened 2026-08-24
refactor: extract signing flag validation

aaronlew02:
#### Summary

This change simplifies `LoadTrustedMaterialAndSigningConfig` by extracting its signing flag validation logic and relevant inputs into a new function `ValidateSigningOptions`.

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -80,7 +80,7 @@ func Attest() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := signcommon.ValidateSigningOptions(cmd.Context(), o.UseSigningConfig, o.SigningConfigPath,
 				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL,
-				o.TlogUpload, o.NewBundleFormat, "",
+				o.TlogUpload, o.NewBundleFormat, o.BundlePath,
 				"", "", "", "", "", ""); err != nil {
 				return err
 			}
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -78,6 +78,13 @@ func Attest() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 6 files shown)


cmd/cosign/cli/attest.go
```diff
@@ -75,6 +75,13 @@ func Attest() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := signcommon.ValidateSigningOptions(cmd.Context(), o.UseSigningConfig, o.SigningConfigPath,
+				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL,
+				o.TlogUpload, o.NewBundleFormat, "",
```
steiza:
```suggestion
				o.TlogUpload, o.NewBundleFormat, o.BundlePath,
```
This is admittedly quite the corner-case (and if this is a bug, then it was pre-existing), but if we have `o.BundlePath`, should we supply it here?

cmd/cosign/cli/sign.go
```diff
@@ -94,6 +94,13 @@ race conditions or (worse) malicious tampering.
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := signcommon.ValidateSigningOptions(cmd.Context(), o.UseSigningConfig, o.SigningConfigPath,
+				o.Rekor.URL, o.Fulcio.URL, o.OIDC.Issuer, o.TSAServerURL,
+				o.TlogUpload, o.NewBundleFormat, "",
```
steiza:
```suggestion
				o.TlogUpload, o.NewBundleFormat, o.BundlePath,
```
Same question as above.

steiza (commented):
Two minor questions (well, really the same minor question twice)

cmd/cosign/cli/attest.go
aaronlew02: Good catch! This bug is pre-existing. Fixed in [`19cb9ca`](https://github.com/sigstore/cosign/pull/5070/commits/19cb9ca220e9ca8aa6c5f1e3fa36b7563c85da75).

cmd/cosign/cli/sign.go
aaronlew02: See above.

steiza (approved):

PR #4727, opened 2026-02-27
Backport bundle detection to sign and attest

steiza:
#### Summary

For #4531; to help folks on older Cosign versions start adopting new bundle formats without making verification more complicated.

Tested with:
```
$ go run cmd/cosign/main.go sign --new-bundle-format=true localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a@sha256:9f596f3fcea4eee0c4d2fb266adc5ed7024441cb7cf9b9a058c29ac501871d2a
...
$ go run cmd/cosign/main.go verify --certificate-oidc-issuer="..." --certificate-identity="..." localhost:1338/demo/sigstore-go-verification-ce4f5bf3233ac2b951c3667b2d19da3a@sha256:9f596f3fcea4eee0c4d2fb266adc5ed7024441cb7cf9b9a058c29ac501871d2a
...
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates
...
```

#### Release Note

* Backported new protobuf bundle detection to subcommands `verify` and `verify-attestation`

#### Documentation

N/A

PR #5075, opened 2026-08-26
Backport: Automatically require signed timestamp with Rekor v2 entries

codysoyland:
Backport of #4666 to release-2.6.

Since v2.6.3 (via the #4727 backport), verify and verify-attestation use new-format Sigstore bundles when an image has them. Rekor v2 entries in those bundles have no integrated timestamp, so keyless verification fails with "threshold not met for verified log entry integrated timestamps: 0 < 1" unless the user passes --use-signed-timestamps. In practice, an image signed with a Rekor v2 bundle verifies on cosign <=2.6.2 (which ignores the bundles) and on >=3.0.5 (which has this fix), but fails by default on 2.6.3-2.6.5.

This is a pure cherry-pick: the only conflict was in verify_bundle_test.go's imports, resolved by taking the upstream version of the file as-is, so both files are identical to main. Tested against an image signed with Rekor v2 bundles: stock v2.6.5 fails as described above, and with this change it verifies.

Assisted by Claude Fable 5

### DIFF
--- a/pkg/cosign/verify_bundle.go
+++ b/pkg/cosign/verify_bundle.go
@@ -21,8 +21,11 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/verify"
 )
 
-// VerifyNewBundle verifies a SigstoreBundle with the given parameters
+// VerifyNewBundle verifies a Sigstore bundle with the given parameters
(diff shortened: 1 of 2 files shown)


steiza (approved):
LGTM!

PR #5076, opened 2026-08-27
Align privacy statement prompt behavior

### LINKED ISSUE #5015: Align privacy statement prompt behavior
**Description**

There is currently ambiguity regarding when the Sigstore public good privacy statement should be presented to users and how that should be determined.

Currently, the following users are prompted:
* **ALL** keyless signers using the interactive OIDC authentication flow (see `internal/auth/auth.go`), even if using non-public-good services.
* Signers who do not use a signing config (provided either from TUF or manually) and do not set `--tlog-upload=false` (see the call to `signcommon.ShouldUploadToTlog` in `sign` and `sign-blob`).
    * These users will no longer be prompted following #5059, which unconditionally initializes `ko.SigningConfig`.
 
Key-based signers using public-good services via a signing config are unprompted.

For the sake of consistency, this issue proposes either:
* removing the privacy statement, or
* prompting all signers using a public-good service.

aaronlew02:
Closes #5015

#### Summary

This change aligns the appearance of the Sigstore public good privacy statement so that all signers using a Sigstore public good transparency log (i.e., Rekor or Fulcio's CT log) are prompted to confirm.

To achieve this:
* The determination of whether a Rekor upload will occur (via `ShouldUploadToTlog`) is expanded to all signers, not just those not using a signing config.
    * **Note:** `attest` and `attest-blob` did not call `ShouldUploadToTlog` due to a regression in #4618 but now do.
* The privacy statement logic for Rekor is extracted from `ShouldUploadToTlog` into a new function `ConfirmPrivacyStatement`, which is called universally after `ShouldUploadToTlog`.
* The privacy statement logic for Fulcio's CT log, called by `GetOAuthFlow` erroneously for all keyless signers, is removed in favor of a public good Fulcio check for keyless signers in `ConfirmPrivacyStatement`.

#### Release Note

* Privacy statement prompt behavior aligned such that **ALL** signers using a public good Rekor or Fulcio service are required to confirm the Sigstore public good privacy statement

### DIFF
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -139,9 +139,13 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		if err != nil {
 			return fmt.Errorf("creating signing config: %w", err)
 		}
-		if !c.TlogUpload {
-			c.SigningConfig = c.SigningConfig.WithRekorLogURLs()
-		}
+	}
+	shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, c.KeyOpts, digest, c.TlogUpload)
+	if err != nil {
+		return fmt.Errorf("should upload to tlog: %w", err)
+	}
+	if !shouldUpload {
+		c.SigningConfig = c.SigningConfig.WithRekorLogURLs()
 	}
 
 	bundleBytes, pubKey, hashAlgProto, err := signcommon.NewAttestationBundle(ctx, c.KeyOpts, c.CertPath, c.CertChainPath, bundleOpts, c.SigningConfig, c.TrustedMaterial)
--- a/cmd/cosign/cli/attest/attest.go
+++ b/cmd/cosign/cli/attest/attest.go
@@ -140,12 +140,12 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 			return fmt.Errorf("creating signing config: %w", err)
 		}
 	}
-	shouldUpload, err := signcommon.ShouldUploadToTlog(ctx, c.KeyOpts, digest, c.TlogUpload)
-	if err != nil {
-		return fmt.Errorf("should upload to tlog: %w", err)
-	}
+	shouldUpload := signcommon.ShouldUploadToTlog(ctx, c.KeyOpts, digest, c.TlogUpload)
 	if !shouldUpload {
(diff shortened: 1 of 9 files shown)


cmd/cosign/cli/attest/attest.go
```diff
@@ -139,9 +139,13 @@ func (c *AttestCommand) Exec(ctx context.Context, imageRef string) error {
 		if err != nil {
 			return fmt.Errorf("creating signing config: %w", err)
 		}
-		if !c.TlogUpload {
-			c.SigningConfig = c.SigningConfig.WithRekorLogURLs()
-		}
+	}
+	shouldUpload := signcommon.ShouldUploadToTlog(ctx, c.KeyOpts, digest, c.TlogUpload)
```
steiza: Minor: `shouldUpload` might be slightly confusing in the container context, as it could refer to uploading the bundle to the registry. But I don't feel strongly enough to insist that we change it.

cmd/cosign/cli/signcommon/common.go
```diff
@@ -118,12 +117,45 @@ func GetKeypairAndToken(ctx context.Context, ko options.KeyOpts, cert, certChain
...
+}
+
+// ConfirmPrivacyStatement prompts the user with the Sigstore privacy statement
+// if the operation will record data to a public transparency log.
+func ConfirmPrivacyStatement(ctx context.Context, ko options.KeyOpts, uploadToRekor bool) error {
+	isKeyless := (ko.KeyRef == "" && !ko.Sk) || ko.IssueCertificateForExistingKey
+	// The privacy statement applies when publishing to the public Rekor
+	// transparency log or the public Fulcio Certificate Transparency (CT) log.
+	if (uploadToRekor && hasPublicGoodRekorURL(ko.SigningConfig)) ||
+		(isKeyless && hasPublicGoodFulcioURL(ko.SigningConfig)) {
```
steiza, line 157: Oh man, nice catch! I definitely would have missed this.

steiza (dismissed):
Nice work!

I cannot wait to fully remove `--tlog-upload` in the near future to simplify things. 😞

cmd/cosign/cli/attest/attest.go
aaronlew02: I agree. I'll think about this for a later change.

aaronlew02: I've renamed `shoudUpload` to `uploadToTlog` across commands.

steiza (approved):

PR #5079, opened 2026-08-27
fix: pass missing OIDC flags to KeyOpts

aaronlew02:
#### Summary

This change passes missing OIDC-related flags to the `KeyOpts` of certain commands.

Previously, these flags were registered on the CLI but not passed into `KeyOpts`:
* `sign-blob` was missing `--oidc-provider`.
* `attest` and `attest-blob` were missing `--oidc-disable-ambient-providers`.


steiza (approved):
I was confused as to why there weren't docgen changes, but it looks like we were just silently dropping these values on the floor 😞. At any rate, nice catch!

PR #5080, opened 2026-08-27
fix(verify): support image indexes in local bundle discovery

aaronlew02:
#### Summary

This change adds support for image indexes (with the kind annotation `dev.cosignproject.cosign/imageIndex`) in `getLocalBundleDescriptors` when discovering bundles in a local layout.

Previously, `getLocalBundleDescriptors` only checked for the `dev.cosignproject.cosign/image` kind annotation.


steiza (approved):
Nice catch! This will be very useful for detecting / verifying local bundles.

PR #5081, opened 2026-08-28
fix: restrict bundle signing and attestation to one artifact

aaronlew02:
#### Summary

This change restricts signing and attestation to a single artifact (blob or image) when a bundle is requested.

Previously, if a bundle was requested for a single signing or attestation operation over multiple artifacts, the requested bundle was overwritten for each artifact.

For `sign-blob`, this restricts users to a single blob input, and for `sign` and `attest`, this prevents users from providing multiple image inputs when requesting a bundle (as well as `--recursive` for `sign`).

`attest-blob` does not require a change, as inputs are already restricted to a single blob.

### DIFF
--- a/cmd/cosign/cli/attest.go
+++ b/cmd/cosign/cli/attest.go
@@ -68,9 +68,12 @@ func Attest() *cobra.Command {
 
 		Args:             cobra.MinimumNArgs(1),
 		PersistentPreRun: options.BindViper,
-		PreRunE: func(_ *cobra.Command, _ []string) error {
+		PreRunE: func(_ *cobra.Command, args []string) error {
 			if o.NewBundleFormat && o.NoUpload && o.BundlePath == "" {
 				return fmt.Errorf("must enable upload to the OCI registry or specify a local --bundle path with --new-bundle-format")
+			}
+			if o.BundlePath != "" && len(args) > 1 {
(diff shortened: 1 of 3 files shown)


steiza (dismissed):
This LGTM!

I always worry about if we're breaking folks who were relying on the previous behavior, but it looks like the previous behavior here was "last bundle wins" and so this was probably already broken for folks - better to let them know upfront that it wont' work.

It looks like this includes the #5080 changes? So want to land this and close #5080? Otherwise if you land #5080 first you'll have to rebase this.

aaronlew02:
> It looks like this includes the https://github.com/sigstore/cosign/pull/5080 changes? So want to land this and close https://github.com/sigstore/cosign/pull/5080? Otherwise if you land https://github.com/sigstore/cosign/pull/5080 first you'll have to rebase this.

@steiza My intention was to land #5080 first since these are separate changes.

steiza:
> @steiza My intention was to land https://github.com/sigstore/cosign/pull/5080 first since these are separate changes.

It looks to me like both pull requests contain commit `f201065`, but either way is fine with me!

aaronlew02:
@steiza I've rebased so that this PR contains only the commit pertinent to its change.

steiza (approved):

PR #5085, opened 2026-09-01
feat: add native Syft JSON support in cosign attest

### LINKED ISSUE #4773: Enhancement: Add native Syft JSON support in cosign attest
I noticed that Syft JSON SBOMs were previously supported in the deprecated [cosign attach sbom](https://github.com/sigstore/cosign/blob/4c9ba21b465b07ed03da06ebb7a5ef3e4e5913fc/doc/cosign_attach_sbom.md) command. With the move to the new SBOM attestations workflow, Syft JSON is no longer a supported --type. Currently, the supported types are:

> slsaprovenance|slsaprovenance02|slsaprovenance1|link|spdx|spdxjson|cyclonedx|vuln|openvex|custom

While --type custom could technically carry a Syft JSON file, it is not automatically recognized as an SBOM by most tools, and users lose the native workflow integration they previously had.

Enhancement request:

* Add native support for Syft JSON as a recognized --type in [cosign attest](https://github.com/sigstore/cosign/blob/4c9ba21b465b07ed03da06ebb7a5ef3e4e5913fc/doc/cosign_attest.md).
* This would simplify workflows for teams that:
   * Use Syft for SBOM generation 
   * Rely on Grype/VEX for vulnerability scanning
   * Want to attach SBOMs directly to container images without generating duplicate formats (CycloneDX/SPDX)

Was there a specific reason Syft JSON was removed as a --type when moving from cosign attach sbom to the new SBOM attestations workflow?

ompushkara:
> Fixes #4773

### DIFF
--- a/cmd/cosign/cli/options/predicate.go
+++ b/cmd/cosign/cli/options/predicate.go
@@ -34,6 +34,7 @@ const (
 	PredicateSPDX      = "spdx"
 	PredicateSPDXJSON  = "spdxjson"
 	PredicateCycloneDX = "cyclonedx"
+	PredicateSyft      = "syft"
 	PredicateLink      = "link"
 	PredicateVuln      = "vuln"
 	PredicateOpenVEX   = "openvex"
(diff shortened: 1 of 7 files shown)


cmurphy (approved):
@steiza @Hayden-IO do you remember any reason we wouldn't want to expand this list?

steiza (changes requested):
Yeah, I think we want to get out of the business of maintaining a list of predicates in `--type`.

Instead, I would recommend that folks use `cosign attest` (or `cosign attest-blob`) with the existing `--predicate` flag (and `--statement` for the in-toto subject).

We aren't actively deprecating `--type`, but that could be a good project for after https://github.com/sigstore/cosign/pull/5109 / when Cosign v4 is out.

PR #5094, opened 2026-09-02
fix(verify): propagate CommonVerifyOptions to VerifyCommand

aaronlew02:
#### Summary

This change propagates `CommonVerifyOptions` to all callers constructing `VerifyCommand`, so that shared flags like `--allow-certificate-chain` and `--trusted-root` are not silently dropped. It also removes duplicate field definitions on `VerifyCommand` to prevent field shadowing from ignoring options set in `CommonVerifyOptions`.


steiza (approved):
LGTM!

PR #5103, opened 2026-09-07
copy annotations by key instead of copying obj

bobcallaway:
#4498 added the title annotation to layer descriptors, however when preserving per-layer annotations (e.g. from an `oci.Signature` layer returned by `cosignstatic.NewAttestation(...)` which includes signature or certificate metadata), `layerDescriptors[i].Annotations` was directly overwritten


steiza (approved):
LGTM!

PR #5111, opened 2026-09-11
fix: write referrers when image index has attestations

aaronlew02:
#### Summary

This change removes an early return in `WriteSignedImageIndexImages` for image indexes with attestations that prevented referrers from being written later in the function.


steiza (approved):
LGTM!

PR #5124, opened 2026-09-15
fix(verify): enforce mutually exclusive key and security key

aaronlew02:
#### Summary

This change enforces the mutual exclusivity of a key and security key in `verify`, matching all other signing and verification commands.


cmd/cosign/cli/verify/verify.go
```diff
@@ -104,6 +104,11 @@ func (c *VerifyCommand) Exec(ctx context.Context, images []string) (err error) {
 		return &options.KeyAndIdentityParseError{}
 	}
 
+	// key and security key are mutually exclusive
+	if options.NOf(c.KeyRef, c.Sk) > 1 {
```
steiza, line 108: I had to check what we were doing in the other `verify*` subcommands:

- `verify-attestation` has the same check
- `verify-blob` we include `c.CertRef`, but I'd just as soon get rid of `c.CertRef` instead of add it here
- `verify-blob-attestation` has the same check

steiza (approved):

PR #5125, opened 2026-09-17
Reorder events in LoadTrustedMaterialAndSigningConfig

aaronlew02:
#### Summary

This change reorders the events in `LoadTrustedMaterialAndSigningConfig` (now named `LoadSigningConfigAndTrustedMaterial`) so that the signing config is loaded first. The trusted root is then loaded, with a preference for a user-provided file; otherwise, the trusted root is loaded from TUF under the same conditions, but only if a signing config with services has been loaded. 

This allows users to:
* Sign offline (i.e., without contacting TUF) by providing a signing config with no services configured
* Provide a trusted root file with `--use-signing-config=false` (e.g., for key-based signing)
* Fail fast with an invalid signing config without contacting TUF for a trusted root

This change also corrects the error returned from a bad fetch of a trusted root from TUF, replacing `Continuing with individual targets` with `Continuing without trusted root`.


cmurphy (approved):

steiza (approved):
Love it!!!

This is a great example of enabling functionality (offline signing) in a flexible way without requiring additional CLI flags. Nice work!

PR #5126, opened 2026-09-18
feat: support selecting PIV cards

### LINKED ISSUE #4493: Add support for multiple PIV tokens
Right now, Cosign doesn't allow to sign when multiple PIV tokens are connected: it errors out and no signatures re generated:

```
signing digest: getting keypair and token: getting signer: open key: found 2 cards, please attach only one
```

This can be problematic in scenarios where you are using a machine in which you don't have physical access.

Having a way to filter the PIV tokens by specifying a `--piv-uuid` flag would be a way to solve this.

What do you think? Would such a contribution welcome?

niallnsec:
(cherry picked from commit ede652bcb7dda714ab266e233c1683edfd1760fe)

#### Summary

Resolves sigstore/cosign#4493 ([issue](https://github.com/sigstore/cosign/issues/4493)).

Adds `--piv-serial` and `--piv-key-sha256` selectors to commands supporting `--sk`, allowing the correct PIV card to be selected when multiple cards are attached. If no selector is supplied, the error now lists each card's reader name, serial number, and slot public-key SHA-256 fingerprint.

The proposed UUID selector was not used because `go-piv` does not expose the PIV CHUID/GUID. The relevant [go-piv/piv-go#67](https://github.com/go-piv/piv-go/pull/67) implementation has remained open for over six years without further maintainer response since September 2020. The serial and public-key selectors use information available through the current API.

Validated with:

- `go test ./pkg/cosign/pivkey ./cmd/cosign/cli/...`
- `go test -tags=pivkey ./pkg/cosign/pivkey`
- `go build -tags=pivkey,pkcs11key ./cmd/cosign`
- `bash cmd/help/verify.sh`
- `make lint`

Card-selection behavior is covered using mock cards; no physical-token tests were run.

#### Release Note

Added `--piv-serial` and `--piv-key-sha256` options for selecting a PIV card when multiple cards are attached, and improved multi-card error diagnostics.

#### Documentation

Updated the generated CLI reference documentation and `--help` output. No separate `docs.sigstore.dev` update is required.


steiza (changes requested):
Hello! Thanks for submitting this PR.

In general, we are trying to greatly reduce the number of flags Cosign supports (see #4696).

But, of course, we still want to find some way of supporting this functionality; ideally without adding additional flags.

What do you think of one of these options:

- Prompt the user to choose among the items returned from `piv.Cards()`? The advantage here is we don't need additional flags, the downside is the signing might be taking place in automation where user input is not available.

- Getting this information from an environmental variable instead? See `pkg/cosign/env/env.go` for some prior work here. This would support automated workflows, although maybe it's not as convenient as a CLI flag.

All that aside, I'm also wondering if we need both the serial number as well as the key hash? Would just the serial number be sufficient? I don't often work with multiple PIV devices on the same machine.

lukaszszafranski (commented):
## Veridical review

AI-assisted review, checked against the current PR head `af36b2616bdc814bd4a8364f6b23d9c446d854ff`. One medium-severity finding is attached to the changed line below.

[Veridical.dev](https://veridical.dev) · Contact: [contact@veridical.dev](mailto:contact@veridical.dev)

cmd/cosign/cli/verify.go
```diff
@@ -113,6 +113,8 @@ against the transparency log.`,
 				SCTRef:                       o.CertVerify.SCT,
 				Sk:                           o.SecurityKey.Use,
 				Slot:                         o.SecurityKey.Slot,
+				PIVSerial:                    o.SecurityKey.PIVSerial,
```
lukaszszafranski: ### Medium: Reject `--key` with `--sk` before selecting a verifier

**Impact:** `cosign verify --key file.pem --sk --piv-serial 123 IMAGE` uses the file-key verifier without opening or checking the selected PIV card. A caller supplying both flags can believe the requested hardware key was enforced when it was not. The signature must still match `file.pem`.

**Evidence:** This command passes both `KeyRef` and `Sk`/`PIVSerial` into `VerifyCommand`. `VerifyCommand.Exec` checks `KeyRef` against certificate identity, but does not reject `KeyRef` with `Sk`. In `LoadVerifierFromKeyOrCert`, the `keyRef != ""` branch returns before the `sk` branch can inspect the card or selector. The blob and attestation verification commands already reject this combination.

**Fix:** Reject `--key` with `--sk` for image verification and add a regression case that supplies a PIV selector alongside `--key`.

[Veridical.dev](https://veridical.dev) · Contact: [contact@veridical.dev](mailto:contact@veridical.dev)

